Issue #15, opened 2020-11-16
Clearly define use-cases of `HTTPMethodModifier`


lschlesinger:
**Is your change related to a problem? Please describe.**

As of now the `HTTPMethodModifier` has the provided scope `.environment`. I find this unintuitive, as it implies that `HTTPMethod`s are set on different components recursively, instead of the more restrictive (and IMO more intuitiv) scope `.nextComponent`. From my understanding the current implementation changes the _default_ HTTP method for all nested components.
However, even _if_ `.environment` is assumed as the (also semantically) correct scope for `HTTPMethodModifier`s, it is currently implemented incorrectly (see #12  ).

Example to reproduce the bug:

```swift
// ...
Group {
    Text("Hi from Test!")
        .httpMethod(.PUT)
    Text("Bye from Test!")
}.httpMethod(.POST)
// ...
```
will actually output:
```
[ INFO ] v1 + PUT -> String with 1 guards.
[ INFO ] v1 + GET -> String with 1 guards. // <- here, the new "group-default" POST should be used
```

After discussing with Paul, we should overthink the use of `HTTPMethodModifier` on `.environment` scope and more generally, if HTTP methods depict a proper abstraction level, since GraphQL, gRPC, and WebSockets exporters do not rely on this abstraction.

**Propose a solution**

We should discuss the following points:

- Proper scope of `HTTPMethodModifier`
- Proper abstraction level for "CRUD" methods (e.g., HTTP verbs, `.read`/`.create`/...)
- Define mapping for deriving exporter-specific actions from the agreed upon abstraction level

PR #17, opened 2020-11-22
Replace the http modifier with a more abstract CRUD modifier

bauer-andreas:
This is a follow up to https://github.com/Apodini/Apodini/issues/15 (@lschlesinger).
In meeting 3 we agreed on replacing the HTTP verbs with some custom CRUD verbs (and discussed to change the scope from `.environment` to `.nextComponent`).
I'm using this PR to start a discussion about a specific realization which will make its way into the main branch (by already providing a suggestion, which is currently also used in the prototype for the shared semantic model builder).

I see four points which still need to be discussed:
* How should we name it (I proposed `Operation` with this PR, but not sure if there is maybe some better name)?
* What should the scope be? The referenced issue suggest changing it to `.nextComponent` but I'm not sure if we agreed on that, thus this PR still uses `.environment`
* What should we define as the default value for the ContextKey? Setting `.READ` as the default makes sense for REST, but for gRPC it could be beneficial to tell the difference between undefined or explicitly defining `.READ` (maybe?).
* Are there any other values we should consider (which are relevant for any of the other interface exporters)?
For REST there is a difference between PUT and PATCH (overwriting the resource vs modifying individual properties of it). In that could be helpful to add another operation `.MODIFY` besides `.UPDATE`?


Sources/Apodini/Modifier/OperationModifier.swift
```diff
@@ -0,0 +1,75 @@
+//
+//  OperationModifier.swift
+//  Apodini
+//
+//  Created by Paul Schmiedmayer on 6/26/20.
+//
+
+public enum Operation: RawRepresentable {
```
PSchmiedmayer: I would suggest to spell the enum cases lowercases as these are not fixed terms like `GET` for HTTP but can be expressed by verbs and `.create` looks a bit more Swift like.

In addition I would suggest to remove the conformance to `RawRepresentable` for the `Operation` enum as the raw value is currently not used and the Swift Runtime uses the enum name in print statements if there is no other information provided by a description property.

Sources/Apodini/Modifier/OperationModifier.swift
```diff
@@ -0,0 +1,75 @@
...
+    init(_ component: ModifiedComponent, operation: Operation) {
+        self.component = component
+        self.operation = operation
+    }
+}
+
+
+extension OperationModifier: Visitable {
+    func visit(_ visitor: SynaxTreeVisitor) {
+        visitor.addContext(OperationContextKey.self, value: operation, scope: .environment) // TODO issue #15, change that to .nextComponent?
```
PSchmiedmayer: Yes, I would agree to change the operation to only affect the next component.

PSchmiedmayer (changes requested):
@Supereg Thanks for the improvement! I just have added a few smaller comments regarding he implementation of the `Operation` type that should make the PR even shorter, otherwise it looks great! I would also support the decision to make the `OperationModifier` only apply to the next component.

PSchmiedmayer (dismissed):
As all tests are now passing and the changes in the `Operation` enum have been applied I approve the PR.

bauer-andreas:
> As all tests are now passing and the changes in the `Operation` enum have been applied I approve the PR.
Thanks.

I'm still going to incorporate what we discussed in the last work meeting:
* A dedicated default `Operation`
* Some documentation detailing the behavior in the different exporters

PSchmiedmayer (dismissed):

PSchmiedmayer:
@Supereg Sounds good to me, then we will leave this PR open until then 👍

bauer-andreas:
> * A dedicated default Operation
> * Some documentation detailing the behavior in the different exporters

Added both.

Documentation/Operation.md
```diff
@@ -0,0 +1,59 @@
...
+
+As the `Operation` enum cannot be properly map to any feature of the technology itself (like HTTP methods can be
+used for HTTP), the gRPC exporter SHOULD incorporate the operation into the naming of the procedure.  
+As `.automatic` is the default value, meaning most likely the user hasn't specified any `.operation`, it should
+be ignored and no changes should be done to the procedure name.
+
+
+### GraphQL
+
+The `Operation` is not really relevant when creating the query schema. Instead the GraphQL exporter
```
PSchmiedmayer: GraphQL has two different types of functions that can be exposed: Queries and Mutations. The Operation could be used to infer if a Component is exposed as a query or mutation.

PSchmiedmayer (dismissed):
I only added a small comment regarding the GraphQL exporter, the PR otherwise looks great!

Documentation/Operation.md
bauer-andreas: Ah messed up here I think. What the sentence was meant to say is that `.automatic` isn't really useful to GraphQL right?
Cause GraphQL can only detect in an incoming query/mutation if that is a read or write request (or update). That information can then be used to match the appropriate `Component.handle`. Like a mutation will end up at the Component which has operation update/create.
Do I understand correctly?

So probably best would be for GraphQL to use `.read` as default for `.automatic`?

bauer-andreas:
Reworked/fixed the instructions for the GraphQL exporter. I hope I got what you wanted in there.

As I pushed another commit you will have the re-approve the PR.

PSchmiedmayer (approved):

Issue #12, opened 2020-11-15
Component tree is incorrectly mapped to ContextNode tree


bauer-andreas:
### Describe the bug

Before we describe the bug in detail we first look at the following provided tree of `Component`s

```swift
var content: some Component {
    Group("swift") {
        Text("Hello Swift! 💻")
        Group("mac") {
            Group("pro") {
                Text("Hello Mac Pro!")
            }.httpMethod(.POST)
            Text("Hello Mac!")
        }
    }
    Text("Hello World! 👋")
}
```

Looking at the provided example, the following output is what we would **expect** from the current `RESTSemanticModelBuilder` (Note I added a additional section at the end to identify the associated `Text` component more easily):

```
[ INFO ] v1/swift + GET -> String with 0 guards => "Hello Swift! 💻"
[ INFO ] v1/swift/mac/pro + POST -> String with 0 guards => "Hello Mac Pro!"
[ INFO ] v1/swift/mac + GET -> String with 0 guards => "Hello Mac!"
[ INFO ] v1 + GET -> String with 0 guards => "Hello World! 👋"
```

So lets look at what is actually produced:

```
[ INFO ] v1/swift + GET -> String with 0 guards => "Hello Swift! 💻"
[ INFO ] v1/mac/pro + POST -> String with 0 guards => "Hello Mac Pro!"
[ INFO ] v1/mac + POST -> String with 0 guards => "Hello Mac!"
[ INFO ] v1 + GET -> String with 0 guards => "Hello World! 👋"
```

We got two problems here. First, the paths of `Group("mac")` got mismatched. Instead of `v1/swift/mac[/pro]` we got `v1/mac[/pro]`. Secondly the route `v1/mac` registered with http method `POST` although it wasn't placed in the `Group("pro")` which is the one having the `httpMethod` modifier.

### To Reproduce

 1. To reproduce the path mismatch: Put any non `Visitable` (meaning any Component for which the `SyntaxTreeVisitor.register` method is called) into a `Group` component followed by any `Component` you want. You observer that any `Component` following the non `Visitable` Component will have a mismatched path.
 2. Add any modifiers to a `Group` component and you observer that those modifiers will also apply to Components on the **same** level.

### So why/how does it happen

First of all we need to understand how a `Group` instance looks like once built by the `ComponentBuilder`.  
For the `Group("mac")` the following is produced:

```swift
[...]

PR #19, opened 2020-11-23
Adjustable Configuration for Apodini

hendesi:
**Implementation**
This pull request contains changes by @hendesi and @awoc to the Apodini framework, which enable the enduser to define and use configurations for the Apodini system. The configuration structure behaves similar to the one of the components. This allows for a customisation by the user without exposing the internals of Apodini to the outside. The user can add various Configuration Object in the 
```
var configuration: some Configuration {
...
}
```

Configuration Objects need to conform to the `Configuration` protocol and need to implement the `func configure(_ app:Application)` method. The user can implement additional functions, which need to return `Self`. These can then be appended to the configuration object similar to the modifiers of components. The configurations are loaded after registering the semantic models and before the start of the server.

Example:

```
var configuration: some Configuration {
        DatabaseConfiguration(.mongo)
            .connectionString("mongodb://localhost:27017/vapor_database")
            .addMigrations(CreateBird())
    }
```

Example additional methods:
```
public func connectionString(_ string: String) -> Self {
        self.connectionString = string
        return self
    }
```

**Problems we faced**
Currently the `var configuration: some Configuration {}`needs to be implemented with at least one Configuration Object, otherwise the server startup will fail due to conformation to `Never`. Since the Configuration structure is very similar to the one of the components, we would face the same problem when removing `var content: some Component`. This is something we need to further investigate as users should still be able to use Apodini despite not implementing configuration.


Sources/Apodini/Configurations/AnyConfigurationCollection.swift
```diff
@@ -0,0 +1,49 @@
+//
```
nerdsupremacist: nit: Can we please remove the headers? Not a big fan of it

Sources/Apodini/Configurations/Configurator.swift
```diff
@@ -0,0 +1,42 @@
...
+//
+
+import Vapor
+
+protocol Configurable {
+    func configurable(by configurator: Configurator)
+}
+
+public class Configurator {
+    private(set) var app: Application
```
nerdsupremacist: nit: The configurator couples the DSL directly to Vapor. We need to start considering ways to decouple it. 
But that seems like it's out of scope for this PR ;)

nerdsupremacist (dismissed):
# Great Job!

Thanks for making this better ;)
I left a couple of nits ("not important though" comments)
Please deal with the lint errors and then we can merge this

## Future Work
I'm thinking that we should consider moving towards additional type safety in the Component DSL. Perhaps we should do that here as well. I'll start a branch with that as soon as we merge this ;)

Sources/Apodini/Configurations/APNSConfiguration.swift
```diff
@@ -0,0 +1,60 @@
...
+        } catch {
+            fatalError("Error setting up APNS")
+        }
+    }
+}
+
+
+public enum APNSAuthentication {
+    case pem(pemPath: String, privateKeyPath: String? = nil, pemPassword: [UInt8]? = nil)
+    case p8(path: String, keyIdentifier: JWTKit.JWKIdentifier, teamIdentifier: String)
```
awoc, line 72: Are too short enum element names really a problem or can we just change the swiftlint rule for this? I think using the file extension as the enum element name makes it clear which case to use

Sources/Apodini/Configurations/AnyConfigurationCollection.swift
```diff
@@ -0,0 +1,38 @@
...
+        self.configurations = [config]
+    }
+}
+
+extension AnyConfigurationCollection: Configurable {
+    func configurable(by configurator: Configurator) {
+        for config in configurations {
+            config.configurable(by: configurator)
+        }
+    }
```
bauer-andreas: Judging by `ConfigurationGroup.configurable` and the current implementation of `Configurator` which is using the `ContextNode` as well ( mimicking what the `SyntaxTreeVisitor` does for `Component`s), the current code might create the same issue described in https://github.com/Apodini/Apodini/issues/12 (which is trying to be fixed with https://github.com/Apodini/Apodini/pull/13).
The issue really is missing `Configurator.enter` and `Configurator.exit` calls in the marked code section.

Is there any intention to use `ContextNode` for future Configurations? Looking into the already provided implementation it doesn't seem like (there isn't really any possibility to use it really).

Sources/Apodini/Configurations/APNSConfiguration.swift
PSchmiedmayer: I think we can make an exception for this specific case on our .swiftLint file or just add a SwiftLint disable command here.

awoc:
@PSchmiedmayer @nerdsupremacist We reduced the complexity and resolved the problem with the missing `configuration` property at server startup. This slightly changed the DSL from `var configuration: some Configuration {}` to `var configuration:  Configuration {}`, therefore, not using an opaque type.

nerdsupremacist (dismissed):

Sources/Apodini/Configurations/APNSConfiguration.swift
```diff
@@ -0,0 +1,74 @@
...
+    ///     - authentication: The APNS authentication
+    ///     - topic: The Bundle identifier of your iOS application
+    ///     - environment: Specifies the APNS environment
+    public init(_ authentication: APNSAuthentication, topic: String, environment: APNSwiftConfiguration.Environment) {
+        self.authentication = authentication
+        self.topic = topic
+        self.environment = environment
+    }
+    
+    public func configure(_ app: Application) {
```
nerdsupremacist, line 25: Depending on what get's merged first: this or #32 we might habe to do some decoupling here as well

nerdsupremacist (dismissed):

Sources/Apodini/Configurations/DatabaseConfiguration.swift
```diff
@@ -0,0 +1,136 @@
...
+    ///
+    /// - Parameters:
+    ///     - migrations: One or more `Migration` objects that should be migrated by the database
+    public func addMigrations(_ migrations: Migration...) -> Self {
+        self.migrations.append(contentsOf: migrations)
+        return self
+    }
+    
+    private func databaseFactory(for type: DatabaseType) throws -> Fluent.DatabaseConfigurationFactory {
+        var databaseFactory: Fluent.DatabaseConfigurationFactory?
```
PSchmiedmayer:
```suggestion
```
This property is not needed, feel free to remove it here.

PSchmiedmayer (changes requested):
I only had a single comment regarding an unused variable. Apart from this the PR looks great, you can merge it after fixing the small issue.

PSchmiedmayer (approved):

PR #20, opened 2020-11-23
Adds Tuple Component and expands maximum of components to 100

### LINKED ISSUE #16: Maximum number of components
**Describe the bug**
Only 10 ```Component```s can be defined with the ```ComponentBuilder```. This means that Apodini can handle at most 10 Endpoints. Trying to add more ```Component```s will result in the following compilation error: **Extra arguments at positions # 11, ... in call**. This error is thrown because the ```ComponentBuilder``` only has functions for a fixed number of arguments, therefore not supporting an arbitrary number of ```Component```s.

**To Reproduce**
This bug can be reproduced by defining more than 10 ```Component```s, e.g.:

```swift
var content: some Component {
    Text("1")
    Text("2")
    ...
    Text("10")
    ...
}
```

**Expected behavior**
The ```ComponentBuilder``` should support a variable number of arguments in order to define as many ```Component```s as the User wants. This would also clean up the ```ComponentBuilder``` by unifying all functions.

 - Swift Version: Apple Swift version 5.3.1 (swiftlang-1200.0.41 clang-1200.0.32.8)
Target: x86_64-apple-darwin19.6.0
 - Xcode Version: 12.2 (12B45b)

nerdsupremacist:
## Summary
This PR seeks to add some extra type information by no longer using type erasure in the Component Builder by default. So I
- Deleted AnyComponentCollection
- Added TupleComponent with the generic parameter that keeps the component type information
- Moved the component builder to a gyb (generate your boilerplate) file and increased the number of maximum blocks to 100.

Closes #16 

## Test Plan
All previous tests pass ;)


PSchmiedmayer (changes requested):
Thanks for the improvement! This will help us a lot to get more type information out of the DSL tree. Using gyb sounds like a good idea!

Unfortunately there seems to be a build issue on linux at the moment when applying your changes. Can you take a look at this before we merge the PR. In addition there are some new changes and documentation on develop that can also help you to resolve your SwiftLint issues 👍

PSchmiedmayer (changes requested):
@nerdsupremacist It seems that the tests about `Group` `Component`s are still failing after replacing the `AnyComponentCollection` with a `TupleComponent`. I suspect that this is due to the changed `visit` function in the `TupleComponent`.
I would also suggest to add some additional regression tests about visiting Group components in the PR too.

nerdsupremacist:
@PSchmiedmayer Ah I see I understood enterCollectionItem wrong. It's fixed now

PSchmiedmayer (approved):

Issue #11, opened 2020-11-12
Concurrent requests on same endpoint cause undefined behavior


theMomax:
**Describe the bug**
@Supereg  and I discovered that the handling `Component`s are long living objects. Per endpoint, only one `Component` instance is created on startup which is then used throughout the whole lifetime of the process. However, Vapor is a multithreaded server, i.e. it creates multiple event loops on startup which can process requests in parallel. Thus multiple parallel requests on the same endpoint are handled by the same `Component` at the same time without synchronization.

Consequences:

Concurrent access to `RequestInjectable`s is not handled: 
* When the `Component`'s `handle` function is called to provide a result for request **A** the `RequestInjectable` property wrapper may hold content belonging to request **B**  
      `inject(A) -> inject(B) -> handle(A) -> ...`
* request **A** may access a `RequestInjectable` which's `disconnect` method was already called by a overlapping request **B** resulting in a server-crash  
      `inject(A) -> inject(B) -> handle(B) -> disconnect(B) -> handle(A) -> crash`

**To Reproduce**
A PoC to reproduce the issue is available on branch [experimental/control-flow](https://github.com/Apodini/Apodini/tree/experimental/control-flow) on commit a5215fec284777bffe96ca669dd686cb9936f098. Start the Apodini TestServer and run the test script (`go run Tests/test_thread_safety.go`). There are also pre-compiled versions available for 64-bit Windows, Linux and macOS. 

The test spawns multiple worker-threads which execute requests on the server's `greet` endpoint. This endpoint takes a HTTP query parameter named `name`. The `Greeter`'s `handle` function returns this parameter's value. Each request issued by the script provides a different value for the `name` parameter. If Apodini is not thread-safe several things might happen:
* The Apodini server crashes
* The response does not match the request's `name` parameter

In all cases this test exits with code 1 and prints out an according error message. The test passes only after all requests have returned with the expected result.

**Expected behavior**
Reducing the `threads` to 1 shows the expected behavior.

**Environment**

 - Swift Version:
    Apple Swift version 5.3 (swiftlang-1200.0.29.2 clang-1200.0.30.1)
    Target: x86_64-apple-darwin19.6.0
 - Xcode Version: Version 12.1 (12A7403)

**Discussion**

We need to discuss the concept of one `Component` instance being responsible of all requests arising from one endpoint. While currently the only concrete issue are `RequestInjectable`s, this concept could cause more problems in the future:
[...]

PR #22, opened 2020-11-24
Fix concurrent requests on same endpoint

theMomax:
This PR solves issue #11. As stated in the issue the only concrete problem have been `RequestInjectable`s, which - as of this PR - are no longer classes but mutating structs. When a `Component` enters the request context, the framework creates a shallow copy of the context (along with its value properties). Then the runtime library is used to detect `RequestInjectable` children, which are then mutated (injected with the request) and overwritten on the `Component`.

Once we have a a more defined view on how control flow (in regards of ObservableObject properties, etc.) work and we implement additional property wrappers, we should define a public `DynamicProperty` protocol to generalize this solution and allows for developers to create their own property wrappers while still being protected by the Apodini framework.

This PR also fixes some missing default decoders that should have been passed to the `enterRequestContext` method.

Tagged Persons: @Supereg @nerdsupremacist

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,8 +1,9 @@
 name: Publish Documentation
 
 on:
-  release:
-    types: [published]
+  push:
+    branches:
+      - develop
 
 jobs:
   deploy_docs:
--- a/.github/workflows/swift.yml
+++ b/.github/workflows/swift.yml
@@ -42,4 +42,4 @@ jobs:
     - name: Build
       run: swift build
     - name: Test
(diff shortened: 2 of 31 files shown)


PSchmiedmayer:
@theMomax I have fixed the build errors on Linux by merging the current state of develop in this branch 👍

PSchmiedmayer (approved):
@theMomax Thanks for the improvements! I like the solution you and @nerdsupremacist have found using the Runtime framework. Feel free to merge the PR into the develop branch.

PR #23, opened 2020-11-24
Improve the @Parameter manifesto and add a basic implementation

PSchmiedmayer:
This pull request improves the `@Parameter` manifesto to include a single `@Parameter` property wrapper and adds a suggestion on how to implement this to the Apodini repository.


PSchmiedmayer:
@Supereg @nerdsupremacist @theMomax @lschlesinger @tgymnich We discussed a few approaches of even using a single Property Wrapper for `@Parameter` to express all three types of `@Parameters` we can express in Apodini. I have updated the manifesto and added an implementation that enables us to eliminate the `@PathParameter` property wrapper. Using the `@Parameter(.path)` as a way to express the exact same functionality and still pass the `@Parameter` across multiple components as described in the end of the manifesto. Let me know what you think.

moritzschuell:
Looks good to me, in general 👍 

The only concern I have for gRPC is the mapping of each `@Parameter` to the appropriate field-tag in the Protobuf message. Based on the existing properties of Parameter, I am not sure how to derive a meaningful number for the field-tag.

In contrast to the `@Body`, `@Parameter`s can represent parts of the Protobuf message (not the whole message, as was the case with `@Body`). Thus, we need a field tag for each Parameter to locate its data in the Protobuf message.

I think we have three options to get this info:
- Add a `fieldTag: Int` to the class `Parameter`, and have the user define it (just like `name` in the current implementation). This would probably be the most convenient choice for us.
- Have the user add `CodingKey`s to the Component that `Parameter`s are placed in. However, this would not really work I guess, because in the `inject` method in `Parameter` we are missing knowledge about the Component that the Parameter is placed in.
- Make some venturous guesses about the field-tag based on the order of `Mirror(..).children`. This would assume that Mirror iterates over children in the same order every time (which I am not sure of). And it would also require us to bloat the `inject(..)` method in the `RequestInjectable` protocol with another parameter that is only relevant to gRPC.

Do you have further suggestions on how we could derive a field tag for the Parameters? Maybe I am missing something 🤓
As indicated, I would probably prefer option 1.

Documentation/PropertyWrapper/Parameter.md
```diff
@@ -3,7 +3,7 @@
 The `@Parameter` property wrapper can be used to express input in different ways: 
 * **Request Content**: Parameters can be part of the requests send to the web service such as the HTTP body or request content of an other protocol.
```
lschlesinger, line 4: typo _sent_

Documentation/PropertyWrapper/Parameter.md
```diff
@@ -3,7 +3,7 @@
...
 * **Lightweight Parameter**: Some middleware types and protocols can expose parameters as lightweight parameters that can be part of a URI path such as query parameterst found in the URI of RESTful and OpenAPI based web APIs.
```
lschlesinger, line 5: typo _parameters_ found...

Documentation/PropertyWrapper/Parameter.md
```diff
@@ -280,10 +280,10 @@ Exposing the `Component` on a web socket interface requires the client to send t
 }
 ```
 
-## Parameters Defining An Endpoint
+## Path Parameters
 
 Some middleware types and protocols can expose parameters as part of the endpoint defining charactersistics. E.g. RESTful and OpenAPI based APIs use the URI path to define endpoints using variables such as `/birds/BIRD_ID` to identify the `Bird` the request is target at using the `BIRD_ID`.
```
lschlesinger, line 285: typo _characteristics_

lschlesinger, line 285: _targeted_  or more easy: "identify the requested `Bird` by using `BIRD_ID` as identifier"

Documentation/PropertyWrapper/Parameter.md
```diff
@@ -184,10 +184,10 @@ Exposing the `Component` on a web socket interface requires the client to send t
...
-* `.automatic`: Apodini infers the type of parameter based on the type information of the wrapped propety. This is the default behaviour of an `@Parameter` property wrapper.
+Some middleware types and protocols can expose parameters as lightweight parameters that can be part of a URI path such as query parameters found in the URI of RESTful and OpenAPI based interfaces. Apodini automatically exposes primitive types as lightweight parameters that conform to [`LosslessStringConvertible`](https://developer.apple.com/documentation/swift/losslessstringconvertible). Complex types such as custom types conforming to `Codable` can not be exposed using lightweight parameters. A lightweight parameter is defined the same way a normal parameter can be defined. The developer using Apodini can force a specific behaviour by passing in an enum value into the property wrapper initializer, e.g. `@Parameter(.lightweight)`. Possible values are:
+* `.automatic`: Apodini infers the type of parameter based on the type information of the wrapped property. This is the default behaviour of an `@Parameter` property wrapper.
```
lschlesinger: I am not sure if this is the right place to introduce (or summarize) the possible options for `@Parameter` as we are in the section ##Lightweight Parameters. Maybe we should move this to the very beginning, as this is a more general paragraph and would be the practical implication of the fact, that "property wrapper can be used to express input in different ways". At least I found it somewhat irritating that the use of `.content` is explained here, as it is not a lightweight parameter.

lschlesinger:
I like, that `@PathParameter` is now replaced by @Parameter(.path) as this is more consistent to the other types of parameters. Maybe a more general remark, I was wondering if `.path` is the right wording - could it be an improvement to find an even more abstract name, e.g. `.identifier`. The property wrapped by `@Parameter(.path)` is used as identifier in the path (at least for RESTful and OpenAPI APIs) but could be used as identifier in a different context by other exporters. Or it might even be the case that a developer does not have the meta-model of a path in mind when defining an "identifying" property. (I know this would entail some annoying refactoring but maybe it's worth thinking about it :) )

Associated with this, is it required for the struct whose property is wrapped with the `@Paramter(.path)` to conform to `Identifiable`? This would go along with wrapping the property, which makes the struct identifiable (and can be used to build a path), with something called `@Parameter(.identifier)`, wouldn't it?

PSchmiedmayer:
@moritzschuell Thanks for your review! I think the first solution would be to follow your last approach to decide the filed number based on the order in the Component for now. The order the properties are delivered to the mirror are the same across different executions. The same applies to the approach currently used in develop that uses the Runtime framework.
 
I would agree that we should go for something similar to the solution described in your first approach. My only concern would be that there would be exporter specific elements in the main Apodini DSL. Ideally this configuration could be decided by the Exporter and if we add an additional Exporter the new Exporter could also add configuration options into the Parameter. I will think about a way to achieve this and might even be able to add an approach to this PR 👍

Documentation/PropertyWrapper/Parameter.md
PSchmiedmayer: I agree with you and will move this section to the beginning of the document.

PSchmiedmayer:
@lschlesinger Thanks for fixing the many typos and smaller mistakes I have made in the initial draft and thanks for your input!
I agree that `.identifier` would be more descriptive and then using `.path` and I will adapt the proposal and implementation to reflect these changes 👍

Documentation/PropertyWrapper/Parameter.md
```diff
@@ -295,7 +295,7 @@ struct Bird: Codable, Identifiable {
 }
 
 struct ExampleComponent: Component {
-    @PathParameter var birdID: Bird.ID
+    @Parameter(.path) var birdID: Bird.ID
 
     // ...
 }
```
bauer-andreas, line 285: I'm unsure what the REST/OpenAPI exporter would do with that path parameter in this example?
The `ExampleComponent` is defined under the path `/api/birds` and the `@Parameter(.path)` is not explicitly included in the PathComponents.
Should the REST/OpenAPI exporter assume a path parameter on the top level then? My assumption was that a `@Parameter` which should be a path component must be explicitly declared as in the example below, or have I mixed up something?

bauer-andreas: Just noticed the short disclaimer below the code 😅
Are we sure that we want to allow this implicit definition of path parameters? I would argue against it as one loses the overview, as the "structure" of the Webservice could now be hidden inside of Components.

bauer-andreas:
The code and documentation contained some last traces of the old naming `.path`, quickly removed those.

moritzschuell:
@PSchmiedmayer I thought a bit about your remark, how we could achieve exporter specific options for the `@Parameter` without having to add it to the main Apodini DSL. My idea would be something like the following:

We would introduce a new 
```swift
protocol ExporterOptions { }
```
that can be extended by enums specific to the different exporters.
For gRPC this could be e.g.
```swift
enum GRPCExporterOptions: ExporterOptions {
    case grpcFieldTag(tag: Int)
}
```
Further, we would allow the user to add those options to the `@Parameter`:
```swift
public struct Parameter<Element: Codable> {
    private var options: [ExporterOptions] = []
    // ...
    public init(_ parameterType: ParameterType = .automatic, _ options: ExporterOptions...) {
        // ...
        self.options = options
    }
    // ....
}
```
Finally, we would adjust the decode method of the SemanticModelBuilder, to allow the `@Parameter` to deliver the given options to the specific SemanticModelBuilder:
```swift
class SemanticModelBuilder {
    // ...
   func decode<T: Decodable>(_ type: T.Type, from request: Vapor.Request, with options: [ExporterOptions]) throws -> T? { /* ... */ }
}
```

The downside I see is that we cannot "force" the user to supply the exporter specific options. The exporter would still have to provide default behavior without any options. Thus, we would still have to implement the implicit approach (via the property order that the reflection sees) as a fallback, for the case that the user does not provide the exporter specific option.

This is just my first idea I would like to discuss 🤔 If this approach sounds feasible, I would be happy to add it to this PR. Although, maybe it's out of scope for this PR.

nerdsupremacist:
@moritzschuell I just noticed that you had a similar idea to mine.
I just built it over here: #29

tgymnich (dismissed):

Documentation/PropertyWrapper/Parameter.md
PSchmiedmayer: That's a good point. The proposal would suggest that the `@Parameter(.path)` parameters which were not explicitly passed into a `Component` are added at the end of the path for RESTful APIs. While this is not explicitly visible in the DSL, I would suggest it is fine that this is sometime defined in the `Component` itself.

PSchmiedmayer:
@moritzschuell @nerdsupremacist That's a great idea! I will take a look at @nerdsupremacist 's PR tomorrow.
I would suggest to merge the proposal into a single PR to have a single point of discussion if this is fine with you @nerdsupremacist ?

Documentation/PropertyWrapper/Parameter.md
```diff
@@ -280,110 +265,115 @@ Exposing the `Component` on a web socket interface requires the client to send t
...
-
-struct ExampleBirdComponent: Component {
-    @PathParameter var birdID: Bird.ID
+struct ExampleComponent: Component {
+    @Parameter(.http(.body)) 
+    var times: Int
+    
+    // Can define different handling for different exporters
+    @Parameter(.http(.path))
+    var id: UUID = true
```
Supereg, line 357: Is this still an outdated example, right? Cause `@Parameter(.http(.path))` was replaced by `@PathParameter`.
Additionally `@Parameters` should be placed on the Handler and `@PathParmeter` are the only ones placed at the Component level (?).

bauer-andreas: Is this still an outdated example, right? Cause @Parameter(.http(.path)) was replaced by @PathParameter.
Additionally @Parameters should be placed on the Handler and @PathParmeter are the only ones placed at the Component level (?).

Sources/Apodini/Request/PathParameter.swift
```diff
@@ -0,0 +1,31 @@
+import Foundation
+
+
+/// A `@PathComponent` can be used in `Component`s to indicate that a part of a path is a parameter and can be read out in a `Handler`
+@propertyWrapper
+public struct PathParameter<Element> {
```
bauer-andreas: the `@PathParameter` still needs to conform somehow to the `PathComponent` protocol.
Otherwise it can't be used as stated in the example:

```swift
// ...
struct TestWebService: WebService, Component {
    @PathParameter var birdID: Bird.ID

    var content: some Component {
        Group("api", "birds", /* >>> */birdID /* <<< */) {
            ExampleBirdComponent(birdID: $birdID)
            Group("nests") {
                ExampleNestComponent(birdID: $birdID)
            }
        }
    }
}
// ...
```

Sources/Apodini/Request/PathParameter.swift
```diff
@@ -0,0 +1,31 @@
...
+        fatalError("""
+            You can never access the wrapped value of a @PathParameter.
+            Please use a `@Parameter` wrapped property within a `Handler` to access the path property.
+        """)
+    }
+    
+    /// Accessing the projected value allows you to pass the `@PathParameter` to a `Handler` or `Component`
+    public var projectedValue: PathParameter {
+        self
+    }
```
bauer-andreas: Doesn't the projected value of a PathParameter need to return an instance of `@Parameter`?
Judging by the initializer inside the `@Parameter` property wrapper I think `self` should be replaced with `Parameter(self)`? Not sure what the intention here is.

```swift
// ...
struct ExampleNestComponent: Handler {
    @Parameter var birdID: Bird.ID
    @Parameter var nestName: String?

    // ...
}

struct TestWebService: WebService, Component {
    @PathParameter var birdID: Bird.ID

    var content: some Component {
        Group("api", "birds", birdID) {
            ExampleBirdComponent( /* >>> */ birdID: $birdID  /* <<< */) // PathParameter doesn't match type Parameter
            Group("nests") {
                ExampleNestComponent( /* >>> */ birdID: $birdID  /* <<< */)
            }
        }
    }
}
// ...
```

Sources/Apodini/Request/Parameter.swift
```diff
@@ -0,0 +1,138 @@
...
+        self.options = PropertyOptionSet(options)
+    }
+    
+    /// Creates a new `@Parameter` that indicates input of a `Component` based on an existing component.
+    /// - Parameter pathParameter: The `@PathParameter` that can be passed in from a a parent component.
+    /// - Precondition: A `@Parameter` with a specific `http` type `.body` or `.query` can not be passed to a seperate componet. Please remove the specific `.http` property option or specify the `.http` property option to `.path`.
+    public init(_ pathParameter: PathParameter<Element>) {
+        self.options = PropertyOptionSet([.http(.path)])
+        self.id = pathParameter.id
+    }
```
bauer-andreas: Can we make the initializer private/internal? I don't think this should ever be invoked by the user right?

bauer-andreas (changes requested):
Left some comments on the appropriate parts of the code which I think don't quite line up with what is described in the markdown file for the `Parameter` PropertyWrapper (or is even causing compile time errors when trying to use them as described in the documentation).

EDIT: Besides that, looks good to me 👍

Sources/Apodini/Request/PathParameter.swift
bauer-andreas: A quick follow up. Wouldn't it make sense that the generic `Element` must conform to `LosslessStringConvertible`.
Thus we may need to create our own version of `Identifiable` where the associated type `ID` also conforms for `LosslessStringConvertible` (instead of `Hashable`).

Documentation/PropertyWrapper/Parameter.md
PSchmiedmayer: Thanks for the feedback! `@Parameter(.http(.path))` could still be used in `Handlers` and just be added at the end of the path as `Handler`s can only be a leafs in the `Component` tree. `@PathParameters` can only be used in `Component`s and can then be passed down to `Handler`s.

To your additional comment: Yes, you are right. I have not yet differentiated between `Component`s and `Handler`s in the .md files. I will do this to clarify it.

Sources/Apodini/Request/PathParameter.swift
PSchmiedmayer: Yes, you are right, it needs to conform to `PathComponent`, I will update this.
Element in general must not need to conform to `LosslessStringConvertible` but we should add a precondition check that this is the case in case you use `.http(.path)` option.

Sources/Apodini/Request/PathParameter.swift
PSchmiedmayer: The current implementation has an initializer for `@Parameter` that gets a `@PathParameter` as an input. Therefore you can not pass `@Parameters` down a component tree but only `@Parameters` can be easily passed down the component tree. I will update the initializer to only get an `UUID` so we can keep the requirement that an `Element` in a `@Parameter` only has to conform to `Codable` and not also `LosslessStringConvertible`.

Sources/Apodini/Request/Parameter.swift
PSchmiedmayer: I will check if we can make this private or internal. This initializer is used when passing down a `@PathParameter` down the component tree using the projected value. I will get back to you on this 👍

PSchmiedmayer: I have updated the initializer in the latest commit, it is now internal.

PSchmiedmayer:
@Supereg I have incorporated your feedback and added a test case 👍

Documentation/PropertyWrapper/Parameter.md
bauer-andreas: So when defining like shown in the example the REST exporter for example should just automatically infer a path parameter without the need to explicitly listing it in the PathComponents, right?

Documentation/PropertyWrapper/Parameter.md
```diff
@@ -298,9 +298,9 @@ struct TestWebService: WebService, Component {
 
     var content: some Component {
         Group("api", "birds", birdID) {
```
bauer-andreas: One nit: To my understanding the `@Parameter` property wrapper now conforms to `PathComponent` but not `@PathParameter`.
So either the code is wrong or the example is still outdated and the projected value needs to be used here? What is the intention? Should the `PathParameter` be passed as a path component or an instance of `Parameter`.

EDIT: judging by the test file it seems like the markdown is outdated and should be changed to using the project value aka the `@Parameter` instance.

bauer-andreas: Other than that I would happily approve ☺️👍

bauer-andreas (approved):

Documentation/PropertyWrapper/Parameter.md
PSchmiedmayer: You are right, this should be `$birdID` as this is the projected value of a `@PathParameter` and returns a `@Parameter`. I have updated the .md file

PR #24, opened 2020-11-25
Allow user to enable HTTP/2 and TLS

moritzschuell:
This PR adds a Configuration for HTTP/2. 
The user can simply use 
```swift
HTTP2Configuration()
```
This will read file-paths from two command line options, `--cert` and `--key`, that allow the user to specify a certificate and a private key. 
Or, the user can also hard-code the required file-paths, if desired:
```swift
HTTP2Configuration()
    .certificate("some/path/cert.pem")
    .key("some/path/key.pem")
```
If the user provides both options, the server will enable HTTP/1, HTTP/2 and TLS. Otherwise, the server will operate as before using only HTTP/1 without TLS.

**Why?**
We need the possibility to use HTTP/2 and TLS to be able to communicate with the gRPC clients that are generated from `.proto` files using the `protoc` compiler. So, if the user wants the exported gRPC interface of Apodini to work with the generated client code, he has to provide both `--cert` and `--key` to the server. 

**Note**
I had to adjust how we detect the app environment from...
```swift
var env = try Environment.detect()
```
... to...
```swift
let environmentName = try Environment.detect().name
var env = Environment(name: environmentName, arguments: ["vapor"])
```
It seems like the auto-detected environment is also parsing the command line arguments, which causes interference with the usage of custom arguments.


awoc:
I think in the future you can also write a `Configuration` once PR #19 is merged which might integrate better into the DSL. But I guess this would be another PR :)

moritzschuell:
Closing this for now, until #19 is merged. I will rework this PR to use the new configuration functionality then.

moritzschuell:
Reopening, I rewrote the HTTP/2 and TLS configuration for the new config approach.

Package.swift
```diff
@@ -35,9 +35,6 @@ let package = Package(
                 .product(name: "APNS", package: "apns"),
                 .product(name: "AssociatedTypeRequirementsKit", package: "AssociatedTypeRequirementsKit"),
                 .product(name: "Runtime", package: "Runtime")
-            ],
-            exclude: [
-                "Components/ComponentBuilder.swift.gyb"
```
awoc: I wouldn't delete this part. This will lead to a warning in the build jobs and in Xcode.

Sources/Apodini/Configurations/HTTP2Configuration.swift
```diff
@@ -0,0 +1,92 @@
...
+    public init() {
+        do {
+            for arg in CommandLine.arguments {
+                if arg.starts(with: "--cert=") {
+                    certData = try Data(contentsOf: URL(fileURLWithPath: arg.substring(from: 7)))
+                } else if arg.starts(with: "--key=") {
+                    keyData = try Data(contentsOf: URL(fileURLWithPath: arg.substring(from: 6)))
+                }
+            }
+        } catch { }
```
awoc: In case of a failure, I think it would be nice to also indicate this to an user in form of a print statement or something else.

awoc (changes requested):
I noted down some minor remarks. Otherwise, it looks good to me 👍

awoc (approved):

hendesi (approved):
I have nothing to add. Looks good to me 👍

PR #25, opened 2020-11-25
Add caching to swift.yml workflow

bauer-andreas:
This little PR adds a caching mechanism to the "Build and Test" workflow (as discussed in the meeting; if this change is still wanted).
Currently only the dependencies are cached leading to action runtimes to be reduced from e.g. ~6,5 minutes to ~2,5 minutes (heavily depending on the workload of the GitHub runner).
Maybe we can investigate in the future why caching the output of `swift build` doesn't really work.

`Package.resolved` is checked in to be used for reliably calculating the cache keys (and fallback/restore keys).

Additionally I added the ` --build-tests` to the `swift build` command, so **all** building is now done in the "Build" step and the "Test" step is only about **executing** tests.

### DIFF
--- a/.github/workflows/swift.yml
+++ b/.github/workflows/swift.yml
@@ -18,12 +18,18 @@ jobs:
     - uses: maxim-lobanov/setup-xcode@v1.1
       with:
         xcode-version: latest
+    - uses: actions/cache@v2
+      with:
+        path: .build
+        key: ${{ runner.os }}-spm-${{ hashFiles('**/Package.resolved') }}
+        restore-keys: |
+          ${{ runner.os }}-spm-
     - name: Check Xcode version
       run: xcodebuild -version
     - name: Check Swift version
       run: swift --version
     - name: Build
-      run: swift build
+      run: swift build --build-tests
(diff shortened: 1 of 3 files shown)


bauer-andreas:
Seems like "swiftlint" workflow is set to be required for merging in the branch protection rules, but the changes in this PR don't trigger that workflow 🙃

PSchmiedmayer:
@Supereg Thanks for pointing out the issue regarding the SwiftLint action, I have fixed this on develop and will merge this into this branch

PSchmiedmayer (approved):

PSchmiedmayer:
@Supereg Now SwiftLint will be triggered even though no Swift files have been changed. Feel free to merge the PR so we can benefit from the build time improvements 👍

bauer-andreas:
> Feel free to merge the PR so we can benefit from the build time improvements 👍

äääh `You’re not authorized to merge this pull request`. I think you gonna have to do that for me 😅
Some sort of branch protection rule on `develop`: `The base branch restricts merging to authorized users.`

PR #26, opened 2020-11-25
vision - communicational patterns

theMomax:
This PR aims to combine all our proposals and discussions on communicational control flow/component types into a comprehensive reading. This reading will consist of the introduction (outlining the topic and the core concept of how Apodini approaches this problem without going into implementation-details), and two chapters defining our vision for a) what tooling Apodini will provide to the developer as well as how this tooling could behave for different exporters and b) how this API can be utilized to implement typical communicational patterns.

I changed the terminology a bit because IMO neither of "control flow" and "component types" really hit the spot or told the whole story. I found "component types" a bit misleading, because our concept does explicitly not differentiate between types of components but provides a unified interface.

This is a WIP, so feel free to comment on the PR if you do not agree with certain paragraphs, find inconsistencies or don't like some terminology. I will continue to work on the second and third chapter until next Tuesday and continuously update the PR along the way.


theMomax:
With this PR I also introduce a labeling-system for our Documentation folder, where every file/headline should be marked with one of the following labels:

proposal
A proposal is a suggestion on the future development of Apodini. PRs containing proposals only can be merged into develop without discussion.  It can be written in the style of any of the following labels.

PRs containing changes on content marked with one of the following labels should only be merged into develop once approved by the community.

concept
A concept is a document that contains little to no implementational-detail but focuses on core principles instead.

documentation
A documentation is a document that describes Apodini in its current state and goes into technical detail. 

vision
A vision is a document that describes how Apodini is supposed to look like in the future and goes into technical detail.

I think this could improve the user-experience when navigating inside the Documentation folder because it makes it easy to tell what impact a text has. The images are currently hosted on my own GitHub.io. In case you like the idea we should move the images to an Apodini-owned repository. Of course, I could also rename/remove/add/recolor labels if wanted.

Documentation/Communicational Patterns/2. Tooling/2.2. Parameter.md
```diff
@@ -0,0 +1,92 @@
...
+
+    var handler: Handler<String> {
+        if formal {
+            FormalGreeter()
+        } else {
+            InformalGreeter()
+        }
+    }
+}
+
```
awoc: Why would the case `formal = true` and `surname = "Skywalker"` result in an error? The Handler would execute `FormalGreeter` which requires `surname` which is specified.

Documentation/Communicational Patterns/2. Tooling/2.4. ObservedObject.md
```diff
@@ -0,0 +1,46 @@
+![document type: vision](https://themomax.github.io/resources/markdown-labels/document_type_vision.svg)
+
+# ObservedObject
```
awoc, line 3: What would be the lifetime of an `ObservableObject`? Will it be created based on the first request to the endpoint and kept in memory until the connection is closed or could it also be a global single instance that can be shared between other `Component`s?

Documentation/Communicational Patterns/2. Tooling/2.5. Environment.md
```diff
@@ -0,0 +1,20 @@
...
+
+If the `wrappedValue` conforms to `ObservableObject`, the `@Environment` behaves exactly the same as [`@ObservedObject`](./2.4.%20ObservedObject.md).
+
+In any other case `@Environment` has no influence on the `Handler`'s lifetime or control flow.
+
+## Connection
+
+Apodini provides a `Connection` through the environment. This object can be accessed using `@Environment(\.connection)`. The `Connection` provides useful protocol-agnostic information about the current state of the communication. The `Connection` could provide a `start: Date` so the service could e.g. close connections with an timeout error.
+
+The most important property of `Connection` is `state: ConnectionState`. This state can be `.open` or `.end`, where `.end` signalizes that the protocol expects the connection to be closed now. The `Handler` will be destructed after the next evaluation that results in a `.send(Request)`. Usage of this special environment object may cause exporters to utilize a client-side stream for implementation of the respective endpoint.
```
awoc: Can `.end` also be used to close the connection to the client from the server or does it only signal when the client closes its connection? Something like `.end(Response)` where this would send the last response from the server and also closes the connection to the client. I think a similar behavior can be useful for server side stream.

awoc (commented):
Thank you for the very detailed documentation! I noted down some open questions that I have.

Documentation/Communicational Patterns/2. Tooling/2.2. Parameter.md
```diff
@@ -0,0 +1,92 @@
...
+struct InformalGreeter: Handler {
+    @Parameter var firstName: String
+
+    func handle() -> String {
+        "Hi \(firstName)"
+    }
+}
+
+struct Greeter: Handler {
```
nerdsupremacist: Do we want handlers to be composable? This seems more like the scenario where Components handle composition and handlers handle computation

Documentation/Communicational Patterns/2. Tooling/2.2. Parameter.md
theMomax: You are completely right, it was supposed to be `formal = false`, thanks! I will fix it.

Documentation/Communicational Patterns/2. Tooling/2.5. Environment.md
theMomax: No, `.end` cannot be used to close the connection. I think that this is a relatively rare requirement. My idea was that usually a client would unsubscribe from a subscription gracefully, or would just close their connection to the server. Both events would be handled at an exporter-level. If a developer really wants to close a stream from the server, they have to cancel all `@Published`s the `Handler` is subscribed to.

Documentation/Communicational Patterns/2. Tooling/2.4. ObservedObject.md
theMomax: The lifetime of an observable object is not managed by Apodini. I see two options for how to achieve this:
* `@ObservedObject` requires an auto-closure for its initial value, which can then be called to get the actual value each time the `Handler` is prepared to be evaluated (i.e. right before/after RequestInjectables are injected). The auto-closure could either return the same object each and every time or initialize a new one for each call.
* The transition from `Component` to `Handler` is made in a way that allows us to safely initialize a new instance of the according `Handler` each time we need one.

Documentation/Communicational Patterns/2. Tooling/2.2. Parameter.md
theMomax: Well...the tooling we came up with in the process of our discussion is heavily based on a SwiftUI-like declarative style. In this style, functionality is offered to the developer in the form of property wrappers. The thing about property wrappers is that they follow the same rules as properties - which are pretty strict.

The key limitation about properties in this context is that there is only one layer of logic allowed. In Swift you have to have all properties initialized at the end of `init`. Before that you cannot read a property's value even if you initialized it in the line above. Thus you cannot base the initial value of one property on another one ("Cannot use instance member x within property initializer; property initializers run before 'self' is available"). The same applies to property wrappers.

That is the reason why composition is a **necessity** for this type of declarative style. You cannot express certain logic (while taking advantage of the functionality exposed via property wrappers) without composition. This, btw, also applies to SwiftUI.

Take the following example: An endpoint that takes an `userId` as an argument and provides a stream of messages that exclusively address this specific user. For the stream you need the `@ObservedObject`. However, in order to get the correct `ObservableObject`, you need the `userId` - which is provided by the `@Parameter`. The property wrapped by `@ObservedObject` depends on the `userId` property.

In order to provide the developer with the logical layers they need, handlers need to be composable!

Documentation/Communicational Patterns/1. Introduction.md
```diff
@@ -0,0 +1,67 @@
...
+ supported by: gRPC, WebSocket
+ 5. **Unsolicited Message**: An unsolicited message that is sent to an _unspecified addressee_ from the web service based on some event
+ schema: `() -?> service-message`
+ supported by: APNS
+
+ ## Apodini's Paradigm
+
+ In general Apodini aims to provide a declarative DSL that is **by default agnostic** of the exported communication protocols. However, the DSL also allows for **customization** of the service's behavior for single exporters in a **protocol-specific** way. This also applies to Apodini's API for handling all of the above communication-patterns.
+
+The problem is that certain classes of client-service communication are not supported by all communication protocols. The developer can create an endpoint that makes use of a more advanced pattern but exporters responsible for a protocol that does not fully support this pattern cannot export the exact functionality as described by the developer.
```
PSchmiedmayer, line 47: I would suggest to use the term `Component` or `Handler` instead of endpoints.

Documentation/Communicational Patterns/2. Tooling/2.1. Tooling.md
```diff
@@ -0,0 +1,127 @@
+![document type: vision](https://themomax.github.io/resources/markdown-labels/document_type_vision.svg)
+
+# Tooling
+
+## Component
+
+A service consists of multiple endpoints. Apodini provides a DSL that allows to define the service's structure using `Component`s. The `Component`-stack is determined on startup and may not change during runtime. `Component`s that provide a `handler` instead of the recursive `content` property define endpoints. Both, input- and output-type of an endpoint are fixed at startup.
```
PSchmiedmayer: Based on the last discussions I thought that `Component`s and `Handler`s are mutually exclusive. How could we enforce that `Component`s can only have a content or a `handle()` method?

Documentation/Communicational Patterns/2. Tooling/2.4. ObservedObject.md
```diff
@@ -0,0 +1,46 @@
...
+
+The `@ObservedObject` wraps an `ObservableObject`. An `ObservableObject` may have multiple `@Published` properties. Each `@Published` property is projected to a `Publisher`. A `Handler` that contains an `@ObservedObject` subscribes to these `Publisher`s and evaluates the `Handler` on each update.
+
+## Lifetime
+
+### Influence
+
+The presence of `@ObservedObject` properties on a `Handler` signalizes exporting the endpoint as a Server-Side Stream makes sense. If the exporter decides to do so the lifetime of the `Handler` has to be extended accordingly. The `Handler` should stay alive until all `Publisher`s projected from observed properties have finished.
```
PSchmiedmayer: I think that the idea of having to wait for all `@Published` properties `Publisher`s are finished is a approach that sounds a bit complicated to me. When we take a look at SwiftUI we don't have the notion of closing an `ObservedObject`'s publisher or the publishers of the properties marked with `@Published`.
I would suggest that the decision of when a `Handler` should close the service side or bidirectional stream should be decided within the handler and within the `handle` method. I would propose to add an additional `.end(Reponse)` `Action` that behaves the same as `.send(Reponse)` for all. communication patters except for service-side and bidirectional streams.

Documentation/Communicational Patterns/2. Tooling/2.4. ObservedObject.md
```diff
@@ -0,0 +1,46 @@
...
+
+An exception is made for `@Published(.weak)`. `Handler`s do not wait for weakly published properties to finish before being destructed.
```
PSchmiedmayer: Looking at possible forms of reuse I would suggest to keep the `Combine` API the same as it can be found on Apple Platforms or OpenCombine and only offer an `@Published` property wrapper and a `.end(Reponse)` `Action`.

Documentation/Communicational Patterns/2. Tooling/2.4. ObservedObject.md
PSchmiedmayer: I think this is a very important question. I would suggest to have the same approach as SwiftUI: `ObservedObjects` that are passed in from the outside could be shared across different components as sort of singletons. If an `ObservedObject` is created in the `Component` it is recreated every time a new connection is established but lives for the complete connection, e.g. in a client side stream the `ObservedObject` is staying around until the stream is closed by a response from the web service.

Documentation/Communicational Patterns/2. Tooling/2.5. Environment.md
PSchmiedmayer: I would argue that the connection is closed independent if the next response from the service is a `.send(Request)` or a `.nothing`. I would suggest that the `.end` `Connection` property indicates that the client has the intention of terminating the connection and this provides a single last chance to respond before the connection is eventually closed, otherwise we would have zombie connections around that simply refuse to send a response back to the client and if they eventually do the client is long disconnected.

PSchmiedmayer: I agree with @awoc that a `.end(Response)` would be reasonable approach to indicate that a connection is closed from the web service and have suggested a similar a idea above that would also fit into the idea that @awoc provided.

Documentation/Communicational Patterns/1. Introduction.md
```diff
@@ -0,0 +1,67 @@
+![document type: concept](https://themomax.github.io/resources/markdown-labels/document_type_concept.svg)
```
PSchmiedmayer: I would suggest to move the resources over to this repository 👍

PSchmiedmayer (changes requested):
@theMomax Thanks for the great summary and the collection of examples and information. This sums up the ideas we had in the last few weeks in a great way and documents them as part of the Apodini repo. I had a comments across the different documentation pages and had a few places where I would suggest changes. The biggest one would probably be the concept of adding a `.end(Response)` action and the clear differentiation between a `Handler` and `Component`.
Thanks for all the work and have a great evening 👋

PSchmiedmayer:
@theMomax Would it be possible that you can also add the label structure you described above also to the documentation folder? I like the classification and I think that a lot of the documents would benefit from it.

Documentation/Communicational Patterns/2. Tooling/2.4. ObservedObject.md
theMomax: I added a couple of sentences at the beginning of ObservedObject -> Lifetime to clarify this topic.

Documentation/Communicational Patterns/1. Introduction.md
theMomax: The reason I did not include these terms here is that this is a `concept` document, which should not include terminology that requires knowledge of Apodini's API. Furthermore, a `concept` should stay valid no matter how our final implementation looks like. If we really are using the `Handler` depends on if we need composable `Handler`s which is not decided yet.

Documentation/Communicational Patterns/2. Tooling/2.4. ObservedObject.md
theMomax: I see your point. I agree that having a `.end(Response)` is the superior approach. The initial counter-argument I saw with this approach was that you cannot have a silent endpoint, that never sends a service-message but only consumes requests (e.g. supported by WebSocket). Thus I created two new cases: `.end` and `.final(Response)`.

Documentation/Communicational Patterns/2. Tooling/2.5. Environment.md
theMomax: The initial counter-argument I saw with this approach was that you cannot have a silent endpoint, that never sends a service-message but only consumes requests (e.g. supported by WebSocket). Thus I created two new cases: .end and .final(Response).

Documentation/Communicational Patterns/2. Tooling/2.1. Tooling.md
theMomax: I don't quite get your point. This is exactly the same pattern as we have it right now on the `Component`'s `content: some Component` property and its `handle() -> Response` function.

Documentation/Communicational Patterns/2. Tooling/2.4. ObservedObject.md
theMomax: duplicate

Documentation/Communicational Patterns/2. Tooling/2.5. Environment.md
theMomax: This is indeed a good point I didn't think about in too much detail. I adapted the paragraph accordingly.

Documentation/Communicational Patterns/1. Introduction.md
theMomax: I would advise against hosting them right inside the Apodini/Apodini repository. Seems kind of dirty to me.

PSchmiedmayer: I think it would be fine so store them in the documentation folder but I would also be fie to have a separate repo in the Apodini organization.

Documentation/Communicational Patterns/2. Tooling/2.1. Tooling.md
```diff
@@ -0,0 +1,133 @@
...
+
+### Dynamic Properties
+
+`Handler`s can use different types of `DynamicProperty` variables. Apodini provides multiple implementations of this protocol that provide different functionality. The implementations are described in the following chapters. Properties on `Handler`s that are a `DynamicProperty` are managed by the Apodini framework. The framework makes sure they are exclusive to one client-service connection. The developer does not have to take care of synchronization. Other properties should be used for static, globally shared content only or references to services that manage access and synchronization themselves.
+
+### Layers of Logic
```
PSchmiedmayer, line 23: I liked the idea of a `Maybe` section for proposals and documentation that @Supereg and @nerdsupremacist had in that proposal: https://github.com/Apodini/Apodini/pull/41#discussion_r537880859 .
I would suggest to move this section into a `Maybe` section.

Documentation/Communicational Patterns/1. Introduction.md
PSchmiedmayer: I see your point, thanks for pointing this out 👍

Documentation/Communicational Patterns/2. Tooling/2.1. Tooling.md
PSchmiedmayer: As far as I understand your description and the discussions we had in the last weeks I assumed that `Component`s specify the static structure of the API using the `content: some Component` and do not include the functionality to handle requests. E.g. a `@PathParameter` is only allowed in a `Component` and not in a context where a `handle() -> Response` function exists.
In contrast to `Component`s, `Handler`s are responsible for dynamic behavior and will require some sort of `Layers of Logic` that would be described in a maybe section that might rely on composing `Handler`s or property wrappers. A `Handler` should not include a `content: some Component` but only a `handle() -> Response` method and can include `@Parameter` or `@Environment` property wrappers.
I think we should make this differentiation clear and be sure that this is also defined by the protocols found in `Apodini`.

Documentation/Communicational Patterns/2. Tooling/2.5. Environment.md
```diff
@@ -6,15 +6,15 @@ The `@Environment` property wrapper provides different functionality depending o
...
 
-The most important property of `Connection` is `state: ConnectionState`. This state can be `.open` or `.end`, where `.end` signalizes that the protocol expects the connection to be closed now. The `Handler` will be destructed after the next evaluation that results in a `.send(Request)`. Usage of this special environment object may cause exporters to utilize a client-side stream for implementation of the respective endpoint.
+The most important property of `Connection` is `state: ConnectionState`. This state can be `.open` or `.end`, where `.end` signalizes that the protocol expects the connection to be closed now. The `Handler` will be destructed after the next evaluation. A change to `Connection.state` is handled the same as a change to `@Parameter`, i.e. **one** client-message that includes updates to `@Parameter`s and `Connection.state` results in **one** evaluation; **one** client-message that does not update any `@Parameter`s but does update `Connection.state` also results in **one** evaluation of the `Handler`.
```
PSchmiedmayer, line 17: Sounds nice, this is very clear and well defined 👍

Documentation/Communicational Patterns/2. Tooling/2.1. Tooling.md
theMomax: Ah, I think I now see why you are confused: A `Component` cannot have a `func handle()` anymore. It can only have a `var handler: some Handler`. It is just one character difference in the text. Is that what caused the confusion on your side?

edit: I added the types to both properties to make this clearer.

Documentation/Communicational Patterns/2. Tooling/2.1. Tooling.md
theMomax: I don't think supporting multiple Layers of Logic is a `Maybe` as we cannot provide a framework that doesn't allow for expressing certain logic. The two options for implementing multiple layers of logic, i.e. **Composition of Handlers** and **Composition of Property Wrappers** are marked as `proposal`.

Documentation/Communicational Patterns/2. Tooling/2.1. Tooling.md
```diff
@@ -4,7 +4,7 @@
 
 ## Component
 
-A service consists of multiple endpoints. Apodini provides a DSL that allows to define the service's structure using `Component`s. The `Component`-stack is determined on startup and may not change during runtime. `Component`s that provide a `handler` instead of the recursive `content` property define endpoints. Both, input- and output-type of an endpoint are fixed at startup.
+A service consists of multiple endpoints. Apodini provides a DSL that allows to define the service's structure using `Component`s. The `Component`-stack is determined on startup and may not change during runtime. `Component`s that provide a `handler: some Handler` instead of the recursive `content: some Component` property define endpoints. Both, input- and output-type of an endpoint are fixed at startup.
```
PSchmiedmayer: I would suggest that the Component should not have a `handler` and a `content` property but I would suggest something similar to this:
```
public protocol Component {
    associatedtype Content: Component = Never
    
    @ComponentBuilder var content: Self.Content { get }
}

protocol Handler: Component where Content == Never {
    associatedtype Response: ResponseEncodable = Never
    
    func handle() -> Self.Response
    func handle() -> Action<Self.Response>
}
```
so there is only one content which is always a `Component` and `Handler`s are a specialization of `Component`s with the requirement that the content must be `Never` so not defined.

PSchmiedmayer (changes requested):

Documentation/Communicational Patterns/2. Tooling/2.1. Tooling.md
theMomax: Yes, I think that would be more elegant. I adapted the paragraph to match this solution.

Documentation/Communicational Patterns/1. Introduction.md
theMomax: They are now hosted on Apodini.github.io

Documentation/Communicational Patterns/2. Tooling/2.1. Tooling.md
theMomax: duplicate

PSchmiedmayer (dismissed):

PSchmiedmayer:
Thanks for all the work, this looks great!

PSchmiedmayer (approved):

PR #27, opened 2020-11-26
Documentation/push notifications

awoc:
This pull request adds documentation for push notifications and cron jobs. The documentation will be extended in the future for plattform specific elements for APNS and FCM.


Documentation/AdditionalFunctionality/PushNotifications.md
```diff
@@ -0,0 +1,133 @@
...
+    case apns
+    case fcm
+}
+```
+
+The web service needs a specific route and `Component` to handle push notification registration. The following example shows how to add a `Device` to the currently authenticated `User`.
+
+```swift
+struct RegisterComponent: Component {
+    @Request(\.user) var user: User  // The logged in user
```
theMomax, line 76: I think we agreed on keeping the number of property wrappers down if possible. Furthermore, `@Request` is protocol specific terminology. The way how the user is identified and authenticated could be based on requests and session-cookies or a WebSocket connection with respective authentication measures in place. I think using `@Environment(\.user)` would be better.

Documentation/AdditionalFunctionality/PushNotifications.md
```diff
@@ -0,0 +1,133 @@
...
+
+## Sending Push Notifications
+
+`Notification`s are structured as follows:
+
+- **alert**: Displayed message on the device which includes title, subtitle, and body.
+- **payload**: Service specific notification settings like sound, badges, images, etc.
+- **data**: Background data.
+
+`Notification`s don't have to include an alert and can just send a silent push notification with background data.
```
theMomax: The code below does not really match this sentence. A `Notification` **requires** all three types of data. Did you mean to make alert optional (`alert: Alert?`)?

Documentation/AdditionalFunctionality/PushNotifications.md
```diff
@@ -0,0 +1,133 @@
...
+
+Events can trigger the sending of push notifications. This builds on top of `Component`s and follows the same approach as **[Service-Side Stream Components](../ComponentTypes/ServiceSideStream.md)** with `ObservableObject`s. `Component`s subscribe to `ObservableObject`s that are exposed to the environment. The handle method of the `Component` will be executed every time `@Published` properties of the `ObservableObject` change. `ObservableObject`s also can be scheduled as **[CronJobs](./CronJob.md)**.
+
+The following example shows a _WeatherService_ as a `CronJob` that fetches the weather every day and publishes it to its subscribers using the `@Published` property wrapper. The _AlertComponent_ listens to changes of the _WeatherService_ and sends out push notification based on the current weather. Every `Device` with the **subscription** _weatherSubscription_ will receive this push notification.
+
+```swift
+struct WeatherService: ObservedObject {
+    @Published var weather = "sunny"
+
+    handle() {
```
theMomax: A `func` is missing here. Also, I don't like the name `handle` as this name has a special place in Apodini and may confuse the reader. Or do you propose that Apodini should call the `handle` method on initialization of any `ObservedObject`?

Documentation/AdditionalFunctionality/PushNotifications.md
```diff
@@ -0,0 +1,133 @@
...
+        // Updates weather
+    }
+}
+
+struct AlertComponent: Component {
+    @Environment(\.notificationCenter) var notificationCenter: NotificationCener
+
+    @Environment(\.) var weatherService: WeatherService
```
theMomax: I think you could just use `@Environment` instead of `@Environment(\.)`.

Documentation/AdditionalFunctionality/PushNotifications.md
```diff
@@ -0,0 +1,133 @@
...
+- Send to a specific device
+- Send to a specific token
+
+## Unsolicited Push Notifications
+
+Events can trigger the sending of push notifications. This builds on top of `Component`s and follows the same approach as **[Service-Side Stream Components](../ComponentTypes/ServiceSideStream.md)** with `ObservableObject`s. `Component`s subscribe to `ObservableObject`s that are exposed to the environment. The handle method of the `Component` will be executed every time `@Published` properties of the `ObservableObject` change. `ObservableObject`s also can be scheduled as **[CronJobs](./CronJob.md)**.
+
+The following example shows a _WeatherService_ as a `CronJob` that fetches the weather every day and publishes it to its subscribers using the `@Published` property wrapper. The _AlertComponent_ listens to changes of the _WeatherService_ and sends out push notification based on the current weather. Every `Device` with the **subscription** _weatherSubscription_ will receive this push notification.
+
+```swift
```
theMomax: Please extend this example. It is not quite clear to me, where this `Component` would be placed inside the `Server` declaration. I especially do not understand how "normal" exporters would interact with this `Component`. What part of Apodini would control the lifetime of this `Component` and observe the `weatherService`? What happens to the return-value of the `Component`'s `handle`?

Documentation/AdditionalFunctionality/PushNotifications.md
```diff
@@ -0,0 +1,133 @@
...
+## Device Registration
+
+Devices can be registered to the `NotificationCenter` by conforming to the `Device` protocol. This protocol consists of the following proerties:
+
+- **type**: This property specifies which push notification service to use.
+- **token**: The device id for one app used by the push notification service
+- **subscriptions**: Subscriptions are used to group `Device`s together. Example: All devices that want to receive news alerts for a specific topic.
+
+```swift
+public struct Device: Content {
```
theMomax: What is `Content`?

theMomax (changes requested):
I think I understand you concept of the `NotificationCenter`. I especially like your idea of the `configuration` property on the server. I think the design of the `Device` could be more elegant in Swift than just using `String`. My biggest concern however is the handling of unsolicited push notifications. It did not become clear to me, how the example component you specified would be handled by Apodini and where it would be used inside the `Server` declaration.

Documentation/AdditionalFunctionality/PushNotifications.md
theMomax: Sorry, I just noticed that `@Request` is already documented. I still do not agree with the terminology, but this discussion should therefore not be part of this PR.

Documentation/AdditionalFunctionality/PushNotifications.md
awoc: `Content` is Vapor specific for decoding and encoding. I removed it from the proposal.

Documentation/AdditionalFunctionality/PushNotifications.md
awoc: Since unsolicited `Component`s are not request based they would be ignored by the exporters in case of APNS. My idea was to either place it inside of the `content` stored property or the `configuration`. It might be confusing to put in `content` with other nested `Component`s like `Group`, which wouldn't really change the lifetime and execution of it since it isn't request based. However, it makes more sense to specify the whole behavior of the server in `content`. Another possibility would be to specify it in `configuration` to clearly separate it from other `Component`s and their control flow. My concept is to create a singleton from a `EventComponent` which is only executed once it is triggered by publishers it listens to.

Documentation/AdditionalFunctionality/PushNotifications.md
awoc: I also added a proposal for `CronJob`s. Therefore, the _WeatherService_ would be of type `CronJob` than `ObservedObject`. At startup this `CronJob` will be created and properties of it can be used in other `Component`s by injecting it with `@Environment`.

Documentation/AdditionalFunctionality/PushNotifications.md
awoc: You're absolutely right, at least one is required.

awoc:
@theMomax  Thank you for your review! I also included the proposal for `CronJob`s and hopefully resolved some of the confusion.

Documentation/AdditionalFunctionality/PushNotifications.md
```diff
@@ -0,0 +1,164 @@
...
+The `NotificationCenter` can also be used to register devices. This is useful to retrieve saved devices and send out push notifications to them. This feature requires a connected **[database](./Database.md)** to Apodini. Using the special modifier `addNotifications()` with the `DatabaseConfiguration` will create a new database table called `NotificationDevice` of type `Device`.
+
+The following example configures a MongoDB database and enables the `NotificationCenter` to save devices to the _apodini_db_ database.
+
+```swift
+var configuration: Configuration {
+    // ...
+    DatabaseConfiguration(.mongo)
+        .connectionString("mongodb://localhost:27017/apodini_db")
+        .addNotifications()
```
PSchmiedmayer, line 31: I would suggest that the modifier should have additional optional parameters to indicate `Model`s that conform to the `Device` protocol so the developer of Apodini could implement their own `Device`.

Documentation/AdditionalFunctionality/PushNotifications.md
```diff
@@ -0,0 +1,164 @@
...
+- **type**: This property specifies which push notification service to use.
+- **token**: The device id for one app used by the push notification service
+- **subscriptions**: Subscriptions are used to group `Device`s together. Example: All devices that want to receive news alerts for a specific topic.
+
+```swift
+public struct Device {
+    public var user: User
+    public var type: PushNotificationService
+    public var token: String
+    public var subscriptions: [String]?
```
PSchmiedmayer: I would maybe think if we can make the subscriptions array a bit more type safe then just a `String` identifying a topic. I don't think this has to be part of the first implementation but would be something that we might want to documental as part of the proposal.

Documentation/AdditionalFunctionality/PushNotifications.md
```diff
@@ -0,0 +1,164 @@
...
+```swift
+struct WeatherService: CronJob {
+    @Published var weather = "sunny"
+
+    func run() {
+        // Updates weather
+    }
+}
+
+struct AlertComponent: EventComponent {
```
PSchmiedmayer: While I see the need and the requirement to enable unsolicited request I think that the idea of marking them as `Component`s or `Handler`s might be a bit misleading. Would it be possible to include all that functionality directly into the `CronJob`? So a `CronJob` would have a `run` method and would be able to store state using `@State` or even observe other objects using `@ObservedObject` or access the environment using `@Environment`. By accessing the environment a `CronJob` could also send out `Notification`s using the `NotificationCener` and could even update other global `@ObservedObject` that might trigger different other components. 

Alternatively they might also be able to directly trigger service or bidirectional streams by behaving like `ObservedObjects`. These two modes under one hood could eliminate the need for `EventComponent `s and would make the act of adding a `Component` or `Handler` in a `content` property that is ignored by all exporters nonexistent.

This functionality could then be described in the CronJobs documentation.

PSchmiedmayer (changes requested):
@awoc Thanks for the detailed documentation and ideas! I can now see a clear vision regrind the implementation of push notifications and action based events. The biggest comment I had was regarding the concept of `EventComponent`s. I would be optimistic that we could pack all the functionality described there into `CronJobs` themselves and would have a similar result. I am interested in hearing your thoughts about this!

Documentation/AdditionalFunctionality/PushNotifications.md
awoc: I'm not really sure how this would look like. FCM comes with the ability to send push notifications to topics/subscriptions which is String based. So for all devices which use FCM I would just use the functionality of FCM and for APNS devices I would individually send messages to the ids/tokens of devices.

PSchmiedmayer: I see, so adding subscriptions to an APNS device would have no effect? Wouldn't it be possible to manually search for all APNS devices that include the subscription and manually send out batches of push notifications?
Regarding the String topics/subscriptions: I see the point that we will eventually map the topic to a string describing the subscription. Nevertheless I would try to avoid having string identifies across an Apodini service. E.g. if you would change the name of a subscription it would require you to manually find all occurrences of the String in the application. Maybe a key path based approach or a wrapper object that is globally accessible and is a single source of truth your be nice.

theMomax: > I see, so adding subscriptions to an APNS device would have no effect? Wouldn't it be possible to manually search for all APNS devices that include the subscription and manually send out batches of push notifications?

I think that's exactly what @awoc meant with his last sentence.

Regarding the string topic: I totally agree with @PSchmiedmayer that the Apodini API should enforce usage of an approach where the existence of topics is checked at compile-time by default. However, there might also be scenarios where the developer wants to create subscriptions at runtime (e.g. one subscription for each PR on a repository). I would suggest creating a `Subscription` protocol that requires a `identifier: String` computed property. The user could then extend `String` to implement it or use an enum e.g.:

```swift
enum Subscriptions {
    case all
    case PR(RepoId, PRId)
}

extension Subscriptions: Subscription {
    var identifier: String {
        switch self {
        case .all:
              return "all"
        case .PR(let repoId, let prId):
              return "pr:\(prId):\(repoId)"
        }
    }
}

```

Documentation/AdditionalFunctionality/PushNotifications.md
awoc: Yes, `CronJob`s can also use `@Environment` and I think we can extend this also to other property wrappers such as `@State`. The only concern I have is using `@ObservedObject` to listen to another object inside of a `CronJob`. This might be confusing because the `run` method would be executed everytime the `CronJob` is scheduled and is also triggered if one `@ObservedObject` changed. That is why I included the `EventComponent`. However, I like the idea that `CronJob`s can update properties of global `@ObservedObject`s.

I agree with you that other `Component`s should use `@ObservedObject` or `@Environment` to subscribe to changes of `CronJob`s, which can be used for server-side streams or bidirectional communication to a client.

PSchmiedmayer: @awoc I would argue that it would even be fine if an `@ObservedObject` can trigger an execution of the `run` function even though I only see very specific use cases for this at the moment. E.g. an Edge Chase of a `CronJob` that would fire once or never and would just be used to observe other `@ObservedObject` without the need to register and clutter the `Component`s DSL as this feature should not be triggered by Requests. Or a `CronJob` that would fire every 24 hours to e.g send a summary of the day or if the user manually requests such a report. 

Maybe It would even make sense to generalize the feature to `Job`s instead of `CronJob`s as cron would be very specific feature to configure them but I could also image some other ways to configure them except for cron Strings.

I think we would need a mechanism to check what has fired. This would not only apply to `Job`s but also to `Handler`s. A sort of way to access the projected value of a property wrapper to check if it has triggered the `handle` method in a `Handler` or the `run` method in a `Job`. E.g.:
```swift
struct ExampleJob: Job {
    @Environment(\.logger) var logger: Logger
    @ObservedObject var test: SomeObject = SomeObject()

    func run() {
        if $test.changed {
            logger.log("The `@ObservedObject` changed.")
        } else {
            logger.log("The Job was triggered.")
        }
    }
}
```
That would enable the developer to differentiate what has triggered the `Handler` or `Job` in the edge cases where this information is important.

awoc: I like the idea to check if an `@ObservedObject` has triggered. This would be especially useful when using multiple `@ObservedObject`s in a `Component` in order to differentiate them.

PSchmiedmayer: Maybe we can incorporate this idea in the PR so we can further refine and specify the behavior of `@ObservedObject`  and `@Environment` with this `.changed` behavior? I would see that `Job`s could benefit from this behavior and I agree that `Handler`s could also greatly benefit from this feature.

awoc: If `Job`s should also have the ability to change properties of `ObservedObjects`, I think it would be good to differentiate between the property wrappers `@Environment` and `@ObservedObject` to have a different meaning. I could envision that `Components` or `Handlers` can change the state of `ObservedObject`s with `@Environment` but don't listen to events that are emitted. On the other hand, the property wrapper `@ObservedObject` is used when `Component`s listen to events which will trigger the `handle` method everytime one property annotated with `@Published` changes. The same would apply to `Job`s. This would only work for globally defined `ObservedObject`s or when declared to child `Component`s with the modifier `.environment()`.

Let's consider the following example: The component `RegisterVisitor` counts every person that visits this specific path and saves this in `VisitorObject`. Everytime `VisitorObject` changes its property `count` the `Summary` job is executed and sends out a push notification. Furthermore, the `Summary` job also sends a summary of all visitors every Friday at 9 am. Global `ObservedObject`s are declared in the `configuration` property with the `Environment` configuration. This would define a key path which can either be used in every `Component` or `Job` using the `@ObservedObject(\.keypath)` or `@Environment(\.keypath)`.


```swift
struct VisitorObject: ObservedObject {
    @Published var count = 0
}

struct Summary: Job {
    @ObservedObject(\.visitorObject) var visitorObject: VisitorObject

    @Environment(\.notificationCenter) var notificationCenter: NotificationCenter

    func run() {
        if $visitorObject.changed {
            notificationCenter.send(alert: Alert(title: "We have a new customer"), to: "visitorTopic")
        } else {
            notificationCenter.send(alert: Alert(title: "This week we had a total of \(visitorObject.count) visitors", to: "visitorTopic"))
        }
    }
}

struct RegisterVisitor: Component {
    @Environment(\.visitorObject) var visitorObject: VisitorObject

    func handle() -> String {
        visitorObject.count += 1
        return "Welcome to our site!"
    }
}

var content: some Component {
    RegisterVisitor()
}

    Schedule(Summary.self, on: "0 9 * * 5")
    Environment(\.visitorObject, VisitorObject())
}
```

PSchmiedmayer: I like this differentiation and I think this sounds reasonable!
The option of have an `@ObservedObject` with a KeyPath to observe an object in the environment and without a key path to observe an object that is created locally or passed in using the initializer creates a nice API. And it makes it very clear if an object should be observed (`@ObservedObject`) and just used (`@Environment`).

The example sounds good and I would suggest to extend the example to include the base Apodini `WebService` and then add it as part of the PR in the documentation folder. 
There are a lot of different ideas in this review thread and it would make sense to summarize everything and as part of documentation files, update the PR and send the PR out to the whole team to give everyone the change to review the changes and get familiar with the proposed ideas 👍

Documentation/AdditionalFunctionality/PushNotifications.md
awoc: I've added some documentation for this. In the beginning the modifier will not work with other implementations that conform to `DeviceProtocol` but only with the default implementation of `Device`.

Documentation/AdditionalFunctionality/Jobs.md
```diff
@@ -0,0 +1,132 @@
...
+        }
+    }
+
+    var content: some Component {
+        RegisterComponent()
+    }
+
+    var configuration: Configuration {
+        Environment(\.visitorObservedObject, VisitorObservedObject())
+        Schedule(SummaryJob.self, on: "0 9 * * 5")
```
PSchmiedmayer: I would suggest to let the developer pass in a concrete instance of a `Job` instead of the type. E.g. the `Job` might have an initializer that we are not aware of and we therefore will not be able to create an instance. Or the `Job` has an initializer that takes a parameter to configure the `Job` and we want to schedule the job twice with different parameters.

Documentation/AdditionalFunctionality/Jobs.md
```diff
@@ -0,0 +1,132 @@
...
+# Jobs
+
+`Job`s allow to schedule repeating tasks and are in contrast to `Component`s not request based.
+Use cases for `Job`s could be running background database operations or fetching data from a remote server.
+In Apodini such tasks are defined by conforming to the `Job` protocol which includes a `run` method.
+This method is executed at scheduled points in time or when triggered by an [ObservableObject](<./../Communicational\ Patterns/2.\ Tooling/2.4.\ ObservedObject.md>).
+
+## Usage
+
+`Job`s can include the `@Environment` property wrappers to use other services in the `run` method.
```
PSchmiedmayer: I would maybe just state that `Job`s can use all property wrappers that `Handler`s can also use that are not directly connected to an incoming request. But e.g. `@State` and `@ObservedObject` also would be possible.

Documentation/AdditionalFunctionality/Jobs.md
```diff
@@ -0,0 +1,132 @@
+![document type: vision](https://apodini.github.io/resources/markdown-labels/document_type_vision.svg)
+
+# Jobs
+
+`Job`s allow to schedule repeating tasks and are in contrast to `Component`s not request based.
```
PSchmiedmayer:
```suggestion
`Job`s allow developers to schedule repeating tasks. In contrast to `Handler`s, `Job`s are not request based.
```

Documentation/AdditionalFunctionality/Jobs.md
```diff
@@ -0,0 +1,132 @@
...
+│  │  │  ┌───────────── month (1 - 12)
+│  │  │  │  ┌───────────── day of the week (0 - 6)
+│  │  │  │  │              (Sunday to Saturday)
+│  │  │  │  │
+│  │  │  │  │
+\* \* \* \* \*
+```
+
+`Job`s are scheduled at server startup from the `configuration` property using the `Schedule` configuration and the cron expression.
+When scheduling `Job`s, Apodini will create a single instance which can be used with `@ObservedObject` in other components.
```
PSchmiedmayer: Maybe even with an `@Environment` in addition to `@ObservedObject`. It might just be that a `Handler` just wants to great an `@Published` property of a `Job` but doesn't need to be triggered on an update of a `Job`.

Documentation/AdditionalFunctionality/PushNotifications.md
```diff
@@ -0,0 +1,127 @@
...
+- **topics**: Used to group devices together. Example: All devices that want to receive news alerts.
+
+A default implementation of this protocol called `Device` can also be used.
+
+```swift
+public struct Device: DeviceProtocol {
+    public var id: String
+    public var type: PushNotificationService
+    public var user: User?
+    public var topics: [Topic]?
```
PSchmiedmayer, line 63: I like the term topic here 👍

Documentation/AdditionalFunctionality/Jobs.md
```diff
@@ -0,0 +1,132 @@
...
+
+Consider the following example which consists of a `Component` called _RegisterComponent_, a `Job` _SummaryJob_, and an `ObservableObject` _VisitorObservedObject_.  
+The _RegisterComponent_ is responsible to count every access to its path and store this information in _VisitorObservedObject_. Using `Environment(keypath, value)` in the `configuration` stored property, which takes as arguments a keypath and the corresponding value, we can declare the _VisitorObservedObject_ as a Singleton. This allows us to use the `ObservableObject` in the web service with the property wrappers `@Environment` and `@ObservedObject` and the keypath. The `Job` will send a weekly push notification based on the total number of visitors and also when someone accessed the _RegisterComponent_ by listening to changes to the _VisitorObservedObject_. To differentiate between these two cases we use the projected value `$visitors.changed` to check if the `Job` was triggered by being scheduled or by the _VisitorObservedObject_.
+
+```swift
+struct TestWebService: WebService {
```
PSchmiedmayer, line 97: A nice example!

Documentation/Communicational Patterns/2. Tooling/2.4. ObservedObject.md
```diff
@@ -8,6 +8,12 @@ The `@ObservedObject` wraps an `ObservableObject`. An `ObservableObject` may hav
 
 The lifetime of the `ObservedObject` itself depends on the way it is created. The developer could pass it in from the outside (e.g. from a lower `Component`) as sort of a singleton which lives until the process dies. If the `ObservedObject` is created inside the `Handler` it is recreated every time a new connection is established but lives for the complete lifetime of the `Handler`.
 
+`ObservedObject`s can be globally declared by defining them in the `configuration` computed property of the web service using `Environment(keypath, value)`. This `Configuration` takes as arguments a keypath that is used to retrieve the `ObservableObject` from a property wrapper `@ObservedObject(\.keypath)` and the corresponding value that should be injected. Locally defined `ObservableObject`s don't require the use of a keypath.
```
PSchmiedmayer, line 11: I would suggest to add a small example here that showcases the configuration syntax.

Documentation/Communicational Patterns/2. Tooling/2.5. Environment.md
```diff
@@ -6,15 +6,17 @@ The `@Environment` property wrapper provides different functionality depending o
 
 ## ObservableObject
 
-If the `wrappedValue` conforms to `ObservableObject`, the `@Environment` behaves exactly the same as an [`@ObservedObject`](./2.4.%20ObservedObject.md) that was passed into the `Handler` from outside.
-
-In any other case `@Environment` has no influence on the `Handler`'s lifetime or control flow.
+If the `wrappedValue` conforms to `ObservableObject`, the `@Environment` can be used to change or read its properties. The `ObservableObject` will be injected to the `Handler` by using the corresponding keypath. This action may trigger other `Handler`s or `Job`s when changing properties annotated with `@Published`.
```
PSchmiedmayer: I would add an explicit statement that an `ObservedObject` injected using `@Environment` doesn't trigger the `handle` method.

PSchmiedmayer (changes requested):
@awoc I like the additions to the proposal, this forms a good picture and seems to be well rounded. I only had a few comments while looking through the documentation but the feedback shouldn't lead to fundamental changes. Apart from the few comments I would think that the PR is soon ready to be merged 🚀

PSchmiedmayer (approved):
@awoc Thanks for all the work! I approve the PR and the proposed changes 👍

theMomax (approved):
Looks good👍🏻. I'm looking forward to seeing this in action!

PR #28, opened 2020-11-26
Resolve Package.resolved

bauer-andreas:
The PR https://github.com/Apodini/Apodini/pull/22 added the new dependency `Runtime`. The PR was based on a version of Apodini where the build cache wasn't yet in place and `Package.resolved` wasn't yet checked in.
https://github.com/Apodini/Apodini/pull/25 was then merged not properly considering the changes made in #22 (my fault), thus Package.resolve was not properly update to include `Runtime` dependency.

**Impact:** Latest action runs would still use a build cache which didn't include an already fetched `Runtime` dependency, thus every run would need to go the extra step of resolving that dependency resulting in slightly longer run times.

This PR restores the consistency between Package.resolved and Package.swift.

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -65,6 +65,15 @@
         }
       },
       {
+        "package": "Runtime",
+        "repositoryURL": "https://github.com/wickwirew/Runtime.git",
+        "state": {
+          "branch": null,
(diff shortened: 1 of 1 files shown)


theMomax (approved):
LGTM👍

Issue #18, opened 2020-11-22
Deriving HTTP endpoints from Apodini DSL


lschlesinger:
# Preface 

This proposal is based on the ideas from [another proposal](https://gist.github.com/theMomax/85c5cf3fbc140ec17b89647b7558929c).
Their terminology is used in the following.
Additionally, we will call a _path parameter_ a path component inside the URL (e.g., `/product/:productId`) and a _query parameter_ a query string appended at the end of the URL (e.g., `?productId=1&...`).

The subject of this proposal is to manifest how HTTP endpoints are inferred from general Apodini concepts.
Ultimately, it would be desirable to create a shared `PathBuilder` for `RESTSemanticModelBuilder` and `OpenAPISemanticModelBuilder`.

# Default Inference (HTTP agnostic)

For the agnostic approach we assume `@Parameter` as the universal `RequestInjectable` for any use case.
The default inference has to cope with this limited information and still derive syntactically correct, i.e., unique, HTTP endpoints, e.g., for building a REST API.

In particular, this approach also assumes the unavailability of `Vapor.PathComponent.parameter`. 
The reason being that application developers may decide to only semantically group handling components without having a "route meta model" in mind (see example below).
In theory, it would be possible to make educated guesses about whether a `@Parameter` property should be mapped to a path parameter, but in practice, using query parameters is simpler (see example below).

## Example

```swift
struct Product: Content {
  var id: String?
  var name: String
  var image: Image?
}

struct Image: Content {
  var id: String?
  var name: String
  var url: String
}

struct CreateProduct: Component {
  
  @Parameter
  var product: Product
  
  func handler() -> Product {
    // some handler logic...
  }
}

struct GetProducts: Component {
  
  @Parameter("searchQuery")
  var searchQuery: String?
  
  func handler() -> [Product] {
    // some handler logic...
  }
}

struct GetSingleProduct: Component {
  
  @Parameter("productId")
  var productId: String
  
  func handle() -> Product {
    // some handler logic...
  }
}

struct CreateImage: Component {
  
  @Parameter("productId")
  var productId: String
  
  @Parameter
  var image: Image
  
  func handler() -> Image {
    // some handler logic...
  }
}

struct GetSingleImage: Component {
  
  @Parameter("productId")
  var productId: String
  
  @Parameter("imageId")
  var imageId: String
  
  func handler() -> Image {
    // some handler logic...
  }
}

struct TestServer: Apodini.Server {
  var content: some Component {
    Group("product") {
[...]

PR #29, opened 2020-11-27
Moving Parameter Options to be defined at Exporter Level

nerdsupremacist:
This PR changes the manifesto and Parameter property wrapper to use an option set defined at the exporter level.
Basically acting as a key store for each exporter.

This has the benefit that we don't have to use abstract concepts such as lightweight, and can be more explicit without coupling the Property Wrapper with HTTP specific knowledge.

This has the downside of making some options less easy to discover, but I think that autocomplete might be good enough for it

### DIFF
--- a/Documentation/PropertyWrapper/Parameter.md
+++ b/Documentation/PropertyWrapper/Parameter.md
@@ -1,10 +1,6 @@
 # `@Parameter`
 
-The `@Parameter` property wrapper can be used to express input in different ways: 
-* **Request Content**: Parameters can be part of the requests sent to the web service such as the HTTP body or request content of an other protocol.
-* **Lightweight Parameter**: Some middleware types and protocols can expose parameters as lightweight parameters that can be part of a URI path such as query parameters found in the URI of RESTful and OpenAPI based web APIs.
-* **Path Parameters**: Parameters can also be used to define the endpoint such as the URI path of the middleware types and protocols that support URI based multiplexing of requests.
-
+The `@Parameter` property wrapper can be used to express input in different ways. 
 The internal name of the `@Parameter` and can be different to the external representation of the component by passing a string into the `@Parameter` property wrapper as first - optional - argument.
 ```swift
 struct ExampleComponent: Component {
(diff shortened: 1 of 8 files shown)


Documentation/PropertyWrapper/Parameter.md
```diff
@@ -283,7 +281,7 @@ Exposing the `Component` on a web socket interface requires the client to send t
 ## Path Parameters
 
 Some middleware types and protocols can expose parameters as part of the endpoint defining characteristics. E.g. RESTful and OpenAPI based APIs use the URI path to define endpoints including variables such as `/birds/BIRD_ID` to identify the requested `Bird` by its identifier `BIRD_ID`.
-The `@Parameter` property wrapper exposes the option to explicitly defining a parameter as defining an endpoint using the `.identifier` option.
+We can infer that a specific parameter is part of this path for example: if it contains the word ID in the name and is at the start of the Component.
```
lschlesinger, line 284: Am I right by the assumption that this would imply the following inference rules (for HTTP):

1. If wrapped property's name contains substring `ID` AND is the first property of the Component => `HTTPParameterMode.path`
2. If not primitive type (or `LosslessStringConvertible`) => `HTTPParameterMode.body`
3. Else => `HTTPParameterMode.query`

Do you think that the potential problems (with 1.) I collected in #18 for this "pattern matching" approach are neglectable (e.g., misspelling, mispositioning)? Could we also consider to use the "conformance to `Identifiable`" (or some other protocol) information as a maybe less error-prone proxy?

nerdsupremacist: Thanks that's a good summarization of what I had in mind.
I think the idea of making it depend on identifiable might be a good idea. 

Mispositioning is I think definitely negligible
I think spelling errors should also be uncommon for something like `id`, but perhaps still somewhat error-prone. I think that we are far from done defining these inference rules, so please continue giving us great ideas on how to make it better and more reliable

PSchmiedmayer (approved):
Thanks for your input @nerdsupremacist , I will merge your input into the main `@Parameter` PR

PR #30, opened 2020-11-29
Protobuffer encoder+decoder

moritzschuell:
This PR adds an encoder and decoder for Protobuffer messages and Swift `Codable` structs. (They conform to `TopLevelEncoder` and `TopLevelDecoder`, respectively. However, this is currently removed due to Linux compatibility.)
I decided to put the functionality into a separate target of the package, because I did not find a sensible place where those files could be placed in the Apodini target itself.
The type mapping between Protobuffer and Swift types is described in the README, as well as the usage with `Codable` structs. Further, the (optional) usage of `ProtoCodingKey` is also described there.

Both, for `ProtoEncoder` and `ProtoDecoder` there are also some tests available.

**Some notes:**
- Single value encoding/decoding is not supported, because Protobuffer data usually only comes in messages. We have to see whether this will become necessary, depending on how the decoding of parameters will finally end up.
- Seems like `Combine` (and thus `TopLevelEncoder` and `TopLevelDecoder`) are not supported on non-apple systems. (non-issue, at least at the moment)


moritzschuell:
@nityanandaz Could you please give this PR a first look / review? 🙂
I think the type mappings in the README would be of particular interest to you, regarding the `.proto` exporting.

Sources/Apodini/Semantic Model Builder/REST/RESTSemanticModelBuilder.swift
```diff
@@ -6,6 +6,7 @@
 //
 
 import Vapor
+import ProtobufferCoding
```
nityanandaz: This import is not necessary and not a dependency I'd expect.

awoc:
I think you have to use [OpenCombine](https://github.com/OpenCombine/OpenCombine) because the official Combine framework from Apple doesn't work on Linux. This will hopefully fix the failing build jobs.

Tests/ProtobufferCodingTests/ProtobufDeccoderTests.swift
```diff
@@ -0,0 +1,166 @@
...
+                default:
+                    throw ProtoError.unknownCodingKey(key)
+                }
+            }
+        }
+    }
+
+    func testDecodeInt32() throws {
+        let positiveData = Data([8, 185, 96])
+        let positiveOriginal: Int32 = 12345
```
nityanandaz: How about `expected` or `expextedValue` instead of `original`?

Tests/ProtobufferCodingTests/ProtobufDeccoderTests.swift
```diff
@@ -0,0 +1,166 @@
...
+        let negativeData = Data([8, 199, 159, 255, 255, 255, 255, 255, 255, 255, 1])
+        let negativeOriginal: Int32 = -12345
+
+        let message1 = try ProtoDecoder().decode(Message<Int32>.self, from: positiveData)
```
nityanandaz: The additional message indicates to me that this should be split into two testMethods.

moritzschuell:
Thanks for the suggestion @awoc ! 👍
I am already looking into this. Maybe it wouldn't even hurt if we omit to implement `TopLevelDecoder` and `TopLevelEncoder` entirely. With the current way of decoding (which is via the specific SemanticModelBuilder) I do not see a strict need to comply to those protocols.

Sources/ProtobufferCoding/ProtoCodingKey.swift
```diff
@@ -0,0 +1,15 @@
...
+//  
+//
+//  Created by Moritz Schüll on 19.11.20.
+//
+
+import Foundation
+
+
+protocol ProtoCodingKey: CodingKey {
+    static func mapCodingKey(_ key: CodingKey) throws -> Int?
```
nityanandaz: I am unsure about the ergonomics and naming of this static method here.

Alternatives:
```
extension ProtoCodingKey {
    var protoRawValue: Result<Int?, Error> {
        return .success(0)
    }
    
    func protoRawValue() throws -> Int? {
        0
    }
}
```

Furthermore, do we really want to allow Optional\<Int\>? Most of the time, coding throws an error, if it is not present, i.e., in `KeyedProtoDecodingContainer.swift`.
I am just currently not aware of the differences between `return nil` or `throw Error`- 🙂

moritzschuell: These are two valid points 👍
Your suggested name `protoRawValue` is definitely more telling. I also agree that we should not use both optional and throws; will get rid of the optionals.

Sources/ProtobufferCoding/Decoder/KeyedProtoDecodingContainer.swift
```diff
@@ -0,0 +1,328 @@
...
+            return decodeRepeatedString(values)
+        }
+        throw ProtoError.decodingError("No data for given key")
+    }
+
+    // swiftlint:disable cyclomatic_complexity
+    func decode<T>(_ type: T.Type, forKey key: Key) throws -> T where T: Decodable {
+        // we need to switch here to also be able to decode structs with generic types
+        // if struct has generic type, this will always end up here
+        if T.self == Data.self, let value = try decode(Data.self, forKey: key) as? T {
```
nerdsupremacist, line 246: Thank you so much for the explanatory comment

nerdsupremacist: BTW this seems like a place for using gyb: https://nshipster.com/swift-gyb/

nerdsupremacist (approved):

Sources/ProtobufferCoding/Decoder/KeyedProtoDecodingContainer.swift
moritzschuell: Yes, indeed. Never used it before, so I'll take a closer look at it.
Thanks for the comment 🙌

PR #31, opened 2020-11-29
Shared Semantic Model

bauer-andreas:
This PR adds a first prototype for the shared semantic model (including a REST exporter working on that model). This is still a draft and not meant to be merged. It should be a place to follow progress and start discussions about the different needs of all the different exporters. ~The branch is based on #17 which is yet to be merged.~

In theory all interface exporters should already be able to work with this shared semantic model, as there are still references maintained to the `ContextNode` of the every registered `Component` (apart from the fact, that most of the Apodini internals still heavily depend on Vapor). Any dependence on the `Context` (and on Vapor) should be removed over time and all requirements should be built right into the model itself.

EDIT (2020-12-01): #17 is now merged.


bauer-andreas:
The shared model is now at a stage where we think it would make sense to get a first version merged.
A first version of the #23 proposal was implemented by @lschlesinger.
`Guards`, `Handlers` (yet to replace handling `Component`s) and `ResponseTransformers` are not yet modeled as #32  and #26 are set to be finished/implemented.
The shared model already generates some relationship information as described in #41 but additional ways mentioned in the proposal are set to be implemented with a separate PR.

Our goal would be to get this version of the shared model merged (hopefully sometime this week) and start adopting it in other exporters, OpenAPI probably being first, but in general anyone could adopt that. This will give us a better insight into the requirements needed by the exporters and allows us to incorporate those step by step.

So would be great if you could have a look at it @PSchmiedmayer @nerdsupremacist or whoever wants to throw in their opinion.

@nerdsupremacist Could you maybe specifically have a look at that following two code snippets. @lschlesinger and me struggled on how to properly check for the `Parameter` type (as the generic `Element` is unknown at that point). Maybe you have a solution on how to better tackle that problem.
https://github.com/Apodini/Apodini/blob/2c7649ed5cc1c59001c71f396c7b1f1862d180eb/Sources/Apodini/Semantic%20Model%20Builder/EndpointParameter.swift#L34-L46
https://github.com/Apodini/Apodini/blob/2c7649ed5cc1c59001c71f396c7b1f1862d180eb/Sources/Apodini/Semantic%20Model%20Builder/EndpointsTree.swift#L98-L107

Sources/Apodini/Components/PathComponent.swift
```diff
@@ -39,29 +41,32 @@ struct StringPathBuilder: PathBuilder {
...
+        self
+    }
+
     func append<P>(to pathBuilder: inout P) where P: PathBuilder {
         pathBuilder.append(self)
     }
 }
 
 /// Used to define parameter in a path
 public struct Identifier<T: Identifiable> {
```
PSchmiedmayer: I think that the concept of an `Identifier` would be no longer needed as we have the `@PathParameter` and therefore I would propose to remove it here.

Sources/Apodini/Request/RequestInjectable.swift
```diff
@@ -11,25 +11,39 @@ import Runtime
...
-    mutating func inject(using request: Vapor.Request, with decoder: SemanticModelBuilder?) throws
+    mutating func inject(using request: Vapor.Request, with decoder: RequestInjectableDecoder?) throws
+}
+
+protocol RequestInjectableDecoder {
+    func decode<T: Decodable>(_ type: T.Type, from request: Vapor.Request) throws -> T?
+}
+
+// swiftlint:disable:next todo
+// TODO Is there ANY better location to place this than on global?
```
PSchmiedmayer: "ANY better location" 😅 ... I think an alternative would be to make the method private and add extensions for different types that need to support this functionality, e.g. `Component.extractRequestInjectables() -> [String: RequestInjectable]`.
If this isn't possible a global scope is fine for me as long as it is not public.

Sources/Apodini/Semantic Model Builder/EndpointsTree.swift
```diff
@@ -0,0 +1,154 @@
...
+            endpoint.treeNode = self
+            endpoints[endpoint.operation] = endpoint
+        } else {
+            var pathComponents = paths
+            if let first = pathComponents.removeFirst() as? _PathComponent {
+                var child = nodeChildren[first.description]
+                if child == nil {
+                    // swiftlint:disable:next todo
+                    // TODO is there a better way to check for type Parameter instead of string comparison?
+                    if let info = try? typeInfo(of: type(of: first)), info.mangledName == "Parameter" {
```
PSchmiedmayer: Maybe you could use something like `let parameterType = try? typeInfo(of: Parameter<Int>.self)` to get the type information of a Parameter type. This would throw a compile error if the `Parameter` type would change.

Sources/Apodini/Semantic Model Builder/EndpointsTree.swift
```diff
@@ -0,0 +1,172 @@
...
+                        let mirror = Mirror(reflecting: first)
+
+                        // swiftlint:disable:next force_cast
+                        let options = mirror.children.first { $0.label == "options" }!.value as! PropertyOptionSet<ParameterOptionNameSpace>
```
PSchmiedmayer: I think we should add some enum with static properties that stores these constants such as `"options"` and `"Parameter"` where they are defined/in the `Parameter` type and use it here. Then we would at least have one file that stores all this context.

Sources/Apodini/Semantic Model Builder/EndpointsTree.swift
PSchmiedmayer: The same would apply to this line here: https://github.com/Apodini/Apodini/blob/2c7649ed5cc1c59001c71f396c7b1f1862d180eb/Sources/Apodini/Semantic%20Model%20Builder/EndpointParameter.swift#L38

Tests/ApodiniTests/SharedSemanticModelBuilderTests.swift
```diff
@@ -0,0 +1,98 @@
...
+            "Hello \(name)"
+        }
+    }
+    
+    struct TestComponent: Component {
+        @PathParameter
+        var name: String
+        
+        var content: some Component {
+            Group("a") {
```
PSchmiedmayer, line 62: I would also suggest to add a test case where there is a `Group` and some other component in the `Group("a")` just to make sure we also cover this edge case.

PSchmiedmayer (changes requested):
@Supereg Thanks for the PR and the new shared Semantic Model! I like the approach and only had a few small comments and suggestions in a few places. Apart from these small comments I think that the code is ready to be merged 🚀

Sources/Apodini/Request/RequestInjectable.swift
bauer-andreas: Solved with 32d549e

Sources/Apodini/Components/PathComponent.swift
bauer-andreas: Solved via 263ffed
The PathBuilder now also properly supports `@Parameter` as it conforms to `_PathComponent` as well.

Sources/Apodini/Semantic Model Builder/EndpointsTree.swift
bauer-andreas: Do I understand you correct, the idea is to add a line like you suggested which directly references `Parameter`, but **isn't used** actually, but serves as a hint when Parameter would change? Or what is the idea here?
Not sure if this solves our problem in the days of automatic refactoring? Most editors (like Xcode) probably support renaming all **used** references, only few IDEs search for string usage as well.

Tests/ApodiniTests/SharedSemanticModelBuilderTests.swift
bauer-andreas: I assume this is solved via a0f36a5 by @lschlesinger

Sources/Apodini/Semantic Model Builder/EndpointsTree.swift
lschlesinger: As far as I've understood, we should do this to avoid the String Comparison:
```
let parameterType = try? typeInfo(of: Parameter<String>.self)
if let info = try? typeInfo(of: type(of: first)), info.mangledName == parameterType?.mangledName { 
/// ...do something
}
```

where String can be replaced by Int, doesn't matter.
I've tested it just now, it works and tests are still succeeding.

Tests/ApodiniTests/SharedSemanticModelBuilderTests.swift
lschlesinger: yes, just added this corner case!

Sources/Apodini/Semantic Model Builder/EndpointsTree.swift
bauer-andreas: Could you maybe illustrate how you want this to look? Just an empty enum with `static let` definitions?

Sources/Apodini/Semantic Model Builder/EndpointsTree.swift
bauer-andreas: ah yeah, my head went to a complete different direction 😂 
@lschlesinger do you want to make those changes to resolve this ☺️?

lschlesinger: solved with 298e00d53fa54ce3e7286d83549ac436a4a2d87b

Sources/Apodini/Semantic Model Builder/EndpointsTree.swift
bauer-andreas: I hope 967e95c is what you had in mind?

PSchmiedmayer (approved):

PR #32, opened 2020-11-29
Decouple Vapor

tgymnich:
### Changes
- Replaced `import Vapor` with `@_implemntationOnly import Vapor` to hide Vapor from out public interface.
- Replaced `ResponseEncodable` with `Encodable`
- Added protocols for `Response` and `Request` and made `Vapor.Response` and `Vapor.Request` conform to those protocols.
- Decoding and encoding of parameters is left to the concrete implementation of `Request` and `Response`


Sources/Apodini/Components/Component.swift
```diff
@@ -16,7 +15,7 @@ public protocol Component {
     /// The type of `Component` this `Component` is made out of if the compoent is a composition of multiple subcomponents.
     associatedtype Content: Component = Never
     /// The type that is returned from the `handle` method when the component handles a request. The returntyp of the `handle` method is encoded into the response send out to the client.
-    associatedtype Response: ResponseEncodable = Never
+    associatedtype Response: Encodable = Never
```
nerdsupremacist, line 19: We might want to consider a different protocol that will give us the information we need for HATEOAS + Documentation for OpenAPI

Sources/Apodini/Request/Request.swift
```diff
@@ -5,27 +5,41 @@
 //  Created by Paul Schmiedmayer on 7/12/20.
 //
 
-import Vapor
+import Foundation
+import Fluent
+
+
+public class Request {
```
nerdsupremacist: Since Requests can look very different for different protocols. 
Perhaps we should go with a protocol:

```
protocol Request { 
    func parameter<T : InputType>(for uuid: UUID) throws -> T
    // ... other requirements like getting access to DB
}
```

@PSchmiedmayer what do you think if we add an httpHeader option to parameter?

Sources/Apodini/Components/Component.swift
tgymnich: lets refine this part in a future PR

moritzschuell:
@tgymnich As discussed, I added a few changes to the decoding / encoding of requests and responses.

Specifically, I added another parameter-specific function (`decodeParameter`) to the `DecoderProtocol`, because decoding the values for the different `@Parameter`s cannot really be done in the `Request` directly (as it was done previously). I think decoding of parameters can only happen in the semantic model builders, as only they will / should know the right decoding mechanism (JSON, Protobuffer, etc.). 

Further, my current approach forwards the `Parameter` instance itself to the `decodeParameter` call to the `SemanticModelBuilder`, because we need some context about which parameter we are actually decoding in the respective call. This allows us to access the (exporter-specific) options that were applied to the `@Parameter` in the `SemanticModelBuilder` (e.g. `name` for JSON, or field-tag for Protobuffers). 
Not sure whether that is the most elegant way though 🤔

Sources/Apodini/Request/Request.swift
```diff
@@ -5,28 +5,69 @@
...
+}
+
+typealias TotallyNotVaporRequest = Vapor.Request
+
+extension TotallyNotVaporRequest: Request {
+    public func parameter<T: Decodable>(for id: String) throws -> T {
+        try self.query.get(at: id)
+    }
+
+    public func bodyParameter<T: Decodable>() throws -> T? {
```
moritzschuell: Do we really need this function? It will only work for JSON encoded body-data, thus the user could not rely on it with every request.

tgymnich: For REST I think we are only dealing with JSON data anyway. If you want to use this for gRPC you might have to implement your own Request / Response types handling the decoding/encoding in a different way.

moritzschuell: Ah, I see. I think I misunderstood parts of your approach.
So by having a specific `Request` implementation for different bodies, having the decoding in the `SemanticModelBuilder`s is unnecessary 👍 Sorry, my bad.
But I am not sure how we would decode multiple `@Parameter`s for Request content (like described [here](https://github.com/Apodini/Apodini/blob/develop/Documentation/PropertyWrapper/Parameter.md#multiple-parameters-for-request-content)) with the current `bodyParameter`. I assume every `@Parameter` of type `.body` would try to call `request.bodyParameter()` in their `inject` method. But in such a case, the `T` type in `bodyParameter` would not match the entire body, but rather one (of the multiple) nested elements contained in the body.

tgymnich: I still need to find a way to map a Parameters UUID to the right Parameter in Vapor.Request.

Sources/Apodini/Request/Request.swift
tgymnich: It would be easier to do this:  `func parameter<T : Codable>(for parameter: Parameter<T>) throws -> T`.

tgymnich: the part with `LosslessStringConvertible` is still a bit ugly though.

Sources/Apodini/Request/Request.swift
```diff
@@ -5,28 +5,64 @@
...
+        case .query:
+            guard let name = parameter.name else {
+                throw Vapor.Abort(.badRequest)
+            }
+            return try query.get(T.self, at: name)
+        case .body:
+            return try content.decode(T.self)
+        default:
+            throw Vapor.Abort(.badRequest)
+        }
```
bauer-andreas: Not a big fan of the rather undescriptive error handling. Not saying this has to be resolved now, but once incorporated into the REST exporter, this should be properly described (especially the one which handles missing query parameters).
Not sure if the other two cases fall into the `badRequest` category?
Would suggest to annotate it with a `// TODO` or a `#warning` so it isn't forgotten.

Sources/Apodini/Components/Component.swift
nerdsupremacist: See #41

Sources/Apodini/Request/Request.swift
```diff
@@ -5,28 +5,64 @@
...
+            return parameter
+        case .query:
+            guard let name = parameter.name else {
+                throw Vapor.Abort(.badRequest)
+            }
+            return try query.get(T.self, at: name)
+        case .body:
+            return try content.decode(T.self)
+        default:
+            throw Vapor.Abort(.badRequest)
```
nerdsupremacist: The idea is that when building the semantic model we decide how to decode it (and infer the best strategy, for now we can default to something else), and store that information paired to the ID of the Parameter.

Sources/Apodini/Request/Request.swift
```diff
@@ -5,28 +5,64 @@
...
+}
+
+typealias VaporRequest = Vapor.Request
+
+extension VaporRequest: Request {
+    public func parameter<T: Codable>(for parameter: Parameter<T>) throws -> T {
+
+        switch parameter.option(for: PropertyOptionKey.http) {
+        case .path where T.self is LosslessStringConvertible:
+            guard let parameter = parameters.get(parameter.id.uuidString) as? T else {
```
nerdsupremacist: The uuid is not the actual name...

Sources/Apodini/Request/Response.swift
```diff
@@ -0,0 +1,33 @@
...
+//
+//  Created by Tim Gymnich on 24.11.20.
+//
+
+import Foundation
+import NIO
+@_implementationOnly import Vapor
+
+
+public protocol Response {
```
nerdsupremacist: Stupid question but why do we need a Response protocol?

nerdsupremacist (changes requested):

Sources/Apodini/Request/Request.swift
bauer-andreas: A first implementation of `@Parameter` and its inference strategies are already part of the shared semantic model #31. It may be worth coordinating how both of them get merged 🤔

Sources/Apodini/Request/Request.swift
tgymnich: 🙈 yeah this is still a bit unrefined. I'll create some custom error type for this.

Sources/Apodini/Request/Response.swift
tgymnich: we use it to create a generic request handler (`Context.swift:24`) that for example takes a `Vapor.Request` and returns a `Vapor.Response`.

The protocol mainly defines how we encode responses.

Sources/Apodini/Request/Request.swift
tgymnich: This is undefined right now since we are not yet handling PathParameters in the RESTSemantic model builder. right?

Sources/Apodini/Request/Request.swift
tgymnich: Added a dedicated error type to fix this.

tgymnich:
Looking at `Configuration` and how Apodini uses Vapor throughout its whole application lifecycle, I doubt we will ever get rid of Vapor. Now I am asking myself does it make sense to hide vapor as good as possible our should we just embrace our vapor dependency? The Request / Response abstractions introduced in this PR are only useful if some exporters actually don't use Vapor.

Sources/Apodini/Request/Request.swift
```diff
@@ -5,28 +5,75 @@
...
+            let body = try? self.body.data?.getJSONDecodable(T.self, at: 0, length: length)
+            let string = self.body.string as? T
+            guard let result = body ?? string else {
+                throw RequestDecodingError.couldNotDecodeParameter(for: parameter.id)
+            }
+            return result
+        default:
+            throw RequestDecodingError.couldNotDecodeParameter(for: parameter.id)
+        }
+    }
```
bauer-andreas, line 53: Just throwing in some thoughts, haven't had time to look at the rest (updated parts) of the PR yet.

IMO we still need to think about how we would incorporate that into the shared semantic model. Not having a `PropertyOptionKey.http` option defined is actually a common case. The inference approach what type of parameter to use is actually decided and stored in the `EndpointParameter`. Additionally the exporter decides on where to pull the parameter information from.

As of the changes made today, the shared model will create a request handler, which is responsible to handle guards, the `handle()` call and response modifiers, thus including the injection of RequestInjectables (and thus Parameters). How exactly those things are handled isn't set in stone.

So we need two abilities here:
* the `.parameter` method needs to be controlled by the shared model. It can match up the uuid with what is stored in its model and then decide on what type of parameter is used.
* The described procedure needs to be able to delegate retrieval of the actual parameter to the interface exporter. I could imagine adding something like `func retrieveParameter<T: Codable>(of endpointType: EndpointParameterType, as type: T, from request: ExporterRequest) -> T` to the `InterfaceExporter` where `ExporterRequest` is what the interface exporter uses as request type e.g. `Vapor.Request`.

Looking at those requirements I think a better approach would be to just simply use a wrapper approach (like really `ApodiniRequest<ExporterRequest>`) for our own Request type instead of making the request type of the exporters (like `Vapor.Request`) to conform to a protocol. The wrapper approach would allow for easily adding new exporters with different Request types in the future.
But then I don't know what discussions or arguments were brought up in the past that lead to the current approach.

Just wanted to throw that in as I can't see the current approach working with what we have/need and I don't want to waste your time.

Reference:
* [How the SharedSemanticModelBuilder creates the request handler](https://github.com/Apodini/Apodini/blob/2cc2d6bda74f990bcd87cf177530fa1e4f49f8e2/Sources/Apodini/Semantic%20Model%20Builder/SharedSemanticModelBuilder.swift#L109-L134)
* [Location where a InterfaceExporter would create its own request handler from that](https://github.com/Apodini/Apodini/blob/2cc2d6bda74f990bcd87cf177530fa1e4f49f8e2/Sources/Apodini/Semantic%20Model%20Builder/EndpointsTree.swift#L67-L69)


EDIT: connects to https://github.com/Apodini/Apodini/pull/61#discussion_r542035322

tgymnich: Thanks for the clarification. This is very helpful. I'll try to make the InterfaceExporter to handle parameters and decoding / encoding tomorrow.

tgymnich:
@nerdsupremacist @PSchmiedmayer @Supereg I've decided to take a step back and tackle this PR in 3 steps:

1.   Retain more type information for `Endpoint` #60 
2.  Replace `ResponseEnocdable` with `Response` #61 
3.  Generalize `Response` / `Request` and delegate encoding / decoding to `SemanticModel`

bauer-andreas:
> @nerdsupremacist @PSchmiedmayer @Supereg I've decided to take a step back and tackle this PR in 3 steps:
> 
> 1. Retain more type information for `Endpoint` #60
> 2. Replace `ResponseEnocdable` with `Response` #61
> 3. Generalize `Response` / `Request` and delegate encoding / decoding to `SemanticModel`

Any reason why you decided to do so and why this is reasonable? I thought we were on a good track with this issue?

nerdsupremacist:
In general I support splitting PRs to make them more digestible. I trust your judgement on this

tgymnich:
closing this in favor of #61

PR #33, opened 2020-11-30
Create ProtobufferBuilder

### LINKED ISSUE #48: Protobuffer generation: handle top-level scalar types
**Describe your use case**
The gRPC exporter assumes that the top-level type is a _Message_.
However, `Apodini.Component`s, such as the following, are just to convenient:

```
struct Greeter: Component {
    @_Request
    var req: Vapor.Request
    
    func handle() -> String {
        do {
            return try req.query.get(at: "name")
        } catch {
            return "World"
        }
    }
}
```

`Greeter.handle` returns a string, which is only a scalar type in the Protobuffer language specification.

**Describe the problem**
ProtobufferBuilder filters scalar types at the top-level.
Therefore, no appropriate message type is built for the `.proto` file.

**Propose a solution**
Detect scalar types in the type tree and wrap them in a generic `ScalarMessage`.

**Describe alternatives you've considered**
Expect the API user to only return _composite_ types.

**Additional context**
- https://developers.google.com/protocol-buffers/docs/proto3#simple
- https://developers.google.com/protocol-buffers/docs/proto3#scalar

### LINKED ISSUE #51: Protobuffer generation: inspect Component properties
**Describe your use case**
Such a `Component`:

```
struct Person: Component {
    let name: String
    let age: Int32

    public func handle() -> String {
        return "Hello \(name)."
    }
}
```

should produce a service like

```
service PersonService {
    rpc handle (Person) returns (/*...*/);
}

message Person {
    string name = 1;
    int32 age = 2;
}

...
```

**Describe the problem**
ProtobufferBuilder does not inspect properties of a `Component`, yet.

**Propose a solution**
Inspect the properties of a `Component` in `ProtobufferBuilder.addService`.

**Describe alternatives you've considered**
/

**Additional context**
- https://github.com/Apodini/Apodini/pull/33#discussion_r534775385

### LINKED ISSUE #63: Protobuffer generation: handle recursive type declarations
ProtobufferBuilder does not terminate (crashes at runtime 😅), if it encounters a recursive type, i.e.:

```
struct Node<T> {
    let value: T
    let children: [Node]
}
```

**Propose a solution**
Enforce a directed acyclic graph.
Save and check some ID of types that have already been visited.

**Additional context**
/

nityanandaz:
# Create ProtobufferBuilder

## :bulb: Proposed solution

ProtobufferBuilder is organized in a separate target.
Its functionality is developed independently of Apodini.

### Problem that is solved

`ProtobufferSemanticModelBuilder: SemanticModelBuilder, InterfaceExporter` exposes a `.proto` definition of the `Apodini.WebService` at `/apodini/proto/`

### Implications

Only a tiny API is exposed: `ProtobufferBuilder.addService`

### Testing

Are there tests included? Yes, of course.

Which situations are tested and which corner cases are missing?
Different Message kinds are tested and even `Apodini.WebService`s.

Enums are currently not supported.
Thus, there might arise new corner cases in the future.

### Reviewer Nudging

A reviewer might only look at the changes in `Sources/Apodini` first.


Tests/ProtobufferBuilderTests/ProtobufferBuilderTests.swift
```diff
@@ -0,0 +1,86 @@
...
+            case array([JSON])
+            case object([String: JSON])
+        }
+        
+        XCTAssertThrowsError(try code(JSON.self))
+    }
+}
+
+extension ProtobufferBuilderTests {
+    func testTypeTwoLevelsDeep() throws {
```
moritzschuell: Not quite sure I understand the output of this case 🤔
Why does this output the following?
```protobuf
message ArrayOfTransaction {
	_ArrayBufferOfTransaction _buffer = 0;
}
message Transaction {
	Int amount = 0;
}
message Account {
	repeated ArrayOfTransaction transactions = 0;
}
```
I would've expected something like this:
```protobuf
message Transaction {
	Int amount = 0;
}
message Account {
	repeated Transaction transactions = 0;
}
```

Tests/ProtobufferBuilderTests/ProtobufferBuilderTests.swift
```diff
@@ -0,0 +1,86 @@
...
+        }
+        
+        struct Transaction {
+            let amount: Int
+        }
+        
+        XCTAssertNoThrow(try code(Account.self))
+    }
+    
+    func testRecursiveType() throws {
```
moritzschuell: Similar to the comments above.
Wouldn't the intuitive output be
```protobuf
message NodeOfInt {
	Int value = 0;
	repeated NodeOfInt children = 1;
}
```

Sources/ProtobufferBuilder/Model.swift
```diff
@@ -0,0 +1,41 @@
...
+extension Message.Property: Comparable {
+    static func < (lhs: Message.Property, rhs: Message.Property) -> Bool {
+        lhs.uniqueNumber < rhs.uniqueNumber
+    }
+}
+
+// MARK: - Service and Service.Method
+
+struct Service: Equatable, Hashable {
+    struct Method: Equatable, Hashable {
```
moritzschuell, line 31: `input` and `output` could each also be a `stream` instead of a single message.
I guess this goes too far for this first PR, but we will also need to support client and server side stream components (as defined in [Apodini's ComponentTypes docu](https://github.com/Apodini/Apodini/tree/feature/parameter/Documentation/ComponentTypes)).

Sources/ProtobufferBuilder/ProtobufferBuilder.swift
```diff
@@ -0,0 +1,54 @@
...
+
+public class ProtobufferBuilder {
+    private var messages: Set<Message> = .init()
+    private var services: Set<Service> = .init()
+    
+    public init() {}
+}
+
+public extension ProtobufferBuilder {
+    func add<T>(_ messageType: T.Type = T.self) throws {
```
moritzschuell: Currently, this ignores any functions the given type might have.
We will also need to check whether the given type conforms to `Component`, and if so populate the `private var services` (that already exists in `ProtobufferBuilder`) with the component and its `handle` method. 

This closely relates to my comment above, as here we would also have to distinguish between the different types of messages (`stream`s or unary) that the method takes as input and output. 

So, throwing in this 
```swift
struct Person: Component {
    let name: String
    let age: Int32

    public func handle() -> String {
        return "Hello \(name)."
    }
}
```
should produce a service like
```protobuf
service PersonService {
    rpc handle (Person) returns (/*...*/);
}

message Person {
    string name = 1;
    int32 age = 2;
}
```

moritzschuell (changes requested):

Sources/ProtobufferBuilder/Model.swift
nityanandaz: I agree that this is out of scope. I created a new issue https://github.com/Apodini/Apodini/issues/44

Sources/ProtobufferBuilder/ProtobufferBuilder.swift
nityanandaz: Conformance to `Component` is inforced in `ProtobufferSemanticModelBuilder`.
`ProtobufferBuilder.services` is now populated.

I created a follow-up issue: https://github.com/Apodini/Apodini/issues/51

Sources/ProtobufferBuilder/ProtobufferBuilder.swift
```diff
@@ -0,0 +1,105 @@
...
+public extension ProtobufferBuilder {
+    /// `addService` builds a Protobuffer service declaration from the type parameter.
+    /// - Parameter type: the type of the service
+    /// - Throws: `Error`s of type `Exception`
+    func addService<T>(of type: T.Type = T.self) throws {
+        guard let serviceNode = try Tree<TypeInfo>.make(type) else {
+            return
+        }
+        
+        let serviceName = try serviceNode.value.compatibleName() + "Service"
```
moritzschuell: As far as I see, currently only one service per `Component` type can be added, right?
So, if we have two `Text`s in the `content` of the service, only one `TextService` will show up in the `.proto` IDL.
I think this is fine for now, but we should think about generating the service names from the `PathComponent`s (or the `.absolutePath` of an `EndpointsTreeNode` in the case of the shared semantic model).

Sources/ProtobufferBuilder/Extensions/Model+TypeInfo.swift
```diff
@@ -0,0 +1,48 @@
...
+                        let postfix = ParticularType(element.type).isPrimitive
+                            ? ""
+                            : "Message"
+                        let typeName = try typeInfo.compatibleName() + postfix
+                        
+                        return Property(
+                            isRepeated: ParticularType(typeInfo.type).isArray,
+                            name: element.name,
+                            typeName: typeName,
+                            uniqueNumber: offset
```
moritzschuell: The first field number / unique number will be 0.
Protobuffer field numbers have to be positive integers.

moritzschuell (commented):
Looks great so far! 🎉

I noticed that properties with the `@Parameter` property-wrapper applied cause the builder to not generate a service for the component. E.g.
```swift
struct Greeter: Component {
    @Parameter
    var name: String
        
    func handle() -> String {
        return "Hello \(name)"
    }
}
    
var content: some Component {
    Greeter()
}
```
produces only
```proto
syntax = "proto3";
```
While removing the `@Parameter` produces correct output.

Sources/ProtobufferBuilder/Extensions/Model+TypeInfo.swift
```diff
@@ -0,0 +1,48 @@
...
+                        )
+                    } catch {
+                        print(error)
+                        return nil
```
lschlesinger: not sure if "ignoring" an erroneous property is the way to go? Maybe throw an exception or check with precondition?

Sources/ProtobufferBuilder/Extensions/ProtobufferBuilder+CustomStringConvertible.swift
```diff
@@ -0,0 +1,30 @@
...
+        
+        let messages = self.messages
+            .sorted(by: \.name)
+            .map(\.description)
+            .joined(separator: "\n\n")
+        
+        let protoFile = [
+            #"syntax = "proto3";"#,
+            services,
+            messages
```
lschlesinger, line 23: I am not that familiar with .proto files but I would expect messages before services, as these types and used in the signatures.

Sources/ProtobufferBuilder/ProtobufferBuilder.swift
```diff
@@ -0,0 +1,105 @@
...
+            }
+        
+        guard let message = messageTree?.value else {
+            return
+        }
+        
+        let voidMessage = Message.void
+        let method = Service.Method(
+            name: "handle",
+            input: voidMessage,
```
lschlesinger: just for my understanding, this is where the list of parameters will be used in the future and voidMessage is just temporary?

Tests/ProtobufferBuilderTests/ProtobufferBuilderIntegrationTests.swift
```diff
@@ -0,0 +1,49 @@
...
+            url: "http://127.0.0.1:8080/apodini/proto",
+            expectedResponse: """
+                syntax = "proto3";
+                
+                service PokemonService {
+                  rpc handle (VoidMessage) returns (PokemonMessage);
+                }
+
+                message PokemonMessage {
+                  int64 id = 0;
```
lschlesinger: Not sure if "internal" properties (that are not annotated with `@Parameter` and therefore irrelevant for requests) should be part of the proto message types?

Sources/Apodini/Semantic Model Builder/gRPC/ProtobufferSemanticModelBuilder.swift
```diff
@@ -0,0 +1,26 @@
...
+        super.init(app)
+        
+        self.app.get("apodini", "proto") { req in
+            self.builder.description
+        }
+    }
+    
+    override func register<C>(component: C, withContext context: Context) where C : Component {
+        print(Result(catching: { try builder.addService(of: C.self) }))
+        print(Result(catching: { try builder.addMessage(of: C.Response.self) }))
```
lschlesinger: I guess you will replace the return type of `handle()` by the type ultimately returned - after being modified through transformers?

Sources/ProtobufferBuilder/Model.swift
```diff
@@ -0,0 +1,35 @@
...
+//  Created by Nityananda on 24.11.20.
+//
+
+import Runtime
+
+// MARK: - Message and Message.Property
+
+struct Message: Equatable, Hashable {
+    struct Property: Equatable, Hashable {
+        let isRepeated: Bool
```
lschlesinger: are you planning to add `optional` in the future?

lschlesinger (commented):
Looks generally good to me, tried to provide useful feedback :)

Sources/ProtobufferBuilder/Extensions/Model+TypeInfo.swift
nityanandaz: I haven't actually encountered erroneous properties, but I have encountered two types of Errors here:

- `couldNotGetTypeInfo(type: Builtin.Int8, kind: Runtime.Kind.opaque)`
- `couldNotGetTypeInfo(type: Builtin.BridgeObject, kind: Runtime.Kind.opaque)`

I use this mechanism to find an _organic_ end to my tree depth.
The above errors are thrown, if Runtime inspects internal implementations of `Swift.String`, for example.

Sadly, I cannot specialize the catch statement with pattern matching, because Runtime.RuntimeError is internal.

Sources/ProtobufferBuilder/Extensions/ProtobufferBuilder+CustomStringConvertible.swift
nityanandaz: The order doesn't matter. I put services first, because I found them more relevant.

Sources/ProtobufferBuilder/ProtobufferBuilder.swift
nityanandaz: Yes, now 0 or 1 Parameters are handled.

I created a follow-up issue: https://github.com/Apodini/Apodini/issues/62

Tests/ProtobufferBuilderTests/ProtobufferBuilderIntegrationTests.swift
nityanandaz: I filter Parameters now 👍

Sources/ProtobufferBuilder/Model.swift
nityanandaz: I do now 🙂

Sources/ProtobufferBuilder/ProtobufferBuilder.swift
nityanandaz: Services are now initialized with the absolute path

Sources/Apodini/Semantic Model Builder/gRPC/ProtobufferSemanticModelBuilder.swift
nityanandaz: ProtobufferSemanticModelBuilder is now an InterfaceExporter and uses `Endpoint.responseType`

Sources/ProtobufferBuilder/EnrichedInfo.swift
```diff
@@ -0,0 +1,56 @@
...
+        case zeroToOne
+        case exactlyOne
+        case zeroToMany
+    }
+    
+    let typeInfo: TypeInfo
+    let propertyInfo: PropertyInfo?
+    let propertiesOffset: Int?
+    
+    var cardinality: Cardinality = .exactlyOne
```
lschlesinger, line 21: what about optional array types, imo we should change `cardinality` to `[Cardinality]` to allow this.

lschlesinger (approved):
Looks good. As discussed, I will move your `Tree/Node`, `EnrichedInfo` and `ParticularType` as well as the `Handle...` code into the `SharedSemanticModel` to make use of it in #35

moritzschuell (approved):
Great work, looks good to me 🥳 Thank you.

PR #37, opened 2020-12-02
Add Support to Debug on Linux via VS Code + Docker

nerdsupremacist:
# Motivation

Sometimes Swift on Linux just behaves a bit differently than on macOS.
Currently we need to push and let the CI work when we want to see if the tests pass on Linux. 
And when it comes to debugging on Linux we have to either use a VM or use print debugging on the CI. Yuck!

This PR gives us the option to test the behavior on Linux from Mac (or even Windows)!
The requirements are:
- VS Code
- Docker

## How it works

This PR adds a Dockerfile along the extensions and settings and the targets for VS Code to be able to run the tests and debug from Linux. When you open the project you will be prompted to open it in the Development Container. When you open the project in the development container VS Code will:
- Build a docker image with Swift and LLDB installed on it. 
- Then install the VS Code server on it
- Install the extensions for VS Code to communicate with LLDB and to communicate with the sourcekit lsp server (for autocomplete)
- Create a volume to share the project files with the container

Screen Shot 2020-12-02 at 10 00 36 AM

You're now ready to set any breakpoints you want. Then go to the Run tab and click "Debug Tests". It will compile the project and the tests and run with a fully functioning debugger ;)

## Demo

Here's some screenshots:

Screen Shot 2020-12-02 at 9 57 34 AM
Screen Shot 2020-12-02 at 9 56 58 AM
Screen Shot 2020-12-02 at 9 57 11 AM


PSchmiedmayer (approved):
Thanks, this will make working with Linux way easier!

PR #39, opened 2020-12-02
Disable conditional_returns_on_newline rule

tgymnich:
I'd like to disable the conditional_returns_on_newline SwiftLint rule.
When I am dealing with a simple error, I want to get it out of the way as fast as possible. I don't want to spent another 2 lines of code returning nil.

```
guard let x = x else { return nil }
```


PSchmiedmayer:
While I could see how this rule might make the source code a few lines longer I actually value the increased readability of return statements on new linkes and therefore I would vote to keep the rule in place.

PSchmiedmayer (commented):
While I could see how this rule might make the source code a few lines longer I actually value the increased readability of return statements on new lines and therefore I would vote to keep the rule in place.

PR #40, opened 2020-12-02
@Environment property wrapper

awoc:
This PR adds @Environment based on the implementation from [OpenSwiftUI](https://github.com/Cosmo/OpenSwiftUI). Keypaths can be added as follows:

```swift
enum SomeClassEnvironmentKey: EnvironmentKey {
    static var defaultValue: SomeClass = SomeClass()
}

extension EnvironmentValues {
    public var someClass: SomeClass {
        get { self[SomeClassEnvironmentKey.self] }
        set { self[SomeClassEnvironmentKey.self] = newValue }
    }
}
```
and then used in a Component:

```swift 
@Environment(\.someClass)
```


Sources/Apodini/Environment/Environment.swift
```diff
@@ -0,0 +1,60 @@
...
+public protocol EnvironmentKey {
+    /// Represents the type of the environment key’s value.
+    associatedtype Value
+    /// The default value of the `EnvironmentKey`
+    static var defaultValue: Self.Value { get }
+}
+
+/// A collection of environment values.
+/// Custom environment values can be created by extending this struct with new properties.
+public struct EnvironmentValues: CustomStringConvertible {
```
nerdsupremacist: nit: I don't think we need `CustomStringConvertible` here

nerdsupremacist (dismissed):

Sources/Apodini/Environment/Environment.swift
```diff
@@ -0,0 +1,61 @@
...
+    public init(_ keyPath: KeyPath<EnvironmentValues, Value>) {
+        content = .keyPath(keyPath)
+    }
+    
+    public var wrappedValue: Value {
+        switch content {
+        case let .value(value):
+            return value
+        case let .keyPath(keyPath):
+            // not bound to a view, return the default value
```
theMomax: View...

theMomax (changes requested):
All of the code seems pretty reasonable to me. I would like to see some examples for how it would be used in the `Documentation` folder though. Also, tests would be great as kind of a proof that it really works.

Other than that I have one question: How would I inject elements other than the `defaultValue` into the `Environment`? In SwiftUI there is a `.environment(x)` modifier available on `View`s that allows to inject `x` for usage in any child-view.

PSchmiedmayer:
I agree that we would define the new behavior of the property wrapper as part of the documentation. In addition it would be great to have one or two tests to verify the behavior or the `@Environment` property wrapper.

@theMomax I think your comment is regarding what would be considered an `@EnvironmentObject` property wrapper in SwiftUI. I would suggest to try to merge the two property wrappers into one in a separate PR. You could inject a value using `.environment(...)` in a component and it is available in all `Handler`s that are registered as part of the `Component` or sub-`Component`s defined by the `content` property. I would suggest to add an option for an `@Envrinment` property wrapper that doesn't get a key path as an input and just tries to match the type based on the current environment context.

PSchmiedmayer (changes requested):
I agree that we would define the new behavior of the property wrapper as part of the documentation. In addition it would be great to have one or two tests to verify the behavior or the `@Environment` property wrapper.

@theMomax I think your comment is regarding what would be considered an `@EnvironmentObject` property wrapper in SwiftUI. I would suggest to try to merge the two property wrappers into one in a separate PR. You could inject a value using `.environment(...)` in a component and it is available in all `Handler`s that are registered as part of the `Component` or sub-`Component`s defined by the `content` property. I would suggest to add an option for an `@Envrinment` property wrapper that doesn't get a key path as an input and just tries to match the type based on the current environment context.

In addition we would need a PR that extend the `@Environment` property wrapper and the exporter in a way that they could listen for changes to wrapped values that conform to `ObservedObjects`.

Sources/Apodini/Environment/Environment.swift
```diff
@@ -0,0 +1,60 @@
...
+}
+
+protocol DynamicProperty { }
+
+/// A property wrapper to inject pre-defined values  to a `Component`.
+@propertyWrapper
+public struct Environment<Value>: DynamicProperty {
+    internal enum Content {
+        case keyPath(KeyPath<EnvironmentValues, Value>)
+        case value(Value)
```
PSchmiedmayer: As far as I can see we never use the `value` case so I would suggest removing it for now.

Sources/Apodini/Environment/Environment.swift
```diff
@@ -0,0 +1,60 @@
...
+    public init(_ keyPath: KeyPath<EnvironmentValues, Value>) {
+        content = .keyPath(keyPath)
+    }
+    
+    public var wrappedValue: Value {
+        switch content {
+        case let .value(value):
+            return value
+        case let .keyPath(keyPath):
+            return EnvironmentValues()[keyPath: keyPath]
```
PSchmiedmayer: Wouldn't it make sense to have some sort of global EnvironmentValues storage for an Apodini web service that can be injected here e.g. from a SemanticModel? It probably would need to be created when the service first get initialized so we can access it in the configuration part of the web service.

Sources/Apodini/Environment/Environment.swift
```diff
@@ -0,0 +1,60 @@
...
+            }
+            return K.defaultValue
+        }
+        set {
+            values[ObjectIdentifier(key)] = newValue
+        }
+    }
+    
+    /// Represents contents of the environment values instance.
+    public var description: String {
```
PSchmiedmayer: I would suggest to remove this property as it is not needed and does not provide any additional value

Sources/Apodini/Environment/Environment.swift
```diff
@@ -0,0 +1,60 @@
...
+    static var defaultValue: Self.Value { get }
+}
+
+/// A collection of environment values.
+/// Custom environment values can be created by extending this struct with new properties.
+public struct EnvironmentValues {
+    var values: [ObjectIdentifier: Any] = [:]
+    
+    /// Initializer of `EnvironmentValues`.
+    public init() { }
```
PSchmiedmayer: I would suggest to make this initializer internal as we should only create one instance of an `EnvironmentValues` per web service.

Tests/ApodiniTests/EnvironmentTests.swift
```diff
@@ -0,0 +1,54 @@
+//
```
PSchmiedmayer, line 1: Thanks for the unit test! There are a few errors in the test coverage report so I would suggest to add additional tests for some edge cases to improve the test coverage.

PSchmiedmayer (changes requested):
Thanks for the unit test. Looking over the functionality and testing it in Xcode with an other example I noticed some visibly improvements that could be made. In addition I would suggest to create a single instance of `EnvironmentValues` per web service and make it available to the Semantic Models so they can be injected in the `@Environment` property wrapper.

Sources/Apodini/Environment/Environment.swift
awoc: I think we can solve this easily with a Singleton and then we can access it like this `EnvironmentValues.shared`

PSchmiedmayer: Sounds good to me 👍

PSchmiedmayer (approved):
Thanks for the great work, looking forward to seeing how the environment is going to be used.

moritzschuell (commented):
Having one global `EnvironmentValues` makes sense to me 👍

However, I think it makes `Environment` unsuitable for injecting the connection state as currently described in the vision, since the connection state is specific to each request (of which we might have multiple, concurrently).
So we'll have to adjust the approach there.

PSchmiedmayer:
@moritzschuell That's a good point, I did not consider this. I would suggest to merge this as a first version of the @Environment property wrapper and focus on a way to inject request based information and context in your PR 👍

theMomax (approved):
Thanks for adding the tests. That made it a lot easier to understand! I agree with @PSchmiedmayer: let’s merge it now and re-iterate on it if needed.

PR #41, opened 2020-12-06
Proposal: Retrieving Relationship Information

bauer-andreas:
This PR adds a proposal/concept on how to retrieve relationship information to be used by the REST and GraphQL exporter.
It is based on the approaches we discussed in the work meeting on Tuesday and was further refined after the meeting with @nerdsupremacist on Friday.

Credits: The example implementation in 4.1. is based on the work from @nerdsupremacist which we developed in the meeting on Friday.

@nerdsupremacist @PSchmiedmayer @tgymnich @sadikekin


Documentation/Relationship/Relationship.md
```diff
@@ -0,0 +1,333 @@
...
+Any manual definition of a relationship MUST always override a inferred relationship if the name collides.
+
+Looking at the example below, a REST exporter would add the link `greeter` pointing to `/greeting/{userid}` on a 
+response returned from `/user/:userId`. As the value of the `userId` `PathParameter` is known it is replaced with
+the actual value. `@PathParameter`s for which no predefined value can be inferred should be formatted according to
+[RFC 6570](https://tools.ietf.org/html/rfc6570) (as well as potential query parameters of the destination). 
+
+```swift
+var content: some Component {
+  Group("user", $userId.relationship("greeter", to: "greeting", $userId)) {
```
nerdsupremacist: I'm not sure about this specific scenario. 
We don't have a way to enforce at compile time, that this path exists.

If I understand this correctly you want to create a link:
```
"_links": {
    "greeting" : "/greeting/12345"
}
```

And while I think this is a cool and valid use case, I'm not sure if we need to address this right away. 

PS: I think we could derive it by the use of the shared parameter `userId`. And not require any explicit path...

nerdsupremacist (commented):
# Love it!

I love it. I have only one comment on the explicit adding of new links with an unrelated path. I have no other points to raise.

@PSchmiedmayer would love to hear your feedback!

@sadikekin please take a look at this and see how this affects the GraphQL exporter. I think we would also benefit from documenting how these relationships are then mapped in GraphQL.

Documentation/Relationship/Relationship.md
bauer-andreas: > I think we could derive it by the use of the shared parameter userId

The use of the `userId` parameter in this example was really only to demonstrate that this mechanism could also perfectly handle path parameters. Yes, in this particular example one could probably somehow create the same result by looking at locations where the same parameter is used multiple times, but the general problem of defining manual additions to sort of "random" locations isn't solved with that.

I agree that the need of providing the full path isn't a very elegant solution, I couldn't really think about a better solution.

This mechanism is really only meant to solve the problem described in [1.](https://github.com/Apodini/Apodini/blob/feature/realtionship/Documentation/Relationship/Relationship.md#1-using-the-path-structure): 
> One of the core concepts of REST (with HATEOAS) is that a client does not need to know anything about the path structure. Instead, just like a web browser, a client follows hyperlinks returned from the web service.
This allows you to change the path structure of your REST service - keeping the same named links at the same places - without breaking compatibility for your clients.

If I may give you another (more abstract) example:

Transforming ...
```swift
var content: some Component {
  Group("a") {
    AHandler()
    Group("b") {
      BHandler()
    } 
  }
}
```
... to ...
```swift
var content: some Component {
  Group("a") {
    AHandler()
  }
  Group("b") {
    BHandler()
  } 
}
```

should be possible without breaking (REST) client compatibility when keeping a link from `a` -> `b`.
This mechanism would allow for that.

But I see your point. We could also just agree on not allowing manual linking at all and force users to do API versioning instead. Or maybe we envision a better idea in the future which solves all this.

bauer-andreas:
> @sadikekin please take a look at this and see how this affects the GraphQL exporter. I think we would also benefit from documenting how these relationships are then mapped in GraphQL.

yes any help with documenting that for GraphQL is appreciated. Obviously as I'm one of the REST guys this document heavily relies on REST concepts to illustrate the points. After our meeting @nerdsupremacist, all of this can be mapped to GraphQL, but I'm simply missing the GraphQL terms to explain it properly 😅

bauer-andreas:
I would be very interested on thoughts about the note in chapter [3.1.](https://github.com/Apodini/Apodini/blob/feature/realtionship/Documentation/Relationship/Relationship.md#31-looking-at-the-return-type-of-handle):

> _Note: As the type of a `@PathParameter` is a `LosslessStringConvertible` it most certainly is either a 
 `String` or an `Int`. Meaning checking if the type of the `User.id` property is equal to `User.ID` type might not be enough.
 Thus it could be necessary for `@PathParemeter`s to specify the type: e.g. `@PathParameter(type: User.self)` where
 the `type` argument must conform to the `Identifiable` protocol (or something similar as `Identifiable.ID` conforms 
 to `Hashable` and we actually need `LosslessStringConvertible`)_

The question is if we may need to force to specify a proper type for `@PathParameter` in order to be sure about linking them to the return types of Handlers.

Documentation/Relationship/Relationship.md
```diff
@@ -0,0 +1,333 @@
...
+
+Any manual definition of a relationship MUST always override a inferred relationship if the name collides.
+
+Looking at the example below, a REST exporter would add the link `greeter` pointing to `/greeting/{userid}` on a 
+response returned from `/user/:userId`. As the value of the `userId` `PathParameter` is known, it is replaced with
+the actual value. `@PathParameter`s for which no predefined value can be inferred should be formatted according to
+[RFC 6570](https://tools.ietf.org/html/rfc6570) (as well as potential query parameters of the destination). 
+
+```swift
+var content: some Component {
```
sadikekin, line 216: In section "2.2 Adding new links", two-way binding usage is pretty smart. I genuinely liked it. However, I am not sure how to adjust this to GraphQL.

Documentation/Relationship/Relationship.md
```diff
@@ -0,0 +1,333 @@
...
+
+It may also be a sensible decision for a user to hide a inferred relationship. To support that
+a modifier can be introduce.
+
+Even if a user supplies a `.relationshipName` modifier, the relationship will still be hidden when `.hideRelationship` is used.
+
+Considering the example below, a request to `/` will not include a relationship pointing to `/test`.
+
+```swift
+var content: some Component {
```
sadikekin: Why do we want to hide a relationship? Instead of hiding it, why don't we directly not create it?

If it is hidden, what d owe do when the user request `/test`, because the group has `Handler()`

Documentation/Relationship/Relationship.md
```diff
@@ -0,0 +1,333 @@
...
+  var id: String
+  var name: String
+  var lastname: String
+  var email: String
+}
+
+struct TestService: WebService, Component {
+  @PathParameter var userId: User.ID
+
+  var content: some Component {
```
sadikekin, line 163: This query translates into a GraphQL engine that is able to respond to the following,
```graphql
query {
    User(id: 123) {
         id
         name
         ....
    }
}
```
Is this correct @nerdsupremacist 
I have one question here, what is the meaning of $postId and where is it defined @Supereg?

Documentation/Relationship/Relationship.md
```diff
@@ -0,0 +1,333 @@
...
+      // could contain other routes giving more information for the given user, e.g. ./posts
+    }
+    Group("me") {
+      MeUserHandler() // handle() returns `User`
+    }
+  }
+}
+```
+
+In such a case, we can search the web service for a `Handler` which also returns `User`, but has a `@PathParameter`.
```
sadikekin: Why do we have a special `/me` path here? Can't we just use `/user/{me_user_id}`?
The mapping would be
```graphql
query {
  me {
    name 
    surname
  }
}
```
What do you think @nerdsupremacist

Documentation/Relationship/Relationship.md
```diff
@@ -0,0 +1,333 @@
...
+A `Article` could then be defined like the following:
+```swift
+struct Article: IdentifiableWithForeignKeys {
+  var id: String
+  var heading: String
+  var content: String
+
+  var writtenBy: String
+
+  static var foreignKeys: some ForeignKeys {
```
sadikekin: Shouldn't we use just an array of User parameters? Like
`var writers : [User]`
Correct me if I am wrong.
I think the GraphQL usage of it will be as in the following
```graphql
query {
  article(name: "Something") {
    writer(name: "One of the writer") {
      name
      surname
    }
    content
  }
}

File: Documentation/Relationship/Relationship.md [hunk 7]
```diff
@@ -0,0 +1,333 @@
...
+Giving an example: When a client sends a request for `/user/532` the a potential REST exporter will return the following:
+```json
+{
+  "id": "532",
+  "name": "Rudi",
+  "lastname": "Raser",
+  "email": "rudi@raser.de",
+  "_links": {
+    "self": "https://example.api/user/532",
+    "post": "https://example.api/post"
```
lschlesinger: Is this intentionally not `user/532/post`?

bauer-andreas: You are right. That should be `"post": "https://example.org/user/532/post"`. Thanks for catching that 👍

Documentation/Relationship/Relationship.md
bauer-andreas: The idea in this **example** is that every `Article` has one author (in the example a `User`). As `User`s are handled by the Handler registered under `/user/:userId` we want to basically "link" to that endpoint. With the above declaration we give Apodini the hint on what the `userId` is and where to find it.
In REST this would result in a `writtenBy` link in the `_links` section.
In GraphQL you could do a query like the following (if I'm not wrong)
```graphql
query {
  article(id: "1237) {
    writtenBy {
      name
      lastname
      email
    }
  }
}
```
----
(@nerdsupremacist)

This basically models 1:n relationships.
But as you suggest, we probably also need to think about n:m relationships, meaning e.g. Articles which are written by multiple authors.

I could see us adding a `ForeignKey` implementation which accepts a KeyPath to a variable holding an array of identifiers.

Assuming the Article example now has `var writtenBy: [String]` declared (with some proper way of defining that as ForeignKey).
I could envision that the REST endpoint would just insert an array of links to all known authors 
```json5
{
  // ...
  "_links": {
    "writtenBy": ["https://example.api/users/123", "https://example.api/users/124"],
    "self": "..."
}
```

And for the GraphQL exporter a query like the following would be possible:
```graphql 
query {
  article(id: "1237) {
    writtenBy {
      edges {
        nodes {
          name
          lastname
          email
        }
      }
    }
  }
}
```

Documentation/Relationship/Relationship.md
bauer-andreas: > I have one question here, what is the meaning of $postId and where is it defined

~Ah yeah, forgot to define the postId pathParameter explicitly. Exactly like the `userId` there would be a `@PathParameter var postId: Post.ID` inserted below `userId`.~

Nevermind, the example is just wrong. Should be `$userId` there.

Documentation/Relationship/Relationship.md
bauer-andreas: The specific example in this section would allow queries like the following 
```graphql
query {
  user(id: "1237) {
    greeter {
      // whatever fields the Greeter handler supports (not sure if you would need to write `edges.node` here 🤔 
    }
  }
}
```

Documentation/Relationship/Relationship.md
bauer-andreas: I think this option may be specifically tailored towards REST.

When having the web service defined like in that example, a request to `/` won't include a link to `/test` in the `_links` section, but a request pointing to `/test` will nonetheless be successful.
So graphql would also just server a query to `query { test { ... } }`.

The more I think about this, the harder it is to find a real use case for this 🤔

Documentation/Relationship/Relationship.md
bauer-andreas: Throwing in a argument against this part of my proposal: 

Regarding the point
> Any manual definition of a relationship MUST always override an inferred relationship if the name collides.

This works for REST, but for GraphQL this would essentially render the whole sub route, which is overwritten by the manual definition, inaccessible. 🤔 So at least that part must be changed, so that Apodini must not allow overriding existing relationship definitions (or name collisions with sub components).

Documentation/Relationship/Relationship.md
bauer-andreas: Thats just how the example is defined. `/me` could return the currently logged in user for example.

Documentation/Relationship/Relationship.md
nerdsupremacist: `edges.node` is about pagination.

If the greeter has any properties you would just list them there ;)

bauer-andreas: Thanks. Understood 👍🏽

Documentation/Relationship/Relationship.md
nerdsupremacist: Maybe we should add a `Maybe` section (like the generic manifesto from Swift: [here](https://github.com/apple/swift/blob/main/docs/GenericsManifesto.md#maybe)).

So that we still document the idea but don't commit to it necessarily ;)

Documentation/Relationship/Relationship.md
nerdsupremacist: Yes, that's correct ;)

Documentation/Relationship/Relationship.md
nerdsupremacist: Exactly. So the idea is that me has some logic that checks the currently authenticated user. And we know that we can get it via `/user/{me_user_id}` as well, we just don't know it's id just yet. But we can still link /me to all the sub routes of `/user/{me_user_id}/*`.

And I think that this Type based way of getting to the links lines up with GraphQL quite nicely

Documentation/Relationship/Relationship.md
nerdsupremacist: Something like `var writers : [User]` would be intuitive and easy, but it doesn't scale well.

For example think of a Post <-> Author relationship:

```swift
struct Post {
    let text: String
    let author: User
}

struct User {
    let name: String
    let posts: [Post]
}
```

results in an endless loop. Plus usually you don't load everything into memory at once, since loading the posts for a user or vice versa could be costly (e.g. loading from a DB). In these cases keeping track of the ids is simpler

Documentation/Relationship/Relationship.md
bauer-andreas: I think that is a good idea. Would suggest to move both `2.2` and `2.3` to the Maybe section.

Documentation/Relationship/Relationship.md
nerdsupremacist: In the case of N:M relationships I think that we can for now model it via additional nested groups

```swift
Group("article", $articleID) {
    ArticleHandler(id: $articleID)
    Group("authors") {
        AuthorsOfArticles(id: $articleID)
    }
}
```

which for REST would lead to

/article/:id -> { ..., "_links": { "self": "/article/:id", "authors": "/article/:id/authors" } } 
/article/:id/authors

And for GraphQL:

```
query {
    article(id: 123) {
        authors {
            name
            email
        }
    }
}
```

Documentation/Relationship/Relationship.md
```diff
@@ -0,0 +1,334 @@
+# Retrieving Relationship Information
+
+This proposal describes how relationship information can be generated from the DSL, introduces some concepts
+to give hints for relationship retrieval and adds mechanisms for manually defined hints.
+
+A potential REST interface exporter will use such relationship information to generate Hypermedia information.
+Specifically it will generate hyperlinks which SHOULD be place in a designated `_links` section.
```
tgymnich:
```suggestion
Specifically it will generate hyperlinks which SHOULD be placed in a designated `_links` section.
```

Documentation/Relationship/Relationship.md
```diff
@@ -0,0 +1,334 @@
...
+### 2.2. Adding new relationships
+
+In order to support rearranging components, a user can also define its own relationships.
+We introduce a `.relationship` modifier for `PathComponent`s, to which one would pass
+a relationship name and `PathComponent`s describing the destination.  
+The destination can of course also include `@PathParameters`.
+
+Any manual definition of a relationship MUST always override an inferred relationship if the name collides.
+
+Looking at the example below, a REST exporter would add the link `greeter` pointing to `/greeting/{userid}` on a
```
tgymnich: Are you intentionally using two notations for path parameters? `../{param}` and `../:param`

Documentation/Relationship/Relationship.md
```diff
@@ -0,0 +1,334 @@
...
+Checking the structure of the Webservice is already pretty powerful, but doesn't allow for relationships between handlers
+which are on the same level or on different paths of the path component tree.  
+This chapter highlights how we can nonetheless retrieve relationship information by looking at type information.
+
+### 3.1 Looking at the return type of `handle()`
+
+For this chapter we will look at the following example web service:
+
+```swift
+struct User: Identifiable {
```
tgymnich, line 148: This should also be `Codable`? right?

Documentation/Relationship/Relationship.md
```diff
@@ -0,0 +1,334 @@
...
+      MeUserHandler() // handle() returns `User`
+    }
+  }
+}
+```
+
+In such a case, we can search the web service for a `Handler` which also returns `User`, but has a `@PathParameter`.
+We could then find the `UserHandler` registered under `/user/:userId`.
+
+A response returned from the `MeUserHandler` would then include a `self` relationship pointing to `/user/{userId}`
```
tgymnich: Are you intentionally using two notations for path parameters? ../{param} and ../:param

Documentation/Relationship/Relationship.md
bauer-andreas: Out of the scope of this proposal. But yeah, whatever makes it able to be encoded and decoded.

Documentation/Relationship/Relationship.md
bauer-andreas: Yes, `./{param}` indicates that this parameter is actually replaced by the exporter as it is known at this point. Used the URI Template format here see [RFC 6570 3.2.2 Simple String Expansion: {var}](https://tools.ietf.org/html/rfc6570#section-3.2.2).

Documentation/Relationship/Relationship.md
bauer-andreas: See other comment.

Documentation/Relationship/Relationship.md
tgymnich: thanks for the clarification 👍

Documentation/Relationship/Relationship.md
bauer-andreas: Thanks, will correct that in the next commit I'll do

Documentation/Relationship/Relationship.md
```diff
@@ -0,0 +1,334 @@
...
+The above described inferring approach will use the string description of the `PathComponent` as the relationship name.
+This may not fit everybody's needs.
+
+Similar to how you can manually define a `@Parameter` name, you can also override a relationship name.
+This proposal suggest a sort of `PathComponentModifier` (similar to `ComponentModifiers`).
+
+Such a definition could then look like the following:
+```swift
+var content: some Component {
+  Group("test".relationshipName("new-name")) {
```
PSchmiedmayer: I have the feeling that the simple list of parameters in a `Group` will get long and complicated. Maybe it would be better to think about introducing a result builder here too so we can have a better grouping and intention approach?

Documentation/Relationship/Relationship.md
```diff
@@ -0,0 +1,334 @@
...
+Any manual definition of a relationship MUST always override an inferred relationship if the name collides.
+
+Looking at the example below, a REST exporter would add the link `greeter` pointing to `/greeting/{userid}` on a 
+response returned from `/user/:userId`. As the value of the `userId` `PathParameter` is known, it is replaced with
+the actual value. `@PathParameter`s for which no predefined value can be inferred should be formatted according to
+[RFC 6570](https://tools.ietf.org/html/rfc6570) (as well as potential query parameters of the destination). 
+
+```swift
+var content: some Component {
+  Group("user", $userId.relationship("greeter", to: "greeting", $userId)) {
```
PSchmiedmayer: I see the necessity to have such a feature but I think that the redundant information, e.g. `"greeting", $userId` two times, could be reduced. I would love to have a type safe approach. In addition the current approach is ok if the relationship and the `Component`s or `Handler`s are in the same `Component` but if they are in different components a change in one `PathComponent` might be missed in some other `Component` that the links are broken. 

I thought about how one could solve the problem. It is not ideal but maybe this would make the approach a bit more type safe and reduce the redundancy:
```swift
struct SomeComponent: Component {
  let greeterRelationship = Relationship("greeter")

 var content: some Component {
  Group("user", $userId.relationship(greeterRelationship)) {
    Handler()
   }
   Group("greeting", $userId) {
     Greeter()
       . relationship(greeterRelationship)
   }
 }
}
```
I could image that the type could be named `Link` or `Relationship`, e.g. `Greeter().link(greeterLink)`.

Such links/relationships could also be passed down the `Component` tree or we might have some sort of global key value storage e.g. using KeyPaths to access links in different other `Component`s.

Documentation/Relationship/Relationship.md
```diff
@@ -0,0 +1,334 @@
...
+It can be observed that the handle function of the `UserHandler` returns a `User` instance.
+User conforms to the protocol `Identifiable`, so it is also known that every `User` instance is uniquely identified by the
+`id` property.  
+Furthermore the `UserHandler` is registered with a path parameter having the same `User.ID` type, meaning
+we can assume that the `/user/:userId` route can be used to query a certain `User` instance
+(meaning `:userId` translates to `User.id` property).
+
+_Note: As the type of a `@PathParameter` is a `LosslessStringConvertible` it most certainly is either a 
+`String` or an `Int`. Meaning checking if the type of the `User.id` property is equal to `User.ID` type might not be enough.
+Thus it could be necessary for `@PathParemeter`s to specify the type: e.g. `@PathParameter(type: User.self)` where
```
PSchmiedmayer, line 197: Yes, this could be a needed extension. I am not sure if we would be able to use the Runtime framework to actually get to type defined by the developer of the user or it it is already mapped to `String` or `Int` when the program is executed. Otherwise explicitly adding this in the `@PathParameter` would be a possible option. We would also need to to the same of the `.http(.path)` option for `@Parameter`s.

Documentation/Relationship/Relationship.md
```diff
@@ -0,0 +1,334 @@
...
+
+In such a case, we can search the web service for a `Handler` which also returns `User`, but has a `@PathParameter`.
+We could then find the `UserHandler` registered under `/user/:userId`.
+
+A response returned from the `MeUserHandler` would then include a `self` relationship pointing to `/user/{userId}` 
+(`userId` is known from the returned `User` instance). More specifically the response will include all relationships
+which are also returned on a request to `user/:userId` (meaning links to potential subroutes).
+
+**Open questions:**
+* This approach heavily relies on the fact that `/me` returns the same type as `user/:userId`. It might be worth
```
PSchmiedmayer: Maybe we could introduce a concept of a `defaultReference`, e.g.: `defaultReference(to: User.self, id: $userId)` that can be attached to `UserHandler(userId: $userId)` 
```swift
 struct TestService: WebService, Component {
   @PathParameter var userId: User.ID
   @PathParameter var postId: User.ID

   var content: some Component {
     Group("user", $userId) {
       UserHandler(userId: $userId)
         .defaultReference(to: User.self, id: $userId)
       // could contain other routes giving more information for the given user, e.g. ./posts
     }
     Group("me") {
       MeUserHandler() // handle() returns `User`
     }
   }
 }
 ```
We would need to have some startup assertions that there is only one `defaultReference ` but this might solve a few of the problems that are posed by your question.

File: Documentation/Relationship/Relationship.md [hunk 16]
```diff
@@ -0,0 +1,334 @@
...
+```swift
+struct Article: IdentifiableWithForeignKeys {
+  var id: String
+  var heading: String
+  var content: String
+
+  var writtenBy: String
+
+  static var foreignKeys: some ForeignKeys {
+    ForeignKey<User>(name: "author", at: \.writtenBy)
```
PSchmiedmayer: I think this idea makes sense to indicate the links in the model. I would suggest to think about a Property wrapper approach. You could also have the same functionality if we have a `@ForeignKey` property wrapper around the existing property. This is the same approach that Fluent uses e.g. with `@Parent`: https://github.com/vapor/fluent-kit/blob/ef2a9a76c48856f0388ee2026d67d2ebde8c571c/Sources/FluentKit/Properties/Parent.swift and https://docs.vapor.codes/4.0/fluent/overview/ . I could image an approach that without a database imported we would have our own property wrapper that indicates the relationship and if @hendesi 's work is integrated and Fluent is imported the Fluent property wrappers will serve the exact same functionality and will be treated the same way as our own property wrappers. This would also make the move to a database a nice transition and the definition of `ForeignKey`s is not done twice in the model.

PSchmiedmayer: I just remember the you mentioned some `Codable` problems with the property wrappers, maybe we can use the fluent approach to solve these problems. They also support `Codable` with their property wrappers.

PSchmiedmayer (changes requested):
Hi @Supereg,
thanks for the great summary of the different discussions in a single compact proposal! I like the general direction and had a few suggestions and comments regarding some approaches to identify references and links. I am looking forward to your and all other's thoughts about my suggestions.

Documentation/Relationship/Relationship.md
bauer-andreas: If I'm not wrong you are proposing the same thing I proposed to @nerdsupremacist on our meeting (for reference my first draft of my proposal https://gist.github.com/Supereg/0c46967264611f0712e82c25624b86f3)

This basically suggested a syntax like 
```swift
struct Article: Identifiable {
    var id: String
    @ForeignKey(for: User.self) var writtenBy: String // String is the same type as User.ID here
    var heading: String
    var content: String
}
```

> I just remember the you mentioned some Codable problems with the property wrappers, maybe we can use the fluent approach to solve these problems. They also support Codable with their property wrappers.

That was one of the reasons why we move to the sort of "Mini-DSL", which could also leave some room for future advancements.

PSchmiedmayer: I really like the idea of having an `ForeignKey` or maybe a bit more descriptive as we are not in a database context without adding a database an `@Relationship` property wrapper that could be used like this: `@Relationship(to: User.self)`. I would argue that property wrappers would allow a easier transition to and combination with web services that use a database. If a developer would use an `@Parent` property wrapper from fluent this would already provide all the functionality needed and 

I see, that's why I remembered `Codable`, that's one topic that @nerdsupremacist and I talked about. I think that fluent solved this problem pretty elegantly using the Codable implementation they provide.

bauer-andreas:
Using the review from @PSchmiedmayer and the discussion from the last work meeting I revised the whole proposal to improve the general structure, removing some REST specific phrasing and incorporating GraphQL query examples.

As the list of changes would get pretty lengthy I would suggest to just look at the diff https://github.com/Apodini/Apodini/pull/41/commits/1f50af323a739d89088a8b19a26c25e75d7dc0a4.

---

**After** you have read the revised proposal the following questions are still to be discussed:

1. The mechanism described in [3.1](https://github.com/Apodini/Apodini/blob/feature/realtionship/Documentation/Relationship/Relationship.md#31-indexing-endpoints-by-their-handler-return-type) heavily relies on the fact that we can determine if the return type of a `Handler` conforms to the protocol `Identifiable`. I played with the Runtime framework a bit thinking that would be possible but it isn't. So I really don't have a clue if that is possible at all.
2. In regards to the property wrapper vs. mini DSL discussion on how to implement manual relationship definitions ([3.3.2](https://github.com/Apodini/Apodini/blob/feature/realtionship/Documentation/Relationship/Relationship.md#332-explicitly)):
As we need the type information at startup it probably won't be possible to use property wrappers. At least that's the current tendency after the research of @nerdsupremacist where he tried to use Runtimes `createInstance` to extract information stored in the property wrappers. Thus the proposal still uses the DSL approach.
3. The relationship inheritance approach described in [3.3.2.2](https://github.com/Apodini/Apodini/blob/feature/realtionship/Documentation/Relationship/Relationship.md#3322-relationship-definition-inherits) works perfectly for REST by just adding the `self` link to the `_links` section. I'm not sure how in graphQL one could access fields of `User`. Would it be feasible to inline those properties into the Me type (with somehow managing name collisions) @nerdsupremacist @sadikekin?


EDIT: some points got addressed in https://github.com/Apodini/Apodini/pull/41#issuecomment-747735489

Documentation/Relationship/Relationship.md
bauer-andreas: The revised proposal introduces a PathComponentBuilder [2.](https://github.com/Apodini/Apodini/blob/feature/realtionship/Documentation/Relationship/Relationship.md#2-manually-defining-relationship-information)

Documentation/Relationship/Relationship.md
bauer-andreas: The revised proposal removed the need of specifying that redundant information. Instead I introduced a approach similar to what you suggested [2.2](https://github.com/Apodini/Apodini/blob/feature/realtionship/Documentation/Relationship/Relationship.md#22-adding-new-relationships).

Would love to hear your feedback on this, as I think specifying `KeyPaths` with a global/centralized Relationship definition is pretty elegant, but not sure if I delivered the best possible solution.

Documentation/Relationship/Relationship.md
bauer-andreas: The problem described here was solved with the mechanism described in [3.3.2.2](https://github.com/Apodini/Apodini/blob/feature/realtionship/Documentation/Relationship/Relationship.md#3322-relationship-definition-inherits).

Additionally I took your suggestions of `defaultReference` to solve a different problem [3.2](https://github.com/Apodini/Apodini/blob/feature/realtionship/Documentation/Relationship/Relationship.md#32-using-defaultreference-to-resolve-ambiguous-return-type-information).

Documentation/Relationship/Relationship.md
bauer-andreas: See https://github.com/Apodini/Apodini/pull/41#discussion_r541122416 (moving discussion to that subthread)

> The revised proposal removed the need of specifying that redundant information. Instead I introduced a approach similar to what you suggested [2.2](https://github.com/Apodini/Apodini/blob/feature/realtionship/Documentation/Relationship/Relationship.md#22-adding-new-relationships).
> 
> Would love to hear your feedback on this, as I think specifying `KeyPaths` with a global/centralized Relationship definition is pretty elegant, but not sure if I delivered the best possible solution.

Documentation/Relationship/Relationship.md
```diff
@@ -0,0 +1,591 @@
...
+```
+
+Apodini will create a Relationship for each Endpoint linking to the **first** Endpoint lying deeper in the path, 
+for every child on the `PathComponent` tree.  
+Considering the example, the `UserHandler` Endpoint (located at `/user/:userId`) would have a relationship to 
+the `StaticInfoHandler` (located at `/user/:userid/static-info`) and the `PostHandler` (located at `/user/:userid/post/:postId`).  
+A relationship to the `DetailedStaticInfoHandler` won't be created, as `StaticInfoHandler` is the **first** Endpoint
+for the `static-info` sub tree.
+
+Not explicitly created, but specifically important for the REST Interface Exporter: A Endpoint will always maintain a
```
tgymnich: nit
```suggestion
Not explicitly created, but specifically important for the REST Interface Exporter: An Endpoint will always maintain a
```

Documentation/Relationship/Relationship.md
```diff
@@ -0,0 +1,591 @@
...
+}
+```
+
+The same `PathComponent` name cannot be used multiple times on the same depth in the `PathComponent` tree.
+Similarly the renamed Relationship MUST NOT collide with any other relationship name on the same depth.
+
+### 2.2. Adding new relationships
+
+To support rearranging components without breaking HATEOAS linking information or the GraphQL query schema,
+a user can also **add** its own relationship definitions.
```
tgymnich: maybe their or his/her would fit better than "its" ?

Documentation/Relationship/Relationship.md
```diff
@@ -0,0 +1,591 @@
...
+
+As we don't want users to manually list every `PathComponent` for the destination of the relationship, the user can create
+instances of `Relationship` wich are then referenced with `.relatable` (marking destination) and
```
tgymnich:
```suggestion
instances of `Relationship` which are then referenced with `.relatable` (marking destination) and
```

Documentation/Relationship/Relationship.md
```diff
@@ -0,0 +1,591 @@
...
+  }
+  Group("me") {
+    MeUserHandler()
+  }
+}
+```
+
+**REST:**  
+A request to the `me` endpoint would then generate a response like the following:
+```json5
```
tgymnich, line 458: did you mean json?

tgymnich (dismissed):
Great work! Can't wait to see this implemented. 
Do you have any idea what the return type of the `handle()` function of a `Component` should return, so you can get enough information out of it. Right now #32 aims to replace `ResponseEncodable` with `Encodable`. Reading this proposal I think this should be fine. I'd love to get your input on this just to make sure.

Documentation/Relationship/Relationship.md
bauer-andreas: solved via 1cb794f

Documentation/Relationship/Relationship.md
bauer-andreas: solved via 1cb794f

Documentation/Relationship/Relationship.md
bauer-andreas: solved via 1cb794f

bauer-andreas:
@PSchmiedmayer @nerdsupremacist would be great to get your feedback on the latest iteration of the proposal, so we can get that thing merged. Are there still any issues you see or changes/additions you think would be good to address?

---

To address some of the discussion points from my previous comment:

1. Mathias linked me to his [AssociatedTypeRequirementsKit](https://github.com/nerdsupremacist/AssociatedTypeRequirementsKit) which will seemingly solve my problem I raised in that point. I got a bit lost with the framework, so I don't have a POC yet, but I hope Mathias can guide be towards the right direction.
EDIT: solved it thanks to the help of Mathias 👍 
2. As a result of the discussion of the Intermediate Presentation, we agreed on introducing some sort of generic DSL on the "data" level, but also on a `Component`/`Handler` level (e.g. specify documentation to be used by the OpenAPI exporter). As this is out of the scope of this proposal I would suggest to leave that section as it is right now and adjusting that in the proposal tackling that more generic DSL approach (once we have a better overview of all the requirements from all the different exporters).  
I documented the decision on "Property Wrappers vs. DSL" in the [Appendix 4.2.](https://github.com/Apodini/Apodini/blob/feature/realtionship/Documentation/Relationship/Relationship.md#42-discussion-dsl-vs-property-wrapper)
3. Point three is still an open question. Not sure if is as easy as I think to model relationship inheritance in GraphQL. So some clarifying of you think this is doable would be great (maybe @nerdsupremacist?).

Documentation/Relationship/Relationship.md
```diff
@@ -0,0 +1,598 @@
...
+      StaticInfoHandler(userId: $userId)
+      Group("detailed") {
+        DetailedStaticInfoHandler(userId: $userId)
+      }
+    }
+  }
+}
+```
+
+Apodini will create a Relationship for each Endpoint linking to the **first** Endpoint lying deeper in the path,
```
PSchmiedmayer: I think this sentence could be misunderstood that Apodini only links to a single endpoint and not all endpoints one layer deeper a RESTful architecture. It might even be that there is a `Group` with no path component that must be ignored in this case. I would suggest to rephrase this in a way that is states that a RESTful exporter will link to all endpoints that are registered for a URI path with one additional path component.
Example for an empty `Groups` could be a `Group` that is only used to e.g.: indicate an authentication requirement.

Documentation/Relationship/Relationship.md
```diff
@@ -0,0 +1,598 @@
...
+    }
+  }
+}
+```
+
+Apodini will create a Relationship for each Endpoint linking to the **first** Endpoint lying deeper in the path, 
+for every child on the `PathComponent` tree.  
+Considering the example, the `UserHandler` Endpoint (located at `/user/:userId`) would have a relationship to 
+the `StaticInfoHandler` (located at `/user/:userid/static-info`) and the `PostHandler` (located at `/user/:userid/post/:postId`).  
+A relationship to the `DetailedStaticInfoHandler` won't be created, as `StaticInfoHandler` is the **first** Endpoint
```
PSchmiedmayer: The the comment above regarding "first".

Documentation/Relationship/Relationship.md
```diff
@@ -0,0 +1,598 @@
...
+
+This chapter imagines ways how a user could manually override or make modifications to relationship information.
+
+To provide such functionality the proposal introduces multiple `PathComponentModifiers`, similar to the already
+existing `ComponentModifiers`. As the `PathComponents` are passed as arguments to a `Group` instance,
+this could quickly get messy in terms of code readability.
+
+Thus this proposal additionally introduces a `PathComponentBuilder` with a corresponding initializer for the `Group` struct:
+
+```swift
```
PSchmiedmayer: I think that this example is a bit too abstract here. I would add a real world example or remove it at this point in the proposal.

Documentation/Relationship/Relationship.md
```diff
@@ -0,0 +1,598 @@
...
+The same `PathComponent` name cannot be used multiple times on the same depth in the `PathComponent` tree.
+Similarly the renamed Relationship MUST NOT collide with any other relationship name on the same depth.
+
+### 2.2. Adding new relationships
+
+To support rearranging components without breaking HATEOAS linking information or the GraphQL query schema,
+a user can also **add** their own relationship definitions.
+
+As we don't want users to manually list every `PathComponent` for the destination of the relationship, the user can create
+instances of `Relationship` which are then referenced with `.relatable` (marking destination) and
```
PSchmiedmayer: I would suggest to maybe make it a bit clearer that this is the target of the relationship, `relatable` is a bit hard to understand on a first read. How about `.target(of: ...)` or `.destination(of: ...)`?

Documentation/Relationship/Relationship.md
```diff
@@ -0,0 +1,598 @@
...
+```swift
+struct TestServiceRelationShips: RelationshipContainer {
+    var greeter = Relationship("greeter")
+}
+
+struct TestService: WebService {
+
+  @PathParameter var userId: User.ID
+
+  var relationshipContainer = TestServiceRelationShips()
```
PSchmiedmayer: I am not sure that the key path `\TestServiceRelationShips.greeter` would be able to identify the instance of `relationshipContainer` shown here. I am not really sure why we would need an other protocol for `RelationshipContrainer` as a normal struct containing and grouping values would also work out as far as I can see. What was the reason behind the `RelationshipContrainer` protocol and what would be its requirements?
I suspect I could pass `Relationships` across components using simple initializers or the `@Environment` property wrapper?

Documentation/Relationship/Relationship.md
```diff
@@ -0,0 +1,598 @@
...
+        .relatable(\TestServiceRelationShips.greeter)
+    }
+    Group("other") {
+      Other()
+        .relatable(otherRelationship)
+    }
+  }
+}
+```
+
```
PSchmiedmayer: I would suggest to add an example for a RESTful response to `/user/{userID}` to showcase how this example would manifest in the response.

Documentation/Relationship/Relationship.md
```diff
@@ -0,0 +1,598 @@
...
+    }
+    Group("user", $userId) {
+      UserHandler(userId: $userId)
+    }
+  }
+}
+```
+
+It will be observed that the `StatusInformationHandler` returns instances of type `StatusInformation`.
+So when another Endpoint references the type `StatusInformation`, we can easily get a reference to the Endpoint returning
```
PSchmiedmayer: At this point in the proposal would be a bit lost as I have no idea how an endpoint could reference the type. We might want to foreshadow a few things here or maybe change the order of explanations.

bauer-andreas:
> Great work! Can't wait to see this implemented.
> Do you have any idea what the return type of the `handle()` function of a `Component` should return, so you can get enough information out of it. Right now #32 aims to replace `ResponseEncodable` with `Encodable`. Reading this proposal I think this should be fine. I'd love to get your input on this just to make sure.

@tgymnich To my current understanding `Encodable` should be fine. I think we can get any type information we need from the return typ of `handle()` by doing extensions to `Component`/`Handler` or by inspection via the Runtime framework.

Documentation/Relationship/Relationship.md
```diff
@@ -0,0 +1,598 @@
...
+```swift
+struct StatusInformation {
+  var uptime: UInt64
+  // ...
+}
+// ...
+var content: some Component {
+  Group("status") {
+    StatusInformationHandler()
+      .defaultReference()
```
PSchmiedmayer: I would suggest to use a common naming across all modifiers. E.g. either relationship, reference or link. I am currently trending towards relationship but all other options would also be fine with me.

Documentation/Relationship/Relationship.md
```diff
@@ -0,0 +1,598 @@
...
+
+The GraphQL Interface Exporter MUST ignore the `hidden` flag and treat it like any other relationship.
+
+Considering the example below, a request to `/` will not include a relationship pointing to `/test`,
+but the endpoint `/test` will nonetheless be reachable when requesting it directly.
+
+```swift
+var content: some Component {
+  Group {
+    "test".hideLink()
```
PSchmiedmayer: I would suggest to use a common naming across all modifiers. E.g. either relationship, reference or link.

Documentation/Relationship/Relationship.md
```diff
@@ -0,0 +1,598 @@
...
+    post(postId: "123") {
+      title
+    }
+  }
+}
+```
+
+## 4. Appendix
+
+### 4.1. Example implementation of `var relationships`
```
PSchmiedmayer: While this is an interesting detail and we should reuse the code I would suggest to put it in a swift file and merge it as part of the proposal or a separated PR so it is right in the project.

PSchmiedmayer (changes requested):
@Supereg Thanks for the great proposal. I carefully read through the proposal once again and have the feeling that is showcases a well rounded idea of retrieving relationship information! Great job! I had a few comments across the proposal that I would suggest to address before merging it but I suspect this is the last round of feedback from my side.

Documentation/Relationship/Relationship.md
bauer-andreas: > I would suggest to rephrase this in a way that is states that a RESTful exporter will link to all endpoints that are registered for a URI path with one additional path component.

Exactly that is not the case. It not just selects the endpoint "[...] with **one additional** path component".
The idea was that to handle cases like this:
```swift
var content: some Component = {
  TestHandler()
  Group("path", "to") {
    Test2Handler()
  }
}
```

With your approach the `/` endpoint (the `TestHandler`) would link to `/path`. This is not really useful, as the user didn't specify a Handler there, so it would like to pretty much nothing.

My idea was that for every sub route (in this case the direct and only sub route is  `"path"`), we search the "sub path tree" for the "first" endpoint we encounter, which is `Test2Handler`.

But yes I agree that we probably need to improve phrasing.

Documentation/Relationship/Relationship.md
bauer-andreas: Not using relationship here was intended. I thought using `hideRelationship` is a bit misleading, as GraphQL will ignore the `hidden` flag and the REST exporter is really the only one using it. So the idea was to make that clear by using different naming here. But maybe there is a better word than `hide`? 
EDIT: maybe someone has an idea? Otherwise I would just go with `hideRelationship`.

Documentation/Relationship/Relationship.md
PSchmiedmayer: I see, so so summarize our tactic I could say "It is linking to the first `Component` layer down the component tree that is (1.) separated by at least one path component and (2.) that contains any `Handler`s in the layer"?
As you said:
> for the "first" endpoint we encounter, which is Test2Handler.
Just to be sure I understand "first" in this context, how would we approach the following example.
```
var content: some Component = {
  TestHandler()
  Group("path", "to") {
    Test2Handler() // Uses HTTP GET
    Test3Handler() // Uses HTTP POST
  }
}
```
Would we link to both endpoints at /path/to or just to the first one?

If I understood this correctly we wouldn't link to `Test3Handler` in this example?
```
var content: some Component = {
  TestHandler()
  Group("path", "to") {
    Test2Handler() // Uses HTTP GET
    Test3Handler() // Uses HTTP POST
  }
}
```

Documentation/Relationship/Relationship.md
PSchmiedmayer: I see. Reading through your explanation I see why `link` would be desirable in this context and I would even prefer `hideLink` over `hideRelationship`. We should somehow document the reasoning behind either choice we make as part of the proposal.

Documentation/Relationship/Relationship.md
bauer-andreas: Your example tackles a pretty different problem the proposal doesn't yet clearly specify how to handle that.
Maybe the issues is also in the current wording of the proposal. A relationship (currently) doesn't care about the `Operation`. And I think that's the right way to do it. After all the hyperlink will just be `/path/to` (we might then want to return in e.g. a `OPTIONS` or `INFO` route what different endpoints are available at that route). This approach also fits well with GraphQL I think. The relationship will exist anyways, and depending on the `Operation` a query or mutation will be possible or not.

I think a more interesting example is the following

```swift
var content: some Component = {
  TestHandler()
  Group("path") {
    Group("to0") {
      Test2Handler()
    }
    Group("to1") {
      Test3Handler()
    }
    Group("to2") {
      Group("to2.1") {
        Test4Handler()
      }    
    }
  }
}
```

How I envision it `TestHandler` would link to all three: `/path/to0`, `/path/to1` and `/path/to2/to2.1`.
First of all, I don't think it would be great to make a decision about order (like to take `Test2Handler` because its written few lines above).
Excluding anyone of them would result in those parts not being discoverable. HTTP wouldn't return links to those endpoints and for GraphQL does parts wouldn't even be accessible.
Would love to hear your opinion on that specific example, if you see it the same way as I?

So yes, the phrasing of this part of the proposal most definitely needs some clarifications (and maybe some more detailed examples) 👍 

So the retrieval algorithm looks something like the following: We search through every sub route of the Endpoint, and should we encounter a path which doesn't hold any endpoints we continue to recursively search all of sub routes for that path the same way.
Code description can be found in the experimental rest branch: https://github.com/Apodini/Apodini/blob/45023b3f2c871d7df14f269ec63e9a49f18f9a39/Sources/Apodini/Semantic%20Model%20Builder/EndpointsTree.swift#L172-L181

PSchmiedmayer: Thanks for the clarifications and the example! I think your approach makes sense and I can follow the algorithm. A agree with the tactic of linking to `/path/to0`, `/path/to1` and `/path/to2/to2.1`.

I think it makes sense that different operation types for RESTful APIs are not yet part of the proposal. Nevertheless I could envision that it might be useful to extend the functionality in the future. to e.g. not only like to ignore the operation type but to link to different operations in a different way. Maybe this would be part of the configuration DSL of a return type of a `Handler`. E.g. I could envision the idea of linking to a route that can delete the current retrieved type, modify the type and so on. But I think the first iteration is already a great step in the right direction. I am confident that the ideas is clearly communicated when you clarify the algorithm as described in the comment above. Tackling different operation types would then be something that we could tackle once we got basic relationships working.

Documentation/Relationship/Relationship.md
bauer-andreas: The idea was that the `var relationshipContainer` is part of the `Component` protocol. Thus more or less the reason for the `RelationshipContainer` protocol. When visiting the `Component` one could collect all the different `RelationshipContainer` and saving their type to be able to match up `KeyPath`s to the different containers.
It would act the same way as a component modifier stored in the Context with scope `.environment`.
It was my approach trying to create some sort of key value storage for `Relationship` and making it easier to reference them.
But I agree that it probably is the least polished thing in this proposal and that there might be cleaner ways of doing this. As this is really a "syntactic sugar" kind of feature, I would suggest, that we leave it out of the proposal for now (see https://github.com/Apodini/Apodini/pull/41#issuecomment-748679123), at least if nobody has a better idea. I think there is currently a lot of infrastructure work ongoing which we might be able to use(?)

bauer-andreas:
With 6d9c1fe I did the following changes (addressing @PSchmiedmayer latest feedback):
* Rephrased how Apodini creates Relationships from the Component structure (Chapter 1.)
* Renamed `.relatable()` to `.destination(of:)` (Chapter 2.1.)
* Added examples for REST and GraphQL showcasing manually added relationships (Chapter 2.1.)
* Removed RelationshipContainer (Chapter 2.2.)
* Explained why we used `hideLink` instead of `hideRelationship` (Chapter 2.3.)
* Better explanation of what we mean with "referencing types" (Chapter 3.1.)
* Renamed `.defaultReference` to `defaultRelationship` (Chapter 3.2.)
* Implementation example in the Appendix (Chapter 4.1.) was moved into sources as `RelationshipDefinition.swift`

bauer-andreas:
https://github.com/Apodini/Apodini/pull/41/commits/06f3c89b7c625428ae7f2e351752c5496fea4ae5 made the following two changes:

* Changed the order of the proposal so that the "Manual additions" chapter appears **after** the chapter about relationship retrieval through type information. I think this better resembles the importance of the latter one.
* Added [3.2.1](https://github.com/Apodini/Apodini/blob/feature/realtionship/Documentation/Relationship/Relationship.md#321-providing-type-hints) chapter which adds the `.relationship(to:,name:)` modifier. Using that you can manually define relationships by just supplying the return type of an Endpoint. This pretty much resembles the `References` relationship definition but for paths where we have not path parameter.

Unless there aren't any other feedback points I consider this PR to be ready to merge.

Documentation/Relationship/Relationship.md
bauer-andreas: Solved. See https://github.com/Apodini/Apodini/pull/41#issuecomment-748679123.

Documentation/Relationship/Relationship.md
bauer-andreas: Resolved. See https://github.com/Apodini/Apodini/pull/41#issuecomment-748679123.

Documentation/Relationship/Relationship.md
bauer-andreas: Resolved. See https://github.com/Apodini/Apodini/pull/41#issuecomment-748679123.

Documentation/Relationship/Relationship.md
bauer-andreas: Resolved. See https://github.com/Apodini/Apodini/pull/41#issuecomment-748679123.
Please state if the new formulation addresses your issue.

Documentation/Relationship/Relationship.md
bauer-andreas: Added. See https://github.com/Apodini/Apodini/pull/41#issuecomment-748679123.

Documentation/Relationship/Relationship.md
bauer-andreas: Resolved. Using `.destination(of:)`. See https://github.com/Apodini/Apodini/pull/41#issuecomment-748679123.

Documentation/Relationship/Relationship.md
bauer-andreas: Example was removed. Now just pointing to the examples in the following chapters. See https://github.com/Apodini/Apodini/pull/41#issuecomment-748679123.

Documentation/Relationship/Relationship.md
bauer-andreas: The phrasing was addressed with the latest changes. Additional a pseudo code implementation was provided, because I think nothing can describe recursive algorithms better than actual code.
See https://github.com/Apodini/Apodini/pull/41#issuecomment-748679123.
I agree that it will be beneficial to revisit that part of the proposal once we have basic relationships working 👍

bauer-andreas:
@PSchmiedmayer I saw you added a [Documentation.md](https://github.com/Apodini/Apodini/blob/develop/Documentation/Documentation.md). Should I add the `Relationship` folder to that file or should the `Relationship.md` be relocated to `AdditionalFunctionality` folder?

PSchmiedmayer (approved):
Thanks for all the thoughts that went into this proposal! I think it now well rounded and showcases a clear vision on how to implement the relationship information. Thanks for all the work and a great job @Supereg! 👍 

Regarding linking this proposal to [Documentation.md](https://github.com/Apodini/Apodini/blob/develop/Documentation/Documentation.md): I like the idea to add a separate `Relationship` folder and link it in [Documentation.md](https://github.com/Apodini/Apodini/blob/develop/Documentation/Documentation.md).

PR #43, opened 2020-12-06
GRPC exporter

moritzschuell:
This PR adds the initial version of the `GRPCSemanticModelBuilder`.
The exporter is based on the `SharedSemanticModelBuilder`.

The exporter supports the following:
- Expose unary endpoints that can be called by auto-generated client code
- Decode Protobuffer messages that are received from auto-generated clients
- Inject decoded Protobuffer message into (multiple) `@Parameter`s
- Encode unary response such that auto-generated clients can understand it

Streaming functionality will be added in a later PR.


Sources/Apodini/Semantic Model Builder/gRPC/GRPCService/GRPCServiceUnary.swift
```diff
@@ -0,0 +1,58 @@
...
+                .flatMap { _ in
+                    request.enterRequestContext(with: component) { component in
+                        var response: Encodable = component.handle()
+                        for responseTransformer in context.get(valueFor: ResponseContextKey.self) {
+                            response = request.enterRequestContext(with: responseTransformer()) { responseTransformer in
+                                responseTransformer.transform(response: response)
+                            }
+                        }
+                        // swiftlint:disable force_cast
+                        return request.eventLoop.makeSucceededFuture(response as! C.Response)
```
bauer-andreas: I'm not sure if we can cast this to C.Response here? It might be possible that response transformers modify the return type. 🤔

bauer-andreas (commented):
Besides the one comment the PR looks good to me. But can't really say anything about the gRPC details, so somebody might want to look over that again.

Sources/Apodini/Semantic Model Builder/gRPC/GRPCSemanticModelBuilder.swift
```diff
@@ -5,5 +5,49 @@
 //  Created by Paul Schmiedmayer on 6/26/20.
 //
 
+@_implementationOnly import Vapor
+@_implementationOnly import GRPC
+import NIOHPACK
+import ProtobufferCoding
```
tgymnich: could you use `@_implementationOnly` here?

tgymnich (dismissed):
LGTM

What would happen if my `Component` wants to return an Int16? Would I get a run time error or a compile time error?

Package.swift
```diff
@@ -34,7 +35,9 @@ let package = Package(
                 .product(name: "FluentMySQLDriver", package: "fluent-mysql-driver"),
                 .product(name: "APNS", package: "apns"),
                 .product(name: "AssociatedTypeRequirementsKit", package: "AssociatedTypeRequirementsKit"),
-                .product(name: "Runtime", package: "Runtime")
+                .product(name: "Runtime", package: "Runtime"),
+                .product(name: "GRPC", package: "grpc-swift"),
```
tgymnich: could you add a comment to describe where and how this package is used

Sources/Apodini/Semantic Model Builder/EndpointsTree.swift
```diff
@@ -33,6 +33,10 @@ struct Endpoint {
     let context: Context
     
     let operation: Operation
+
+    /// The type name of the associated component.
+    /// Used by the gRPC exporter as the default method name.
+    let componentName: String
```
nityanandaz: Note to myself: I just use "handle" for the method name.

Sources/Apodini/Semantic Model Builder/gRPC/GRPCError.swift
```diff
@@ -0,0 +1,23 @@
...
+ //
+ //
+ //  Created by Moritz Schüll on 05.12.20.
+ //
+
+ import Foundation
+
+ /// Errors used by any GRPC specific functionality of Apodini
+ /// and the GRPCSemanticModelBuilder.
+ enum GRPCError: Error {
```
nityanandaz, line 12: If you always have the same associated value, I would consider:
```
struct GRPCError: Error {
   enum Code { ... }
   let code: Code
   let message: String
}
```

Tests/ProtobufferCodingTests/ProtobufEncoderTests.swift
```diff
@@ -10,69 +10,124 @@ import XCTest
 @testable import ProtobufferCoding
 
 class ProtobufEncoderTests: XCTestCase {
```
nityanandaz: Shall we commit to always write Protobuffer instead of abbreviations?

Sources/Apodini/Semantic Model Builder/SharedSemanticModelBuilder.swift
```diff
@@ -27,7 +27,7 @@ class WebServiceModel {
 }
 
 class SharedSemanticModelBuilder: SemanticModelBuilder {
-    private var interfaceExporters: [InterfaceExporter]
+    var interfaceExporters: [InterfaceExporter]
```
nityanandaz: We should at least make the properties constant with `let`. `webService` and `rootNode` included.

Sources/Apodini/Semantic Model Builder/gRPC/GRPCRequest.swift
```diff
@@ -0,0 +1,176 @@
...
+
+/// Used by the `RequestWrapper` as the
+/// `ProtoCodingKey` for the wrapped value
+/// that should be decoded. Default is 1.
+/// Each thread needs its own field-number, because we might
+/// be decoding multiple requests at the same time.
+private var fieldNumber = ThreadSpecificVariable<FieldNumber>()
+
+class FieldNumber {
+    public var tag = 1
```
nityanandaz: This can be private, no?

Sources/Apodini/Semantic Model Builder/gRPC/GRPCError.swift
moritzschuell: Thanks for the suggestion @nityanandaz. 
Where do you see the benefits in this suggested refactoring?

Tests/ProtobufferCodingTests/ProtobufEncoderTests.swift
moritzschuell: Good point! We definitely should stick to one version.
Let's go for writing it out, as you suggested 👍

moritzschuell:
> LGTM
> 
> What would happen if my `Component` wants to return an Int16? Would I get a run time error or a compile time error?

@tgymnich You would get a runtime error. Using `Int16` in your component will work as long as you do not try to either encode or decode something into / from the `Int16`. If you try to, an exception will be raised by the encoder / decoder.

tgymnich:
> @tgymnich You would get a runtime error. Using Int16 in your component will work as long as you do not try to either encode or decode something into / from the Int16. If you try to, an exception will be raised by the encoder / decoder.

@moritzschuell maybe you can catch this runtime error at compile time using a protocol, that all grpc encodable types conform to. This is just an idea for later. You don't have to consider it for this PR.

tgymnich (approved):

Sources/Apodini/Semantic Model Builder/gRPC/GRPCError.swift
nityanandaz: It depends on your intention. Do you always want to require a message, even for future cases? Then you wouldn't need to repeat the associated value, too. Or do you expect cases where the case itself is sufficient?

I think explicit messages for user-facing errors are a good idea.

Sources/Apodini/Semantic Model Builder/gRPC/GRPCService/GRPCServiceUnary.swift
```diff
@@ -0,0 +1,42 @@
...
+@_implementationOnly import Vapor
+
+// MARK: Unary request handler
+extension GRPCService {
+    func createUnaryHandler(requestHandler: EndpointRequestHandler<GRPCInterfaceExporter>)
+    -> (Vapor.Request) -> EventLoopFuture<Vapor.Response> {
+        { (request: Vapor.Request) in
+            let promise = request.eventLoop.makePromise(of: Vapor.Response.self)
+            request.body.collect().whenSuccess { _ in
+                let response: EventLoopFuture<Encodable> = requestHandler.callAsFunction(request: request)
```
bauer-andreas: Small nit, the point of `callAsFunction` is that you can use a RequestHandler instance like the following. This was a follow up change introduced in #85.

For [reference](https://docs.swift.org/swift-book/ReferenceManual/Declarations.html#ID622).

```suggestion
                let response: EventLoopFuture<Encodable> = requestHandler(request: request)
```

Sources/Apodini/Modifier/GRPCMethodModifier.swift
```diff
@@ -0,0 +1,38 @@
...
+
+extension GRPCMethodModifier: SyntaxTreeVisitable {
+    func accept(_ visitor: SyntaxTreeVisitor) {
+        visitor.addContext(GRPCMethodNameContextKey.self, value: methodName, scope: .nextComponent)
+        component.accept(visitor)
+    }
+}
+
+extension Handler {
+    /// Explicitly sets the name of the gRPC service that is exposed for this `Component`
```
bauer-andreas: Nit, correct docs to `Handler` instead of `Component`

Sources/Apodini/Modifier/GRPCServiceModifier.swift
```diff
@@ -0,0 +1,38 @@
...
+
+extension GRPCServiceModifier: SyntaxTreeVisitable {
+    func accept(_ visitor: SyntaxTreeVisitor) {
+        visitor.addContext(GRPCServiceNameContextKey.self, value: serviceName, scope: .nextComponent)
+        component.accept(visitor)
+    }
+}
+
+extension Handler {
+    /// Explicitly sets the name of the gRPC service that is exposed for this `Component`
```
bauer-andreas: `Handler` instead of `Component` in docs

Sources/Apodini/Semantic Model Builder/Model/Endpoint.swift
```diff
@@ -58,6 +58,10 @@ struct Endpoint<H: Handler>: AnyEndpoint {
     
     let operation: Operation
 
+    /// The type name of the associated component.
+    /// Used by the gRPC exporter as the default method name.
+    let componentName: String
```
bauer-andreas: Nit: Is it necessary to embed this information here? First of all, we might want to change this to `handlerName` or `endpointName` as of the latest refactorings. Is this value supposed to always represent the string representation of the Handler type (`"\(H.self)"`)?
The gRPC exporter should be able to generate this string itself in its `export<H: Handler>(...)` implementation, as it has access to the generic type of `Endpoint` right?
If this property is indeed necessary, I would suggest to at least remove the init parameter, as the value can just be set in the initializer, removing the need for the model builder to pass it as an argument.

Sources/Apodini/Semantic Model Builder/gRPC/GRPCService/GRPCService.swift
```diff
@@ -0,0 +1,66 @@
...
+//
+//  Created by Moritz Schüll on 05.12.20.
+//
+
+import Foundation
+@_implementationOnly import Vapor
+@_implementationOnly import ProtobufferCoding
+
+/// Used by the `GRPCSemanticModelBuilder` to expose
+/// `handle` functions of `Components`.
```
bauer-andreas: Nit: `GRPCInterfaceExporter` vs. `GRPCSemanticModelBuilder`; `Handler` vs. `Components`

Tests/ApodiniTests/GRPCSemanticModelBuilderTests.swift
```diff
@@ -0,0 +1,144 @@
+//
+//  GRPCSemanticModelBuilder.swift
```
bauer-andreas: Nit: To be consistent, maybe renaming this test file to `GRPCInterfaceExporterTests.swift` (and the class accordingly)?

Sources/Apodini/Semantic Model Builder/SharedSemanticModelBuilder.swift
```diff
@@ -38,10 +38,10 @@ class WebServiceModel {
 }
 
 class SharedSemanticModelBuilder: SemanticModelBuilder, InterfaceExporterVisitor {
-    private var interfaceExporters: [AnyInterfaceExporter] = []
+    var interfaceExporters: [AnyInterfaceExporter] = []
```
bauer-andreas: Is there a reason for removing the private modifier? I couldn't find one in the code.

bauer-andreas (approved):
This PR looks good to me 👍 I had some very minor stuff I think is worth changing before merging, some address inconsistencies create as a result of splitting Components and Handlers, not sure if I caught all of them though.

Sources/Apodini/Semantic Model Builder/SharedSemanticModelBuilder.swift
moritzschuell: Thanks for pointing this out. No, there is no reason (anymore) 👍

Sources/Apodini/Semantic Model Builder/gRPC/GRPCInterfaceExporter.swift
```diff
@@ -31,11 +31,12 @@ class GRPCInterfaceExporter: InterfaceExporter {
             serviceName = builder.build()
         }
 
+        print("\(H.self)")
```
bauer-andreas: ooppsie debug wuuppsie. 🙃

Sources/Apodini/Semantic Model Builder/Model/Endpoint.swift
moritzschuell: You are right, this info can be generated from the `H: Handler` 👍
This seems to be a left-over from my move to the shared semantic model before the introduction of the generic `H: Handler` for the `export(..)` function.
I removed it.

Sources/Apodini/Semantic Model Builder/gRPC/GRPCInterfaceExporter.swift
moritzschuell: Ah damn 🤣🤣

bauer-andreas:
Feel free to merge from my side 🥳🎉 great work 👍

PR #45, opened 2020-12-07
Graphql Semantic Builder Info md File

sadikekin:
This PR contains the mapping of the Aponni DSL to GraphQl exporter.


sadikekin:
Hey @nerdsupremacist, Do you have any comments about this documentation?

Documentation/SemanticBuilders/GraphQLSemanticBuilder.md
```diff
@@ -0,0 +1,183 @@
...
+
+type Book {
+  author: Author!
+  name: String!
+  genre: String!
+  description: String!
+}
+
+type Query {
+  Book() : Book
```
nerdsupremacist: The field should probably have the same casing as in the group. And I think the Book is non null:

```graphql
book: Book!
```

Documentation/SemanticBuilders/GraphQLSemanticBuilder.md
```diff
@@ -0,0 +1,183 @@
...
+    UserHandler(userId: $userId)
+}
+```
+
+</td>
+<td>
+
+```graphql  
+type User {
+    userId: Int!
```
nerdsupremacist: the fields on user should probably be named like in the User Struct:

```graphql
type User {
    id: Int!
    name: String!
}
```

nerdsupremacist (changes requested):

nerdsupremacist (approved):

PR #46, opened 2020-12-07
NotificationCenter for APNS and FCM

awoc:
# *NotificationCenter for APNS and FCM*

Currently there is no way to send push notifications from a `Component`. Also, a `Configuration` for `FCM` is missing.

## :bulb: Proposed solution

The `NotificationCenter` handles the management of push notification enabled devices with a database and also the sending to APNS and FCM. This class comes with a lot of methods which offer an abstraction for both APNS and FCM. 

To use the `NotificationCenter` in Apodini, either APNS, FCM or both need to be enabled by using the corresponding `Configuration`. Furthermore, a configured database is needed to store and manage `Device`s using the `NotificationCenter`.

```swift
var configure: Configuration {
  APNSConfiguration(.pem(pemPath: "apns.pem"),
                                   topic: "bundleIdentifier",
                                   environment: .sandbox)
  FCMConfiguration("fcm.json")
  DatabaseConfiguration(.defaultMongoDB("mongodb://localhost:27017/apodini_db"))
    .addNotifications()
}
```

When using `.addNotifications()` the `NotificationCenter` will create three database tables: 

- Device: Stores every device with token and which service to use (APNS or FCM)
- Topic: Used to group devices together
- Device+Topic: Pivot for Many-to-Many relation between Device and Topic

The `NotificationCenter` offers a DTO to easily communicate with the API. Currently, custom schemas are not supported.

After the server configuration the `NotificationCenter` can be used in a `Component`.

```swift
struct Sample: Component {
  @Environment(\.notificationCenter) notificationCenter: NotificationCenter

  // ...
}
```

### Problem that is solved

This PR solves the problem to send push notifications from any `Component` by injecting the `NotificationCenter` using the `@Environment` property wrapper. Then any method of the `NotificationCenter` can be used in the `handle` method of a `Component`.

### Implications

This PR doesn't change the current structure of the project and shouldn't have any effect on other exporters. However, this implementation heavily relies on Vapor and there is no clear way how to import Vapor using `@_implementationOnly`. Furthermore, it adds dependencies for `APNS` and `FCM`.

### Testing

Tests cover mostly the database management of the `NotificationCenter` and also the structure of generated push notifications. The sending of messages to FCM and APNS cannot be covered by Unit Tests.

### Reviewer Nudging

It's not really easy to test the sending of push notifications since you need a sample App to send them to. If you want you can use my sample project on BitBucket and I can provide the certificates for APNS and FCM.


Sources/Apodini/Notifications/NotificationCenter.swift
```diff
@@ -0,0 +1,235 @@
+// swiftlint:disable first_where
+//
+//  NotificationCenter.swift
+//  
+//
+//  Created by Alexander Collins on 12.11.20.
+//
+
+import class Vapor.Application
+import struct Vapor.Abort
```
tgymnich, line 10: can we use `@_implementationOnly` here?

tgymnich (approved):
LGTM

Sources/Apodini/Notifications/NotificationCenter.swift
awoc: Unfortunately this doesn't work because I use `Application` directly which is the same problem as with `Configuration`s

PR #50, opened 2020-12-10
Connection for client-side streams

moritzschuell:
Builds on #40.

This PR adds a `Connection` that can be added to components via the `@Environment`:
```swift
struct Greeter: Component {
    @Environment(\.connection)
    var connection: Connection
        
    func handle() -> Action<String> {
        switch connection.state {
        case .open:
            // ...
        case .end:
            return .final("Hello world!")
        }
    }
}
```

The `Connection` can be injected into a component via `.withEnvironment(...)`.

An example, how a client-streaming request handler might use this functionality is `createClientStreamRequestHandler` in `Context`. The created handler can be exposed like this:
```swift
let handler = context. createClientStreamRequestHandler(withComponent: component, using: self)
app.on(.POST, "testservice", "method", body: .stream, use: handler)
```

For every frame on the stream, the request handler will be called. The request handler then can collect the frame's data from the body and decide how to forward it to the component's `handle` (e.g. in the gRPC case, this could mean collecting multiple frames until a whole message was received, and only then calling the component's `handle`).

This does not yet cover the `@State` properties as defined in [3.3. Client-Side Stream](https://github.com/Apodini/Apodini/blob/develop/Documentation/Communicational%20Patterns/3.%20Pattern%20Implementation/3.3.%20Client-Side%20Stream.md#client-side-stream).


Sources/Apodini/Components/Component.swift
```diff
@@ -26,8 +26,24 @@ public protocol Component {
     
     /// A function that is called when a request reaches the `Component`
     func handle() -> Self.Response
+    /// The default implementation simply wraps results from
+    /// `handle() -> Self.Response` into `.final(..)`.
+    func handle() -> Action<Self.Response>
```
moritzschuell: @nerdsupremacist I am having a hard time trying to get rid of the second `handle` function, as we discussed 🤔
If I make `Action: ResponseEncodable`, and only have the `handle() -> Response` function, I still do not see how we could retrieve the real type of the returned Action's generic. Even if I would be able to check the returned `Response` for conformance to `Action` with the code you showed me.
But as far as I see, we need this type-info to be able to unwrap the value from `.final(..)` and `.send(..)`.

After thinking this through again, I think we should discuss (in the next work meeting) whether having specialized types of handlers for different kinds of communicational patterns is an option. In my opinion, it would be a quite useful thing, as it removes the ambiguity of the returned `Response`s type, and it would also hinder the developer to overwrite both of the `handle` functions (as is possible to do in my current version).

theMomax (dismissed):
This looks very nice!

I don't see a problem with having two versions of the `handle()` function. If the developer overrides both, that is totally fine. We only ever call `handle() -> Action<Self.Response>`. They will notice pretty quickly that the other one isn't executed if `handle() -> Action<Self.Response>` is provided.

I would heavily argue against having "specialized types of handlers for different kinds of communicational patterns". We've had the discussion before (which took quite some time) and came up with the vision we're pursuing right now. Please refer to **Apodini's Paradigm** from our [concept on communicational patterns](https://github.com/Apodini/Apodini/blob/develop/Documentation/Communicational%20Patterns/1.%20Introduction.md) for more information _why_ we chose our current approach.

The only problem I can see right now are missing tests, but codecov blocks this PR anyways until this is fixed. In terms of the code you have provided so far and its functionality, I approve.

PSchmiedmayer:
It is great to see that `Connection` and `Action` start to take shape, good job!
While I was just reading over this PR for some context for #56 I just noticed that the PR includes a lot of Vapor dependencies. I would suggest to focus on importing Vapor as implementation only to finally get all the Vapor decoupling going. I have the feeling that the Vapor PRs are open for a very long time and must be merged as soon as possible and all other part of the project should start hiding Vapor as implementation only as discussed in the intermediate presentations. (CC: @tgymnich)

Sources/Apodini/Components/Component.swift
nerdsupremacist: Hmmmm I see. Those are some interesting concerns

My (not so informed) opinion:
1: I personally think that we need to go a different route from `ResponseEncodable`. 
Because stating that an action is response encodable is just wrong semantically speaking.
2: As for the generic type of action, we can probably work around that with an internal protocol like this:
```
protocol ActionProtocol {
    static var encodedType: Any.Type { get }
    var value: Any? { get }
    var isFinal: Bool { get }
}
```
I think we can keep this protocol internal if we decide to only use to build the semantic model
3: The idea of a different kind of handler also works for me and seems intuitive enough

Sources/Apodini/Components/Component.swift
```diff
@@ -25,8 +25,24 @@ public protocol Component {
...
+    /// The default implementation simply wraps results from
+    /// `handle() -> Self.Response` into `.final(..)`.
+    public func handle() -> Action<Self.Response> {
+        .final(handle())
+    }
+
+    /// Either this function or `handle() -> Action<Self.Response>`
+    /// have to be overwritten.
+    public func handle() -> Self.Response {
+        fatalError("Either handle() -> Encodable or handle() -> Action<Encodable> have to be implemented")
```
nerdsupremacist: This indirection of a handle method calling the other, and therefore having both implemented by an extension makes our DSL more prone for issues where the return type of handle is not Encodable and therefore the handle method is not inferred properly by swift

This is why I don't like the idea of having two handle methods, since it reverts one of our biggest wins from splitting Components and Handlers (before we even did the split 😂 )

Sources/Apodini/Request/Action.swift
```diff
@@ -0,0 +1,23 @@
+//
+//  Action.swift
+//
+//
+//  Created by Moritz Schüll on 10.12.20.
+//
+
+@_implementationOnly import Vapor
```
nerdsupremacist: nit: I think you can remove Vapor and just import Foundation

nerdsupremacist (dismissed):
I like this PR a lot. It's a great step forward. Great job!
I'm approving to unblock.

We should discuss in the Work Meeting ideally with @theMomax @Supereg on how to deal with returning action's (hopefully without two functions)

Sources/Apodini/Components/Component.swift
bauer-andreas: Should be solved with c1d8e5a

Sources/Apodini/Request/Action.swift
```diff
@@ -18,14 +18,17 @@ public enum Action<Element: Encodable>: ApodiniEncodable {
...
     func accept(_ visitor: ApodiniEncodableVisitor) {
         visitor.visit(action: self)
     }
 
     public func encode(to encoder: Encoder) throws {
         switch self {
-        case .nothing:
+        case .nothing,
+             .end:
             fatalError("Tried encoding Action.nothing!")
```
bauer-andreas: A little nit here: In the case that we would reach the race condition that `.end` is encoded, the fatalError would still print "Tried encoding .nothing" which is a bit misleading I think, should we ever need to track down such an issue.
Maybe just make a extra case or put `\(self)` into the string.

moritzschuell:
Rerequesting review from @theMomax, @nerdsupremacist, @Supereg.

@supereg helped to get rid of the second `handle` method via a visitor that gives us back info about the returned type of the handler. 
Also, I added a "tiered" approach to the `@Environment` values as we discussed in the last team-meeting. `@Environment` can now hold request-specific as well as global values. It will try to lookup in this order (first request-specific, then global) for a given key-path. If neither is present, the default value is returned.

Sources/Apodini/Environment/Connection.swift
```diff
@@ -0,0 +1,39 @@
+//
+//  Connection.swift
+//  
+//
+//  Created by Moritz Schüll on 09.12.20.
+//
+
+import Foundation
+@_implementationOnly import Runtime
```
theMomax: Unused import?

Sources/Apodini/Environment/Environment.swift
```diff
@@ -37,14 +39,46 @@ public struct EnvironmentValues {
...
+    /// Sets properties of the `Component` annotated with
+    /// `@Connection` to the given connection value.
+    /// - parameters:
+    ///     - connection: The `Connection` that should be injected.
+    func withEnvironment<Value>(_ value: Value, for keyPath: WritableKeyPath<EnvironmentValues, Value>) -> Self {
+        var selfRef = self
+        do {
+            let info = try typeInfo(of: type(of: self))
+
+            for property in info.properties {
```
theMomax, line 72: I think we'll have to rework this at some point in the future. If implemented this the method only works on the handling component. I guess it would be best to first store the value+keypath in some Context Key, so the shared semantic model can take care of injection later on (this would also make it work if used on a non-handling parent-component).

Sources/Apodini/Environment/Environment.swift
```diff
@@ -37,14 +39,46 @@ public struct EnvironmentValues {
...
+    mutating func setValue(_ value: Value, for keyPath: WritableKeyPath<EnvironmentValues, Value>) {
+        self.dynamicValues[ObjectIdentifier(keyPath)] = value
+    }
+}
+
+extension Component {
+    /// Sets properties of the `Component` annotated with
+    /// `@Connection` to the given connection value.
+    /// - parameters:
+    ///     - connection: The `Connection` that should be injected.
```
theMomax: This documentation seems wrong to me...this function is not specific to `Connection`, is it?

Sources/Apodini/Semantic Model Builder/SharedSemanticModelBuilder.swift
```diff
@@ -118,15 +118,70 @@ class SharedSemanticModelBuilder: SemanticModelBuilder {
...
+                        let visitor = ActionVisitor(request: request,
+                                                    promise: promise,
+                                                    responseModifiers: responseModifiers)
+                        switch response {
+                        case is ApodiniEncodable:
+                            // is an Action
+                            // use double dispatch to access
+                            // wrapped element
+                            // swiftlint:disable:next force_cast
+                            (response as! ApodiniEncodable).accept(visitor)
```
theMomax: I think this force-cast could be avoided using `case let apodiniEncodableResponse as ApodiniEncodable`.

Sources/Apodini/Semantic Model Builder/SharedSemanticModelBuilder.swift
```diff
@@ -118,15 +118,70 @@ class SharedSemanticModelBuilder: SemanticModelBuilder {
...
+            case let .send(element):
+                let result = transformResponse(element)
+                promise.succeed(result)
+            case let .final(element):
+                let result = transformResponse(element)
+                promise.succeed(result)
+                try request.eventLoop.close()
+            case .end:
+                // send back an empty response
+                promise.succeed("")
```
theMomax: I think the exporter should be allowed to handle the `Action` itself. One exporter might be able to close a connection without sending an extra message while others indeed have to send an empty message. I would thus make `ActionVisitor.promise` (and as a consequence the return-value of `RequestHandler`) an `EventLoopFuture<Action<Encodable>>`.

Sources/Apodini/Semantic Model Builder/SharedSemanticModelBuilder.swift
```diff
@@ -118,15 +118,70 @@ class SharedSemanticModelBuilder: SemanticModelBuilder {
...
+                try request.eventLoop.close()
```
theMomax: Are you sure this is necessary/correct? From my understanding this would close the entire `EventLoop` (which might be handling thousands of requests/connections and not only the one you are handling here). Same applies to L164. If `ActionVisitor.promise` would contain the whole `Action` (that might contain the transformed response), the exporter could handle closing the connection according to their logic.

Tests/ApodiniTests/SharedSemanticModelBuilderTests.swift
```diff
@@ -149,4 +166,18 @@ final class SharedSemanticModelBuilderTests: XCTestCase {
...
+            .withEnvironment(Connection(state: .open), for: \.connection)
+        let requestHandler = SharedSemanticModelBuilder.createRequestHandler(with: component,
+                                                                             guards: [],
+                                                                             responseModifiers: [ { transformer } ])
+        let name = "Craig"
+        let request = RESTRequest(Vapor.Request(application: app, on: app.eventLoopGroup.next())) { _ in name }
+        let response = try requestHandler(request).wait()
+        let responseString = try XCTUnwrap(response as? String)
+        XCTAssertEqual(responseString, "✅ Hello \(name) ✅")
+    }
```
theMomax, line 242: I think you should add a test on how `requestHandler` behaves when `.final` or `.end` is returned.

theMomax (changes requested):
I noted down some smaller clean-up suggestions. However, I also have some concerns about `ActionVisitor.visit<Element: Encodable>(action: Action<Element>)`. Maybe I'm wrong, but the lines I'm concerned about are not covered by tests either, so please check it out again (I added some in-code comments with more details).

Sources/Apodini/Semantic Model Builder/SharedSemanticModelBuilder.swift
moritzschuell: That's a good point. Reading this, I do not think that closing the `EventLoop` is actually the correct thing to do here. I just assumed that closing the `EventLoop` is the intended behavior of the `.end` case.

I removed this.
Instead, the `RESTInterfaceExporter` now sends back an empty message with status `200 OK` (both, for `.nothing` and `.end`). But I am not sure whether that is sufficient for a `.end`?

Sources/Apodini/Semantic Model Builder/SharedSemanticModelBuilder.swift
moritzschuell: I like the idea of delegating the handling of the `Action` to the exporter. This would also keep us from having to make assumptions about how the proper reaction to `.end` looks like.
Unfortunately, `Action<Encodable>` doesn't work (because `Value of protocol type 'Encodable' cannot conform to 'Encodable'` 💩). I used the `AnyEncodable` instead, which makes it a bit uglier but good enough in my opinion.

theMomax (approved):
Looks good! I think it's a really good step forward to have `Action` set up and `Connection` defined. Thanks a lot for pushing this topic forward👍🏻.

PR #49, opened 2020-12-10
Add codecov

tgymnich:
Added CI workflow to upload code coverage reports to [codecov.io](https://app.codecov.io/gh/apodini/apodini/)

PR #52, opened 2020-12-10
Create codecov.yml

tgymnich:
Ignore tests in code coverage report.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,23 @@
+ignore:
+- "/Tests/*"
+
+codecov:
+  require_ci_to_pass: yes
+
+coverage:
+  precision: 2
+  round: down
+  range: "70...100"
+
+parsers:
+  gcov:
+    branch_detection:
+      conditional: yes
+      loop: yes
+      method: no
+      macro: no
+
+comment:
+  layout: "reach,diff,flags,files,footer"
+  behavior: default
+  require_changes: no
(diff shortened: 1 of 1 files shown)


codecov.yml
```diff
@@ -0,0 +1,23 @@
+ignore:
+- "Tests"
+
+codecov:
+  require_ci_to_pass: yes
+
+coverage:
+  precision: 2
+  round: down
+  range: "70...100"
```
bauer-andreas, line 11: I assume this force the code coverage to be between 70-100? Otherwise the PR won't be mergeable or what is the intention?

bauer-andreas (commented):
So this is a followup to #49  right?
It basically adds the configuration file for codecov?

codecov.yml
tgymnich: correct. its just the default value.

tgymnich: > So this is a followup to #49 right?
> It basically adds the configuration file for codecov?

yes. Codecov didn't ignore the tests by itself, so I had to add this file

bauer-andreas: Does the CI run enforce a certain code coverage? Or does the CI job just fail when the code coverage decreases like in this case?

tgymnich: it just fails. Codecov checks are not "required" Should we set the range to 51% (current coverage) or leave it a 70 to motivate us to write tests?

bauer-andreas (dismissed):

bauer-andreas (approved):

PR #53, opened 2020-12-12
Shared Semantic Model R2

bauer-andreas:
This is a minor revision to the shared semantic model.

This PR contains two changes:

* d24c571: I found a way to completely eliminate the use of the Runtime framework, the idea really just came randomly to my mind 😅. It uses a basic Visitor pattern to get to the type information. As you @lschlesinger are the code owner of that part of the shared model, would be great if you could leave a code review.  
This commit also fixes an issue where a PathParameter would always be passed as a string representation to the PathBuilder.

* 6a6dc97 improves on the `InterfaceExporter` interface. When discussing the needs of other exporters like gRPC with @moritzschuell, we found that the exporter doesn't really need to get a reference to the individual nodes of the EndpointsTree. Thus the `InterfaceExporter.export`  was changed to be called for every endpoint in the tree.

### DIFF
--- a/Sources/Apodini/Request/RequestInjectable.swift
+++ b/Sources/Apodini/Request/RequestInjectable.swift
@@ -12,11 +12,29 @@ import Runtime
 
 protocol RequestInjectable {
     mutating func inject(using request: Vapor.Request, with decoder: RequestInjectableDecoder?) throws
+    func visit(_ visitor: RequestInjectableVisitor)
+}
+
+extension RequestInjectable {
+    func visit(_ visitor: RequestInjectableVisitor) {
+        visitor.register(self)
+    }
 }
 
 protocol RequestInjectableDecoder {
     func decode<T: Decodable>(_ type: T.Type, from request: Vapor.Request) throws -> T?
 }
+
+protocol RequestInjectableVisitor {
+    func register<Injectable: RequestInjectable>(_ requestInjectable: Injectable)
+
+    func register<Element>(_ parameter: Parameter<Element>)
+}
+extension RequestInjectableVisitor {
+    func register<Injectable: RequestInjectable>(_ requestInjectable: Injectable) {}
+    func register<Element>(_ parameter: Parameter<Element>) {}
+}
+
 
 private func extractRequestInjectables(from subject: Any) -> [String: RequestInjectable] {
     Mirror(reflecting: subject).children.reduce(into: [String: RequestInjectable]()) { result, child in
--- a/Sources/Apodini/Request/RequestInjectable.swift
+++ b/Sources/Apodini/Request/RequestInjectable.swift
@@ -12,12 +12,12 @@ import Runtime
 
 protocol RequestInjectable {
     mutating func inject(using request: Vapor.Request, with decoder: RequestInjectableDecoder?) throws
-    func visit(_ visitor: RequestInjectableVisitor)
+    func accept(_ visitor: RequestInjectableVisitor)
 }
 
 extension RequestInjectable {
-    func visit(_ visitor: RequestInjectableVisitor) {
-        visitor.register(self)
+    func accept(_ visitor: RequestInjectableVisitor) {
+        visitor.visit(self)
     }
 }
 
@@ -26,13 +26,13 @@ protocol RequestInjectableDecoder {
 }
 
 protocol RequestInjectableVisitor {
-    func register<Injectable: RequestInjectable>(_ requestInjectable: Injectable)
+    func visit<Injectable: RequestInjectable>(_ requestInjectable: Injectable)
 
-    func register<Element>(_ parameter: Parameter<Element>)
+    func visit<Element>(_ parameter: Parameter<Element>)
 }
 extension RequestInjectableVisitor {
-    func register<Injectable: RequestInjectable>(_ requestInjectable: Injectable) {}
-    func register<Element>(_ parameter: Parameter<Element>) {}
+    func visit<Injectable: RequestInjectable>(_ requestInjectable: Injectable) {}
+    func visit<Element>(_ parameter: Parameter<Element>) {}
 }
 
 
--- a/Sources/Apodini/Semantic Model Builder/EndpointParameter.swift
+++ b/Sources/Apodini/Semantic Model Builder/EndpointParameter.swift
@@ -5,7 +5,6 @@
 //  Created by Lorena Schlesinger on 06.12.20.
 //
 
-import Runtime
 import Foundation
 
 struct EndpointParameter {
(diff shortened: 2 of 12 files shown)


Sources/Apodini/Request/RequestInjectable.swift
```diff
@@ -12,12 +12,30 @@ import Runtime
 
 protocol RequestInjectable {
     mutating func inject(using request: Vapor.Request, with decoder: RequestInjectableDecoder?) throws
+    func visit(_ visitor: RequestInjectableVisitor)
```
lschlesinger: I know this naming schema `visit(visitor: ...)` is similar to the `SyntaxTreeVisitor` but I got a little confused why we do not use the more common `accept(visitor: ...)` method. Is it for a certain reason? Just from the patterns lecture and wikipedia, I would assume, that the `Visitor` has a `visit()` method whereas the `Visitable` has an `accept()` method.

bauer-andreas: Yeah, I pretty much adopted the naming from the `SyntaxTreeVisitor`. 😅 But yes you are correct. We may want to change that to be consistent with the literature. ~I got a issue open right know for the SyntaxTreeVisitor to fix a spelling mistake in the class name, we may just incorporate that naming change there and adjust this PR as well?~

bauer-andreas: solved via e7383a3

bauer-andreas: for the sake of not including two different things in one PR I already merged #54 and we may just create an additional PR to change the naming there as well.

Sources/Apodini/Semantic Model Builder/EndpointsTree.swift
```diff
@@ -50,11 +53,26 @@ struct Endpoint {
...
+    var relationships: [EndpointRelationship] {
+        treeNode.relationships
+    }
+
+    init(description: String, context: Context, operation: Operation, guards: [LazyGuard], requestInjectables: [String: RequestInjectable], handleMethod: @escaping () -> ResponseEncodable, responseTransformers: [() -> (AnyResponseTransformer)], handleReturnType: ResponseEncodable.Type, parameters: [EndpointParameter]) {
+        self.description = description
+        self.context = context
+        self.operation = operation
+        self.guards = guards
+        self.requestInjectables = requestInjectables
```
moritzschuell, line 70: I think @hendesi found a problem in how the `requestInjectables` are stored in the `Endpoint`.
By storing them as `[String: RequestInjectable]` they are not recognized by the `inject` method's `as? RequestInjectable`.

Credits to @theMomax

Sources/Apodini/Semantic Model Builder/EndpointParameter.swift
```diff
@@ -23,57 +22,86 @@ struct EndpointParameter {
...
+    mutating func append<T>(_ parameter: Parameter<T>) {
+        result = PathParameterAnalyzingResult(
+                parameterMode: parameter.option(for: .http)
+        )
+    }
+
+    mutating func append(_ string: String) {}
+
+    /// This function does two things:
+    ///   * First it checks if the given `_PathComponent` is if type Parameter. If it is it returns
```
lschlesinger, line 99: > is _of_ type Parameter.

Sources/Apodini/Semantic Model Builder/InterfaceExporter.swift
```diff
@@ -7,7 +7,13 @@ import Vapor
 protocol InterfaceExporter: RequestInjectableDecoder {
     init(_ app: Application)
 
-    func export(_ node: EndpointsTreeNode)
+    func export(_ endpoint: Endpoint)
+
+    func finishedExporting(_ root: EndpointsTreeNode)
```
moritzschuell: Will there be a real use-case for this function?
From my gRPC POV I can't imagine one (except logging "I'm done exporting" 😁)

moritzschuell (commented):
Looks great 👍
Just left a question and a remark that refers to the recent discussion in the RocketChat group.

Sources/Apodini/Semantic Model Builder/InterfaceExporter.swift
bauer-andreas: Yeah felt a bit dirty to me as well 😅 The use case for REST is that, in the case that the user didn't register a handler under the `/` path (not sure if they should actually), the REST exporter should automatically create a route handler there to be able to respond with links to all 'entry points'.

like the example:
```swift
var content: some Component {
  Group("user") {
    // ...
  }
  Group("articles") {
    // ...
  }
}
```

In order to be 100% HATEOS "compliant" a client should be able to do a request to `GET /` and be able to get links to `/user` and `/articles`. The REST exporter basically needs the information, are there any endpoints under the root path.
By the time of writing, I could see that we could remove the `node` argument, and thus fully hide `EndpointsTreeNode`s from the exporters. the `export` calls could just save the information of the presence of a `/` endpoint. But in any case there would need to be a `finishedExporting`. But the method is optional, as it provides a empty default implementations as I'm aware, that not many exporters need that.

bauer-andreas: Nevermind, I need the information, to which endpoints the root node links. Those must be retrievable with some other mechanism in order to remove that parameter.

lschlesinger:
Looks very good to me! From my POV this can be merged.

lschlesinger (dismissed):

Sources/Apodini/Semantic Model Builder/EndpointsTree.swift
bauer-andreas: Yeah, just chatted with @theMomax as well. This is a bigger "conceptual" problem. As the current shared model doesn't keep a reference to the component 🤔
I'll fix that in a follow up PR if that is fine?

moritzschuell: Alrighty 👍

Sources/Apodini/Semantic Model Builder/InterfaceExporter.swift
bauer-andreas: @moritzschuell Would you mind having a look at fe34827 which tries to completely hide the EndpointsTree to the exporters by creating a WebService model to store said information.

moritzschuell: Looks good to me 🙂

moritzschuell (approved):

lschlesinger (approved):

PR #55, opened 2020-12-12
Fix in Endpoint for "external" `.path` Parameters

lschlesinger:
The PR adds a path parameter also if it is defined outside of Handler without explicit declaration inside Handler

The following corner case was not handled correctly:

```swift
struct TestHandler4: Component {
    func handle() -> String {
        "Hello Test Handler 4"
    }
}

struct TestComponent2: Component {
    @PathParameter
    var name: String

    var content: some Component {
        Group($name) {
            TestHandler4()
        }
    }
}
```
The corner case is now part of the `SharedSemanticModelBuilderTests`.

The `PathParameter` was included in the path but not listed as `EndpointParameter` in the `parameters` property of an `Endpoint`. 
 As a solution, one should retrieve `EndpointParameter`s from both, `RequestInjectables` (in case they are declared in a `Handler`) AND `_PathComponent`s included in the `Endpoint`s `absolutePath` (in the described case).

### DIFF
--- a/Sources/Apodini/Semantic Model Builder/EndpointParameter.swift
+++ b/Sources/Apodini/Semantic Model Builder/EndpointParameter.swift
@@ -30,14 +30,26 @@ struct EndpointParameter {
             }
     }
     
+    static func create(from pathComponent: _PathComponent, label: String) -> EndpointParameter? {
+        guard let info: TypeInfo = try? typeInfo(of: type(of: pathComponent)) else {
+            return nil
+        }
+        return create(from: info, source: pathComponent, label: label)
+    }
+    
     static func create(from requestInjectable: RequestInjectable, label: String) -> EndpointParameter? {
         guard let info: TypeInfo = try? typeInfo(of: type(of: requestInjectable)) else {
             return nil
         }
+        return create(from: info, source: requestInjectable, label: label)
+    }
+    
+    /// TODO: Why does this not work with `create<T>(from source: T, ...) { typeInfo(of: type(of: source)) ... }`
+    private static func create<T>(from info: TypeInfo, source: T, label: String) -> EndpointParameter? {
         /// Parameter<String> serves as representative `parameterType` of any Parameter<T> as  `mangeldName` of all Parameter<T> is `Parameter`
         let parameterType = try? typeInfo(of: Parameter<String>.self)
         if info.mangledName == parameterType?.mangledName {
-            let mirror = Mirror(reflecting: requestInjectable)
+            let mirror = Mirror(reflecting: source)
             // swiftlint:disable:next force_cast
             let id = mirror.children.first { $0.label == ParameterProperties.id }!.value as! UUID
             let name = mirror.children.first { $0.label == ParameterProperties.name }?.value as? String
@@ -75,5 +87,15 @@ struct EndpointParameter {
                 parameterType: parameterType)
         }
         return nil
+    }
+}
+
+extension EndpointParameter: Equatable, Hashable {
+    static func == (lhs: EndpointParameter, rhs: EndpointParameter) -> Bool {
+        lhs.id == rhs.id
+    }
+    
+    func hash(into hasher: inout Hasher) {
+        hasher.combine(id)
     }
 }
--- a/Sources/Apodini/Semantic Model Builder/EndpointsTree.swift
+++ b/Sources/Apodini/Semantic Model Builder/EndpointsTree.swift
@@ -50,7 +50,7 @@ struct Endpoint {
     }()
     
     /// All `@Parameter` `RequestInjectable`s that are used inside handling `Component`
-    lazy var parameters: [EndpointParameter] = EndpointParameter.create(from: requestInjectables)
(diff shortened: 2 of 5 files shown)


Sources/Apodini/Semantic Model Builder/EndpointParameter.swift
```diff
@@ -30,14 +30,26 @@ struct EndpointParameter {
...
+    }
+    
     static func create(from requestInjectable: RequestInjectable, label: String) -> EndpointParameter? {
         guard let info: TypeInfo = try? typeInfo(of: type(of: requestInjectable)) else {
             return nil
         }
+        return create(from: info, source: requestInjectable, label: label)
+    }
+    
+    /// TODO: Why does this not work with `create<T>(from source: T, ...) { typeInfo(of: type(of: source)) ... }`
```
lschlesinger, line 47: @Supereg Do you have a better idea for this? It does not seem to work with a straightforward generic function, as the `TypeInfo` will then always be the original `protocol` (e.g., `RequestInjectable` or `_PathComponent`) rather than `Parameter<.>`.

bauer-andreas:
@moritzschuell also noticed this edge case being unhandled when we talked about how gRPC could use the shared model.
My idea was that we just ignore `@PathParameters` contained in the path which don't have a `@Parameter` in a Handler somewhere. Instead we just print a warning if a `Endpoint` has a `@PathParameter` in its path but doesn't have a `@Parameter` in its `Handler`. I would argue against the way you handled that for two reasons.
- **1:** We would have a Parameter which isn't handled anyways. We can't inject that into anything. So the web service would be broken anyways and the developer needs to fix that. A warning (or maybe even an error?) would be an important hint.
- **2:** We don't have a name for it. The `@Paramter` definition inside the handler holds all the important information (as to the current defined proposal!). Either the property name or a manual defined `.name` inside the `@Parameter` struct is used. 

How would different exporters handle such invalid definitions if we would go for a warning:
  * REST exporter would include the PathParameter in its path registration, as it just takes the PathComponents and passes them to Vapor. In a request, the actual value for the PathParameter would then get lost/be ignored, as the handler has no `@Parameter` to inject it.
  * OpenAPI is similar to how REST would register a `@PathParamter` as the open api documentation would just include it. As there is no name to be used in the documentation, maybe we can somehow mark them as "unnamed".
  * gRPC will ignore any path parameter when generating its procedure name anyways. So basically gRPC just ignores those.
  * In theory GraphQL is similar to REST that it would expect a path parameter in the query schema but ignoring it in the Handler. The problem is that the parameter name must be know for the query schema as it is supplied by the client. So the only real option I see for GraphQL is to error on such an occasion.
  * I have no idea about Websockets as I don't know how they handled regularly defined `@PathParamerters`

---
### Appendix

Two things which came to my mind when writing this which I think still need to be handled in the shared model.

* **Name collisions:**
  We need to ensure that parameter names don't collide with each other. Though, different exporters have different needs for     that. E.g. for REST a name must only be unique for the given parameter type: query, content and for path it isn't really relevant (only for OpenAPI to get an unambiguous documentation). For gRPC the namespace is independent from the parameter type. So maybe we can provide a mechanism where the exporter specifies the namespace when retrieving the parameters (like a Getter passing a namespace definition) or we document, that this must be handled by the exporter.
 
* **Create a custom PathComponent representation:**
  Having the 'raw' `@Parameter` in the path isn't the best solution I think, as it introduces complication handling those in the exporters. I think we could somehow incorporate a Model for that as well, where we can encode additional information, like e.g. like a link the `EndpointParameter`. This could make life easier to retrieve e.g. naming information. Additionally exporters like GraphQL could easily detect if there is no associated `EndpointParameter` and thus throw an error.

Sources/Apodini/Semantic Model Builder/EndpointParameter.swift
bauer-andreas: > It does not seem to work with a straightforward generic function, as the TypeInfo will then always be the original protocol (e.g., RequestInjectable or _PathComponent) rather than Parameter<.>.

Such Type information for generics is always determined at **compile time** (at the location of the call). As create is called few lines above for `RequestInjectable` or `_PathComponent` those types are used.

In #53 I incorporated the visitor pattern for RequestInjectables to preserve the that type information. I could remove the dependency on the Runtime framework here.

lschlesinger: Perfect, I'm looking at it right now.

lschlesinger:
Thank you @Supereg, valid points that you've mentioned! Let's leave this PR as it is for now and discuss this on/after Tuesday :)

bauer-andreas:
Sounds good 👍 though if we can discuss this earlier it would also be good. As next meeting (+ work meeting) is schedule on the 22nd

bauer-andreas:
Just created a PR addressing that issue (see the above PR where this Draft got mentioned). I would propose moving the discussion over to that and closing this Draft?

lschlesinger:
addressed by #104

PR #56, opened 2020-12-12
DynamicProperty and Dynamics

theMomax:
This PR contains the implementation of two new internal functions:

```swift
func execute<Element, Target>(_ operation: (Target, _ name: String) -> (), on element: Element)

func apply<Element, Target>(_ mutation: (inout Target, _ name: String) -> (), to element: inout Element)
```

These functions provide an abstraction of the code that is used by  `RequestInjectable`s (more specifically `enterRequestContext`) to find and modify all `RequestInjectable` properties on an element.

The two new functions execute the given `operation`/`mutation` on any property of type `Target` that lives on `element`. The difference between `execute` and `apply` is that the callback given to `apply` may manipulate the given `Target` and thus the whole `element`. This also means that `execute` is more lightweight because it requires less runtime reflection.

The core functionality of `enterRequestContext` is now expressed as follows:

```swift
apply({ (ri: inout RequestInjectable) in
    do {
        try ri.inject(using: self, with: decoder)
    } catch {
         fatalError("Injecting into element \(e) failed.")
    }
}, to: &element)
```

This PR also defines two new public types:
* The empty protocol `DynamicProperty` can be implemented by any struct.
* The struct `Dynamics` holds a list of named objects.

Both, the properties of any `DynamicProperty` and the objects stored by `Dynamics` are taken into account by `apply` and `execute` recursively, i.e. the `operation`/`mutation` is called on any object of type `Target` that lives either directly on the `element` or  somewhere on a `DynamicProperty` or `Dynamics` where the n-th parent is `element` and all parents n-1...1 are either of `DynamicProperty` or `Dynamics`.

This enables nesting of Apodini's property wrappers like the following:

```swift 
struct Greeter: Component {
        @A var name: String?
        var d: Dynamics = ["surname": Parameter<String?>()]
        
        func handle() -> String {
            let surnameParameter: Parameter<String?>? = d.surname
            
            return (name ?? "Unknown") + " " + (surnameParameter?.wrappedValue ?? "Unknown")
        }
    }
    
    @propertyWrapper
    struct A: DynamicProperty {
        @Parameter var name: String?
        
        var wrappedValue: String? {
            return name
        }
    }
```

~~The API of `Dynamics` is still pretty unpolished (one could imagine usage of `@propertyWrapper` or `@dynamicMemberLookup` to make unwrapping the parameter inside the `handle` easier. Thus, this PR is a WIP.~~
`Dynamics` now
* provides string-based, generic `@dynamicMemberLookup` (mostly relevant for easier access for tests)
* conforms to `ExpressibleByDictionaryLiteral`
* confroms to `Collection`
* provides `func typed<T>(_ type: T.Type) -> Dynamics<T>` that allows to easily get a `Collection` that only contains those elements which conform to type `T`

edit: update `Dynamics` API


Sources/Apodini/Request/DynamicProperty.swift
```diff
@@ -0,0 +1,208 @@
...
+    
+                mutation(&t, name)
+                self.elements[name] = t
+            case var dp as DynamicProperty:
+                assert((Mirror(reflecting: element).displayStyle) == .struct, "DynamicProperty \(name) on Dynamics must be a struct")
+                
+                dp.apply(mutation)
+                self.elements[name] = dp
+            case var dyn as Dynamics:
+                // no assertion needed because Dynamics is defined by Apodini
```
nerdsupremacist, line 199: I'd still add the assertion though. To catch it in tests when something is changed

Sources/Apodini/Request/Dynamics.swift
```diff
@@ -0,0 +1,26 @@
+//
+//  Dynamics.swift
+//  
+//
+//  Created by Max Obermeier on 10.12.20.
+//
+
+import Foundation
+
+public struct Dynamics {
```
nerdsupremacist: Please make it a sequence or collection... To make iterating over the injectables easier

Sources/Apodini/Request/Dynamics.swift
```diff
@@ -0,0 +1,26 @@
...
+
+    internal var elements: [String: Any]
```
nerdsupremacist: I would add a generic parameter here and constrain it to dynamic property.
Since the dynamic property protocol has no associated types worst case the developer can use `@Dynamics var dynamics: [String : DynamicProperty]`
 
Will make it harder to make a mistake and add something that can't be injected in here:

```swift
public struct Dynamics<Element : DynamicProperty> : Collection {
    var elements: [String : Element]
}
```

Sources/Apodini/Request/Dynamics.swift
```diff
@@ -0,0 +1,26 @@
...
+    
+    init(_ elements: (name: String, Any)...) {
```
nerdsupremacist: I'd add an initializer from a dictionary, and conform to `ExpressibleByDictionaryLiteral`

Sources/Apodini/Request/Dynamics.swift
theMomax: I also thought about adding a generic, but constraining it to `DynamicProperty` makes no sense to me as e.g. `@Parameter` is no `DynamicProperty`.

Sources/Apodini/Request/Dynamics.swift
theMomax: done

Sources/Apodini/Request/Dynamics.swift
theMomax: I think it makes sense to make the `Dynamics`' contents accessible as a `Collection`, however, a user is probably only interested in elements of a specific type. Thus I wouldn't make `Dynamics` itself a collection but provide a generic computed property/function that returns a `Collection` that only contains the elements which conform to that generic type.

Sources/Apodini/Request/DynamicProperty.swift
```diff
@@ -0,0 +1,214 @@
...
+/// `DynamicProperty` allows for wrapping Apodini's property wrappers while maintianing their functionality. By
+/// conforming a `struct` to `DynamicProperty` you this `struct`'s properties discoverable to the Apodini
+/// runtime framework. This can be used to e.g. compine two property wrappers provided by the Apodini framework
+/// into one that combines their functionality
+/// - Precondition: Only structs can be a DynamicProperty
+public protocol DynamicProperty { }
+
+
+// MARK: Execute
+func execute<Element, Target>(_ operation: (Target, _ name: String) -> Void, on element: Element) {
```
PSchmiedmayer: I would suggest to move all global functions into the `RequestInjectable` or `Injectable` protocol and would suggest to constraint all Injection functionality at one point. I think it is highly beneficial for the whole project if all injection work stays within one point in the project and the shared semantic model builds on top of that functionality. I would suspect that having these methods on a global level could lead to alternative versions of `RequestInjectable` or `Injectable` and therefore lead to more complicated handling in the shared semantic model that should serve as the basis for all the work the different exporters build on top of.

In addition I don't see any usage of `execute` in the current implementation and only see the usage of apply when calling `inject` on a RequestInjectable. Therefore I would suggest to move all the functionality in the inject method of a `RequestInjectable` or `Injectable` protocol. I think this would make it clear for the project that all injectable content must use `RequestInjectable` or `Injectable` and as far as I can see we don't have a specific use case for an alternative for `RequestInjectable` or `Injectable` at the moment, especially after the work that @tgymnich does to decouple `RequestInjectable` or `Injectable` from Vapor.

Sources/Apodini/Request/DynamicProperty.swift
```diff
@@ -0,0 +1,214 @@
...
+}
+
+func apply<Element, Target>(_ mutation: (inout Target) -> Void, to element: inout Element) {
+    apply({(target: inout Target, _: String) in
+        mutation(&target)
+    }, to: &element)
+}
+
+// MARK: DynamicProperty Implementation
+private extension DynamicProperty {
```
PSchmiedmayer: There is a high code duplication between this implementation as a private extension of `DynamicProperty` and the global function that I would suggest to reduce. I would propose to move all the functionality in a single `RequestInjectable` or `Injectable` protocol extension as detailed above and it might even be reasonable to move it to an `inject` function that can be used by the semantic model.
The same code duplication could also be seen in the `apply` function.

Sources/Apodini/Request/Dynamics.swift
```diff
@@ -0,0 +1,100 @@
...
+//
+
+import Foundation
+
+/// Dynamics provides same functionality as `DynamicProperty`, but for elements that are only known
+/// at statup and not compile-time. That is, `Dynamics` stores named elements and makes them
+/// discoverable to the Apodini runtime framework.  This can be used to e.g. delay the decision, which
+/// `Parameter`s are exported to startup-time
+@dynamicMemberLookup
+public struct Dynamics<Element> {
```
PSchmiedmayer: Similar to the comment made by @nerdsupremacist I would suggest to constraint `Dynamics` or `Injectables` to a base protocol that all injectable elements conform to, such as `RequestInjectable` or a more generalized `Injectable` implementation.
I think that `Injectables` as proposed by @hendesi would fit better to the naming of `RequestInjectable` or `Injectable` and therefore would suggest to rename `Dynamics` to `Injectables`. I think this more clearly details the functionality of injecting multiple different elements conforming to `RequestInjectable` or `Injectable`.

Sources/Apodini/Request/DynamicProperty.swift
```diff
@@ -0,0 +1,214 @@
...
+        } catch {
+            fatalError("Applying operation on all properties of \((try? typeInfo(of: Target.self))?.name ?? "Unknown Type") on DynamicProperty \((try? typeInfo(of: Self.self))?.name ?? "Unknown Type") failed.")
+        }
+    }
+}
+
+
+// MARK: Dynamics Implemenation
+
+private protocol Traversible {
```
PSchmiedmayer: I would suggest to move these requirements to `Injectable` or `RequestInjectable` as one constraint that must be implemented by a type that can be `Injectable` or hide the functionality in a single `inject` method that does all the work for all types conforming to `Injectable` or `RequestInjectable`.  `DynamicProperty` or `DynamicParameter` would be an extension of that functionality.

Sources/Apodini/Request/DynamicProperty.swift
```diff
@@ -0,0 +1,214 @@
...
+    func execute<Target>(_ operation: (Target, _ name: String) -> Void)
+    
+    mutating func apply<Target>(_ mutation: (inout Target, _ name: String) -> Void)
+}
+
+extension Dynamics: Traversible {
+    func execute<Target>(_ operation: (Target, _ name: String) -> Void) {
+        for (name, element) in self.elements {
```
PSchmiedmayer: As `Dynamics` conforms to collection you can remove `.element`
```suggestion
        for (name, element) in self {
```

Sources/Apodini/Request/DynamicProperty.swift
```diff
@@ -0,0 +1,214 @@
...
+            
+                dynamics.execute(operation)
+            default:
+                break
+            }
+        }
+    }
+    
+    mutating func apply<Target>(_ mutation: (inout Target, _ name: String) -> Void) {
+        for (name, element) in self.elements {
```
PSchmiedmayer:
```suggestion
        for (name, element) in self {
```

Sources/Apodini/Request/Dynamics.swift
```diff
@@ -0,0 +1,100 @@
...
+    public subscript(position: Dictionary<String, Element>.Index) -> (String, Element) {
+        self.elements[position]
+    }
+}
+
+public extension Dynamics {
+    /// This function provides a  copy of the `Dynamics` which only contains elements that
+    /// conform to type `T`
+    /// - Complexity: O(n)
+    func typed<T>(_ type: T.Type) -> Dynamics<T> {
```
PSchmiedmayer: I think you can remove the second implementation below by adding a default argument to this function here:
```suggestion
    func typed<T>(_ type: T.Type = T.self) -> Dynamics<T> {
```

Sources/TestWebService/main.swift
```diff
@@ -42,15 +42,22 @@ struct TestWebService: Apodini.WebService {
     }
     
     struct Greeter: Component {
```
PSchmiedmayer, line 44: I would suggest to add a test case that showcases how `Injectables`  or `Dynamics` can be dynamically created in the initializer as shown on slide 7 in @hendesi 's slides: https://confluence.ase.in.tum.de/display/SERVERSIDESWIFT2021/Intermediate+Presentation?preview=/66506934/66513049/Intermediate_Presentation_Felix_Desiderato.pdf

Sources/Apodini/Request/Dynamics.swift
PSchmiedmayer: I think it would result in a nice interface if `Dynamics` or `Injectables` would also work as a property wrapper as shown in the slides of @hendesi in the intermediate presentation: https://confluence.ase.in.tum.de/display/SERVERSIDESWIFT2021/Intermediate+Presentation?preview=/66506934/66513049/Intermediate_Presentation_Felix_Desiderato.pdf
I think the syntax would offer a very nice abstraction and the developer using Apodini could simply work on a dictionary of the type `[String: Injectable]` or `[String: RequestInjectable]` or even a more concrete type if the developer e.g. only uses `@Parameter`s and could therefore use a dictionary of that type `[String: Parameter]`.

Sources/Apodini/Request/DynamicProperty.swift
```diff
@@ -0,0 +1,214 @@
+//
```
PSchmiedmayer: While I see the point of marking a type as discoverable by Apodini I would suggest to focus this implementation on what we currently name `RequestInjectable` and extend the functionality of `RequestInjectable` with the functionality found in `execute` and `apply`. Maybe even a more general term like `Injectable` might be more suiting as we might sometimes also inject information from a connection state (`@Environment(\.connection)`) or from the global environment. I think this would also solve some of the comments and discussions below about constraining `Dynamics` or `Injectables` by a specific protocol: https://github.com/Apodini/Apodini/pull/56#discussion_r543137013

Sources/Apodini/Request/DynamicProperty.swift
```diff
@@ -0,0 +1,214 @@
...
+
+import Foundation
+import Runtime
+
+/// `DynamicProperty` allows for wrapping Apodini's property wrappers while maintianing their functionality. By
+/// conforming a `struct` to `DynamicProperty` you this `struct`'s properties discoverable to the Apodini
+/// runtime framework. This can be used to e.g. compine two property wrappers provided by the Apodini framework
+/// into one that combines their functionality
+/// - Precondition: Only structs can be a DynamicProperty
+public protocol DynamicProperty { }
```
PSchmiedmayer: If we have to keep this protocol around I would suggest to rename it to `DynamicParameter`.  While it might use some other property wrappers such as `@Environment` or `@ObservedObjects` I see that it behaves like a complex `@Parameter` for the implementations that developers using Apodini write. As we do not have the concept of a property in Apodini but `Parameter`s I would think that this would be a more descriptive name.
Alternatively I would image that `@Parameter` conforms to `Injectable` which is public and we use an internal protocol `_Injectable` to implement the inject functionality for the internal implementations that also include `execute` and `apply`.

PSchmiedmayer (changes requested):
@theMomax Thanks for the implementation showcasing how to implement the dynamic injecting of different elements into a `Handler`! You have already provided a comprehensive logic on how these dynamic injections can work. In general my review suggest to move the different functionality into a modified `RequestInjectable` protocol e.g.: `_Injectable` and expose a public protocol `Injectable` that `@Parameter`s conform to and allow you to constraint `Dynamics` or `Injectables`. In addition I added a few comments to make the naming a bit clearer and some smaller suggestions on a code level to improve the implementation a bit. I am looking forward to your thoughts about the comments and suggestions! :)
I am looking forward to seeing this feature in Apodini and have a great feeling that this will enable a lot of cool functionality for Apodini developers and elements in ApodiniFoundation such as the work by @hendesi !

In addition to the code changes I would also suggest to add a documentation about the new elements and functionality in the documentation folder that showcases some examples and how to use these features once we agreed on a sable API as part of this PR.

I would also be interested in the input from @hendesi by comparing this implementation to the ideas you have presented in the intermediate presentation that would also fit nicely in the ideas proposed here and @Supereg, @moritzschuell, and @tgymnich regarding the thought about `Injectable` or `RequestInjectable` and how these concepts would fit into the general architecture if injecting content into the `Handler`s of an Apodini project.

Sources/Apodini/Request/DynamicProperty.swift
theMomax: True, that was an artifact from the initial version (where `Dynamics` was no `Collection`) I forgot to remove.

Sources/Apodini/Request/Dynamics.swift
theMomax: Yes, that works. Thanks for the hint!

Sources/TestWebService/main.swift
theMomax: I cannot do it exactly as suggested by @hendesi. I think the `Dynamics`' `wrappedValue` should be get-only, as its properties are to be fixed at startup. The closest I can get is the following (with the only difference being the underscore in `self._dynamics`):

```swift
struct HandlingComponent: Component {
    @Dynamics var dynamics: [String: Param<String>]
        
    init() {
        self._dynamics = ["name": Param<String>()]
    }
        
    func handle() -> String {
        dynamics["name"]
    }
}
```

Sources/Apodini/Request/Dynamics.swift
theMomax: I made `Dynamics` an `@propertyWrapper` with get-only `wrappedValue` that exposes the `Dynamics`' `elements`.

> the developer using Apodini could simply work on a dictionary of the type [String: Injectable] or [String: RequestInjectable]

I do not get why you think that would be useful or even possible. `RequestInjectable` is an internal protocol and provides no interface to the user of Apodini. I also don't get what a user of Apodini should do with an `inject` function on those elements if the more abstracted `Injectable` protocol were to be made public.

Sources/Apodini/Request/DynamicProperty.swift
theMomax: I heavily disagree. `RequestInjectable` is by far not the only use-case for `apply`. I use different protocols within my implementation of the WebSocket exporter and also the shared semantic model builder will use the execute function to find all `@Parameter`s at startup (it currently uses `Mirror` without support for `DynamicProperty` or `Dynamics`).

There will be way more use-cases (that do not fit a `Injectable` protocol) once we start working on `ObservableObject` and similar topics. My goal here was to provide abstracted **infrastructure-code** that leverages most of the work from working with protocols like `RequestInjectable`. As you said, this PR is all about enabling the user to **mark a type as discoverable by Apodini** and help the Apodini-internal developers access those types.

As of right now I don't see what the advantage of an `Injectable` protocol would be. Please help me see it. From my point of view `Injectable` does **not** cover all use-cases but the user of Apodini **should be able to** combine different use-cases (e.g. combine `@State` and `@Parameter` or `@Parameter` and `@ObservedObject`).

Sources/Apodini/Request/DynamicProperty.swift
theMomax: Again: `DynamicProperty` has nothing to do with `@Parameter` in the first place. A user of Apodini could combine a `@Environment(\.database)` and an `@ObservedObject` (publishing user-ids)  into a new property wrapper which publishes the complete user-object as fetched from the database.

Sources/Apodini/Request/DynamicProperty.swift
theMomax: I would also like to have a bit more abstraction inside `DynamicsProperty.swift`, but I would not combine this with `RequestInjectable `. `RequestInjectable ` is a `Target`, not the infrastructure, not `Traversible`. Where would e.g. `Parameter` (which is the primary `RequestInjectable`) find the `Target`?

Sources/Apodini/Request/Dynamics.swift
theMomax: In my opinion, `Dynamics` as well as `DynamicProperty` code for making things discoverable to the Apodini framework.

I don't want to constrain Apodini to only be able to discover `RequestInjectable`s (or a more generic `Injectable`s). The only thing I could imagine is constraining it to a `public protocol Discoverable { }` with no requirements that is implemented by all of Apodini's property wrappers.

However, a `DynamicProperty` can also contain properties of any type (and there is nothing we can do about that), so why should we constrain `Dynamics`. It doesn't make any difference for us because we will only search for elements of those types we are interested in. It only gets more complicated for the programmer using Apodini, because they have to use `Dynamics<Any>` and must use type-assertions when using the elements.

Sources/Apodini/Request/DynamicProperty.swift
theMomax: I agree on the part about the code-duplication. I tried to reduce this a couple of times, but failed because of corrupted memory access. I would be happy to address this problem in a pair-programming session with you or @nerdsupremacist.

Sources/Apodini/Request/DynamicProperty.swift
theMomax: I replaced the remaining usage of `Mirror` in `RequestInjectable.swift` by usage of `execute`. I advise against moving/merging this PR's functionality with `RequestInjectable` or `Injectable` because of one simple reason: I don't think `RequestInjectable` will be able to cover all use-cases on the long run.

I think the shared semantic model is already quite constraining right now. Especially when it comes to downgraded functionality because of constraints from the underlying protocols and the communicational patterns they support in the future, I think some exporters will need more flexibility. This flexibility will most likely include extending Apodini's property wrapper with exporter-specific protocols.

Maybe this can even be unified inside the shared semantic model, using one or two additional protocols, but then we still need `apply` and `execute` as stand-alone functions that are not bound to `RequestInjectable`.

PSchmiedmayer: Thanks for showcasing how `execute` can be used in the shared semantic model!

I see your point that `RequestInjectable` might not be sufficient for all use cases, especially `ObservedObjects` in the long run and we might need to add additional protocols that do not work based on injecting content from a request into a property wrapper as currently implemented with `RequestInjectable`. This was the idea of `Injectable` as described in the response of a thread above (https://github.com/Apodini/Apodini/pull/56/files#r546223763).

While I agree that the shared semantic model might not cover all edge cases for different exporters I think that it is one of the highest priorities of the current development project to move it into that direction. I think that it is critical for the long term maintainability of the project to move as much common functionality into parsing the Syntax Tree and generating a shared semantic model as possible. The past few weeks and presentations have shown me that a lot of code duplication and redundant work is happening if every exporter tries to find a strategy to deal with common problems on their own. Identifying, collecting, and injecting is a common functionality that will be shared across all exporters even though some exports might make different decisions based on different "Injectables". This is the motivation and idea that I tried to express with `Injectable` and I have the feeling that I didn't do a great job explaining it in the previous comments, sorry about that.

After reading through your comments and your completely valid points I would suggest to have a feeling that moving these `execute` and `apply` into elements of the shared semantic model might also make sense even though I would still also see them defined in extensions of `Injectable` as this could solve the code duplication problem we see in a thread below (https://github.com/Apodini/Apodini/pull/56/files#r546220170). I do greatly value all your ideas and additions to the parsing of the Apodini property wrappers and would like to encourage you that all the exporters can immediately benefit from your work by focusing on bringing the semantic model to a point where it can deal with the different constraints and can formulate a shared basis for all exporters, including your Web Socket exporter. I took a look at your branch and I have seen very promising solutions and ideas over there and it would be highly beneficial for the project if everybody your benefit from a common basis where your solutions also play a major role in instead of diverting the web socket exporter more and more from all other exporters. I hope I have made a better job in stating my intentions behind the different comments in this PR :)

Sources/TestWebService/main.swift
PSchmiedmayer: I like this solution of using the dictionary literal to initialize the `@Dynamics`/`@Injectables` property wrapper! I think this is a good solution and still makes the `@Dynamics`/`@Injectables` read only which is a good idea 👍

Sources/Apodini/Request/DynamicProperty.swift
PSchmiedmayer: As pointed out in this thread I have the feeling that I could have been more clear with my idea of `Injectable` and this might have lead to the heavy disagreement from your side, I am sorry about not describing by idea more clearly.

I don't suggest that an `Injectable` protocol should expose any public information to the developer that uses Apodini. Similar to your `DynamicProperty` protocol (and going beyond that) I would see `Injectable` as a base protocol that `@Parameter`, `@Environment`, `@ObservedObject`, `@Dynamics`/`@Injectables` and all custom property wrappers must conform to. 
This protocol should offer offer an user of Apodini a clear understanding of what could be put into `@Dynamics`/`@Injectables` by default and provides a good overview of what should be inspected by the shared semantic model to inject information and what other types should simply be ignored in the identification, collection, and injection process performed by the shared semantic model. `@RequestInjectable` would group some property wrappers that might require information that is injected based on a request while other property wrappers such as `@State`, `@ObservedObject` or others also conform to `@Injectable` but do not contain the same functionality that might be required by `RequestInjectable`. 
E.g. I tried to motivate this possibility by the `@Environment(\.connection)` property wrapper example that might benefit from `RequestInjectable` but could also use some other mechanisms that go beyond that and might be defined in some other internal protocols that further refine `Injectable` but can build on top of its internal functions that are defined in extensions. 

This thinking was also the reason why I proposed to move the internal `execute` and `apply` functions into extension of `Injectable` to make them available to all possible property wrappers that a shared semantic model and exporters building on top of the shared semantic model can rely on. As pointed out in the reply to a thread below (https://github.com/Apodini/Apodini/pull/56/files#r546219973) I could also imagine moving them right into the shared semantic model but as far as I can see having them as extension of `Injectable` could also solve the code duplication problem described in a thread below (https://github.com/Apodini/Apodini/pull/56/files#r546220170).

Sources/Apodini/Request/DynamicProperty.swift
PSchmiedmayer: Thanks for pointing this out. I agree with you that elements such as `@Environment` and `@ObservedObject` should be part of such a construct which I have pointed out above. I see your point that `DynamicParameter` might not be the best solution to convey such a complex setup and therefore might not be easy to understand. Thinking through the different comments you have made and while trying to structure my thoughts around the different approaches and protocols I have the feeling we could simply replace this protocol by an `Injectable` protocol as described in a thread above (https://github.com/Apodini/Apodini/pull/56/files#r546223303) or a protocol such as `public protocol DynamicInjectable: Injectable{}` in case we need to differentiate between an `Injectable` or an `DynamicInjectable ` but as far as I can see this differentiation might not be needed base don your implementations so far.

Sources/Apodini/Request/DynamicProperty.swift
PSchmiedmayer: I tried to clarify a few of my ideas around `Injectable` in the two threads above (https://github.com/Apodini/Apodini/pull/56/files#r546223303) and (https://github.com/Apodini/Apodini/pull/56/files#r546223763). If such a protocol would contain these functions as internal extension this could also eliminate the code duplication problem that we face here right now. I would suspect that the global functions won't be needed any more and everything could just be part of an `Injectable` extension.

Sources/Apodini/Request/Dynamics.swift
PSchmiedmayer: As detailed in my response to a thread above I think that I did not do a good job explaining my intention behind the `Injectable` protocol. It's vision closely aligns with what you have described as `Discoverable` in the massage above. The response in the thread above (https://github.com/Apodini/Apodini/pull/56/files#r546223303) also aligns with your comment to make all property wrappers conform to that protocol.

Consequently I would also suggest to move to a more descriptive name such as `Injectables` to indicate that it is a collection of `Injectable`s that wrap around properties to inject specific information (e.g. `@Environment` from a global storage, `@State` from a state storage across different requests, `@ObservedObjects` from some global environment or state storage across different requests and mainly `@Parameters` that get information injected from requests). Constraining the content of `@Injectables` to this protocol makes it clear that we expect the developer to add any existing types that can act as property wrappers and e.g. `@PathParemter` that should not conform to `Injectable` or a simple `Int` or `String` that would need to be wrapped by an `@Parameter` property wrapper or is part of a dictionary literal expressing it as `Parameter<Int>()`.

My and I suspect also @nerdsupremacist 's intention behind this move was to use to type system to hint to the developer using Apodini what the developer can use as part of the `@Injectables`/`@Dynamics` property wrapper.   Some custom `Injectable`s/`Dynamic`s can of course have properties that are not marked with an Apodini or custom property wrapper but the enclosing type must conform to `Injectable`.


Regarding your last comment about type erasure using Any: I don't think it will be necessary to make `Injectables`/`Dynamics` have a generic type and just use the `Injectable` protocol in the type definitions:
```suggestion
public struct Injectables {
    internal var elements: [String: Injectables]
```
You could have a typed version like `TypedInjectables<Element: Injectable>` that provides you a concretely typed version of an `Injectables` instance when calling the `typed` method defined below. This would also enable you to reduce the need of `Any` in your test cases as far as I can see the implications on the examples in the test cases below.

PSchmiedmayer: Thanks a lot, I think this offers a nice API as already mentioned in my comment in the thread below (https://github.com/Apodini/Apodini/pull/56/files#r546222581)!
I tried to showcase why this would be useful and what the reason behind the `Injectable` protocol is in the comment thread above (https://github.com/Apodini/Apodini/pull/56/files#r546220357) and the initial intention behind `Injectable` and the motivation why I don't think that there is a need for a public `inject(...)` method in the thread above (https://github.com/Apodini/Apodini/pull/56/files#r546223303). I hope this clears up my motivation on why I would suggest to use an `[String: Injectable]` dictionary in `Injectables`/`Dynamics`. I agree with you that constraining it to `[String: RequestInjectable]` would not be useful as `RequestInjectable` is and should stay internal.

PSchmiedmayer (commented):
@theMomax Thanks for all the comments and inputs regarding my comments on this PR. I tried to clarify a few of the intentions behind my initial comments as I could have been more clearer with some of the ideas behind `Injectable` that you strongly disagreed with. I hope the comments showcase the how my ideas and suggestions could improve the PR and how close some of your ideas to what I have proposed. Let me know what you think and I am looking forward to your comments and how the different aspects of the PR can bring the project, the shared semantic model and all exporters forward. 👍

theMomax:
Thanks a lot @PSchmiedmayer for your clarification on the `Injectable` protocol! I think with this new input we made a big step towards a solution👍🏻. My strong disagreement on the `Injectable` protocol was based on the naming and the lack of explanation. I thought this would be more related to `RequestInjectable` than the `Discoverable` I came up with in the comment.

I'm trying to streamline this discussion a bit with this comment summing up the rather messy discussion above:

1. We agree that the generic of `Dynamics<Element>` (or whatever name it will have in the end) should be constrained to an empty protocol (`Injectable`/`Discoverable`) which is implemented by `@Parameter`, `@State`, `@ObservableObject` and `@Environment`.
2. We agree that the code-duplication in `DynamicProperty.swift` (`apply` and `execute`) is too high at the moment. We will try to solve this problem before merging the PR.
3. We still have to come to an agreement regarding the naming of what is currently `Dynamics`, `Injectable`/`Discoverable` and possibly also `DynamicProperty`.
4. Partly dependent on the naming, the location of the code of `apply` and `execute` may have to be changed.

theMomax:
Regarding point 3:

I like the term **dynamic** as it gets to the gist what all these elements have in common: they are dynamically managed by the Apodini framework, i.e. they provide functionality that is not statically expressed in code on the handling component/handler. The protocol `DynamicProperty` also exists in SwiftUI with equal functionality.

**Injection**, however, sounds to me like it is only one use-case of dynamically handling an object. I do not see in what sense an `@ObservedObject` that is provided with a default initializer would be _injected_. It is found and subscribed to, but that is no injection.

Thus I would suggest the following naming:
* `DynamicProperty` remains the same
* `Dynamics` either remains as is, or becomes `DynamicCollection` (which I would prefer as it is a bit more expressive and is less disruptive in spoken language)
* The new, empty protocol (`Injectable`/`Discoverable`) is called `Dynamic` or `DynamicallyManaged`

Please let me know what you think! I'd also be open for `Injectable` if you have convincing arguments (against my objections regarding use-cases and SwiftUI-similarity or **for** the term **Injection**)!

theMomax:
Regarding point 4:

I think @PSchmiedmayer you mentioned moving the `apply` and `execute` globals into an extension to `Injectable` (or did I get something wrong)?

I don't quite understand how this would work as the base-element `apply` or `execute` has to be called on is the handling component/handler, which is no `Injectable`/`Dynamic`. Would you suggest `Component` or the future `Handler` should be `Injectable`/`Dynamic`? I see how this could be justified (as `Component`s are indeed dynamically managed by Apodini). Would this also mean `DynamicProperty` and `Dynamic`/`Injectable` would be merged? This would open the door for having multiple layers of handlers (and thus logic). However, I don't know if we want to get started with that quite yet. At least I can't say I know right now what consequences might arise form this.

Maybe I just misunderstood...or I'm just missing something obvious here😄.

hendesi (commented):
@theMomax Good job! I implemented a first draft of the query builder using `@Dynamics` and it seems to work quite well, beside some issues that are not really related to the implementation of `@Dynamics`. I will propose my solution in a different pull request, but speaking for my use case, your implementation really helped a lot. Thanks! 👍 

If you want to have a look: [feature/database-components](https://github.com/Apodini/Apodini/commit/caa6cebf49b2a942dc07ff108c171c6757a66320)

PSchmiedmayer:
@theMomax Thanks for the summary of the various discussions we had in the review threads. And it is great to see that the ideas of this proposal already provide some benefit to @hendesi's work 🚀

I agree with the points you have mentioned in [your comment](https://github.com/Apodini/Apodini/pull/56#issuecomment-748652488). In addition to the four points mentioned above I would add that we should focus in bringing as much related functionality as possible regarding this functionality into the existing shared semantic model and should try to focus on extending it to cover the different edge cases you mentioned above so all exporters building on top of it could immediately benefit from the additions made in this PR and future improvements regrind other property wrappers as well as advancements to `Component`s and `Handler`s that we will introduce.

Regarding Point 3 and [your detailed comment regarding point 3](https://github.com/Apodini/Apodini/pull/56#issuecomment-748664787):
After reading through your comments I agree that `DynamicProperty` is a suiting name for the functionality. I must be honest that I am still not really a fan of the notion of `Dynamic` and `Dynamics` even through I agree that `DynamicCollection` would be a better wording for `Dynamics`. I agree that the notion of `injecting` a `ObservedObject` that is initialized In a `Handler` could not really be considered injection in its pure meaning. On the other hand I don't really see how `Dynamic` or `DynamicallyManaged` really describes `@State`, `@Parameters`, or `@Environment` that do not have a dynamic behavior but just get injected once based on a request or stay constant across different executions as some properties annotated with `@Environment` might do.
While thinking about different names, I stumbled upon naming one name that could describe the notion of all different property wrappers could just be `Property`. I think this closely describes the notion of what is summarized under the `Property` protocol. A collection of types conforming to `Property` could just be named `Properties` and be annotated with `@Properties`. And `DynamicProperty` then perfectly describes a dynamic composition of different types conforming to `Property` in a new struct. What do you think about this idea?

Regarding Point 4 and [your detailed comment regarding point 4](https://github.com/Apodini/Apodini/pull/56#issuecomment-748666266): 
You are right that I proposed to either make `apply` or `execute` available in extensions of `Injectable`/`Dynamic`/`Property` or move it into the shared semantic model. While reading though your ideas and comments I have realized that it might indeed bring little benefit to reduce the functionality of `apply` and `execute` and only provide it as a means to deal with `Injectable`/`Dynamic`/`Property` as this protocol doesn't yet offer us the abstractions that `inject` provides us in `RequestInjectable`. I nevertheless would still consider moving them out of the global context and into the area of the shared semantic model a beneficial move as this would motivate all exporters to implement this injection/or-what-might-be-relevant-for-other-property-wrappers-functionality in a single place where all exporters could benefit form edge cases discovered and fixed in the shared semantic model. Would you consider this a desirable approach? 
I am not sure it it is beneficial at the moment to extend `Handler` and `Component` to conform to `Injectable`/`Dynamic`/`Property` but this might be a nice way to experiment with the stacked `Handler` idea and if this is already build in out of the box in the shared semantic model that could handle the logic of filling all property wrappers with the desired content, this functionality could be provided to all exporters for free.

theMomax:
@PSchmiedmayer thank you very much for you detailed response! I think we're almost there!

I really like the name `@Properties`! So this is our final terminology then:
* `DynamicProperty` remains as is
* `Dynamics`/`Injectables` becomes `Properties`
* All of Apodini's property-wrappers used on `Handler`s/handling `Component`s shall implement the empty protocol `Property`

I will change the names accordingly. I will also make `DynamicProperty` conform to `Property`, so that structs implementing `DynamicProperty` can automatically be used in `Properties`.

We agree that we don't want to officially support nested-`Handler`s yet. Thus we do not make `Component`/`Handler` a `DynamicProperty` by default.

I will move the whole code closer to the shared semantic model and hide it from the rest of the Apodini package.

PSchmiedmayer:
@theMomax Thanks a lot, this sounds great!!

PSchmiedmayer (commented):
@theMomax The PR looks very good to me! Thanks for all the work that went into introducing the feature to Apodini.
The only element that I have found was that `Properties` were not constrained to only include types conforming to `Property`. I experimented a bit with this and pushed my suggestions to the following branch: https://github.com/Apodini/Apodini/compare/feature/dynamic-property-and-dynamics...feature/dynamic-property-and-dynamics-pauls-suggestions
I think this would make it even clearer where types conforming to `Property` should be used and it would reduce the need for `Any` on most of the cases where `Properties` are used. In addition I introduced a way to access `Properties` using the projected value and added a special type that is able to reflect specific type requirements. Let me know what you think. I pushed the implementation to the branch to reduce the number of comments in the PR and to make it easier to integrate it in the branch if you agree. Feel free to merge the branch into your branch to include the changes in the PR. Apart from this I really like the solutions we came up with and would suggest to merge the branch after getting your input on the changes in the. https://github.com/Apodini/Apodini/tree/feature/dynamic-property-and-dynamics-pauls-suggestions branch 👍

theMomax:
@PSchmiedmayer your suggestion looks quite well! There is one thing I don't get though: What is the advantage of having a `projectedValue` that returns `self`? What would be the difference to using the underscore (`_variableWrappedByProperties`)?

PSchmiedmayer:
@theMomax You are right, that's a good point. I forgot about the possibility to access a property wrapper using an underscore, as using the projected value is so heavily used in SwiftUI. This is also a feasible approach and I have removed the `projectedValue` from `Properties` in the https://github.com/Apodini/Apodini/tree/feature/dynamic-property-and-dynamics-pauls-suggestions branch.

theMomax:
Ok...let's get this merged then🎉

PSchmiedmayer (approved):
Looks great to me, let's merge it 🚀

PR #57, opened 2020-12-12
Fix: RequestInjectables wouldn't be properly injected when using SharedSemanticModelBuilder

bauer-andreas:
As discovered by @theMomax and @hendesi the request handler of the `RESTInterfaceExporter` had an issue with RequestInjectables not being injected at all. The `Component` instance wasn't made available in the shared model, and thus the request handler passed the endpoint instance to `enterRequestContext`.

This was now changed so that the SharedSemanticModelBuilder provides the implementation for the request handler and thus is able to pass the `Component` instance to `enterRequestContext` to restore functionality.

### DIFF
--- a/Sources/Apodini/Semantic Model Builder/Context.swift
+++ b/Sources/Apodini/Semantic Model Builder/Context.swift
@@ -20,7 +20,8 @@ class Context {
     func get<C: ContextKey>(valueFor contextKey: C.Type = C.self) -> C.Value {
         contextNode.getContextValue(for: contextKey)
     }
-    
+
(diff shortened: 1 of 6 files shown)


theMomax (approved):
The changes regarding functionality of RequestInjectables look very good and seem to work. However, I have little insight on how your changes could influence the usability of the shared semantic model for other exporters but the one for REST. Since this is rather urgent, I still approve. You may decide on your own if you prefer to have a second reviewer who is more familiar with this aspect of the PR.

hendesi:
Solved the injection issue for me. Thanks for the quick fix! 👍

PR #59, opened 2020-12-13
Corrected the contentType of EndpointParameter to match what is required by the Parameter generic

bauer-andreas:
Minor fix. The headline describes it. `@Parameter` is defined as `Parameter<Element: Codable>`. This is now reflected in the `EndpointParameter` model as well.

This might be relevant to solve parameter retrieval in #32 (my [thoughts](https://github.com/Apodini/Apodini/pull/32#discussion_r541814050))

EDIT: requesting review from @lschlesinger as she is the code owner for that


lschlesinger (approved):
👍

PR #60, opened 2020-12-13
Retain HandleReturnType for Endpoint

tgymnich:
I went ahead and added a generic type parameter to `Endpoint` in order to retain the type information gathered from `Component`. Further I've added `AnyEndpoint` as a type erased endpoint.

### DIFF
--- a/Sources/Apodini/Semantic Model Builder/EndpointsTree.swift
+++ b/Sources/Apodini/Semantic Model Builder/EndpointsTree.swift
@@ -19,8 +19,34 @@ struct EndpointRelationship { // ... to be replaced by a proper Relationship mod
     var destinationPath: [_PathComponent]
 }
 
+protocol AnyEndpoint {
+    /// Response type ultimately returned by `handle()` and possible following `ResponseTransformer`s
+    var responseType: ResponseEncodable.Type { get }
+    /// Type returned by `handle()`
+    var handleReturnType: ResponseEncodable.Type { get }
+
+    var description: String { get }
+
+    /// The reference to the Context instance should be removed in the "final" state of the semantic model.
+    /// I chose to include it for now as it makes the process of moving to a central semantic model easier,
+    /// as implementing exporters can for now extract their needed information from the context on their own
+    /// and can then pull in their requirements into the Semantic Model.
+    var context: Context { get }
+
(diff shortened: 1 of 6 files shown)


bauer-andreas:
Would you mind elaborating on why this change is needed? I don't really get what you are trying to do, as the type erasured `AnyEndpoint` is stored anyways in the endpoints dict, and the endpoint is also passed to the exporters with its type exerasured(?)

EDIT: maybe I just wait for your other issues to be created(?) ✌️

tgymnich:
@Supereg I have no concrete plan where to go from here. But having more type information available is always a great starting point.

I forgot to mention this a NFC (Non functional change)

Sources/Apodini/Semantic Model Builder/EndpointsTree.swift
```diff
@@ -37,11 +63,12 @@ struct Endpoint {
...
     /// Type returned by `handle()`
-    let handleReturnType: ResponseEncodable.Type
+    var handleReturnType: ResponseEncodable.Type { HandleReturnType.self }
     /// Response type ultimately returned by `handle()` and possible following `ResponseTransformer`s
-    var responseType: ResponseEncodable.Type
-    
+    let responseType: ResponseEncodable.Type
+
     /// All `@Parameter` `RequestInjectable`s that are used inside handling `Component`
     var parameters: [EndpointParameter]
```
lukaskollmer, line 73: is there a reason why this is mutable?

lukaskollmer (approved):

Sources/Apodini/Semantic Model Builder/EndpointsTree.swift
tgymnich: Good catch. There is no reason for this to be mutable

bauer-andreas:
> @Supereg I have no concrete plan where to go from here. But having more type information available is always a great starting point.

If I'm being honest, I don't see the point that we should merge a change, which adds something we currently don't use and for which we don't know if we will (be able to) ever use it? I would love to hear an argument why this makes sense, as it might just be me not being able to see the bigger picture you have in mind. But even then I would argue to include it in that PR which needs that change?
But feel free to merge if that's benefits your future work 👍

PR #61, opened 2020-12-13
Replace ResponseEncodable with Encodable

tgymnich:
I replaced `ResponseEncodable` with `Encodable` to become more independent of Vapor


Sources/Apodini/Semantic Model Builder/SharedSemanticModelBuilder.swift
```diff
@@ -106,27 +106,29 @@ class SharedSemanticModelBuilder: SemanticModelBuilder {
...
+                            request.enterRequestContext(with: component, using: coder) { component in
+                                var response: Encodable = component.handle()
                                 for responseTransformer in responseModifiers {
-                                    response = request.enterRequestContext(with: responseTransformer(), using: decoder) { responseTransformer in
+                                    response = request.enterRequestContext(with: responseTransformer(), using: coder) { responseTransformer in
                                         responseTransformer.transform(response: response)
                                     }
                                 }
-                                return response.encodeResponse(for: request)
+                                return try! coder.encode(AnyEncodable(value: response), request: request)
```
bauer-andreas: I kind of thought we already discussed that in the last work meeting. Do we need to do the encoding in the request handler? I think removing this step here and just leaving it as the exporters responsibility would make the most sense? Especially as you are anyways delegating the encoding process to the exporter via the `ResponseEncoder` interface.
The problem I see with the `ResponseEncoder` interface is that the information, where the response data belongs to, is lost. With this approach we can't do HATEOAS.

To illustrate my point:
I've build a experimental (and I really mean **experimental**) rest exporter over at the `rest` branch. Similar to this PR I also move to `Encodable` (as it was proposed by you in #32).
Here is an excerpt from the request handler hooked up to Vapor: 
https://github.com/Apodini/Apodini/blob/e54306dab10b52634c8aae4d1a9acb6048511d5f/Sources/Apodini/Semantic%20Model%20Builder/REST/EndpointHandler.swift#L50-L61

For HATEOS we need to alter the what is encoded into the response, and the information added is dependent on the current endpoint **and** request (injecting query parameters from the request etc...) handled.
Using the `ResponseEncoder` we loose that ability.

I hope I could better explain my points I made in the last work meeting with that code snippet. 


EDIT: the argument also connects to the points made here https://github.com/Apodini/Apodini/pull/32#discussion_r541814050. With that in mind: How the Request and the Response looks like, is Exporter business. And from the perspective, that Apodini should in theory be able to support any new interface exporters in the future, I really think handling encoding here is really restricting.

bauer-andreas (changes requested):
PR is fine except of the `ResponseEncoder`, I don't think that lines up with our requirements (see code comment).

@nerdsupremacist as you were previously against `Encodable`, would you mind sharing what you had in mind as an alternative?

Sources/Apodini/Semantic Model Builder/SharedSemanticModelBuilder.swift
tgymnich: This looks good so far. The only problem I see is, that we cannot make the `requestHandlerBuilder` generic (because its a closure).
```
fileprivate var requestHandlerBuilder: (RequestInjectableDecoder & ResponseEncoder) -> (Vapor.Request) -> EventLoopFuture<Vapor.Response>
```
Further when creating the `createRequestHandlerBuilder` we don't know anything about the `Exporter` yet, so we cannot pass the desired `Request` type.

bauer-andreas: > This looks good so far. The only problem I see is, that we cannot make the requestHandlerBuilder generic (because its a closure).

Why would we need a generic? currently they generic in the `encode` method isn't useful anyways, cause you always pass a `AnyEncodable` to `.encode`, so type information is lost always (and even makes it weird because nobody expects a wrapped type here, creating complications if one would use Reflection on the response.

> Further when creating the createRequestHandlerBuilder we don't know anything about the Exporter yet, so we cannot pass the desired Request type.

The exporter is passed as the first argument for the outer closure. Anyways what information do we need from the Request (request used here would be replaced with our own Apodini request type right?)?

tgymnich: Nevermind i can do everything we need before i call the handler

tgymnich:
@Supereg I had do disable some tests related to databases but everything else should be ready to go

Sources/Apodini/Request/Database.swift
```diff
@@ -28,7 +27,7 @@ struct _Database: RequestInjectable {
     init() { }
     
     
-    mutating func inject(using request: Vapor.Request, with decoder: RequestInjectableDecoder? = nil) throws {
-        self.database = request.db
+    mutating func inject(using request: Request) throws {
+        self.database = request.database
```
bauer-andreas: Nit: `request.database` is optional. Thus if the database is not present and the user would access is nonetheless it would encounter `fatalError("You can only access the database while you handle a request")`. I think this error message would be a bit misleading. Maybe we can check for this in the `inject`?

bauer-andreas (dismissed):
Aside from by nit pick, looks good to me.

What I currently don't like as much is, when delegating the parameter retrieval to the exporter, that the exporter only gets a UUID as a reference (`func parameterDecoder(for request: Vapor.Request) -> (UUID) -> Codable?`). In my vision the shared model would be able to do the lookup for the UUID and provide the exporter with the parameter model like `func parameterDecoder(parameter: EndpointPrameter, for request: Vapor.Request) -> Codable?`.

But I think that is out of the scope of this PR and I'm happy to get the migration to `Encodable` merged. 👍

bauer-andreas (approved):

Sources/Apodini/Request/Database.swift
tgymnich: done

lukaskollmer:
@tgymnich I'm not 100% sure yet but i think this PR introduced a regression the tests didn't catch: by replacing `ResponseEncodable` with `Encodable`, we lost Vapor's `EventLoopFuture: ResponseEncodable` extension, meaning our `handle()` functions can't return `EventLoopFuture`s anymore.

Relevant test case: [CustomComponentTest.swift](https://github.com/Apodini/Apodini/blob/0811646d1c87cb51b4817402a2228d2968e201fb/Tests/ApodiniTests/CustomComponentTest.swift#L16-L31).

The invalid return type caused the test component's `Response` type to default to `Never`, which we didn't catch because this specific test case directly invokes `handle()` instead of going via the witness table.

PSchmiedmayer:
@tgymnich @lukaskollmer Can you create an GitHub issue for this problem? I think that having EventLoopFutures as possible return types is very important, especially for database access and all operations that are more complex. We should also extend the test cases to cover this regression in the future! 👍

tgymnich:
@lukaskollmer nice catch 👍. I've created #72.

hendesi:
@tgymnich As @PSchmiedmayer already stated this completely breaks any database functionality I'm currently implementing as well as the changes of #70. I would highly appreciate it if you could tackle this asap. Please consider tagging me in upcoming vapor decoupling prs, so we can prevent this in the future. 👍

PR #66, opened 2020-12-18
Split up Component protocol

lukaskollmer:
[link to nicely rendered markdown text](https://github.com/Apodini/Apodini/blob/7adefd51fe0f5911a0113245dcecea3453009216/Documentation/Components/Component%20Types.md)


Documentation/replacing-component.md
```diff
@@ -0,0 +1,266 @@
+# Proposal: Replacing `Component`
```
PSchmiedmayer: I would suggest to phrase a second version of this document as "vision" or "documentation" (with the corresponding implementation in the PR) as previously described in a PR by @theMomax . I have added a small markdown file to give an overview of the different types of documentation we have in the documentation folder: https://github.com/Apodini/Apodini/blob/develop/Documentation/Documentation.md

Documentation/replacing-component.md
```diff
@@ -0,0 +1,266 @@
...
+
+
+- Author: Lukas Kollmer
+- Date: 2020-12-16
+- Draft Implementation: [experimental/dsl-node-protocols][branch_url]
+
+
+## Summary
+
+- We should replace `Component` with a collection of several, more fine-grained protocols, which match more closely to the different [kinds of components](https://github.com/Apodini/Apodini/tree/develop/Documentation/Communicational%20Patterns/3.%20Pattern%20Implementation) we eventually want to support
```
PSchmiedmayer: Based on the ideas as described in the "Communication Patters" section of the documentation we have decided to have a single protocol that reflects all different communication protocols and the exporters can decide how to export a specific `Handler` based on the different property wrappers. The differentiation between `Handler` and `Component` was described [here](https://github.com/Apodini/Apodini/pull/26#discussion_r538396896) and I would suggest to stick with these terms for now as we have already used them in our discussions and it seemed like the differentiation between these two ideas has already been manifested in most of the documentation.

Documentation/replacing-component.md
```diff
@@ -0,0 +1,266 @@
...
+- Instead of a single protocol, there should be a hierarchical structure of several protocols, similar to [how Swift models numeric types](https://files.lukaskollmer.de/embed/swift_numerics.png)
```
PSchmiedmayer: I think that this would be valuable but from looking at your ideas about deployment targets and our last meeting I would go for a more linear approach for now (e.g. `Component`/`EndpointProvidingNode ` -> `Handler`/`EndpointNode` -> `CallableHandler`/`CallableEndpointNode` or similar more fine grained solutions when talking about communication and references between `Handler`s).

Documentation/replacing-component.md
```diff
@@ -0,0 +1,266 @@
...
+- When defining a concrete component, the user would make that component conform to the protocol matching the component's intended use-case
+- We use the type system to ensure that components can only appear in places in the DSL where they actually make sense
+- For example, the leaves in the DSL's syntax tree must always be endpoints (ie components w/ a `handle()` function)
+- If possible, an ill-formed `WebService` shouldn't compile in the first place
```
PSchmiedmayer: This summarizes the problems in a nice way!
As far as I have seen we so far would differentiate between the static structure of a web service (`Component`s/`EndpointProvidingNode`s) and dynamic behavior at the leafs of the `Component`/`EndpointProvidingNode` tree: `Handler`/`EndpointNode`. I would suspect that there is doing to be additional functionality from your side that will be needed to enable the communication between different components for dynamic deployments.

Documentation/replacing-component.md
```diff
@@ -0,0 +1,266 @@
...
+}
+```
+
+
+This component, despite compiling just fine and not causing a single warning or error, will never work as intended.  
+The reason for this is that `User` doesn't conform to `ResponseEncodable`, meaning that swiftc will resolve `GetUser.Response` to `Never` (the default value), which will cause Apodini to invoke the `handle` function implemented in the `Response == Never` protocol extension.  
+The result is that, when a request is sent to the component's endpoint, it will crash at runtime.
+
+
+**Generally speaking, if we can use the type system to statically detect and reject ill-formed programs, instead of having to perform these checks at runtime, we should always do so.**
```
PSchmiedmayer: Agreed!

Documentation/replacing-component.md
```diff
@@ -0,0 +1,266 @@
...
+Consider the following example:
+
+```swift
+struct User: Codable {
+    let id: ID
+    let name: String
+    let age: Int
+}
+
+struct GetUser: Component {
```
PSchmiedmayer: This is an interesting edge case that I have not considered yet. I suspect this would also happen in this implementation for the `Handler`: https://github.com/Apodini/Apodini/pull/26#discussion_r538396896 ... so I suspect we would need to remove the `Never` default value?

Documentation/replacing-component.md
```diff
@@ -0,0 +1,266 @@
...
+}
+```
+
+- `EndpointNode`s are nodes which expose some functionality to the client.  
+  All leaves in the `WebService` must be `EndpointNode`s.
+- `EndpointProvidingNode`s are nodes which sit above the leaves.  
+  They cannot provide any functionality of their own, but they must eventually lead to one or more `EndpointNode`s.
+
+
+**Note** The names for these protocols aren't fixed, I just went with the most descriptive thing i could think of.
```
PSchmiedmayer: I would suggest to use the names described in this comment here: https://github.com/Apodini/Apodini/pull/26#discussion_r538396896

Documentation/replacing-component.md
```diff
@@ -0,0 +1,266 @@
...
+
+## Proposed replacement
+
+
+### The new protocol structure
+
+My current implementation in [experimental/dsl-node-protocols][branch_url] replaces `Component` with the following two protocols:
+
+```swift
+protocol EndpointNode {
```
PSchmiedmayer: I originally suggested in https://github.com/Apodini/Apodini/pull/26#discussion_r538396896 that the type providing the handle function also conforms to the type building the static structure of the web service (`EndpointProvidingNode`) to enable a clear hierarchy of protocols. Out of curiosity, what would be the advantage of using a completely separate protocol with no connection in this case?

Documentation/replacing-component.md
```diff
@@ -0,0 +1,266 @@
...
+
+- `EndpointNode`s are nodes which expose some functionality to the client.  
+  All leaves in the `WebService` must be `EndpointNode`s.
+- `EndpointProvidingNode`s are nodes which sit above the leaves.  
+  They cannot provide any functionality of their own, but they must eventually lead to one or more `EndpointNode`s.
+
+
+**Note** The names for these protocols aren't fixed, I just went with the most descriptive thing i could think of.
+
+
```
PSchmiedmayer: It would be great if you could extend this proposal to cover a few of the ideas we had in our last meeting by going beyond splitting it up into two protocols but trying to envision an API that would enable the communication between `Handler`s/`EndpointNode`s and how that would fit into the two existing protocols and would go beyond them. I don't think that we would need an implementation for a "vision" document but this could already provide a nice discussion on how to extend these concepts in the future.

Documentation/replacing-component.md
```diff
@@ -0,0 +1,266 @@
...
+
+### How does it affect the current implementation?
+
+Switching to this new structure required surprisingly few changes.
+In most cases it was sufficient to replace `Component` with `EndpointNode`, since the semantic model builder already operates only on endpoint nodes, and never on non-endpoint nodes. All tests still pass.
+
+- `Text` is now an `EndpointNode`
+- `EmptyComponent` is removed, since user-exposed empty components were invalid DSL components to begin with.
```
PSchmiedmayer: That's a good point 👍

Documentation/replacing-component.md
```diff
@@ -0,0 +1,266 @@
...
+
+Switching to this new structure required surprisingly few changes.
+In most cases it was sufficient to replace `Component` with `EndpointNode`, since the semantic model builder already operates only on endpoint nodes, and never on non-endpoint nodes. All tests still pass.
+
+- `Text` is now an `EndpointNode`
+- `EmptyComponent` is removed, since user-exposed empty components were invalid DSL components to begin with.  
+  (It still is possible to define a noop component, by typealiasing `Content` or `Response` to `Never`.)
+- Some tests needed to be updated (since a component's nested handlers can't be accessed via `component.content as? SomeHandlerType`)
+- `AnyComponent` is replaced by `AnyEndpointNode` and `AnyEndpointProvidingNode`
+- Neither `content` nor `handle()` have a default implementation anymore.
```
PSchmiedmayer: Wound make sense to me

Documentation/replacing-component.md
```diff
@@ -0,0 +1,266 @@
...
+
+This directly benefits both the user (ie the person implementing a web service), and us:
+
+- The user will, in some cases, get compile-time errors when trying to compile incorrectly implemnented components, or ill-formed DSL constructs
+- We can, when dealing with DSL nodes, always know for a fact their specific type and functionalities and won't have to do any dynamic or runtime-metadata based checks to determine that information
+
+
+
+
+### Issues with the current implementation in [experimental/dsl-node-protocols][branch_url]
```
PSchmiedmayer: Would this be solved if `Handler`/`EndpointNode` is also conforming to `Component`/`EndpointProvidingNode` as described here https://github.com/Apodini/Apodini/pull/26#discussion_r538396896? How would this work with protocols that further refine `Handler`/`EndpointNode` e.g. to enable communication between `Handler`s?

Documentation/replacing-component.md
```diff
@@ -0,0 +1,266 @@
...
+```
+
+Basically, we have to support all potential permutations/mixings of endpoint- and non-endpoint nodes in the function builder.
+This also means that the number of generated functions isn't `O(n)` anymore, but `O(n^2)`.  
+(it's been a while since i took GAD, so i could be wrong here...)
+
+
+
+
+**Certain use cases necessitate code duplication**
```
PSchmiedmayer: I have the feeling that this would also be solved by making `Handler`/`EndpointNode` conform to `Component`/`EndpointProvidingNode`. I suspect it would still be possible to write modifiers that only work on `Handler`s/`EndpointNode`s which might even be disable as e.g. an `operation(...)` (https://github.com/Apodini/Apodini/blob/develop/Documentation/Modifiers/Operation.md) modifier wouldn't make too much sense on a `Group`.

PSchmiedmayer (changes requested):
@lukaskollmer Thanks for the good summary over the problems and possible solutions regarding our current `Component` problem! I have added a few comments and notes throughout the proposal. I think it would be great to extend the document to also provide a vision for communication between `Handler`s/`EndpointNode`s. I don't think we need an implementation for this now but a general vision would make the proposed changes and split up more future proof. Looking forward to the next iteration of the proposal, I think this is going to be available addition to Apodini that we should add very soon to the Apodini source code!

Documentation/replacing-component.md
lukaskollmer: Yes I know, I just wanted to point out the fact that the new design would allow for this kind of specialised protocol hierarchy.

Documentation/replacing-component.md
lukaskollmer: My idea here was that having them be separate would prevent accidentally passing one where we expect the other.
Currently, you can use "components with a handle function" and "components with a content property" interchangeably, this wouldn't be allowed anymore

Documentation/replacing-component.md
lukaskollmer: Probably. The thing is, if we have `Handler` conform to `Component`, we end up in the same situation we currently have: all handlers might also be components, but we don't really know for sure (at least not statically).

The way i see it, the only reason for having `Handler` conform to `Component` would be being able to register endpoints on intermediate paths:
```
- GET /v1/foo
- GET /v1/foo/bar
```

This is already possible without having `Handler` conform to `Component`, by simply having a `Handler` and a `Component` on the same level w/in the DSL (in fact we're already doing this in the TestWebService).


Another issue is that, if handlers were allowed to provide further components, the reasonable expextation would be that the `content` returned by the handler introduces a new path component (ie, `content` returns a `Group`).
Otherwise, if a handler's content property returned another handler, that handler might conflict with the handler itself.

Documentation/replacing-component.md
lukaskollmer: This specific example would in fact work as expected now, since https://github.com/Apodini/Apodini/commit/0811646d1c87cb51b4817402a2228d2968e201fb.

It'd still break if `User` isn't `Codable` anymore.

Yes, removing the default type would fix this

Documentation/replacing-component.md
lukaskollmer: I added the implementation to this PR

Documentation/replacing-component.md
lukaskollmer: I added an explanation of the handler identification API, as well as a very short section on inter process communication (short mainly because I haven't thought this out enough to have some kind of definitive vision document)

Documentation/replacing-component.md
lukaskollmer: **Update:** this now works as expected. Modifiers are now defined once, and can conditionally be enabled to work on 
all `Component` types, or just `Handler`s.

<hr/>

I tried this (having `Handler` conform to `Component`) and the issue is that, since the `Modifier` protocol itself needs to inherit from `Component`, you lose type information after applying one modifier to a component.

Example from the GuardTests:
```swift
Text("Hallo")            // type: Handler
    .operation(.read)    // type: Component
    .guard(TestGuard())  // type: Component
    .operation(.create)  // type: Component
```

The first modifier returns an `OperationModifier<Text>` which, via conformance to `Modifier`, conforms to `Component` but not to `Handler`, meaning we're "downgrading" the `Handler` type to a mere `Component`.

Ideally, we'd be able to fix this by writing something along the lines of
```swift
extension Modifier: Handler where ModifiedComponent: Handler {
    public typealias Response = ModifiedComponent.Response
}
```
but that doesn't compile since you can't declare protocol inheritance in an extension.

Documentation/Components/Component Protocol Structure.md
```diff
@@ -0,0 +1,275 @@
+# Documentation: `Component` protocol structure
```
PSchmiedmayer: I would suggest to morph the current version of the .md file to a documentation like format that is less about the reasoning of this PR and more about the structure that will be found in the Apodini project after this PR is merged. I think this description is a perfect PR description but would not provide too much value for the long run if it is merged into the develop branch. I think the description of `Component`s, `Handler`s, and `IdentifiableHandler`s doesn't need to be to extensive but should provide a user of Apodini some guideline on how to choose which protocol.

Sources/Apodini/Components/Component.swift
```diff
@@ -7,37 +7,84 @@
...
+
+/// A `Handler` which can be uniquely identified
+public protocol IdentifiableHandler: Handler {
+    /// The type of this handler's identifier
+    associatedtype EndpointIdentifier: AnyEndpointIdentifier
     
-    /// A function that is called when a request reaches the `Component`
-    func handle() -> Self.Response
+    /// This handler's identifier
+    var endpointId: EndpointIdentifier { get }
```
PSchmiedmayer: I would suggest to name this property handlerId and the corresponding types to be renamed to `HandlerIdentifier` as we do not use the terminology endpoint in the public facing API and only in the internal APIs and the semantic model.
```suggestion
    var handlerId: HandlerIdentifier { get }
```
The only element that would speak against this would be that the identifier will be eventually be used to identify an endpoint but I would be fine with using the term `HandlerIdentifier` in the `Endpoint` struct or using a type alias in there to name it `EndpointIdentifier` in the context where we start referring to it as an endpoint. What do you think @lukaskollmer ?

Sources/Apodini/Components/Component.swift
```diff
@@ -7,37 +7,84 @@
...
+                content.visit(visitor)
+                visitor.exitCollection()
+            }
+            HandlerVisitorHelperImpl(visitor: visitor)(self)
         }
     }
 }
+
+
+private func LKAssertTypeIsStruct<T>(_: T.Type) {
```
PSchmiedmayer: I would suggest to remove the `LK` prefix.
```suggestion
private func AssertTypeIsStruct<T>(_: T.Type) {
```

PSchmiedmayer (changes requested):
@lukaskollmer I really like the new iteration of the PR!

The only bigger comment I had was regarding the structure of the markdown file that will eventually end up in the develop branch. Apart from this and a few naming comments the PR looks great to me. I think splitting up `Component`s and `Handler`s is a great next step and adding a way to identify `Endpoints` to use this information in determining a dependency between `Handler`s, communicate between `Handler`s and thinking about smart ways to deploy a Apodini web service sounds great to me.

I would be good if a second person of the Apodini team would also be able to take a look at the PR, maybe you could assign two additional random reviewers from the @Apodini/apodini team 👍

Sources/Apodini/Components/Component.swift
lukaskollmer: Yeah the reason I chose `EndpointIdentifier` over `HandlerIdentifier` was precisely because at the end of the day we think of the web service's exported interface as a set of endpoints, not handlers.
Not sure about the typealias, since we'd just introduce a second name for the same thing (in the context where we're referring to it as an endpoint, the `HandlerIdentifier` type would also still be available), and it might cause confusion as to why there's both `HandlerIdentifier` and `EndpointIdentifier`.

I'm fine with it either way, but I'll rename it so that the terminology matches.

Sources/Apodini/Components/Component.swift
lukaskollmer: good catch i missed that one

Sources/Apodini/Components/ComponentBuilder.swift.gyb
```diff
@@ -7,34 +7,38 @@
...
+        generic_names_in = ['T{}'.format(idx) for idx in indices]
+        generics = ', '.join(['{}: Component'.format(name) for name in generic_names_in])
+        generic_names_out = ['T{}'.format(idx) for idx in indices]
+        param_names = ['arg{}'.format(idx) for idx in indices]
+        params = ', '.join(['_ {}: {}'.format(param, type) for param, type in zip(param_names, generic_names_in)])
+        tuple_ctor_args = param_names
+}%
+    public static func buildBlock<${generics}>(${params}) -> some Component {
+        return TupleComponent<(${', '.join(generic_names_out)})>((${', '.join(tuple_ctor_args)}))
     }
```
eldcn, line 42: An empty line would make the linter happier here :)

Sources/Apodini/Semantic Model Builder/EndpointsTree.swift
```diff
@@ -18,14 +18,18 @@ struct EndpointRelationship { // ... to be replaced by a proper Relationship mod
...
     /// This is a reference to the node where the endpoint is located
     // swiftlint:disable:next implicitly_unwrapped_optional
     fileprivate var treeNode: EndpointsTreeNode!
     
     /// Description of the associated component, currently included for debug purposes
     let description: String
     
+    /// An identifier which uniquely identifies this endpoint (via its handler)
+    /// across multiple compilations and executions of the web service.
+    let identifier: AnyHandlerIdentifier
```
eldcn: I really like the current approach of uniquely identifying an endpoint over different compilations and executions of the web service. I have also started to make some thoughts in this regard, because for my thesis I will be detecting api changes between different versions of the web service. However, something that I can observe from the PR, and perhaps might be interesting also for you. Currently, this property does uniquely identify the endpoint, **IFF** the order of registering the ~~components~~ handlers, remains the same over multiple compilations and executions of the web service.

Sources/Apodini/SyntaxTreeVisitor/SyntaxTreeVisitor.swift
```diff
@@ -67,4 +81,26 @@ class SyntaxTreeVisitor {
             fatalError("Tried exiting a ContextNode which didn't have any parent nodes")
         }
     }
+    
+    
+    func formHandlerIndexPathForCurrentNode() -> HandlerIndexPath {
+        let rawValue = currentNodeIndexPath
```
eldcn, line 87: `reduce` + `dropLast`?

Sources/Apodini/Components/ComponentBuilder.swift.gyb
lukaskollmer: good catch!

Sources/Apodini/SyntaxTreeVisitor/SyntaxTreeVisitor.swift
lukaskollmer: what do you mean?

Sources/Apodini/Semantic Model Builder/EndpointsTree.swift
lukaskollmer: Yeah you're right in that regard. My primary use case here is being able to "match" the same instance of a handler across multiple simultaneously running instances.
For example, if your web service (consisting of 3 handlers A, B, C) is deployed to two instances, and you want to remotely invoke a specific handler in the other instance, we need a way to find the correct handler in that other process.
This will also be used to determine which instances of your deployed system should implement which handlers.

As currently implemented, the identifiers are only "stable" as long as the web service's structure remains unchanged.

Another thing is that the identifier's `rawValue` is in itself pretty meaningless: it's either some user-provided string, for `IdentifiableHandler`s, or an IndexPath-like string for all other handlers. (eg: `0:1:0:3`, which would indicate "the fourth child of the first child of the second child of the root node".)
The obvious issue here is that these IndexPath-like identifiers can change even if the exported interface doesn't change (eg: you wrap a component in a Group w/out any path parameters, or you add additional children to a group).

Sources/Apodini/SyntaxTreeVisitor/SyntaxTreeVisitor.swift
eldcn: I was thinking for something like: `var rawValue = currentNodeIndexPath.reduce("") { "\($0)\(max($1, 1) - 1)):" }` and then dropping the last character. You could spare iterating three times over the array. And after that, either marking the function as private since you use it only in STV or directly do the construction inside `register(handler:)`

Sources/Apodini/Semantic Model Builder/EndpointsTree.swift
eldcn: Thanks for the detailed explanation. I can see that the current approach satisfies your use case. Just wanted to point that out, if that would have been relevant for your work as well. Good job overall  👏

Sources/Apodini/SyntaxTreeVisitor/SyntaxTreeVisitor.swift
lukaskollmer: You're right that there's unnecessary iterations over the array (I got it down to 2 now, plus the string concatenation).

Not a fan of reduce though, at least in this specific case. I was trying not to be too clever w/ the code and i think piping it through the `map` and `joined(separator:)` calls is easier to read than using a `reduce`-based approach.

(Also, not to be too pedantic, but I think that, from a technical point of view, your solution, while requiring only 1 iteration over the array, would require a lot more expensive string constructions (since it always copies the result into the next result), and then also a full iteration over the String's characters, as well as another string concatenation at the end (String.dropLast returns an `[Character]`, which would need to be reassembled into a String). I could be wrong though. Also, since we probably won't have components nested more than a few dozen levels deep at most, i doubt the difference would ever be noticable.)

Sources/Apodini/Semantic Model Builder/EndpointsTree.swift
lukaskollmer: My assumption here is that, if a web service is deployed as a system of multiple instances, all of these must always be deployed together. That way we eliminate the issue of identifier mismatches between multiple deployments.

I think it should be possible to track `IdentifiableHandler`s across compilations when the structure of the web service changed. As long as the handler's custom identifier remains that same, it should be sufficient to compare the IndexPath to see if it was moved w/in the DSL, and then based on the path components of the generated Endpoint determine if the move somehow changed the exported interface.

PSchmiedmayer (approved):
@lukaskollmer Thanks for updating your PR. I like the new version of the `Component Types.md` file describing the difference between `Handler`s and `Component`s, it looks good to me!

I would suggest to fix all the SwiftLint warnings that remain in this branch to get all merge checks green. 

I saw [your comment](https://github.com/Apodini/Apodini/pull/81#issuecomment-752383660) on #81 and would suggest to either fix it in this branch or create a short fix that is agreed on with @Supereg . As you have also requested some reviews from other Apodini members I would wait for one review from one of them but from a vision perspective and what I could check in the PR I can approve it 👍

Sources/Apodini/Modifier/ResponseModifer.swift
```diff
@@ -95,7 +89,7 @@ extension ResponseModifier: SyntaxTreeVisitable {
 }
 
 
-extension Component {
+extension Handler {
     /// A `response` modifier can be used to transform the output of `Component`'s response to a different type using a `ResponseTransformer`
```
bauer-andreas: Nit: `Handler` instead of `Component`

Sources/Apodini/Modifier/GuardModifier.swift
```diff
@@ -33,11 +33,12 @@ struct GuardContextKey: ContextKey {
...
+    public typealias ModifiedComponent = C
     
-    let component: C
+    public let component: C
     let `guard`: LazyGuard
     
+    public var content: some Component { EmptyComponent() }
     
     init<G: Guard>(_ component: C, guard: @escaping () -> G) {
         precondition(((try? typeInfo(of: G.self).kind) ?? .none) == .struct, "Guard \((try? typeInfo(of: G.self).name) ?? "unknown") must be a struct")
```
bauer-andreas: Maybe we can reuse the `AssertTypeIsStruct` defined in `Component.swift` here as well. I think similar code is also found for ResponseTansformers and for RequestInjectibles (probably located in `Traversable.swift` as recently refactored by Max).

Sources/Apodini/Modifier/OperationModifier.swift
```diff
@@ -47,7 +48,8 @@ extension OperationModifier: SyntaxTreeVisitable {
 }
 
 
-extension Component {
+
+extension Handler {
     /// A `operation` modifier can be used to explicitly specify the `Operation` for the given `Component`
     /// - Parameter operation: The `Operation` that is used to for the component
     /// - Returns: The modified `Component` with a specified `Operation`
```
bauer-andreas: Nit: 3x replace `Component` with `Handler` in docs

Sources/Apodini/Modifier/ResponseModifer.swift
```diff
@@ -65,25 +65,19 @@ struct ResponseContextKey: ContextKey {
...
     public typealias Response = T.TransformedResponse
     
-    let component: C
+    public let component: H
     let responseTransformer: () -> (T)
     
     
-    init(_ component: C, responseTransformer: @escaping () -> (T)) {
+    init(_ component: H, responseTransformer: @escaping () -> (T)) {
         precondition(((try? typeInfo(of: T.self).kind) ?? .none) == .struct, "ResponseTransformer \((try? typeInfo(of: T.self).name) ?? "unknown") must be a struct")
```
bauer-andreas: As above, reuse `AssertTypeIsStruct` if possible.

bauer-andreas (approved):
Took me some time to review the PR 😅
I got some small nit picks I think worth adjusting 🙃 Additionally I think there are some SwiftLint problems which need to be fixed, seemingly some fatalErrors without a message.
So you probably need to make sam adjustments anyways.
Lastly the naming inconsistencies described in #81 will also be incorporated into this PR right (https://github.com/Apodini/Apodini/pull/81#issuecomment-752700505)?

But you can consider the PR to be approved from my side. Love to finally see the separation of Components and Handlers being incorporated into the project. 👍

PR #70, opened 2020-12-21
Proposal: Database Read Component using QueryBuilder 

hendesi:
### Workflow & Motivation

This PR aims to add a query builder which is independent of the used exporter and the query parameters that come with the request. The main goal is to create a `Read` Component which is not constrained to a single query, but is able to execute any query which is defined by the parameters of the request. The functionality of this component depends on the work of [dynamics](https://github.com/Apodini/Apodini/pull/56) by @theMomax. As `Dynamics` is not merged yet and there is some functionality of the query builder needs to be further defined, I would consider this PR a WIP and can used as a place for further discussions. This approach is similar to the one I worked on with @nerdsupremacist and presented on the intermediate presentation, but as for now, I think I could simply it a bit.

The `QueryBuilder` has a `func info(for type: Model.Type) -> [ModelInfo]` which returns an array of `ModelInfo`, where every element represents one property of the database model. This property contains field key and its type. The `Read` Component uses this info to initialise the `@Dynamics`:

```
public init() {
        var dynamicValues: [String: Param] = [:]
        let infos = QueryBuilder.info(for: Model.self)
        for info in infos {
            dynamicValues[info.key.description] = Param(context: info)
        }
        _dynamics = Dynamics(wrappedValue: dynamicValues)
}
```

The `Param` is a simple property wrapper which expects a `String` as a type and a `ModelType`. I reworked the `func decode` of the `RequestInjectableDecoder` to allow the exporter access to the respective model type of the parameter. 
When the `Read` is called, it build an `[FieldKey: String] ` directory using the wrapped values of the `Param`. 

```
let queryInfo: [FieldKey: String] = dynamics
            .reduce(into: [FieldKey: String?](), { result, entry in
                result[Model.fieldKey(for: entry.key)] = entry.value.wrappedValue
            })
            .compactMapValues({ $0 })
```
It passes the directory to the `QueryBuilder` and executes the query using `queryBuilder.execute(on: database)` which returns the results of the query in a `EventLoopFuture`. Internally the `QueryBuilder` creates a `Fluent.Query` for every entry of the directory and accumulates them to a single query. This depends highly on the `Fluent` framework.

```
public func execute(on database: Fluent.Database) -> EventLoopFuture<[Model]> {
        
        func accumulateFilters(queryBuilder: inout Fluent.QueryBuilder<Model>, key: FieldKey, value: String, method: DatabaseQuery.Filter.Method) {
            if let intValue = Int(value) {
                queryBuilder.filter(key, method, intValue)
            } else if let boolValue = Bool(value) {
                queryBuilder.filter(key, method, boolValue)
            } else {
                queryBuilder.filter(key, method, value)
            }
        }
        
        var queryBuilder = Model.query(on: database)
        for (key, value) in parameters {
            accumulateFilters(queryBuilder: &queryBuilder, key: key, value: value, method: .equal)
        }
        return queryBuilder.all()
    }
```

### Points of Discussion / Problems
- Currently the Querybuilder has to cast the type to see if it can be an integer or a bool. Although we can find the type at runtime, I cannot see a way which would allow use to use a specific `Param` like `Param<Int>` instead of having all parameters conforming to type String
- Maybe the exporters need more information on how to decode. Currently the info they get is the FieldKey and that they should cast whatever they find to String.

The PR currently contains also some components for the other CRUD operations. They can either be eventually merged with the proposals or I could create separate PRs for these.


Sources/Apodini/Components/Create.swift
```diff
@@ -0,0 +1,143 @@
...
+//  
+//
+//  Created by Felix Desiderato on 27.11.20.
+//
+
+import Foundation
+import Fluent
+import Vapor
+
+public protocol DatabaseInjectionContext {
```
nerdsupremacist: I love having clearly separate files if you don't mind. Please split up the files a bit

Sources/Apodini/Components/Create.swift
```diff
@@ -0,0 +1,143 @@
...
+    var key: FieldKey { get }
+    var type: Any.Type { get }
+}
+
+public struct Create<T: DatabaseModel>: Component where T.IDValue == UUID {
```
nerdsupremacist: noob question: why does IDValue have to be a UUID?

Sources/Apodini/Components/Create.swift
```diff
@@ -0,0 +1,143 @@
...
+    @Dynamics var dynamics: [String: Param]
+    
+    public init(_ dummy: Parameter<String>) {
+        self._dummy = dummy
+        var dynamicValues: [String: Param] = [:]
+        let infos = QueryBuilder.info(for: Model.self)
+        for info in infos {
+            dynamicValues[info.key.description] = Param(context: info)
+        }
+        _dynamics = Dynamics(wrappedValue: dynamicValues)
```
nerdsupremacist: Love it!

Sources/Apodini/Components/Create.swift
```diff
@@ -0,0 +1,143 @@
...
+    var database: Fluent.Database
+    
+    @Parameter var object: T
+    
+    @Parameter var id: T.IDValue
+    
+    public func handle() -> EventLoopFuture<T> {
+        return T.find(id, on: database).flatMapThrowing({ model -> T in
+            model?.update(object)
+            return model!
```
nerdsupremacist: please throw an error if the item does not exist

Sources/Apodini/Components/Create.swift
```diff
@@ -0,0 +1,143 @@
...
+
+@propertyWrapper
+struct Param: RequestInjectable {
+    
+    private var context: DatabaseInjectionContext
+
+    var _value: String?
+    
+    var wrappedValue: String? {
+        if context.key == FieldKey("id") || context.key == FieldKey("name") {
```
nerdsupremacist: Why the special case here? My instinct tells me this is debugging code, but maybe I'm overseeing something

Sources/Apodini/Request/Body.swift
```diff
@@ -0,0 +1,35 @@
...
+//  Created by Paul Schmiedmayer on 6/26/20.
+//
+
+import NIOFoundationCompat
+import Vapor
+import Foundation
+
+
+@propertyWrapper
+public struct Body<Element: Codable>: RequestInjectable {
```
nerdsupremacist: I thought we removed this. Maybe you added it again while rebasing?

nerdsupremacist (commented):
Great job! A lot of cool concepts here! and I love the interesting use of the new dynamics...
I'd like to merge this right after merging the dynamics PR.
Let's work on this during the Work Meeting ;)

I'd love to clean up the different property wrappers you implemented here, but I guess that most of them are necessary until all the param + vapor stuff is sorted out properly (is it fixed already I'm not sure)

hendesi:
@nerdsupremacist Thanks for the feedback! Your instinct was right, most of it was debug code or some code relicts that somehow stayed there during the rebase. I removed them, moved different structs to different files and cleaned the code up a bit in general 👍

theMomax:
This is a pretty decent feature👍🏻. However, I think there is still quite some work to do in order to clean up the implementation:

In general I think that these database-related `Components` belong into a "Foundation" package and should thus not make use of **internal** elements (e.g. `RequestInjectable`).

As far as I can see you are not alone with the problem you encountered with `RequestInjectable` being not sufficient for your needs. I had the same problem with the old `RequestInjectable` (thanks to @tgymnich's revision it is now sufficient for me) and I think the REST-exporter is also struggling to get all the required information. I think we should have a bigger discussion on how to tackle this. An abstracted way of your requirements should also be covered by the result.

I think we should also adapt `@Parameter` to fit your needs to replace your new `Param`.

All of these changes to the core of Apodini should of course be rather abstracted and database-agnostic.

PSchmiedmayer (changes requested):
@hendesi I have looked over your PR but as far as I can see most of the elements that I would have commented on are dependent on the PR of @theMomax which should be merged very soon and is already addressed by some of the PRs that are linked to this PR such as #84 which looks like a good approach to me 👍

For now I would leave a detailed review of the PR but let me know as soon as the other PRs are merged and you have a comprehensive version of this PR that introduced the components and includes the functionality of the other PRs. 

In general I would suggest to add some documentation to the respective components to make clear what they can be used for and how a developer can use them in an Apodini project. In general I think that the ideas you have are going in the right direction and I am looking forward to what will be possible with the different database components. 
In addition I would suggest to move the different components such as `Read`, `Create`, ... to a separate Apodini target that we should call `ApodiniFoundation` similar to how the protocol buffer decoder is in a seperate target. This will make it more clear what public functionality will be needed to enable the usage of Apodini for other developers and will lead to a clear and well usable public interface of the Apodini project.

hendesi:
@PSchmiedmayer I moved the database files into a separate target. I'd like to get rid of the different fluent database drivers (e.g. `.product(name: "FluentMongoDriver", package: "fluent-mongo-driver")` in the Apodini target and move them to the Database target. This would impact the `DatabaseConfiguration` in which the Environment value of the database is set. 
```
if let database = app.databases.ids().map({ app.db($0) }).first {
                EnvironmentValues.shared.database = database
}
```
Since `shared` is currently internal, it's not possible to move the config out of main Apodini target. I personally don't think we should change the scope of shared. But this would leave us with having the driver dependencies in main Apodini.

PSchmiedmayer:
@hendesi I agree that the shared values should only be accessible internally but I would envision an API that would be public to manually add new values to the `EnvironmentValues.shared` instance. I think an addition only public API would be fine and in long term I could see how these shared environment values could move the the own `Application` type introduced in #82 which would also be available to any `Configuration` instances.

hendesi:
After a lot of SwiftLint fixing and test adding, I think I implemented the requested changes. 
Changes I made: 
- Added description
- Moved Database into separate target
- Added tests for database handlers and querybuilder
- Removed unused parameter option

Re-requesting review again from @PSchmiedmayer @theMomax @nerdsupremacist 

@PSchmiedmayer I like the solution of offering a public API for developers to add environment values by themselves. But I would move this to a separate PR and try to get this one merged first if you agree 👍

Sources/ApodiniDatabase/DatabaseModel.swift
```diff
@@ -0,0 +1,15 @@
+import Fluent
+import Vapor
+
+///A protocol all Models that are used with `ApodiniDatabase` need conform to
+public protocol DatabaseModel: Content, Model where IDValue == UUID {
```
PSchmiedmayer: What is the reason that the `IDValue` is constrained to `UUID`. What would happen if the ID value is something else?

Sources/ApodiniDatabase/Handlers/Create.swift
```diff
@@ -0,0 +1,30 @@
...
+    public func handle() -> T? {
+        // swiftlint:disable all
+        let result = try! object
+            .save(on: database)
+            .map { _ in
+                self.object
+            }
+            .wait()
+        
+        return try! T
```
PSchmiedmayer: What is the reasoning behind fetching the type from the database? Would it be fine to store `object` to a local property and return it after saving the element to the database. I think as `object` is an instance of a class type the ID should also be updated after the save call.

Sources/ApodiniDatabase/Handlers/Create.swift
```diff
@@ -0,0 +1,30 @@
...
+    private var object: T
+
+    public func handle() -> T? {
+        // swiftlint:disable all
+        let result = try! object
+            .save(on: database)
+            .map { _ in
+                self.object
+            }
+            .wait()
```
PSchmiedmayer: I would suggest to wait for the merging of #94 or #98 or a mix between the two ideas to we don't introduce blocking code in develop. It would also be great to get your input in the ideas as well as the #97 as your code would be one of the main users of asynchronous code.

Sources/ApodiniDatabase/Handlers/Create.swift
```diff
@@ -0,0 +1,30 @@
...
+/// It uses the database that has been specified in the `DatabaseConfiguration`.
+public struct Create<T: DatabaseModel>: Handler {
+    @Apodini.Environment(\.database)
+    private var database: Fluent.Database
+    
+    @Parameter
+    private var object: T
+
+    public func handle() -> T? {
+        // swiftlint:disable all
```
PSchmiedmayer: Please don't disable all SwiftLint warnings. Please only disable a single rule and provide a detailed description on why the rule is disabled in a comment [as described in this message](https://github.com/Apodini/Apodini/pull/83#issuecomment-752128774)

In general I think it would be good to support throwing handlers that would be able to forward throwing errors to the client. Vapor supports this mapping of errors to networking errors and correct responses to the client. I think this would be a great addition to Apodini and the database code would be a prime example of the usage of throwing errors.

Sources/ApodiniDatabase/Handlers/Delete.swift
```diff
@@ -0,0 +1,28 @@
...
+/// A `Handler` that deletes the object with the given `IDValue` in the database, if it exists. If not an error is thrown.
+/// It uses the database that has been specified in the `DatabaseConfiguration`.
+public struct Delete<Model: DatabaseModel>: Handler {
+    @Apodini.Environment(\.database)
+    private var database: Fluent.Database
+
+    @Parameter(.http(.path))
+    var id: UUID
+    
+    var idParameter: Parameter<UUID> {
```
PSchmiedmayer: As far as I can see the `idParameter` is used for testing purposes only. You could achieve the same result using reflection (`Mirror`) in the test case which would reduce the internal interface of the type.

Sources/ApodiniDatabase/Handlers/Read.swift
```diff
@@ -0,0 +1,39 @@
...
+        var dynamicValues: [String: Parameter<String?>] = [:]
+        let infos = QueryBuilder.info(for: Model.self)
+        for info in infos {
+            dynamicValues[info.key.description] = Parameter<String?>(.http(.query))
+        }
+        _dynamics = Properties(wrappedValue: dynamicValues)
+    }
+
+    public func handle() -> String {
+//    public func handle() -> [Model] {
```
PSchmiedmayer: I think the current implementation doesn't return the actual `Models` as this functionality is commented out.

Sources/ApodiniDatabase/Handlers/Read.swift
```diff
@@ -0,0 +1,39 @@
...
+import Apodini
+
+/// A Handler that returns an array of `DatabaseModel` which all conform to the query parameters of the request. It accepts all parameters that represent properties of the specified `DatabaseModel`.
+/// It uses the database that has been specified in the `DatabaseConfiguration`.
+/// An Example using a REST request:
+/// Group("api", "birds") {
+///     Read<Bird>()
+/// }
+/// Sending a request to  ~/api/birds/birds?age=19&name=Foo would return an array of `Bird` object that have an age of 19 and the name Foo.
+public struct Read<Model: DatabaseModel>: Handler {
```
PSchmiedmayer: I like the flexibility of the `Read` component using `@Properties` 🚀

Sources/ApodiniDatabase/Handlers/Update.swift
```diff
@@ -0,0 +1,37 @@
...
+    @Apodini.Environment(\.database)
+    private var database: Fluent.Database
+    
+    @Parameter
+    private var object: Model
+    
+    @Parameter(.http(.path))
+    var id: UUID
+    
+    var idParameter: Parameter<UUID> {
```
PSchmiedmayer: See my comment in the `Delete` `Handler`

Sources/ApodiniDatabase/Handlers/Update.swift
```diff
@@ -0,0 +1,37 @@
...
+        _id
+    }
+    
+//    public func handle() -> EventLoopFuture<T> {
+    public func handle() -> String {
+        // swiftlint:disable all
```
PSchmiedmayer: Please rework the usage of `// swiftlint:disable` across the PR

Sources/ApodiniDatabase/Handlers/Update.swift
```diff
@@ -0,0 +1,37 @@
...
+@_implementationOnly import Vapor
+import Apodini
+
+/// A Handler that updates a `DatabaseModel` with the given id with the new object of the request.
+/// It uses the database that has been specified in the `DatabaseConfiguration`.
+public struct Update<Model: DatabaseModel>: Handler {
+    @Apodini.Environment(\.database)
+    private var database: Fluent.Database
+    
+    @Parameter
```
PSchmiedmayer: I think updating would be an other great use cases of `@Properties` and a dynamic creation of these properties. E.g. we could achieve an `Update` component that can update a `Model` with a request that just provides one property of the `Model` and leaves all other elements like they are. I think this would be a very sophisticated usage of reflection and the runtime information and `@Properties` but would showcase the great flexibility of our approaches. I would be fine with this not being part of this PR but a subsequent PR. If this is the case pleas document the feature using a GitHub issue.

Sources/ApodiniDatabase/Query/QueryBuilder.swift
```diff
@@ -0,0 +1,124 @@
...
+    }
+    
+    init(type: Model.Type, parameters: [FieldKey: String]) {
+        self.type = type
+        self.parameters = parameters
+    }
+    
+    private func extract(from queryString: String) -> [FieldKey: String] {
+        var foundParameters: [FieldKey: String] = [:]
+        let extractedQueryString = queryString.split(separator: "?")[1]
```
PSchmiedmayer: I would suggest to check the count or possible indexes before accessing a collection using a fixed index.

Sources/ApodiniDatabase/Query/QueryBuilder.swift
```diff
@@ -0,0 +1,124 @@
...
+                }
+            }
+        } catch {
+            fatalError("failed to infer info")
+        }
+        return modelInfo
+    }
+    
+    // swiftlint:disable:next todo
+    //TODO: Find a better way to do this
```
PSchmiedmayer: I would suggest to implement some small visitor pattern with an extension to `FieldProperty` that would allow you to access the type and the corresponding information using type checks instead of string representations, let me know if you need some more input on this 👍

PSchmiedmayer (changes requested):
@hendesi Thanks for the updated PR for the Database components. I really like how you were able to split out the implementation into a separate target.

Regarding [your comment](https://github.com/Apodini/Apodini/pull/70#issuecomment-755298583): 
> I like the solution of offering a public API for developers to add environment values by themselves. But I would move this to a separate PR and try to get this one merged first if you agree 👍
I think it is fine to create a separate PR for this. Please create a GitHub issue for this task so we will be able to keep it in mind and track its progress.

I had a detailed look at the code and added some comments throughout the files. I think the biggest comment impacting this PR is my suggestion to wait for the merger of #94 or #98 or a mix between the two ideas. It be great to get your input and review of the PRs and what your comments regarding the different approaches are.
Apart from this I had some suggestions for some improvements and changes to the usage of the `swiftlint:disable` comments. Feel free to let me know if you have any questions or I can help you with anything. 

I am looking forward to seeing these features and `Handlers` in action! 🚀

Sources/ApodiniDatabase/Handlers/Create.swift
hendesi: I think im not quite sure what you mean. This code is like it is due to `EventLoopFutures` not being a possible return type. Once this is possible again `Create` would return an `EventLoopFuture<Model>, where `Model` is the saved object. If we would store `object` to a local property and return it after the database save, we would again have some kind of blocking. But I agree with you that the returned object should have the saved id, I'll add that

Sources/ApodiniDatabase/DatabaseModel.swift
hendesi: This was something I've thought about before we introduced the new Parameter retrieval. The reason was that for `Model.IDValue` needs to conform to `LosslessStringConvertible`, which it does now anyway. So I'll remove that 👍

Sources/ApodiniDatabase/Handlers/Create.swift
PSchmiedmayer: Sorry for the confusing, I could have been clearer with my comment. Here is an example of what I have meant with leaving out the fetching the type from the database.

Fluent automatically updates the ID property on the save call. So you could have a code similar to this one that does both the saving and the returning the object in a single call as seen in the following code block:
```swift
object
    .save(on: database)
    .transform(to: object)
    .wait()
```

Once we have merged #94 or #98 or a mixture of both you can simply remove the `wait()` call and the method should return the instance include the ID that has been created from the database.

[The Vapor template does something similar but uses `map` instead of `transform`](https://github.com/vapor/template/blob/0f3b552f66aee092b457a542d9a83dd5578d05b6/Sources/App/Controllers/TodoController.swift#L20) which has the same effect without the need for a closure in this context.

hendesi: Thanks for clarifying! I agree, that looks a lot cleaner. I will update the `Create` Handler accordingly.

Sources/ApodiniDatabase/Handlers/Read.swift
hendesi: Some of the handlers currently return `String` because calling `wait()` on the `EventloopFuture` returned by the database operation results in an failed precondition as one is not allowed to call `wait()` on an `eventloop`.

```
Precondition failed: BUG DETECTED: wait() must not be called when on an EventLoop.
```

As you already suggested, it would beneficial to wait the merger of #94 or #98. If `EventloopFuture` works again, I would of course change it. So if you agree, I'd suggest we leave it like that for now.

Sources/ApodiniDatabase/Query/QueryBuilder.swift
hendesi: Its definitely a good idea to move away from checking string representations! 👍 Would your approach also allow to have some type info already on compile time? To have a real benefit, we would need to have the type info at compile time such that `Read` can initialise its `@Properties` with `Parameter<Type` and stop using `Parameter<String>` for everything

Maybe this is also something @nerdsupremacist could provide some input on 👍

Sources/ApodiniDatabase/DatabaseModel.swift
```diff
@@ -0,0 +1,15 @@
...
+    ///Has to be overwritten to use `Update` handler.
+    func update(_ object: Self)
+}
+
+internal extension DatabaseModel {
+    static func fieldKey(for string: String) -> FieldKey {
+        // swiftlint:disable:next force_unwrapping
+        Self.keys.first(where: { $0.description == string })!
```
theMomax: Please provide an explanation if you do a force-unwrap. E.g.:
```suggestion
        if let key = Self.keys.first(where: { $0.description == string })! {
            return key
        } else {
            fatalError("Unexpectedly found nil, this shouldn't have happened as <reason>")
            exit(1)
        }
```

Sources/ApodiniDatabase/Handlers/Read.swift
```diff
@@ -0,0 +1,39 @@
+import Foundation
+import Fluent
+import Apodini
+
+/// A Handler that returns an array of `DatabaseModel` which all conform to the query parameters of the request. It accepts all parameters that represent properties of the specified `DatabaseModel`.
```
theMomax: Sorry for being that picky here, but since there is an open change-request anyways, please add line-breaks after ~150 characters on block-comments.

Sources/ApodiniDatabase/Query/QueryBuilder.swift
theMomax: I agree it would be really beneficial to have proper type-info here! While it does not make that much of a difference for HTTP url parameters, it does for other exporters which rely on JSON or protobuf. Also, it would look a bit ugly in the OpenAPI docs.

theMomax (commented):
In general, this looks very promising. It is really unfortunate how you suffer from `EventLoopFuture`s being dysfunctional currently. Except for smaller code-style issues, the only thing remaining is the missing type-info. I hope @nerdsupremacist can help you with that and I would also be happy to support you during the next work-meeting.

hendesi:
>  I hope @nerdsupremacist can help you with that and I would also be happy to support you during the next work-meeting.

Thanks! I think this could be really beneficial 👍

Sources/ApodiniDatabase/Query/QueryBuilder.swift
PSchmiedmayer: @hendesi Yes, I could image that a visitor pattern based approach would be able to address this issue and would be able to get the type info on compile time. @nerdsupremacist or I can take a look at this in the work meeting and provide some input on how to implement this 🚀

Sources/ApodiniDatabase/Handlers/Read.swift
PSchmiedmayer: #98 should be merged pretty soon so I would be optimistic to have `EventLoopFuture` as possible `Handler` return types by tomorrow's team meeting 👍

PSchmiedmayer:
@hendesi We have merged #98 which now enables `EventLoopFutures` on the develop branch 🚀

Sources/ApodiniDatabase/Query/QueryBuilder.swift
hendesi: @PSchmiedmayer Thanks! I already implemented the visitor pattern on my local branch and it works quite well. Might have some additional questions which we could discuss in the work meeting 👍

hendesi:
@PSchmiedmayer @nerdsupremacist @theMomax 
I added the following improvements to this PR:

- Added a visitor pattern to infer the types in the `Querybuilder` instead of using a string. 
- Modified the update handler to enable single properties updates
- Readded the eventloop returns of the handlers

Could you please take another look at this? So we can get this merged 👍

Sources/Apodini/Configurations/DatabaseConfiguration.swift
```diff
@@ -37,6 +37,7 @@ public final class DatabaseConfiguration: Configuration {
             let factory = try databaseFactory(for: self.type)
             databases.use(factory, as: databaseID)
             app.migrations.add(migrations)
+            print(migrations[0].name)
```
theMomax: Is this intended as a startup log or is it a lost debug log?

Sources/Apodini/Semantic Model Builder/REST/RESTInterfaceExporter.swift
```diff
@@ -165,7 +164,16 @@ class RESTInterfaceExporter: InterfaceExporter {
...
             }
 
-            return try request.content.decode(Type.self, using: JSONDecoder())
+            #warning("""
+                     A Handler could define multiple .content Parameters. In such a case the REST exporter would
+                     need to decode the content via a struct containing those .content parameters as properties.
+                     This is currently unsupported.
+                """)
+            do {
+                return try request.content.decode(Type.self, using: JSONDecoder())
```
theMomax: This change seems a bit weird to me as it has nothing to do with Database components. Did this slip in from a merge?

Sources/Apodini/WebService.swift
```diff
@@ -4,10 +4,9 @@
 //
 //  Created by Paul Schmiedmayer on 7/6/20.
 //
-
+import Fluent
```
theMomax: This can be removed since there is no other change in that file, right?

Sources/ApodiniDatabase/Handlers/Update/Updater.swift
```diff
@@ -0,0 +1,58 @@
...
+                        guard let visitable = child.value as? UpdatableFieldProperty,
+                              let label = child.label
+                        else { continue }
+                        let fieldKey = Model.fieldKey(for: label.trimmed())
+                        if let value = properties[fieldKey] {
+                            if visitable.accept( ConcreteUpdatableFieldPropertyVisitor(updater: value) ) {
+                                _ = model.update(on: database).transform(to: model)
+                                return model
+                            } else {
+                                fatalError("the updater was unable to update single properties of")
```
theMomax: Error message seems to be incomplete.

Sources/ApodiniDatabase/Handlers/Update/Updater.swift
```diff
@@ -0,0 +1,58 @@
...
+    }
+    
+    func executeUpdate(on database: Database) -> EventLoopFuture<Model> {
+        if let model = model {
+            return Model.find(modelID, on: database)
+                .unwrap(orError: Abort(.notFound))
+                .map { foundModel -> Model in
+                    foundModel.update(model)
+                    _ = foundModel.update(on: database).transform(to: model)
+                    return model
```
theMomax: That means as soon as one provides the `model`, all `properties` are ignored, right? Just from the endpoint's API I would have guessed `properties` to not be ignored, but to overwrite their counterparts on `model` if `model` is provided. I isn't that important, but I expected it to behave differently. Maybe just emphasize this a bit more on the `Update`'s docs.

Tests/ApodiniTests/DatabaseTests/QueryBuilderTests.swift
```diff
@@ -0,0 +1,18 @@
...
+    func testQueryString() throws {
+        let expectedParameters: [FieldKey: TypeContainer] = [
+            Bird.fieldKey(for: "name"): TypeContainer(with: "Swift"),
+            Bird.fieldKey(for: "age"): TypeContainer(with: "5")
+        ]
+        let queryBuilder = QueryBuilder(type: Bird.self, parameters: expectedParameters)
+        
+        let birds = try queryBuilder.execute(on: app.db).wait()
+        XCTAssert(birds.count == 1)
+        XCTAssert(birds[0].name == "Swift")
```
theMomax, line 16:
```suggestion
        XCTAssert(birds[0].name == "Swift")
        XCTAssert(birds[0].age == 5)
```

Sources/ApodiniDatabase/Context/TypeContainer.swift
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    var description: String {
+        self.typed().debugDescription
+    }
+}
+
+extension TypeContainer: LosslessStringConvertible {
+    public init?(_ description: String) {
+        self = .noValue
```
theMomax: What is this used for? What implications does it have that you always return `.noValue`?

theMomax (commented):
Mostly looks really good! I noted a few smaller things. I dislike the huge switch statements, but I guess they are necessary. Only bigger thing: Optionals aren't supported, are they?

Sources/Apodini/Semantic Model Builder/REST/RESTInterfaceExporter.swift
hendesi: This is due to having two Parameter properties in the `Update` handler of which one is nil. Currently there is no support for requests contain more than one body parameter. If the request contains something like
```
let parameters: [String: TypeContainer] = [
            "name": TypeContainer(with: "FooBird")
]
```
It would try and decode it as a `Bird` which would fail and throw a `DecodingError`. But as this can also be decoded as `Parameter<[String: TypeContainer]>, it should fail silently

Sources/ApodiniDatabase/Handlers/Update/Updater.swift
hendesi: > That means as soon as one provides the model, all properties are ignored, right?

Yes that is correct. I'll update the docs of the `Update` handler to clarify things a bit there.

Sources/Apodini/Semantic Model Builder/REST/RESTInterfaceExporter.swift
theMomax: Ah, makes sense, thanks for the explanation!

Sources/ApodiniDatabase/Context/TypeContainer.swift
hendesi: Since all query parameters need to conform to `LosslessStringConvertible`, but are then internally rated as `.lightweight` and there for not initialised using `init(_ description: String)`, there is currently no need to implement some decision pattern here to init the right type.
But I see your point, this should be handled eventually.

Sources/Apodini/Configurations/DatabaseConfiguration.swift
hendesi: lost debug log, should be removed 👍

hendesi:
> I dislike the huge switch statements, but I guess they are necessary. 

@theMomax  I agree. But its something which is somewhat necessary in the code, at least as far as I know. If we can come up with something more elegant, we can definitely update this section. But for now I think, it is good enough for a first version.

I addressed all of your comments. Could you please have a final look? 👍

Sources/ApodiniDatabase/Context/TypeContainer.swift
theMomax: Ok, just add a comment with this rationale to the code please.

Sources/ApodiniDatabase/Handlers/Update/Updater.swift
```diff
@@ -41,7 +41,7 @@ internal struct Updater<Model: DatabaseModel> {
                                 _ = model.update(on: database).transform(to: model)
                                 return model
                             } else {
-                                fatalError("the updater was unable to update single properties of")
+                                fatalError("An error occurred while updating single parameters the model ")
```
theMomax: "of" missing 🙃

theMomax (approved):
Looks good from my point of view!

hendesi:
@PSchmiedmayer  @nerdsupremacist Could you please have another look at this? 
I would really appreciate it if we can get this merged soon  👍

Sources/Apodini/Semantic Model Builder/REST/RESTInterfaceExporter.swift
```diff
@@ -165,7 +164,16 @@ class RESTInterfaceExporter: InterfaceExporter {
...
+            } catch {
+                return nil
+            }
```
PSchmiedmayer: I would suggest to add a `try?` operator here:
```suggestion
        return try? request.content.decode(Type.self, using: JSONDecoder())
```

Sources/Apodini/Semantic Model Builder/REST/RESTInterfaceExporter.swift
```diff
@@ -165,7 +164,16 @@ class RESTInterfaceExporter: InterfaceExporter {
                 return nil
             }
 
-            return try request.content.decode(Type.self, using: JSONDecoder())
+            #warning("""
+                     A Handler could define multiple .content Parameters. In such a case the REST exporter would
+                     need to decode the content via a struct containing those .content parameters as properties.
+                     This is currently unsupported.
+                """)
```
PSchmiedmayer: Please remove warning found in the source code. I would suggest to add a detailed issue about the problem and maybe add a comment here that describes the issue and references the GitHub issue.

Sources/ApodiniDatabase/Context/DatabaseInjectionContext.swift
```diff
@@ -0,0 +1,22 @@
...
+///A struct implementing `DatabaseInjectionContext` and containing a fieldkey and the expected type for that key.
+struct ModelInfo: DatabaseInjectionContext, Equatable {
+    ///A concrete `Fluent.FieldKey`
+    var key: FieldKey
+    ///A concrete type for that fieldkey
+    var value: TypeContainer
+    
+    static func == (lhs: ModelInfo, rhs: ModelInfo) -> Bool {
+        lhs.key == rhs.key
+    }
```
PSchmiedmayer: As far as I can see the conformance to `Equatable` is not need at the moment.

Sources/ApodiniDatabase/Context/TypeContainer.swift
PSchmiedmayer: I would suggest adding a comment and using a `fatalError(...)` with a message that would clearly indicate why this initializer could never be called.

Sources/ApodiniDatabase/DatabaseModel.swift
```diff
@@ -0,0 +1,23 @@
+import Fluent
+import Vapor
```
PSchmiedmayer: Please make sure that Vapor is only implemented as `@_implementationOnly`.
As far as I can see an Foundation import would be sufficient:
```suggestion
import Foundation
```
If you remove the `description` property you can even remove the import.

Sources/ApodiniDatabase/DatabaseModel.swift
```diff
@@ -0,0 +1,23 @@
...
+    ///Has to be overwritten to use `Update` handler.
+    func update(_ object: Self)
+}
+
+internal extension DatabaseModel {
+    static func fieldKey(for string: String) -> FieldKey {
+        if let key = Self.keys.first(where: { $0.description == string }) {
+            return key
+        } else {
+            fatalError("Unexpectedly found nil. Failed to find a Fieldkey under the given string.")
```
PSchmiedmayer: I would suggest to add the name of the key using string interpolation to the error description to make it easier to identify the issue when encountering the error.

Sources/ApodiniDatabase/DatabaseModel.swift
```diff
@@ -0,0 +1,23 @@
...
+        }
+    }
+    
+    static var description: String {
+        NSStringFromClass(Self.self)
+    }
```
PSchmiedmayer: As far as I can see the property is not needed. I would suggest removing it.

Sources/ApodiniDatabase/Handlers/Delete.swift
```diff
@@ -0,0 +1,20 @@
...
+    private var database: Fluent.Database
+
+    @Parameter(.http(.path))
+    var id: Model.IDValue
+    
+    public func handle() -> EventLoopFuture<UInt> {
+        Model.find(id, on: database)
+            .unwrap(orError: Abort(.notFound) )
+            .flatMap { $0.delete(on: database ) }
+            .transform(to: HTTPStatus.ok.code )
```
PSchmiedmayer, line 18: I think just returning a `Void` would be sufficient. We will need some way to indicate a HTTP status code for a response. I would suggest to add an GitHub issue to document the need for such a feature and maybe think about a way to return a response as well as an appropriate status code.
I would suggest to just remove nothing in the case the deletion was successful.

Sources/TestWebService/main.swift
```diff
@@ -122,6 +121,7 @@ struct TestWebService: Apodini.WebService {
     }
 
     @PathParameter var userId: Int
+    @PathParameter var dummy: String
```
PSchmiedmayer: As far as I can see this `@PathParameter` can be removed.

Sources/TestWebService/main.swift
```diff
@@ -158,6 +158,10 @@ struct TestWebService: Apodini.WebService {
             Random()
         }
     }
+    
+    var configuration: Configuration {
+        HTTP2Configuration()
+    }
```
PSchmiedmayer: I am not sure if this added configuration is relevant for the PR

Sources/TestWebService/main.swift
```diff
@@ -4,9 +4,8 @@
 //
 //  Created by Paul Schmiedmayer on 7/6/20.
 //
-
 import Apodini
-
+import ApodiniDatabase
```
PSchmiedmayer: Feel free to add some examples `ApodiniDatabase` examples in the `TestWebService` 👍

Sources/ApodiniDatabase/Handlers/Read/QueryBuilder.swift
```diff
@@ -0,0 +1,104 @@
...
+import Apodini
+import Foundation
+import Fluent
+@_implementationOnly import Vapor
+@_implementationOnly import Runtime
+
+internal struct QueryBuilder<Model: DatabaseModel> {
+    internal var debugDescription: String {
+        parameters.debugDescription
+    }
```
PSchmiedmayer: Doesn't seem to be used, I would consider removing it.

Sources/ApodiniDatabase/Handlers/Read/QueryBuilder.swift
```diff
@@ -0,0 +1,104 @@
...
+    
+    private let type: Model.Type
+
+    private var fieldKeys: [FieldKey] {
+        type.keys
+    }
```
PSchmiedmayer: Doesn't seem to be used, I would consider removing it.

Sources/ApodiniDatabase/Handlers/Read/QueryBuilder.swift
```diff
@@ -0,0 +1,104 @@
...
+        for (key, value) in parameters {
+            queryBuilder.filter(key: key, method: .equal, codableValue: value)
+        }
+        return queryBuilder.all()
+    }
+    
+    @discardableResult
+    internal static func fieldKeys<Model: DatabaseModel>(for type: Model.Type) -> [FieldKey] {
+        type.keys
+    }
```
PSchmiedmayer: Doesn't seem to be used, I would consider removing it.

Sources/ApodiniDatabase/Handlers/Read/Read.swift
```diff
@@ -0,0 +1,40 @@
...
+/// A Handler that returns an array of `DatabaseModel` which all conform to the query parameters of the request.
+/// It accepts all parameters that represent properties of the specified `DatabaseModel`.
+/// It uses the database that has been specified in the `DatabaseConfiguration`.
+/// An Example using a REST request:
+/// Group("api", "birds") {
+///     Read<Bird>()
+/// }
+/// Sending a request to  ~/api/birds/birds?age=19&name=Foo would 
+///return an array of `Bird` object that have an age of 19 and the name Foo.
+public struct Read<Model: DatabaseModel>: Handler {
```
PSchmiedmayer, line 14: Looks nice to me! I like the longer documentation on top. It would be interesting to also have a `ReadOne` or `ReadSingle` or similar `Handler` that only returns a single instance based on an id passed in using a path parameter.

Sources/ApodiniDatabase/Handlers/Update/Updater.swift
```diff
@@ -0,0 +1,58 @@
...
+            result[Model.fieldKey(for: entry.0)] = entry.1
+        }
+        self.model = model
+        self.modelID = modelId
+    }
+    
+    func executeUpdate(on database: Database) -> EventLoopFuture<Model> {
+        if let model = model {
+            return Model.find(modelID, on: database)
+                .unwrap(orError: Abort(.notFound))
```
PSchmiedmayer: This seems to be a duplicated from the else case. I would suggest to split up the method into two smaller ones that split up after this point to make the method more readable and then you can also document the decision mentioned in the [comment by @theMomax below](https://github.com/Apodini/Apodini/pull/70/files#r559168012) in each of the method documentations.

Sources/ApodiniDatabase/Handlers/Update/Updater.swift
PSchmiedmayer: This call returns before the model is actually updated. I would suggest to use a `flatMap` here and return the future from the `.transform(to: model)` call:
```suggestion
                .flatMap { foundModel -> EventLoopFuture<Model> in
                    foundModel.update(model)
                    return foundModel
                        .update(on: database)
                        .transform(to: model)
                }
```

Sources/ApodiniDatabase/Handlers/Update/Updater.swift
```diff
@@ -0,0 +1,58 @@
...
+                .map { model -> Model in
+                    for child in Mirror(reflecting: model).children {
+                        guard let visitable = child.value as? UpdatableFieldProperty,
+                              let label = child.label
+                        else { continue }
+                        let fieldKey = Model.fieldKey(for: label.trimmed())
+                        if let value = properties[fieldKey] {
+                            if visitable.accept( ConcreteUpdatableFieldPropertyVisitor(updater: value) ) {
+                                _ = model.update(on: database).transform(to: model)
```
PSchmiedmayer: This method also returns before the model is actually saved to the database. Similar to above you will need to use `flatMap` here. Otherwise errors when saving an element to the database would get lost in the void.

Sources/ApodiniDatabase/Handlers/Update/Updater.swift
```diff
@@ -0,0 +1,58 @@
...
+                    return model
+                }
+        } else {
+            return Model.find(modelID, on: database)
+                .unwrap(orError: Abort(.notFound))
+                .map { model -> Model in
+                    for child in Mirror(reflecting: model).children {
+                        guard let visitable = child.value as? UpdatableFieldProperty,
+                              let label = child.label
+                        else { continue }
```
PSchmiedmayer: This method is overall very hard to read. Maybe this functionality can be moved to an extension of the `Model` class? In addition it is hard to see the guard statement here. I would suggest a different formatting:
```suggestion
                        guard let visitable = child.value as? UpdatableFieldProperty,
                              let label = child.label else { 
                            continue
                        }
                        
```

Sources/ApodiniDatabase/Handlers/Update/Updater.swift
```diff
@@ -0,0 +1,58 @@
...
+                        }
+                    }
+                    return model
+                }
+        }
+    }
+}
+
+fileprivate extension String {
+    func trimmed() -> Self {
```
PSchmiedmayer, line 64: I would suggest to add a more descriptive name and some documentation and examples on why this is necessary.

Sources/ApodiniDatabase/Handlers/Update/Updater.swift
```diff
@@ -0,0 +1,58 @@
...
+                        guard let visitable = child.value as? UpdatableFieldProperty,
+                              let label = child.label
+                        else { continue }
+                        let fieldKey = Model.fieldKey(for: label.trimmed())
+                        if let value = properties[fieldKey] {
+                            if visitable.accept( ConcreteUpdatableFieldPropertyVisitor(updater: value) ) {
+                                _ = model.update(on: database).transform(to: model)
+                                return model
+                            } else {
+                                fatalError("An error occurred while updating single parameters of the model ")
```
PSchmiedmayer: Please add a more descriptive error message that e.g. includes the `fieldKey` and more context information.

Sources/ApodiniDatabase/PropertyVisitors/UpdatableFieldProperty.swift
```diff
@@ -0,0 +1,37 @@
...
+    associatedtype Value
+
+    func visit<Model, V>(_ property: inout FieldProperty<Model, V>) -> Value
+}
+
+/// A protocol to make the `FieldProperty` of a database model updatable at run time.
+/// It returns `true` if the update of the given `FieldProperty` was successful, else `false`
+protocol UpdatableFieldProperty {
+    func accept<Visitor: UpdatableFieldPropertyVisitor>(_ visitor: Visitor) -> Visitor.Value where Visitor.Value == Bool
+}
```
PSchmiedmayer: As far as I can see there is no other implementation then the one defined in this scope and all use `Bool` as the concrete type for `Value`. Therefore I would suggest to remove the `associatedtype` completely and only use concrete types.

Sources/ApodiniDatabase/PropertyVisitors/UpdatableFieldProperty.swift
```diff
@@ -0,0 +1,37 @@
...
+
+/// A concrete implementation of the `UpdatableFieldPropertyVisitor` that updates a given `FieldProperty` and returns an `true` if the update was successful
+struct ConcreteUpdatableFieldPropertyVisitor: UpdatableFieldPropertyVisitor {
+    typealias Value = Bool
```
PSchmiedmayer: Can be removed if the comment above is applied.

Sources/ApodiniDatabase/PropertyVisitors/VisitableFieldProperty.swift
```diff
@@ -0,0 +1,57 @@
...
+
+/// A protocol to make the `FieldProperty` of a database model accessible at run time.
+/// It returns an `AnyCodable` object which contains the value associated with this `IDProperty`
+protocol VisitableIDProperty {
+    func accept<Visitor: IDPropertyVisitor>(_ visitor: Visitor) -> Visitor.Value where Visitor.Value == TypeContainer
+}
+
+/// A protocol all field property visitors have to conform to. It returns whatever has been specified as `Value`.
+protocol FieldPropertyVisitor where Value: Codable {
+    associatedtype Value
```
PSchmiedmayer: The `associatedtype` is not needed in the current implementation as far as I can see as you only use the concrete type `TypedContainer` everywhere. I would suggest removing it to reduce the complexity of reading the code and steering the type checker a bit less 😅

Sources/ApodiniDatabase/PropertyVisitors/VisitableFieldProperty.swift
```diff
@@ -0,0 +1,57 @@
...
+    
+    func visit<Model, V>(_ property: IDProperty<Model, V>) -> Value where Model: DatabaseModel
+}
+
+/// A concrete implementation of the `FieldPropertyVisitor` that visits a given `FieldProperty` and returns an `AnyCodable` object
+struct ConcreteTypeVisitor: FieldPropertyVisitor {
+    typealias Value = TypeContainer
+    
+    func visit<Model, V>(_ property: FieldProperty<Model, V>) -> Value where Model: Fields, V: Decodable, V: Encodable {
```
PSchmiedmayer: You can combine the `V: Decodable, V: Encodable` requirement into a single one. In addition you don't need the `typealias` here if you remove the `associatedType` above.
```suggestion
    func visit<Model, V>(_ property: FieldProperty<Model, V>) -> TypeContainer where Model: Fields, V: Codable {
```

Sources/ApodiniDatabase/PropertyVisitors/VisitableFieldProperty.swift
```diff
@@ -0,0 +1,57 @@
...
+        TypeContainer(with: property.value)
+    }
+}
+
+/// A concrete implementation of the `IDPropertyVisitor` that visits a given `IDProperty` and returns an `AnyCodable` object
+struct ConcreteIDPropertyVisitor: IDPropertyVisitor {
+    typealias Value = TypeContainer
+    
+    func visit<Model, V>(_ property: IDProperty<Model, V>) -> TypeContainer where Model: DatabaseModel, V: Decodable, V: Encodable {
```
PSchmiedmayer: Same comment as found in `ConcreteTypeVisitor`.

PSchmiedmayer (changes requested):
@hendesi I have taken a closer look at the PR and had a few comments here and there about simplifications and some `EventLoopFuture` improvements in the code. 
Sometimes I have added a comment about unused code, feel free to let me know if I missed some usages of the code or if this will be come essential for some additional features that build on top of the existing implementation.

In addition I would suggest to take a look at the [automated comments from SonarCloud]( https://sonarcloud.io/project/issues?id=Apodini_Apodini&pullRequest=70&resolved=false&types=CODE_SMELL) that might also be interesting and helpful for improving the PR.

Thanks for already decoupling the implementation into a separate target 🚀
The overall functionality and feature set seems great to me, really looking forward to seeing this feature in Apodini. Thanks for all the work that went into the PR and the long wait until `EventLoopFutures` were added back into develop.

Sources/ApodiniDatabase/Handlers/Delete.swift
hendesi: Not sure if I'm missing something, but it's currently not possible to return `EventloopFuture<Void>` as a `Response`, because Void can't conform to `ResponseTransformable`

hendesi:
@PSchmiedmayer Thanks for the detailed feedback! I think I addressed all of it. 👍 
I have only one [comment](https://github.com/Apodini/Apodini/pull/70#discussion_r560055814) regarding the `Delete` handler.
Other than that, I definitely want to add a `ReadSingle` Handler, that is a good idea. But I would vote for moving this to a separate pr. The focus should be to have a functional version of the database in develop and not to put every feature into one PR 😅 .

Sources/ApodiniDatabase/Handlers/Delete.swift
PSchmiedmayer: Ah, I see. We should probably add an implementation to transform `Void` to a response `.final(Void())`. I would suggest to create an issue about this to discuss if the decisions to return a `.final(Void())` would be correct. An alternative would be to return `.nothing` but this would not create any response to the client which would be needed here, therefore I would not suggest to use `.nothing` here.

Sources/ApodiniDatabase/Context/TypeContainer.swift
```diff
@@ -184,5 +184,6 @@ extension TypeContainer: LosslessStringConvertible {
         // As query parameters are currently internally used a `.lightweight` and therefore not initialized using this init,
         // there is currently no mapping for this and will always be defaulted to `.noValue`.Will be added in the future.
         self = .noValue
```
PSchmiedmayer: You can remove this assignment to `self` as the initializer will not terminate due to the `fatalError` below.

Sources/TestWebService/main.swift
PSchmiedmayer: I there is no example in the `TestWebService` I would suggest to remove the import.

Sources/ApodiniDatabase/Handlers/Delete.swift
PSchmiedmayer: Ah, I see. We should probably add an implementation to transform `Void` to a response `.final(Void())`. I would suggest to create an issue about this to discuss if the decisions to return a `.final(Void())` would be correct. An alternative would be to return `.nothing` but this would not create any response to the client which would be needed here, therefore I would not suggest to use `.nothing` here.

Sources/ApodiniDatabase/Context/DatabaseInjectionContext.swift
```diff
@@ -0,0 +1,18 @@
...
+///A Protocol which provides info about the expected the type of a `FieldKey`.
+protocol DatabaseInjectionContext {
+    ///A `Fluent.FieldKey`
+    var key: FieldKey { get }
+    ///The expected type for the fieldkey
+    var value: TypeContainer { get }
+}
+
+///A struct implementing `DatabaseInjectionContext` and containing a fieldkey and the expected type for that key.
+struct ModelInfo: DatabaseInjectionContext, Equatable {
```
PSchmiedmayer: I would suggest to remove the conformance to `Equatable` as there is no implementation in the type anymore and it doesn't seem to be sued anywhere.

PSchmiedmayer (approved):
@hendesi Thanks for incorporating all the feedback! I agree that the `SingleRead` should be a separate PR. I would suggest to create multiple issues from this PR:

1. The issue with multiple currently conflicting parameters for the `RESTInterfaceExporter` that you previously commented using a warning
2. The issue with returning `Void` from a `handle()` function and how to return different status codes then `200` as raised by the [thread and issue in the `Delete` component](https://github.com/Apodini/Apodini/pull/70/files/00759a1d3bf1f57b10ed377218010dd67d9848e0#r559817626).

Apart from this I only had a few smaller comments. I would already approve the PR as I don't think there is a need for an additional review after the issues are created and the small comments are addressed 👍

PSchmiedmayer:
In addition please take a look at the issues detailed by SonarCloud that might be resolved before merging the PR: https://sonarcloud.io/project/issues?id=Apodini_Apodini&pullRequest=70&resolved=false&types=CODE_SMELL

PR #73, opened 2020-12-21
Make EventLoopFuture conform to Encodable

### LINKED ISSUE #72: [BUG] EventLoopFuture should be Encodable
**Describe the bug**

By replacing `ResponseEncodable` with `Encodable`, we lost Vapor's `EventLoopFuture: ResponseEncodable` extension, meaning our handle() functions can't return `EventLoopFutures` anymore.

**Relevant test case:** `CustomComponentTest.swift`

The invalid return type caused the test component's Response type to default to Never, which we didn't catch because this specific test case directly invokes handle() instead of going via the witness table.

**Expected behavior**

- `EventLoopFuture conforms to `Encodable`
- Unit Tests should test for this

tgymnich:
Make `EventLoopFutur` conform to `Encodable`

Fixes #72


tgymnich:
This approach has some issues, since we have to await the future when we want to encode it. I think this is the best we can do with Encodable, so we might need a different protocol / function

Sources/Apodini/Components/Component.swift
```diff
@@ -41,3 +41,10 @@ extension Component {
         }
     }
 }
+
+extension EventLoopFuture: Encodable where Value: Encodable {
+    public func encode(to encoder: Encoder) throws {
+        let encodable = try self.wait()
```
lukaskollmer: this is a blocking operation, right?

can we safely block the thread, or would that impact other parts of the server (ie not being able to process other incoming requests)?

tgymnich: Yeah its blocking. I need to come up with another function that somehow returns an `EventLoopFuture`

lukaskollmer: just a guess, but if we configure vapor in a way that each request is guaranteed to get its own thread, wouldn't we get away with blocking here?

nerdsupremacist:
In general I think that making EventLoops conform to encodable is not the right approach.
Maybe using encodable is just not the right idea.

A possible idea would be to have our own protocol like I did here for my thesis: https://github.com/nerdsupremacist/GraphZahl/blob/develop/Sources/GraphZahl/Resolution/OutputResolvable/OutputResolvable.swift

This one is a protocol that:
- Returns the type information needed for the documentation
- Resolves the object into something that can be encoded (https://github.com/nerdsupremacist/GraphZahl/blob/develop/Sources/GraphZahl/Resolution/OutputResolvable/Output.swift)

It has some other functions and properties in there but those are not important. Maybe we can use this as an idea of how to move away from encodable and towards a custom solution

bauer-andreas:
> This approach has some issues, since we have to await the future when we want to encode it. I think this is the best we can do with Encodable, so we might need a different protocol / function

My original idea was that we could create a "new" encode function which returns a EventLoopFuture so we don't need to block. I think this is a bit hacky, as we probably need to somehow extends the Encodable protocol similar to this https://github.com/Apodini/Apodini/commit/c1d8e5a77048fed4e4a23e41f0d995eea48a2083.
But this should allow encoding a EventLoopFuture without blocking **if** we call the right encode function. We might just disallow calling the sync encode function on EventLoopFutures.

> In general I think that making EventLoops conform to encodable is not the right approach.
> Maybe using encodable is just not the right idea.

@nerdsupremacist I get where you coming from, and I think for the long term this might be the way to got (even If I haven't completely understood the linked examples and how exactly we would incorporate it).

Though I think we need a solution for the short term, as this issues has a rather big impact on the work of others.
I could see two scenarios:
* Either we revert the change and move back to ResponseEncodable and wait for our custom solution
* or we make EventLoopFuture conform to Encodable as outlined in this PR and wait for our custom solution.

With the changes outlined above it could make sense to go for the second option? In any case a decision here is necessary.
CC: @hendesi

hendesi:
I agree with @Supereg. We should make a decision here. I like the solution proposed by @nerdsupremacist, but as this needs some time to implement and as this blocks, as already mentioned, some other work, I would vote for reverting the change and postponing this to a later point. 
If you are interested, we could also have a meeting about this issue this week.

tgymnich:
Why not apply this patch instead of reverting the whole thing. Reverting is going to be a whole lot of work.

Sources/Apodini/Components/Component.swift
```diff
@@ -5,6 +5,7 @@
 //  Created by Paul Schmiedmayer on 6/26/20.
 //
 
+import Foundation
```
lukaskollmer, line 8: Is this necessary?

Sources/Apodini/Components/Codable.swift
```diff
@@ -0,0 +1,37 @@
...
+//  Created by Tim Gymnich on 22.12.20.
+//
+
+import Foundation
+import NIO
+
+
+extension EventLoopFuture: Encodable where Value: Encodable {
+    public func encode(to encoder: Encoder) throws {
+        let encodable = try self.wait()
```
bauer-andreas, line 14: This would still block the thread right? 

I sketched my idea as described above: https://github.com/Apodini/Apodini/compare/async-encodable-eventloopfuture

I isn't fully cleaned up, and it still lacks implementation using that part in the RequestHandler. (ResponseTransformer where never made for this I think, at least they would have been passed a EventLoopFuture and I would guess this wasn't intended?). 

ut something like this should work to keep it non blocking. And a similar approach is currently used for the implementation of `Action<...>`, meaning a clean implementation is possible for sure.

PSchmiedmayer:
@hendesi @Supereg @tgymnich @lukaskollmer @nerdsupremacist 
I just checked out this PR, I must have missed it last week when looking over all open PRs. 
I agree with @tgymnich that reverting the change is not a good idea and that we should find a good solution that continues the path to decoupling Vapor from the public API but also introduces us a way to use `EventLoopFutures` for now. 
While I agree with @nerdsupremacist that we will eventually need a better solution then a simpel `ApodiniEncodable`/`RequestEncodable` in the future I vote for merging a non blocking idea like the one of @Supereg (`ApodiniEncodable`) that is very similar to Vapor's approach but under our control and part of our API into develop to unblock the work of others relying on `EventLoopFutures` and to also bring some attention to #82 👍

PSchmiedmayer (changes requested):
@hendesi @Supereg @tgymnich @lukaskollmer @nerdsupremacist 
I just checked out this PR, I must have missed it last week when looking over all open PRs. 
I agree with @tgymnich that reverting the change is not a good idea and that we should find a good solution that continues the path to decoupling Vapor from the public API but also introduces us a way to use `EventLoopFutures` for now. 
While I agree with @nerdsupremacist that we will eventually need a better solution then a simpel `ApodiniEncodable`/`RequestEncodable` in the future I vote for merging a non blocking idea like the one of @Supereg (`ApodiniEncodable`) that is very similar to Vapor's approach but under our control and part of our API into develop to unblock the work of others relying on `EventLoopFutures` and to also bring some attention to #82 👍

tgymnich:
@Supereg could you create a PR for your approach? @PSchmiedmayer I am already looking into @nerdsupremacist's idea.

bauer-andreas:
@tgymnich alrighty 👍 I see what I can do

bauer-andreas:
> @Supereg could you create a PR for your approach? @PSchmiedmayer I am already looking into @nerdsupremacist's idea.

Closing this PR in favor of #94 (made it a bit easier to rebase onto latest develop).

EDIT: @tgymnich please check if this branch can be safely deleted.

PR #74, opened 2020-12-22
WebSocket Exporter based on Shared Semantic Model

theMomax:
# *WebSocket Exporter based on Shared Semantic Model*

The WebSocket exporter hasn't yet made it into the develop branch. There only was the PR #47, which featured a stand-alone solution (i.e. one that was not based on the shared model). #47 was aborted in favor of this PR.

This PR features a couple of changes:

### :recycle: Current situation
Vapor's WebSocket API did not provide a good abstraction for Apodini's endpoint-based architecture with each endpoint having defined input- and output-types.

### :bulb: Proposed solution
I introduce a `Router` protocol:

```swift
/// A `Router` provides an endpoint-based, typed abstraction of a WebSocket connection. It uses
/// a spcific `Input` for each `register`ed endpoint to maintain state and possibly also check
/// validity of incoming messages. Each endpoint is identified by its `identifier`.
public protocol Router {
    /// Register a new endpoint on the given `identifier` using the given `opener` when a
    /// new client connects. Closing the connection may be requested by the client (a completion is sent
    /// on the input publisher) and can be executed by the server (a completion is sent on the `output`).
    func register<I: Input, O: Encodable>(
        _ opener: @escaping (AnyPublisher<I, Never>, EventLoop, Database?) ->
            (default: I, output: AnyPublisher<Message<O>, Error>),
        on identifier: String)
}
```
The whole WebSocketInfrastructure target is designed to provide a useful and usable abstraction independently of Apodini.

The `Router` protocol is implemented by the `VaporWSRouter`. This router exposes a single WebSocket endpoint only: (by default) `apodini/websocket`. Each client only ever opens one WebSocket connection. This seems to contradict the `Router` protocol's requirement of being able to `register` multiple routes for different `identifier`s. This contradiction is resolved by the protocol that is run on the single WebSocket connection:

This protocol consists of two concepts: **endpoint** and **context**. One `context` lives on exactly one `endpoint`. A `context` is identified by an `UUID`. A `endpoint` is identified by the `identifier` string passed to the `Router`'s `register` method.

Either client or server can create a new context by sending the following message:
```
{
    "context": "<UUID>",
    "endpoint": "<identifier>"
}
```

A client can send input to a `context`:
```
{
    "context": "<UUID>",
    "parameters": {
        "<name1>": <value2>,
        "<name2>": <value2>,
        ...
    }
}
```
The server can send output or error-messages to a `context`:
```
{
    "context": "<UUID>",
    "content": <Content>
}
{
    "context": "<UUID>",
    "error": <Error>
}
```
Either client or server can close a context:
```
{
    "context": "<UUID>"
}
```

This concept of `context` and `endpoint` allows for multiple parallel 'requests' on one endpoint and connection without requiring the client to open a new WebSocket connection for each and every request. I came up with this protocol. We definitely could change it (there are endless possibilities) and we will probably at least iterate on it a couple of times, but I really would like to get this merged as a first implementation.

#### Problem that is solved
Vapor's WebSocket API is quite low-level. Something higher-level was required to reduce code-complexity in the `WebSocketInterfaceExporter`.

#### Implications
None, pretty much.

### :heavy_plus_sign: Additional Information
🤷‍♂️

#### Related PRs
#47

#### Testing
The basic round-trip of a request-response message is covered by `WebSocketExporterTests.testWebSocketConnection`.

#### Reviewer Nudging
First, get familiar with the `Router` protocol and make sure you understand the communication-protocol featured by the `VaporWSRouter`. If you really want to go into all the details, follow the `ConnectionResponsible` initializer (called in `VaporWSRouter.registerRouteToVapor`) and work your way up from there.

### :recycle: Current situation
Again: the WebSocket exporter hasn't yet made it into the develop branch.

### :bulb: Proposed solution

The current implementation has the following characteristics:
- It utilizes the `SharedSemanticModelBuilder`
- It can export `Handler`s that only contain `@Parameter` property wrappers (i.e. Request-Response)
- It can differentiate between a parameter being absent, explicitly `null` or provided.
- This first version does not work with `Action.nothing` or `Action.send` and it will not auto-close the `context` on `Action.final`
- This first version does not synchronize/serialize evaluation of `Handler`s which return an `EventLoopFuture` -> undefined behavior.

#### Problem that is solved
As opposed to #47, a lot of the code was moved/re-implemented in the shared model which will reduce code-duplication in the future. As opposed to develop, a WebSocket exporter exists🎉.

#### Implications
None.

### :heavy_plus_sign: Additional Information
When implementing the exporter I put a lot of effort into moving as much logic as possible into the shared model. For more detail see the next sections.

#### Related PRs
#47

#### Testing
The exporter is tested by `WebSocketInterfaceExporterTests`.

#### Reviewer Nudging
Since a lot of the low-level logic is done in the WebSocketInfrastructure target and the SharedSemanticModelBuilder, the exporter itself is rather short. You can find it in `WebSocketInterfaceExporter.swift`. This part should be rather easy.

### :recycle: Current situation
The `SharedSemanticModelBuilder` did provide a stateless `RequestHandler`, which was basically just a closure. With the WebSocket exporter I also wanted to include a `Mutability` option (with values `.constant` or `.variable`). In order to validate that `constant` `@Parameter`s are not altered after the first request, a stateful mechanism was required. The `ConnectionContext` is the basis for this stateful validation.

### :bulb: Proposed solution

`Endpoint.createRequestHandler` was replaced by

```swift
func createConnectionContext<I: InterfaceExporter>(for exporter: I) -> AnyConnectionContext<I>
```

with the `ConnectionContext` defined as follows:

```swift
protocol ConnectionContext {
    associatedtype Exporter: InterfaceExporter
    
    mutating func handle(
        request exporterRequest: Exporter.ExporterRequest,
        eventLoop: EventLoop
    ) -> EventLoopFuture<Action<AnyEncodable>>
}
```

The `ConnecitonContext.handle` allows access to the internal `RequestHandler`'s functionality, but it is `mutating`! Thus, the `ConnectionContext` can first do stateful operations (e.g. validation) before handling the request.

#### Problem that is solved
The shared model could not handle state for a connection (i.e. across multiple requests).

#### Implications
All usages of the `RequestHandler` had to be refactored. This is straightforward for request-response exporters:

```swift
let requestHandler = endpoint.createRequestHandler(for: self)
...
requestHandler(request)
```
was replaced by

```swift
var context = endpoint.createConnectionContext(for: self)
...
context.handle(request)
```

#### Related PRs
This could provide a basis for handling `@Connection`(#50) and `@State`(https://github.com/Apodini/Apodini/tree/8694945d89722342ed5eb76dd6220b020a73a841) inside the shared model.

#### Testing
As this code is a replacement  for the `RequestHandler`, all the previous tests cover the new code.

#### Reviewer Nudging
The `ConnectionContext` itself isn't that interesting. Before getting started with `InternalConnectionContext` I suggest you fist read the next section on `Validator`.

### :recycle: Current situation
Input validation was previously handled via the `ParameterRetrievalDelegation` inside the injection process of `RequestInjectable`. This had four disadvantages:
* it was stateless
* errors could not be handled -> fatalError on invalid user-input
* it required the `Handler` to be available (not blocked by another request)
* it made the injection process even more complicated

### :bulb: Proposed solution
A stateful `Validator` (which lives on the `ConnectionContext`) validates the `request` before the injection-process starts. This implementation is called `EndpointValidator` The `EndpointValidator` returns a `ValidatedRequest` that can be injected without having to handle errors.

```swift
/// A `Validator` is a stateful construct that checks incoming `request`s for a specific connection
/// handled by a specific `InterfaceExporter`.
protocol Validator {
    associatedtype Exporter: InterfaceExporter
    associatedtype Input
    associatedtype Output
    
    /// Checks if the `request` is valid by the `Validator`'s criteria.
    /// - Throws: A `ValidationError` is thrown if `request` is invalid for the current state.
    /// - Invariant: The function must not mutate `self` if an error is thrown.
    mutating func validate(_ request: Exporter.ExporterRequest, with input: Input) throws -> Output
    
    /// Resets the `Validator` to its state before the most recent call to `validate`. This can be
    /// used to reset the `Validator` if execution was aborted due to some logic that is **not**
    /// covered by this `Validator`.
    mutating func reset()
}
```

The `InternalConnectionContext` uses the `EndpointValidator` in its `handle` function:

```swift
mutating func handle(
        request exporterRequest: I.ExporterRequest,
        eventLoop: EventLoop
    ) -> EventLoopFuture<Action<AnyEncodable>> {
        do {
            let validatedRequest = try validator.validate(exporterRequest, with: eventLoop)
            
            return self.requestHandler(request: validatedRequest)
        } catch {
            return eventLoop.makeFailedFuture(error)
        }
    }
```

#### Problem that is solved
* Validation can have state
* `Mutability` is validated
* throwing fatalError if injection fails is fine now, as this should never happen on a validatedRequest.

#### Implications
All usages of `ApodiniRequest` had to be refactored to use the `Validator` and `ValidatedRequest`.

#### Related PRs
The concepts introduced by #81 were reworked and extended.

#### Testing
The basics were already covered by `ParameterRetrievalTests`.

#### Reviewer Nudging
Now you can start reviewing at `InternalConnectionContext`. Work your way through the `EndpointValidator` afterwards.


Sources/Apodini/Properties/Options/Mutability.swift
```diff
@@ -0,0 +1,27 @@
...
+    /// The `@Parameter` cannot  be updated once it has been set. Default values can still be overridden.
+    case constant
+}
+
+extension PropertyOptionKey where PropertyNameSpace == ParameterOptionNameSpace, Option == Mutability {
+    static let mutability = PropertyOptionKey<ParameterOptionNameSpace, Mutability>()
+}
+
+extension AnyPropertyOption where PropertyNameSpace == ParameterOptionNameSpace {
+    /// An HTTP specific option that indicates how the `@Parameter` property wrapper should be interpreted by interface exporters using HTTP to identify and multiplex components.
```
moritzschuell: nit: This does not belong here, right?

moritzschuell (approved):
Looks great 🥳 Thank you!
Was quite hard to digest, to be honest. I mostly focused on the parts in the Apodini target.

Sources/Apodini/Properties/Options/Mutability.swift
theMomax: you're right, thanks🙃

PR #77, opened 2020-12-22
Tame Codecov

tgymnich:
- Changed total coverage range from 70% - 100% to: 60% - 90% to allow for some files to not be covered.
- Rounding up to avoid -0.01% coverage blocking anyone
- Allow PR coverage to be 15% smaller than total project coverage.

Feel free to suggest adjustments to this

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -8,7 +8,7 @@ codecov:
 coverage:
   range: "60...90"
   precision: 2
-  round: down
+  round: up
   status:
     # The project status measures overall project coverage and compares it against the base of the pull request
     project:
--- a/codecov.yml
+++ b/codecov.yml
@@ -6,9 +6,24 @@ codecov:
   require_ci_to_pass: yes
 
 coverage:
+  range: "60...90"
   precision: 2
   round: down
-  range: "70...100"
+  status:
+    # The project status measures overall project coverage and compares it against the base of the pull request
+    project:
+      default:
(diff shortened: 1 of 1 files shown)


codecov.yml
```diff
@@ -6,9 +6,24 @@ codecov:
...
   precision: 2
-  round: down
-  range: "70...100"
+  round: up
+  status:
+    # The project status measures overall project coverage and compares it against the base of the pull request
+    project:
+      default:
+        # Choose a minimum coverage ratio that the commit must meet to be considered a success
+        target: auto
```
tgymnich, line 17: auto = total project coverage

codecov.yml
```diff
@@ -6,9 +6,24 @@ codecov:
   require_ci_to_pass: yes
 
 coverage:
+  range: "60...90"
   precision: 2
-  round: down
-  range: "70...100"
+  round: up
```
bauer-andreas, line 11: Isn't `up` the exact opposite of what we want?
Our goal is e.g. for an exact coverage of `45.153123%` to be round to `45.15%`, right? According to the [docs](https://docs.codecov.io/docs/coverage-configuration) this is done using `round: down`.

tgymnich: Depends on the way you look at it. You can look at Repository coverage or PR coverage. Which makes me believe that it does not matter. So no hard feelings there. I'm fine with it either way.

bauer-andreas (approved):

Issue #80, opened 2020-12-23
[QUESTION] Do we have a defined strategy how to determine if a parameter is optional?


theMomax:
### The Question

Do we have a defined strategy how to determine if a parameter is optional? What is this strategy?

### Background

I came across this question when I tried to use an `@Parameter` with an optional value with my own WebSocket exporter. The `@Parameter` internally stores an optional of `Element`. If this value is `nil` when the `@Parameter`'s `wrappedValue` is accessed, this results in a `fatalError`. The `@Parameter` does not check if `Element` is `ExpressibleByNilLiteral`. In that case `nil` would still be a valid option.

Basically the exporter currently decides if the `@Parameter`'s `element` is

* an `Optional<Element>(Optional<SomeType>(nil))` (which results in no fatal error)
* or an `Optional<Element>(nil)` (which results in fatal error)

when the request does not provide a value for an `@Parameter<Element> where Element: ExpressibleByNilLiteral`.

### Possible Solutions

All exporters have to use the same strategy. I currently see four strategies:

#### A: Optionality is declared by usage of `Optional`
```swift
// optional
@Parameter var firstName: String?
// required
@Parameter var surname: String
```
Pros: 
* feels natural

Cons:
* programmer cannot have a required parameter that can be `null`

#### B: Optionality is declared by Option
 ```swift
// optional
@Parameter(.necessity(.optional)) var firstName: String?
// required (default)
@Parameter var surname: String
```
Pros:
* programmer can have a required parameter that can be `null` (and is expressed as an `Optional` in code)

Cons:
* feels less natural
* more to write
* access to `wrappedValue` of `@Parameter` where `Element` is **not** `ExpressibleByNilLiteral` still results in fatal error

#### C: Optionality changes default Option
 ```swift
// optional (can be set to null explicitly)
@Parameter var firstName: String?
// required (cannot be set to null at all)
@Parameter var surname: String
// optional (cannot be set to null explicitly, invalid access results in fatal error)
@Parameter(.necessity(.optional)) var firstName2: String
// required (must be set to null explicitly)
@Parameter(.necessity(.required)) var surname2: String?
```
Pros:
* first two options feel natural
* programmer can have a required parameter that can be `null`

Cons:
* third variation can still result in fatal error
* more complicated to handle internally

#### D: Optionality is declared by usage of `Optional` with override
 ```swift
// optional (can be set to null explicitly)
@Parameter var firstName: String?
// required (cannot be set to null at all)
[...]

PR #81, opened 2020-12-25
Rework parameter retrieval

bauer-andreas:
**tldr**:  
I struggled using the current way of retrieving parameters in the REST exporter. Thus I reworked the parameter retrieval process. I order for that to work I need to make some infrastructure changes on how we model requests and how the request handler works (we are now moving away from the closure thingy).
When reworking the shared models I added proper documentation.

This PR contains some **breaking changes** of the `InterfaceExporter` API. Thus mentioning everybody who is currently integrating with the shared model builder. 
@lschlesinger @tgymnich @moritzschuell @theMomax @hendesi @sadikekin
 Additionally this PR may be important to the @hendesi as to how to integrate his "dynamic" approach to Parameters in his `ReadComponent`.

I do a lot of "magic" to somehow gain back type information, any feedback from @nerdsupremacist would be welcome to highlight parts where something can be made cleaner.

# Rework parameter retrieval

Since we merged the Parameter manifesto #23 there was no clear vision on how the retrieval process of Parameters actually work. The PR #61, which primarily dealt with replacing `ResponseEncodable` with `Encodable`, added a basic parameter retrieval process. This was done using a `Request` protocol and the required method `func parameter<T: Codable>(for parameter: UUID) throws -> T?` (where `T` was the generic `Element` from the `@Parameter` and the UUID `parameter` was the id of the `@Parameter). However the PR lacked any implementation for this (e.g. for REST).

When trying to implement the parameter retrieval for REST I encountered several problems/inconveniences:
* As the `id` is passed as the only argument, every Exporter would always need to do a `UUID` lookup to find the appropriate `EndpointParameter`. This creates code duplication.
* For optional parameter definitions like `@Parameter var id: String?` the generic type `T` passed to `parameter` would be `Optional<String>`, making it impossible to decode the parameter as one can't access the wrapped type. (At least for Vapor based exporters, as the require the wrapped type).  
* As the method is returning `T?` it creates the confusion, that `T` is probably a non optional type and that a `Optional.none` should be returned if the parameter is not found. This is not the case as `T` could be `Optional<String>` and returning `nil` will always signify "There was no parameter injected into `@Parameter`". After running `.inject` a parameter value must **always** be injected (the value might be nil) or a error must be thrown to signify some sort of bad request/arguments.
* For parameter types conforming `LosslessStringConvertible` it is impossible to retrieve that type and init an instance of `T` using the `LosslessStringConvertible` initializer (relevant for REST path parameters). At this point of execution the type information is lost.
* Checking if a required parameter has a value needs to be done by the exporter, creating code duplication. Additionally it creates the possibility, that different exporters treat require/optional differently.

Other smaller issues with the code base right now:
* @theMomax: The `EndpointParameter` currently doesn't save the parameter type in a generic way, complicating exporting Endpoints with parameters.

## :bulb: Proposed solution

When thinking about how I would like parameter retrieval to work, I envisioned the following func implemented by a `InterfaceExporter`.
```swift
func retrieveParameter<Type: Decodable>(_ parameter: EndpointParameter<Type>, for request: ExporterRequest) throws -> Type??
```

The retrieval function is passed two things, the parameter - describing what and how to retrieve - and the request, describing the source to retrieve the parameter from.  
Some details:
* `ExporterRequest` is basically anything the Exporter wants a request to be. Could be a `String`, but can also be something like `Vapor.Request`.
* For optional parameters like `@Parameter var name: String?` the `Type` generic holds the wrapped type of the Optional. In the preceding example this would be `String`.
* The return type is `Type??` (see https://github.com/Apodini/Apodini/pull/81#discussion_r548919891):  
  * Returning `nil` will signify that a value for the parameter could not be found. Setting the default value or erroring on required parameters is handled automatically.
  * Returning `Type??(nil)` (or using the shortcut `.null`) will signify that the "explicit nil" value was supplied (e.g. via JSON property holding `null`). On parameter definitions like `@Parameter var id: String? = "0"` this will overwrite the default value. Again, checking illegal value assignments is done automatically.
  * Otherwise you would return the value found for the parameter.
* The `func` can throw any errors (or Apodini defined default errors). Those errors will be automatically encoded in a proper error response to the user. This is currently not implemented and WIP done by @tgymnich.

As we need to preserve the `ExporterRequest` type to be able to pass it back to `InterfaceExporter.retrieveParameter` 
I needed to remodel our own `Request` type. The new `Request` struct is basically a wrapper around whatever is passed to the `RequestHandler` by the `InterfaceExporter`.  

As our current closure based request handler wasn't able to preserve this type information, I needed to remodel that as well.

### Problem that is solved

Problems described in "current situation" are solved. This PR now ships a working parameter retrieval process for the REST exporter.

As the `EndpointParameter` now adds the parameter type as a generic type, and with the addition of `InterfaceExporter.exportParameter` (refer to the docs), the problem of missing type information is solved.  

The Shared Semantic Model still doesn't check for name collisions on parameters. Different exporters have different
allowed namespaces for the different parameter types, thus this depends heavily on the exporter.  
My plan would be to ship some sort of "Namespace definition" in a future PR so we can reduce code duplication.

As the shared semantic model build creates its own `Request` instance, this is easily extensible to hold `@Evironment` values or manage Database connections.

As the new `RequestHandler` is now much more powerful (by moving away from closures), implementing additions like #50 will be much simpler to do. In general the previous closure based request handler was always painful to deal with when changing something. This should be resolved.

### Implications

As the InterfaceExporter API and the shared model get more and more complex I created two new sub directories in `Semantic Model Builder`: `InterfaceExporter` and `Model` with the following relocations:
* `./InterfaceExporter.swift` moved to `./InterfaceExporter/InterfaceExporter.swift`
* `./EndpointsTree.swift` moved and renamed to `./Model/Endpoint.swift`
* `EndpointParameter.swift` moved to `./Model/EndpointParameter.swift`

Listing the most important refactorings:
* `EndpointParameter`:
  * Add the generic type `Type`, which holds the Wrapped type for Optionals (as described above)
  * Added new `AnyEndpointParameter` protocol when you need to pass endpoint parameters in a type erasured way
  * The `name` property now properly removes the leading `_` when using the property label as parameter name 
  * `contentType` was renamed to `propertyType`
  * Added `necessity` to be able to check if a parameter is required or optional
  * Added `nilIsValidValue` to check if the parameter is defined as a optional type
  * Added `pathId` returning the path parameter identifier for .path parameters. A `@PathParameter` model, which is to be incorporated (see https://github.com/Apodini/Apodini/pull/55#issuecomment-743771845), can build on top of this.
* `Endpoint`:
  * Added the generic type `C`, which is the `Component` type modeled by the endpoint.
  * Added the property `component` which holds the originally registered `Component`. Right now its unclear if that has any use to be exposed to the exporter. Feel free to share any thoughts on this.
  * Added `func createRequestHandler(for exporter:)` to create new request handler instances (Refer to the docs on how to use them)
  * Added `exporterParameters(on exporter:)` which calls `InterfaceExporter.exporterParameter` for every parameter on the endpoint. Refer to the docs on how to use them.
* `InterfaceExporter`:
  * New associatedtype `ExporterRequest`, which needs to conform to the protocol `ExporterRequest`. This defines the type of request used by the Exporter (e.g. Vapor.Request or anything else).
  * New `func exporterParameter(parameter:) -> ParameterExporterOutput` which is called as described above. Refer to the docs or the example implementation in the REST exporter on how to use it.
  * New `func retrieveParameter(...)` as described above.
  * Signature of `func export(_ endpoint: Endpoint)` changed to `export<C: Component>(_ endpoint: Endpoint<C>) -> EndpointExportOutput` adding newly gained type information.

### Related PRs

* #23 Being the original PR introducing the Parameter proposal
* #61 Introducing the initial version of parameter retrieval
<br>

* #80 There are some open questions on what basis we consider a parameter to be optional or required
* #50 The changes done to request handling may be made after this PR is merged
* #70 Those requirements (probably) need some coordination with this PR

**Breaking Changes affecting open PRs:**
* #43 gRPC Exporter
* #74 WebSocket Exporter
* #35 (Draft) OpenAPI Exporter

### Testing

The current test cases were reworked to adapt to the new concepts and structures introduced.
Also I tried to introduce a bit of reuse when dealing if mocked requests, exporters or endpoints.

Added some test cases testing ...
* ... optional `@Parameter` definition
* ... parameter retrieval in the `RESTInterfaceExporter`

### Reviewer Nudging

I think a good entry point for review is by starting with the example of the `RESTInterfaceExporter` and then exploring how the newly added parts of the shared model works, how request handler creation is done.  
Then one could explore how RequestInjectibles are handled, specifically how the `.inject` of the `Parameter` works and then try to understand how the parameter retrieval is delegated back to the `InterfaceExporter`.

Hit me up if you need any help.


Sources/Apodini/Semantic Model Builder/InterfaceExporter/InterfaceExporter.swift
```diff
@@ -0,0 +1,98 @@
...
+    ///
+    /// If the value couldn't be found on the `ExporterRequest` for the given `EndpointParameter`, the `interfaceExporter`
+    /// must return `nil`. Checking for required parameters is done automatically.
+    ///
+    /// - Parameters:
+    ///   - parameter: The `EndpointParameter` describing the parameter for which the value should be retrieved.
+    ///   - request: The `ExporterRequest` as defined by the `InterfaceExporter`
+    /// - Returns: The retrieved value or nil if `ExporterRequest` didn't contain a value for the given `EndpointParameter`.
+    /// - Throws: Any Apodini Error or any other error happening while decoding.
+    func retrieveParameter<Type: Decodable>(_ parameter: EndpointParameter<Type>, for request: ExporterRequest) throws -> Type?
```
theMomax: As we discussed (also in context of [this](https://github.com/Apodini/Apodini/issues/80#issuecomment-751294746) message on #80), the exporter needs to know if an explicit `null` is a valid option. At least this holds true if we want to enable exporters to do parameter validation (which I would argue to be a reasonable approach until the shared semantic model is more mature).

This information could be provided via a property (enum/bool) on `EndpointParameter` which signalizes if `Type` was unwrapped from an `Optional<Type>`.

In any case the exporter must be able to express in the `retrieveParameter` if the value is `nil` (i.e. not present -> does not override default or previous value in client-stream) or `null` (i.e. explicitly null -> does override). I would suggest the following solution (return type is changed to `Optional<Optional<Type>>`:
```suggestion
    func retrieveParameter<Type: Decodable>(_ parameter: EndpointParameter<Type>, for request: ExporterRequest) throws -> Type??
```

Where .some(nil) would be the explicit null. Exporters that don't differentiate _not present_ and _null_ return `nil` or an instance of `Type` as before. Only exporters that make this differentiation have to take care about the **null** `.some(nil)`-case. The special meaning could be stressed using an extension on `Optional`:
```swift
extension Optional where Wrapped: ExpressibleByNilLiteral {
    static var null: Self {
        .some(nil)
    }
}
```

theMomax (changes requested):
As discussed, there is one problem regarding parameter retrieval which shouldn't be too difficult to fix (see comment). Other than that this PR is pretty nice! I love the abstractions you provide. Especially parameter retrieval becomes a lot easier to implement.

Sources/Apodini/Semantic Model Builder/InterfaceExporter/InterfaceExporter.swift
bauer-andreas: Resolved via 98051ce

bauer-andreas:
I made the changes discussed with @theMomax (https://github.com/Apodini/Apodini/pull/81#discussion_r548919891), which changes the return type of `InterfaceExporter.retrieveParameter` to a double optional, to be able to make the destination between "no value found" and "nil as the value", which is also referred to "explicit nil".
The 98051ce also simplified the `@Parameter` property wrapper again, moving all the logic into the `Request`.

The original PR statement was updated to reflect those latest changes.

Requesting rereview from @theMomax and review from @moritzschuell and @lschlesinger as their PRs is also affected by this breaking changes.

Sources/Apodini/Semantic Model Builder/InterfaceExporter/InterfaceExporter.swift
```diff
@@ -0,0 +1,105 @@
...
+    /// - Returns: `EndpointExportOutput` which is defined by the `InterfaceExporter`.
+    func export<C: Component>(_ endpoint: Endpoint<C>) -> EndpointExportOutput
+
+    /// This optional method can be defined to export a `EndpointParameter`.
+    /// It is called for every `EndpointParameter` on an `Endpoint` when calling `Endpoint.exportParameters(...)`.
+    /// `Endpoint.exportParameters(...)` returns an array of `ParameterExportOutput`, being what is returned by this method.
+    ///
+    /// - Parameter parameter:
+    /// - Returns: `ParameterExportOutput` which is defined by the `InterfaceExporter`
+    func exportParameter<Type: Codable>(_ parameter: EndpointParameter<Type>) -> ParameterExportOutput
```
moritzschuell, line 52: Do you have a larger use-case for this method in mind?
The current example of the `RESTInterfaceExporter` returning the name and then only using it for logging seems a bit artificial to me, and I am not sure what to use this for instead.
The added type information for the parameter is great, but we should only add it if we really need it.

moritzschuell (commented):
The code looks good to me 🎉 Just left one question below.

I do not see any problems regarding the gRPC exporter, quite the contrary, handing over the whole `EndpointParameter<Type>` to the parameter-retrieval is the adjustment that I also had to make in my branch so far. So, the change is welcome from my side 👍

Regarding your point on keeping the `component` as a property in the `Endpoint`, I think it might come in handy for the gRPC response encoding in the future. So, I would like to keep it in the endpoint for now and remove it later, if not needed.

Sources/Apodini/Semantic Model Builder/InterfaceExporter/InterfaceExporter.swift
theMomax: An exporter might need a complete model of the endpoint‘s input to process incoming messages. This would be the place to create this model I guess. My current version of the websocket exporter would use this function (at least until the shared model takes over input validation).

bauer-andreas: > Do you have a larger use-case for this method in mind?

As Max said, I think this a great way of building your internal Parameter models. Its the only way of accessing the generic type of the EndpointParameter. As somebody might need to access that type information in a generic way, when exporting its interface,  I came up with that as a solution.
But its optional to implement, meaning there is a empty default implementation. So if you don't need it, don't use it.

That was the thought behind it.

> The current example of the RESTInterfaceExporter returning the name and then only using it for logging seems a bit artificial to me, and I am not sure what to use this for instead.

It is 😅. This was only there to demonstrate how it is called and how you can access the returned values.

moritzschuell: This makes sense. Thanks for the clarification.
Based on this, I think we can utilize the `exportParameter` for the gRPC exporter. Just didn't occur to me while looking through the PR 😁

bauer-andreas: 👍

hendesi (commented):
So far the changes look good to me. The changes by #70 seem to work with the small adjustment that every parameter used for the query builder is now an optional, which it should be anyway, so good job 👍.

I have some thoughts on how the database could be added to the `ApodiniRequest`. We could either return the database stored in the environment (once #84 is merged). This would disable `func database(_ id: DatabaseID?) -> Database?` for now as `DatabaseConfiguration` only supports one database for now. 

Otherwise we could either extend the `ExporterRequest` like this: 
```
protocol ExporterRequest {
    var database: Database? { get }
}
``` 
And extend `Vapor.Request`:
```
extension Vapor.Request: ExporterRequest, WithEventLoop {
    var database: Database? {
        self.db
    }
}
```
This would be a pretty easy fix I guess and allows us to keep `func database(_ id: DatabaseID?) -> Database?`. But introduces another mandatory property to the protocol.

bauer-andreas:
> I have some thoughts on how the database could be added to the `ApodiniRequest`. We could either return the database stored in the environment (once #84 is merged). This would disable `func database(_ id: DatabaseID?) -> Database?` for now as `DatabaseConfiguration` only supports one database for now.
> 
> Otherwise we could either extend the `ExporterRequest` like this: [...]
@hendesi 

I think storing it in environment is the way to go. I don't like the approach of making the exporter responsible to supply the database. This kind works currently, as all of our exporters are based on Vapor, but is a bit limiting if our vision is to allow all sorts of exporter types.
We could extend `ExporterRequest` as an intermediate step (in the way you described it) if you think this is helpful/necessary and then move to the solution via #84.

lschlesinger (dismissed):
Thanks for the great explanation of the PR - it really helped a lot! The changes look very good to me, I've tested it with the OpenAPIExporter, it worked fine!

Sources/Apodini/Request/Request.swift
```diff
@@ -6,29 +6,151 @@
 //
 import Foundation
 import NIO
-import protocol Fluent.Database
+import protocol FluentKit.Database
+import struct FluentKit.DatabaseID
+
+protocol ApodiniRequest: CustomStringConvertible, CustomDebugStringConvertible {
```
tgymnich: Why did you rename `Request` to `ApodiniRequest`?

Sources/Apodini/Request/Request.swift
```diff
@@ -6,29 +6,151 @@
...
+        storedEndpoint
+    }
+
+    var eventLoop: EventLoop
+
+    var defaultDatabase: Database? {
+        database(nil)
+    }
+    /// Use to inject a database in our test cases
+    var databaseRetrieval: ((DatabaseID?) -> Database)?
```
tgymnich: wouldn't it be better to create a dedicated mock request struct for our tests?

Sources/Apodini/Request/Request.swift
```diff
@@ -6,29 +6,151 @@
...
-    func parameter<T: Codable>(for parameter: UUID) throws -> T?
-    /// The EventLoop associated with this request
     var eventLoop: EventLoop { get }
-    /// database
-    var database: Fluent.Database? { get }
+
+    var defaultDatabase: Database? { get }
+    func database(_ id: DatabaseID?) -> Database?
+
+    func retrieveParameter<Element: Codable>(_ parameter: Parameter<Element>, of element: Element.Type) throws -> Element
```
tgymnich: I do not understand the use of the argument `of element: Element.Type`.  Why is this needed?

Sources/Apodini/Request/Request.swift
bauer-andreas: I made `Request` a struct and needed a name for the protocol, and that's what I came up with. I didn't want to use something like `AnyRequest`. But happy for any naming suggestions.

Sources/Apodini/Request/Request.swift
bauer-andreas: Well yes, as the whole database stuff is being reworked anyways I thought this kind of "fast and dirty" way will do for now.

Sources/Apodini/Request/Request.swift
bauer-andreas: Oh, I See. Might be some leftover parameter from a previous state where the generic wasn't used with the Parameter.
This can indeed be removed.

hendesi:
> > I have some thoughts on how the database could be added to the `ApodiniRequest`. We could either return the database stored in the environment (once #84 is merged). This would disable `func database(_ id: DatabaseID?) -> Database?` for now as `DatabaseConfiguration` only supports one database for now.
> > Otherwise we could either extend the `ExporterRequest` like this: [...]
> > @hendesi
> 
> I think storing it in environment is the way to go. I don't like the approach of making the exporter responsible to supply the database. This kind works currently, as all of our exporters are based on Vapor, but is a bit limiting if our vision is to allow all sorts of exporter types.
> We could extend `ExporterRequest` as an intermediate step (in the way you described it) if you think this is helpful/necessary and then move to the solution via #84.

Although I hope #84 doesn't take too long to merged, I would prefer to have at least one working solution on develop, which would not be the case once this PR is merged. So I would vote for the intermediate solution. Once this is done, this pr seems to be good to go, at least from my point of view 👍

Sources/Apodini/Request/Request.swift
bauer-andreas: Solved via 37ebdfa

Sources/Apodini/Request/Request.swift
bauer-andreas: As the protocol is "public internal" API I exchanged the naming. Via e2981f5

Sources/Apodini/Request/Request.swift
bauer-andreas: Obsolet as of 45f826d

bauer-andreas:
> Although I hope #84 doesn't take too long to merged, I would prefer to have at least one working solution on develop, which would not be the case once this PR is merged. So I would vote for the intermediate solution. Once this is done, this pr seems to be good to go, at least from my point of view 👍

@hendesi Resolved via 45f826d. Would be great if you give it a final review 👍

theMomax (dismissed):
LGTM👍

bauer-andreas:
Having received an approval from @theMomax and @lschlesinger, and having incorporated the Feedback from @hendesi and @tgymnich, I assume I can merge that PR.
Or does anybody still need to have a closer look to eat, especially the last two mentioned?

bauer-andreas:
Needed to resolve conflicts with the latest changes pushed by #56.
Requesting re-review from @theMomax in order to merge.

theMomax (dismissed):
With #56 we renamed the `Request` folder to `Properties` in order to establish a clear difference between user-facing elements and the internal semantic model and its enabling code. You added some new code to `Sources/Apodini/Properties/Request.swift`. I think everything in this file except for `_Request` would better fit somewhere into the `Semantic Model Builder` folder (e.g. a new file `Request.swift`).

I'll still approve...if you agree, just ping me once you've pushed and I'll approve again ASAP.

bauer-andreas:
@theMomax Created `ApodiniRequest.swift` (you can't have two files with the same name lol) to separate the Request protocol from its actual internally used implementation.

theMomax (approved):
Thanks for the quick fix and nice job in general👍

lukaskollmer:
@Supereg Is there a reason why the `SyntaxTreeVisitable.visit` function was renamed to `accept`, but is still called `visit` in the Component extension (and in the Visitor)?

bauer-andreas:
@lukaskollmer I see, Component seems to declare an accept method for a visitor (currently called "visit" as you stated) but doesn't conform to the SyntaxTreeVisitable protocol. This weirdness slipped through as I seemingly only refactored the method name of the protocol.

I assume just restoring consistently by renaming it to "accept" should be fine?

lukaskollmer:
@Supereg I assumed that `Component`'s non-conformance to the visitor protocol was intentional. (@PSchmiedmayer is that the case?) If no, we probably should make  `Component` inherit from `SyntaxTreeVisitable`.

I'll rename it to `accept` in #66, is that ok with you?

(on a side note, iirc there is some corner case with the way swift handles protocol conformance where casting an object to one of the protocols its type conforms to can cause the dispatcher to pick up an unexpected implementation when calling one of that protocol's functions. not sure if this even is applicable here, i could be absolutely wrong. my assumption was that something like this was the reason `Component` doesn't inherit from `SyntaxTreeVisitable`.)

bauer-andreas:
> I assumed that Component's non-conformance to the visitor protocol was intentional. 

Seems like that, and it seems like it makes sense. However not quite sure how the definition of `Component.accept` doesn't end in an infinite loop 😅 

> I'll rename it to accept in #66, is that ok with you?

I was seconds away of creating a PR for that, already pushed the branch, but you can of course also include that with your PR. 👍 

Looking at your PR, as you also changed some stuff around with accept, I think it is easier if you just include the change in your PR 👍

PSchmiedmayer:
As the `SyntaxTreeVisitable` protocol is internal we are not able to make `Component` conform to the protocol. This implementation of `accept` provides a default implementation for `Component` that either forwards the visitor to a custom `accept` implementation provided by conforming to the `SyntaxTreeVisitable` or forwards the `SyntaxTreeVisitor` to the content of the `Component` in case the content is not of type `Never`.

Each `Component` that needs to provide a custom `accept` implementation **must** conform to `SyntaxTreeVisitable` and **must** provide a custom `accept` implementation. We require that each Component that conforms to `SyntaxTreeVisitable` provides its own custom `accept` implementation to avoid an endless loop in the `accept` function.


To be honest I am confident that the `enterCollection()` and `exitCollection()` calls are not needed. I have experimented around with the implementation and as far as I can see they can be removed. I will have also added some documentation to the function and have created a PR for this: #88

PR #82, opened 2020-12-27
Decouple Vapor.Application

tgymnich:
# Decouple Vapor.Application

We are currently still using `Vapor.Application` for our config and application lifecycle.

## :bulb: Proposed solution
Create a similar `Application` structure and use this to create `Vapor.Application`s for our exporters.

### Problem that is solved
Vapor is decoupled.

### Implications
Some of the advanced Vapor features are lost like environments, ... . However those can easily be (re-)added if need be.

### Related PRs
#61


PSchmiedmayer:
@tgymnich Apart from #83 is there anything that would block this PR from moving out of the draft status that I or anyone could help with? 

As far as I can see the NotificationCenter and PushNotification part of Apodini seem to impose most of the problems. @awoc do you have any input on this. I think in general moving them to an `ApodiniFoundation` would at least remove the exposure of Vapor from Apodini and move it to `ApodiniFoundation` but we would nevertheless need to find a way to decouple the functionality that is found in `ApodiniFoundation` from Vapor and ideally only rely on Fluent if possible. I also mentioned the idea of `ApodiniFoundation` in #70 in [this comment](https://github.com/Apodini/Apodini/pull/70#pullrequestreview-559574737).

@tgymnich Would it be possible to configure the CI so that warnings will also result in a failed build. This would force everyone to only use `@_implementationOnly` in the future as it already throws a warning if there is an import without `@_implementationOnly`.

I think it is important to decouple Vapor sooner then later, maybe even in 2020 😄

tgymnich:
@PSchmiedmayer there is nothing blocking this PR.

awoc:
I think it should be possible to decouple APNS from Vapor. The [APNS](https://github.com/vapor/apns) abstraction for Vapor is basically doing the same thing as in this PR and uses the `APNSwiftClient` protocol from [APNSwift](https://github.com/kylebrowning/APNSwift). 

[FCM](https://github.com/MihaelIsaev/FCM), on the other hand, is a different story. This library uses Vapor directly and doesn't offer a protocol like `APNSwiftClient` for initialization. I tried adding FCM to this PR, however, due to access rights of variables a lot of code from the library needs to be copied. Furthermore, [FCM](https://github.com/MihaelIsaev/FCM) uses [Client](https://docs.vapor.codes/4.0/client/) from Vapor to send requests to Firebase. Currently, there is no HTTP Client in Apodini, but I think we need one in the future anyway. Maybe the best way for [FCM](https://github.com/MihaelIsaev/FCM) is to create a fork and tailor it to our needs.

tgymnich:
@awoc using vapor internally is not a problem. For the FCM config you can create your own config like I did for HTTP2 and uses this to initialize your http client for FCM

tgymnich:
> Currently, there is no HTTP Client in Apodini, but I think we need one in the future anyway. Maybe the best way for FCM is to create a fork and tailor it to our needs.

@awoc having looked at this some more I think this is a great idea.

PSchmiedmayer:
@tgymnich @sadikekin @nityanandaz It would be great if we could merge the decoupling sooner than later. I know that there has been some activity on develop in the last few days and I don't want to make this push this PR out even more. 

@tgymnich What do you think about [the idea mentioned above](https://github.com/Apodini/Apodini/pull/82#issuecomment-752144462):
> @tgymnich Would it be possible to configure the CI so that warnings will also result in a failed build. This would force everyone to only use @_implementationOnly in the future as it already throws a warning if there is an import without @_implementationOnly.

Sources/Apodini/Configurations/Application/Application+FCM.swift
```diff
@@ -0,0 +1,36 @@
...
+
+    /// Firebase Cloud Messaging
+    public struct FCM {
+        struct ConfigurationKey: StorageKey {
+            // swiftlint:disable nesting
+            typealias Value = FCMConfiguration
+        }
+
+        /// FCM Configuration
+        public var configuration: FCMConfiguration? {
```
awoc: We could rename the struct to `FCMApplication` or something else. This would allow us to import the `FCMConfiguration` from the library like this `FCM.FCMConfiguration`, which would solve the naming issue with our configuration struct `FCMConfiguration`.

awoc (commented):
I tested the push notification feature and everything works well.  I really like the new configuration of the `application` property inside of the `NotificationCenter` :+1:

However, I noticed when both configuring `APNS` and `FCM` at the same time that the server will crash because both configurations will set the `application` property of the `NotificationCenter` and the assertion with the error message`"Application has already shut down"` will fail.

Also, a minor point. Can we rename `FCMConfig`  back to `FCMConfiguration` this would make the naming of configurations consistent with others like `APNSConfiguration` or `HTTP2Configuration`.

Sources/Apodini/Configurations/Application/Application+FCM.swift
tgymnich: fixed it

tgymnich: Setting `NotificationCenter.shared.application = app` inside of `Webservice` should mitigate setting the app twice

Sources/Apodini/Configurations/Application/Application+Vapor.swift
```diff
@@ -0,0 +1,45 @@
...
+        for id in app.databases.ids() {
+            if let config = app.databases.configuration(for: id) {
+                self.databases.use(config, as: id)
+            }
+        }
+        // HTTP2
+        self.http.server.configuration.supportVersions = Set(app.http.supportVersions.map { version in
+            switch version {
+            case .one: return Vapor.HTTPVersionMajor.one
+            case .two: return Vapor.HTTPVersionMajor.one
```
nityanandaz: Is this intended or a typo?

Sources/Apodini/Notifications/NotificationCenter.swift
```diff
@@ -25,19 +24,49 @@ import FCM
...
+/// NotificationCenter
 public class NotificationCenter {
-    internal static let shared = NotificationCenter()
-    internal var application: Application?
-    
-    private var app: Application {
-        guard let app = application else {
-            fatalError("The `NotificationCenter` is not configured. Please add the missing configuration to the web service.")
+    // swiftlint:disable implicitly_unwrapped_optional
+    @VaporApplication
```
nityanandaz: I guess, `@_implementationOnly `has some issues: `(X) Cannot use class 'Application' here; 'Vapor' has been imported as implementation-only` when I remove the property wrapper. Is this a known issue that we can refer?

Tests/ApodiniTests/ModifierTests/ConcatenatedResponseTransformerTests.swift
```diff
@@ -52,10 +52,20 @@ final class ConcatenatedResponseTransformerTests: ApodiniTests {
...
+                .with(exporter: restExporter)
+                .with(exporter: GRPCInterfaceExporter.self),
+            GraphQLSemanticModelBuilder(app),
+            WebSocketSemanticModelBuilder(app)
+        )
         
-        TestWebService.main(app: app)
-        
-        try app.test(.GET, "/v1/") { res in
+        try restExporter.app.test(.GET, "/v1/") { res in
```
nityanandaz: I use `Vapor.Application.test` myself and I really like its functionality. However, this doesn't look encouraging for new tests.
We should probably create some kind of `Apodini.Application.test`, but I wouldn't require it for this PR. I can try implementing it myself :)

nityanandaz (commented):
Very impressive overall.

Sources/Apodini/Configurations/Application/Application+Vapor.swift
tgymnich: typo 😄 Good catch!

Sources/Apodini/Notifications/NotificationCenter.swift
tgymnich: Correct! The `@VaporApplication` is a property wrapper to workaround this.

Tests/ApodiniTests/ModifierTests/ConcatenatedResponseTransformerTests.swift
tgymnich: well the test is rest specific anyway so this is just fine. For other exporter we will have to add similar tests.

Sources/Apodini/Configurations/Application/Application+Vapor.swift
tgymnich: fixed in 14ade673a5b156a11a53d59f84d281ed54896351

Tests/ApodiniTests/ModifierTests/ConcatenatedResponseTransformerTests.swift
tgymnich: I just decided to use one shared vapor application behind the scenes so this is no longer an issue

PSchmiedmayer (changes requested):
@tgymnich Thanks for the work regarding the decoupling of the Vapor Application instance. I have taken a closer look at the current version of the PR and noticed a few elements that I want to share in a review.

## Review
I looked over the diff of the code noticed a large code duplication between the current `Vapor.Application` implementation in the Vapor repo and the implementation of `Apodini.Application`. 
I am currently not convinced that maintaining this large code duplication would be desirable in the long run. I would suggest reducing the scope of `Apodini.Application` by a lot. 
In addition, most of the functionality related to the startup, event loops, and so on is not really needed during the configuration. All other aspects of Apodini and maybe even the test cases could continue using the Vapor application as it is only used internally.

## Proposed Ideas
I could inversion the `Configuration`s using some sort of builder pattern. I don't think that passing an `Vapor.Application` instance nor a very similar and duplicated `Apodini.Application` instance to the `Configuration` API is an approach that would be desirable. The `Configuration`s can do more then just modifying this builder and e.g., add elements to the shared `EnvironmentValues` that could be exposed publicly as described in [this comment](https://github.com/Apodini/Apodini/pull/70#issuecomment-754609847).
I think it would be interesting to see how generic we can make it so the builder would be able to create an `Vapor.Application` based on the context it has collected. Most of the elements that currently happen in the the `public func configure(_ app: Vapor.Application) {. /*...*/ }` should be expressible by some structures or properties in the builder, e.g.: `var autoMigrate: Bool` and so on ...

## General Comments
It would be great to also get some input from @hendesi and @awoc as this work mainly affects their parts of Apodini and moving the different `Configurations` and elements into separate targets would also affect the requirements for the configuration API. Maybe you have some ideas on how we could improve this? 
As said in the previous comments I think that decoupling Vapor from the public interface and moving all different dependencies to `@_implementationOnly` will be important to being able to evolve the Apodini interface independent of its dependencies and potentially swap out dependencies in the long run.
In Addition it would also be great to see some improvement in the code coverage of the `Configuration` elements of Apodini as this is currently the aspect of the main Apodini target that is covered the least: https://codecov.io/gh/Apodini/Apodini/tree/develop/Sources/Apodini

awoc:
@PSchmiedmayer I don’t think that creating different targets and using the current implementation for `Configuration`s would influence the behavior for them, at least for me. However, I agree that your proposed idea would be better and would make it easier to customize the `Application`.

Also, I created a PR #101 to increase the code coverage for `Configuration`s and #108 to move `Notifications` and `Jobs` to different targets.

PSchmiedmayer:
@awoc Great, thanks for the two pull request! Happy to see the additional test coverage and split up into smaller targets that build on top of the core Apodini functionality 👍

nityanandaz (approved):
Looks good to me! 🕵️

PSchmiedmayer (approved):
@tgymnich I looked over all the changes and experimented a bit with the implementation in Xcode. After our discussion in the last team meeting I agree with the proposed solution and have nothing to add regarding the implementation 🚀
Thanks for all the work, decoupling Vapor, and reducing the number of warnings by a lot!! 💪

PR #83, opened 2020-12-27
Tame SwiftLint

tgymnich:
# Tame SwiftLint

SwiftLint complains about nested types and optional collections in #82.
This PR makes SwiftLint complain less.

unblocks #82

# Example
```
struct SomeType {
   struct AnotherType {
       struct YetAnotherType {
      }
   }
}
```
```
let dictionary: [String: Int]?
```

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -53,8 +53,6 @@ only_rules:
     # Discouraged direct initialization of types that can be harmful. e.g. UIDevice(), Bundle()
   - discouraged_optional_boolean
     # Prefer non-optional booleans over optional booleans.
-  - discouraged_optional_collection
-    # Prefer empty collection over optional collection.
   - duplicate_imports
     # Duplicate Imports
   - dynamic_inline
@@ -364,7 +362,7 @@ line_length: # Lines should not span too many characters.
 
 nesting: # Types should be nested at most 2 level deep, and statements should be nested at most 5 levels deep.
   type_level:
-    warning: 2 # warning - default: 1
+    warning: 4 # warning - default: 1
   statement_level:
     warning: 5 # warning - default: 5
     


bauer-andreas:
Would you mind sticking to the PR template and giving code examples what was previously disallowed and would now be allowed by your PR?
I think this would make it easier for other people to reason about this change.

.swiftlint.yml
```diff
@@ -364,7 +362,7 @@ line_length: # Lines should not span too many characters.
 
 nesting: # Types should be nested at most 2 level deep, and statements should be nested at most 5 levels deep.
   type_level:
-    warning: 2 # warning - default: 1
+    warning: 4 # warning - default: 1
```
nityanandaz, line 365: This feels like bargaining, but how about 3? 😄

3 would suffice here: https://github.com/Apodini/Apodini/blob/24c8c7dac9e71ba01fe827d31b68fc876e5b10d4/Sources/Apodini/Configurations/Application/Application%2BCore.swift#L32, no?
I have not found 4 levels.

The comment needs adjustment, too.

tgymnich: I'm not sure about this but I think if we allow 3 levels and use 3 levels there will be a warning. This is why I went with 4

bauer-andreas: I think in the general case restricting type nesting to max two levels is appropriate, isn’t it? 
Wouldn’t it make more sense to use a ‚//swiftlint:disable[:next]‘ comment if nesting makes sense in your specific case?
I don’t think we want to „allow“ that in general?

again, as this is about code Style, this is a bit of a philosophical discussion, kind of 😅

tgymnich: I disagree. There is no reason we shouldn't allow nesting types in general. The main concern behind this rule is that we would end up with too long expressions like this `TypeA.TypeB.TypeC.doStuff()` but this is not how we use nesting of types in general. Our nested types are mostly used inside their parent type and won't result in long expressions. Type nesting helps to keep our code completion clean and helps to categorize/sort and abstract functionality inside our types.

I am fine with disallowing type nesting for `public` types but applying this restriction to `public`, `internal`, `private` and `fileprivate` types is too much collateral damage for my taste.

Also there is still code review to catch such extremely rare edge cases

PSchmiedmayer:
@tgymnich I have taken a look at your PR and would agree that the SwiftLint errors regarding nested types and optional collections in #82 are valid examples where we can make exceptions for these rules. I agree with [your comment](https://github.com/Apodini/Apodini/pull/83#discussion_r549404603) that long expressions should be avoided and that this rule should only apply to public parts of the API and it might be desirable to have a different limit for types that are marked internal or lower.
Nevertheless I think that the general idea of these rules to raise awareness for the two types of code smells is important and should stay in the project. I think It is better to get one more warning and to include an `//swiftlint:disable` comment with a reasoning behind it to explain the exception then lowering the bar for the project as a whole. In general it would be desirable to at least inform the developers about type nesting and optional collections, make then aware and provide a reasoning or restructure the code in case they encounter an error.
I think adding `//swiftlint:disable` comments and providing your reasoning behind it (as far as I can see all the exceptions regarding this rule in #82 are valid 🚀) would unblock the PR which is very important for the project and should be merged soon to completely decouple vapor from the public interface and make the Apodini API way nicer and cleaner. Thanks for the work in that area!

PR #84, opened 2020-12-27
Added Database To @Environment

hendesi:
# Added database to @environment

Currently the database can be access by `@_Database` which is a separate property wrapper and `RequestInjectable`. 
Current usage:
```
@_Database
var database: Fluent.Database
```

## :bulb: Proposed solution
Use the `@Environment` property introduced by #40. This changes the code to:
```
 @Apodini.Environment(\.database)
        var database: Fluent.Database
``` 

### Problem that is solved
This change removes the `Database` property/file as we want to move away from separate property wrappers and make use of the existing `@Environment` infrastructure. Furthermore it allows a way to provide a default database of the new request introduced by #81. The current approach would no longer work with the refactoring by #81.

### Implications
No real implications beside the refactoring from `@_Database` to `@Apodini.Environment(\.database)`.

--

### Related PRs
PRs affected by the change:
#70 
#81 

### Testing
No new tests, but I refactored tests that use the database.

### Reviewer Nudging
One could start looking at the `DatabaseConfiguration` where the environment value is set and then at `CustomComponentTests` to see the usage.


tgymnich (changes requested):
Could you also remove the database property from `Request`?

tgymnich (dismissed):
👍

hendesi:
@tgymnich Could you please approve this again? Github automatically removes review after a new commit

tgymnich (approved):

PR #35, opened 2020-11-30
Create OpenAPI Schema for basic use cases

lschlesinger:
# Create OpenAPI Schema for basic use cases
Detailed information on the OpenAPI Specification (OAS) to be found [here](https://swagger.io/specification/).

## :recycle: Current situation
As of now, no code related to the creation of OAS is in develop.

## :bulb: Proposed solution
The most relevant parts of the OAS were addressed first, leading to a `document` that contains:
- **`info`**: Rudimentary information in `info`
- **`server`**: Server address
- **`paths`**: Relative paths to the individual endpoints (i.e. each `Handler`), their operations, parameters (query and path), requestBody and responses (specified per HTTP status code)
- **`components/schemas`**: Definition of JSON schemas for objects (in contrast to primitive types) that can be referenced within the `paths - requestBody` or `paths - responses` section

As an example:
```swift
    struct SomeStruct: Vapor.Content {
        var someProp: Int
        var optionalInt: Int?
        var optionalString: String?
        var reqDouble: Double
    }

    struct SomeComp: Handler {
        @Parameter
        var name: String?

        func handle() -> SomeStruct {
            SomeStruct(someProp: 4, reqDouble: 5.0)
        }
    }
    
    struct User: Codable {
        var id: Int
        var name: String?
    }

    struct UserHandler: Handler {
        @Parameter var userId: Int

        func handle() -> User {
            User(id: userId)
        }
    }

    @PathParameter var userId: Int
    
    var content: some Component {
        Group("complexHandler") {
            SomeComp()
        }
        Group("user", $userId) {
            UserHandler(userId: $userId)
                    .operation(.update)
        }
    }
}
```

Will provide the following (for demonstration purposes somewhat condensed) OAS as output for the `document` sections `paths` and `components`:

```json
  "paths": {
    "/v1/user/{userId}": {
      "put": {
        "responses": {
          "200": {
            "content": {
              "application/json": {
                "schema": {
                  "$ref": "#/components/schemas/User"
                }
              }
            },
            "description": "OK"
          },
        "parameters": [
          {
            "description": "@Parameter var userId: Int",
            "in": "path",
            "name": "userId",
            "required": true,
            "schema": {
              "type": "integer"
            }
          }
        ]
      }
    },
    "/v1/complexHandler": {
      "get": {
        "responses": {
          "200": {
            "content": {
              "application/json": {
                "schema": {
                  "$ref": "#/components/schemas/SomeStruct"
                }
              }
            },
            "description": "OK"
          },
[...]

PR #85, opened 2020-12-30
Use callAsFunction in the EndpointRequestHandler

bauer-andreas:
# Use callAsFunction in the EndpointRequestHandler

Currently the `EndpointRequestHandler` exposes a `handleRequest` method an exporter would call like:
Like.
```swift
requestHandler.handleRequest(request: ...)
```

## :bulb: Proposed solution

Using `callAsFunction` we add a little bit of syntactic sugar and the above becomes this:
```swift
requestHandler(request: ...)
```

Additionally the left over `RequestHandler` type alias was removed.

### Problem that is solved

There was no problem 🙃

### Implications

`handleRequest` was renamed to `callAsFunction` imposing the need of refactoring for exporters which already adopted changes of #81 (@theMomax, @lschlesinger, @moritzschuell).
We could have added backwards compatibility, but I don't think it is worth it for this change.

### Related PRs

* #81 originally introducing the `EndpointRequestHandler`

**PRs impacted by this change:**
* #74 WebSocket exporter
* #35 OpenAPI exporter
* #43 gRPC exporter

### Testing

Nothing to test.

### Reviewer Nudging

I assume this is an easy one.

### DIFF
--- a/Sources/Apodini/Semantic Model Builder/InterfaceExporter/RequestHandler.swift
+++ b/Sources/Apodini/Semantic Model Builder/InterfaceExporter/RequestHandler.swift
@@ -7,15 +7,15 @@ import protocol NIO.EventLoop
 import protocol FluentKit.Database
 
 class EndpointRequestHandler<I: InterfaceExporter> {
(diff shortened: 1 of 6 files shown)


awoc (approved):
Looks good 👍

hendesi (approved):
👍

PR #88, opened 2020-12-30
Improve the `SyntaxTreeVisitor` and test more edge cases

PSchmiedmayer:
# *Improve the `SyntaxTreeVisitor` and test more edge cases*

- The current version of the `accept` function of the `Component` protocol and the `SyntaxTreeVisitor` do not include any documentation.
- In addition the current way we generated the identifies of Endpoints includes additional layers introduced by `TupleViews`. Manually calling `enter` and `exit` calls in the `SyntaxTreeVisitor` might cause an unbalanced call of these functions when parsing the Apodini DSL

## :bulb: Proposed solution
- We simplify the `SyntaxTreeVisitor`'s `enter` and `exit` API to a closure based syntax
- We add documentation to the `SyntaxTreeVisitor` to clearly state is purpose
- We restructure the Tests to remove the `PrintVisitor` that doesn't test functionality and use test assertions and expectations

### Problem that is solved
The `accept` function of the `Component` protocol is better documented and the behavior of the `SyntaxTreeVisitor ` protocol is more clearly defined using a closure based API. The `SyntaxTreeVisitor` is better tested. The renaming of the `Scope.nextComponent` to `Scope.nextHander` reflects the change to `Handler`s reflected in #66 .

### Implications
- `Scope.nextComponent` has been renamed to `Scope.nextHander`
- No other implications except a better test coverage and a clearer API to work with the `SyntaxTreeVisitor ` that doesn't affect the `SharedSemanticModel` or the interface exporters.

### Related PRs
Discussion in #81 

### Testing
Multipel tests were added and existing tests were restructured


bauer-andreas (approved):
Thanks adding the documentation. 👍 I think it really helps a lot here.

In regards to the `enterCollection` and `exitCollection`: I made some thought experiments and agree that removing those calls there does indeed work and to my understanding is actually the correct way of handling that. But the review of @lukaskollmer is probably crucial for that part of the PR.

bauer-andreas:
EDIT: judging by the reduction of the code coverage, execution must have been different than with the `enterCollection` and `exitCollection` calls. Not sure if that is important or not, but maybe worth looking for the reason(?). Cause I wouldn't expect the [coverage diff](https://codecov.io/gh/Apodini/Apodini/pull/88/changes) if the output would be the exact same(?).

lukaskollmer:
I'd leave the `{enter|exit}Collection` calls in. We probably should improve the naming here, but this is required for computing index paths for components.

Currently I'm using:
- `{enter|exit}Collection` to keep track of the current depth into the web service data structure (depth not in terms of path components or the exported interface, but simply as in "how many levels of `.content` are we into the DSL"). That's why all visits to a component's content are surrounded by a pair of enter/exitCollection calls
- `{enter|exit}CollectionItem` to keep track of the index of the current component relative to its parent's children

The reason why this change didn't cause any tests to fail is simply because the component index path tests i wrote all use a web service defined using `Group`s (which conforms to `SyntaxTreeVisitable` and therefore doesn't use the default implementation changed here. [`Group.accept`](https://github.com/Apodini/Apodini/blob/882f4e1e9e0a3d3d802653f243408875e8befe22/Sources/Apodini/Components/Group.swift#L27) in fact also wraps its `content.accept` call in enter/exitCollection calls.)
I'll add some more test cases to cover components which use the default `accept` implementation.

---

**Edit** come to think of it we might actually need to go one step further and surround the `content.accept` call here with enter/exitCollectionItem calls as well. I'll do some testing to see if this is necessary

lukaskollmer (approved):
@PSchmiedmayer The changes to the visitor look really nice, especially the new approach to generating index paths, and that we now ignore the implicitly inserted TupleComponents (also thank you for figuring out a way to get rid of the implicit 0 root level entry and the `max` call to adjust the indices).

One thought I had, regarding the enter/exit APIs was to maybe replace them with a block-based API, which would ensure that the enter- and exit calls are always balanced:

```swift
// in the Visitor class

private func enterContent() { ... }
private func exitContent() { ... }

func enteringContent(_ block: () -> Void) {
    enterContent()
    block()
    exitContent()
}
```

The downside is that the caller now has to use a closure (and the nesting can get a bit annoying when entering both a new content and an item), but the upside is that we can guarantee that every `enter` call always has its corresponding `exit` call. Also wouldn't be much of an issue in terms of additional performance overhead since these closures would be implicitly noescape.

PSchmiedmayer:
@lukaskollmer Thanks for your input! Great to see that you like the updated `SyntaxTreeVisitor`!
I really like the block based API approach you described in your comment, I will incorporate it into this PR 👍

PR #90, opened 2021-01-01
Jobs

awoc:
# Jobs

Currently there is no mechanism to define cron jobs in Apodini and run them at server startup or from an Handler.

## :bulb: Proposed solution

This PR will add a `Scheduler` that is responsible for the handling and scheduling of `Job`s. Cron Jobs can be created by conforming to the `Job`s protocol which only defines a `run` method and by passing in a crontab String to specify the moment in time the `Job` is executed. The `run` method will be triggered at the specified points in time. `Job`s can either be created at server startup using the `Schedule` configuration or from a `Handler`.

The `Scheduler` also requires a key path conforming to `ApodiniKeys`. Using this key path, the `Job` can be injected with `@Environment` and also be cancelled using the `Scheduler`. The `Scheduler` can be used in a Handler with the Environment property wrapper: `@Environment(\.scheduler)`.

### Example

The following example showcases features of this PR.

```swift
struct TestServer: WebService {
    struct CancelHandler: Handler {
        @Environment(\.scheduler) var scheduler: Scheduler

        func handle() -> String {
            scheduler.dequeue(\ApodiniKeys.testJob)
            return "Cancelled"
        }
    }

    struct RunHandler: Handler {
        @Environment(\ApodiniKeys.testJob) var testJob: TestJob

        func handle() -> String {
            testJob.run()
            return "Executed"
        }
    }

    struct TestJob: Job {
        func run() {
            print("Hello World \(Date())")
        }
    }

    struct KeyStore: ApodiniKeys {
        var testJob: TestJob
    }

    var content: some Component {
        RunHandler()
        group("cancel") {
            CancelHandler()
        }
    }

        // Will be executed at 16:05 every Friday.
        Schedule(TestJob(), on: "5 16 * * 5", \ApodiniKeys.testJob)
    }
}
```

### Implications

This PR shouldn't have any implications on the existing implementation.

- The parsing of crontabs uses the [SwifCron](https://github.com/MihaelIsaev/SwifCron) library.
- Scheduling only uses SwiftNIO

### Related PRs

- The `Schedule` configuration also requires decoupling from Vapor #82. However, this should be straight forward.

### Testing

Tests covers the `Schedule` configuration and the creation of `Job`s in the `Scheduler` class. The correct execution at is not covered by the tests.

### Reviewer Nudging

Maybe you can define sample Jobs and test the correct scheduling. [Crontab Guru](https://crontab.guru/#5_16_*_*_5) is a great website to test crontab strings.


Tests/ApodiniTests/JobsTests.swift
```diff
@@ -0,0 +1,62 @@
...
+        XCTAssertThrowsError(try Scheduler.shared.enqueue(FailingJob(), with: "* * * * *", \KeyStore.failingJob, on: app.eventLoopGroup.next()))
+        XCTAssertThrowsError(try Scheduler.shared.enqueue(TestJob(), with: "* * * *", \KeyStore.testJob, on: app.eventLoopGroup.next()))
+        XCTAssertThrowsError(try Scheduler.shared.enqueue(TestJob(), with: "A B C D E", runs: 5, \KeyStore.testJob, on: app.eventLoopGroup.next()))
+    }
+    
+    func testEnvironmentInjection() throws {
+        let scheduler = EnvironmentValues.shared.scheduler
+        let job = TestJob()
+        try scheduler.enqueue(job, with: "*/10 * * * *", \KeyStore.testJob, on: app.eventLoopGroup.next())
+        let environmentJob = EnvironmentValues.shared.values[ObjectIdentifier(\KeyStore.testJob)] as! TestJob
```
nityanandaz:
```suggestion
        let environmentJob = try XCTUnwrap(
            EnvironmentValues.shared.values[ObjectIdentifier(\KeyStore.testJob)] as? TestJob
        )
```

Yesterday I learned: `XCTUnwrap`. Added benefit: Tests fail gracefully instead of crashing 😄 And you can remove `swiftlint:disable ...` at the top.

Tests/ApodiniTests/JobsTests.swift
```diff
@@ -0,0 +1,62 @@
...
+        environmentJob.num = 42
+        XCTAssert(environmentJob.num == job.num)
+    }
+    
+    func testScheduling() throws {
```
nityanandaz: I would split this into two test methods, each of which better describes the scenario, e.g.:
- testEveryMinute
- testZeroRuns

Sources/Apodini/Jobs/Scheduler.swift
```diff
@@ -0,0 +1,120 @@
...
+    ///
+    /// This method throws an exception if the `Job` uses request based property wrappers or the crontab cannot be parsed.
+    ///
+    /// - Parameters:
+    ///     - job: The background running task conforming to `Job`s.
+    ///     - with: Crontab as a String.
+    ///     - runs: Number of times a `Job` should run.
+    ///     - keyPath: Associates a `Job` for later retrieval.
+    ///     - on: Specifies the event loop the `Job` is exectured on.
+    public func enqueue<K: ApodiniKeys, T: Job>(_ job: T,
```
nityanandaz, line 36:
```suggestion
    /// - Throws: If the `Job` uses request based property wrappers or the crontab cannot be parsed.
    public func enqueue<K: ApodiniKeys, T: Job>(_ job: T,
```

No? 🙂

Same for `Schedule.dequeue`. Maybe it's just a matter of taste.

Sources/Apodini/Jobs/Scheduler.swift
```diff
@@ -0,0 +1,120 @@
...
+    /// - Parameter keyPath: Associatesd key path of a `Job`.
+    public func dequeue<K: ApodiniKeys, T: Job>(_ keyPath: KeyPath<K, T>) throws {
+        guard let config = jobConfigurations[ObjectIdentifier(keyPath)] else {
+            throw JobErrors.notFound
+        }
+        
+        config.scheduled?.cancel()
+    }
+    
+    private func schedule<T: Job>(_ job: T, with config: JobConfiguration, on eventLoop: EventLoop) {
```
nityanandaz: This is totally up to you, but I like this trick:

> Alternatively, you can mark an extension with an explicit access-level modifier (for example, private) to set a new default access level for all members defined within the extension.

From https://docs.swift.org/swift-book/LanguageGuide/AccessControl.html#ID25

You can group all private members in such an extension: `private extension Scheduler { ... }` and omit the individual access-level modifiers.

Sources/Apodini/Configurations/Schedule.swift
```diff
@@ -0,0 +1,50 @@
...
+    }
+    
+    /// Initializes the `Schedule` configuration.
+    ///
+    /// - Parameters:
+    ///     - job: The background running task conforming to `Job`s.
+    ///     - on: Crontab as a String.
+    ///     - runs: Number of times a `Job` should run.
+    ///     - keyPath: Associates a `Job` for later retrieval.
+    public init(_ job: T, on cronTrigger: String, runs: Int, _ keyPath: KeyPath<K, T>) {
```
nityanandaz, line 33: You could use default parameter values and save the other initializer:

```suggestion
    public init(_ job: T, on cronTrigger: String, runs: Int? = nil, _ keyPath: KeyPath<K, T>)
```

nityanandaz (changes requested):
I really like your documentation comments and tight access-levels!

Sources/Apodini/Jobs/Scheduler.swift
awoc: I totally agree, I think this makes it clearer

Sources/Apodini/Configurations/Schedule.swift
awoc: I prefer to have to separate initializers because I think if you want to use `runs` than it doesn’t make sense to declare it as `nil`.

Tests/ApodiniTests/JobsTests.swift
awoc: This is really useful, thank you :D

Sources/Apodini/Jobs/Scheduler.swift
awoc: Thank you for this tip! I didn’t know of this feature

Sources/Apodini/Configurations/Schedule.swift
nityanandaz: Makes sense 👍

nityanandaz (approved):
I am very intrigued by the new possibilities 😄

PR #91, opened 2021-01-01
Combine Build and Test steps to reduce Action runtime

bauer-andreas:
# Combine Build and Test steps to reduce Action runtime

Previously the GitHub workflow would always have two steps `Build` and `Test`.
`Build` compiled the whole project, including tests, and `Test` just ran all unit tests.

With the introduction of code coverage measurement #49, the command line option `--enable-code-coverage` was added to `swift test` execution. SPM needs to recompile the whole project in order to provide code coverage measurement data (I assume adding some sort of instructions to the binary, idk). This basically results in compiling the project twice in every execution of the action.

## :bulb: Proposed solution

Combining `Build` and `Test` steps into one. For consistency reasons this is also done to the linux based jobs.

### Problem that is solved
Reduced waiting time for actions (specifically the job running on macOS, at it is the only one reporting code coverage).

### Implications
--

--

### Related PRs
--

### Testing
--

### Reviewer Nudging
--

### DIFF
--- a/.github/workflows/swift.yml
+++ b/.github/workflows/swift.yml
@@ -26,9 +26,7 @@ jobs:
       run: xcodebuild -version
     - name: Check Swift version
       run: swift --version
-    - name: Build
-      run: swift build --build-tests
-    - name: Test
+    - name: Build & Test
       run: swift test --enable-code-coverage
     - name: Convert coverage report
       run: xcrun llvm-cov export -format="lcov" .build/debug/ApodiniPackageTests.xctest/Contents/MacOS/ApodiniPackageTests -instr-profile .build/debug/codecov/default.profdata > coverage.lcov
@@ -51,7 +49,5 @@ jobs:
         key: ${{ runner.os }}-${{matrix.linux}}-spm-cachev2-${{ hashFiles('**/Package.resolved') }}
     - name: Check Swift version
       run: swift --version
-    - name: Build
-      run: swift build --build-tests --enable-test-discovery
-    - name: Test
+    - name: Build & Test
       run: swift test --enable-test-discovery


.github/workflows/swift.yml
```diff
@@ -26,9 +26,7 @@ jobs:
       run: xcodebuild -version
     - name: Check Swift version
       run: swift --version
-    - name: Build
-      run: swift build --build-tests
```
tgymnich, line 30: wouldn't keeping the two steps and using `swift build --build-tests --enable-test-discovery` make it more obvious where the build failed?

tgymnich (changes requested):

.github/workflows/swift.yml
bauer-andreas: Well yes, that's the reason why it was that way. But that's the sacrifice we have to make in oder to not double compile. I haven't found a `swift build` argument were it compiles in codecov instructions.

tgymnich: I mixed up --enable-code-coverage and  --enable-test-discovery 🙈. I meant `swift build --build-tests --enable-test-discovery --enable-test-discovery` which doesn't work 😅

tgymnich (approved):

.github/workflows/swift.yml
bauer-andreas: Was a bit confused of your comment but now I understand where it came from 😂👍

bauer-andreas:
okay, how the hell did this change the code coverage, arrrg

bauer-andreas:
_The definition of insanity is doing the same thing over and over again, but expecting different results._

lol. And it worked. Merging fast, don't what the codecov gods to change their minds

PR #92, opened 2021-01-02
Improve the Test Coverage

PSchmiedmayer:
# *Improve the Test Coverage*

We currently have a test coverage of about  66% and some edge cases and possible bugs might not be detected by the test suite at the moment.

## :bulb: Proposed solution
This PR tries to increase the test coverage by proposing useful test that cover areas that are currently not tested. Bugs encountered during the writing of the tests are fixed.

### Problem that is solved
This PR reduces the probability of future regressions across the Apodini API. Smaller bugs that have been encountered have been solved in this PR as stated in the _Implications_ section.

### Implications
This PR has no impact on overall Apodini structure and public interface
There are small changes for bugs that were encountered when writing test cases:
- **Guards**: The behavior of resetting `Guard`s was broken. Added an documented `allActiveGuards` method to only get active `Guard`s for an `Endpoint` in the shared semantic model. All interface exporters that currently build on top of the shared semantic model shouldn't be affected.
- **Environment**: `Environment` now conforms to `Property`
- **PathParameter**: The `init(_ parameter: Parameter<Element>)` was removed as it doesn't add additional functionality as `PathParameters` can be directly passed using the direct access to the property wrapper (underscore) as demonstrated in the test case.

### Testing
Added a lot of new tests 🚀

### Reviewer Nudging
It would be important that the tests don't make too strict assumptions about the behavior of the Apodini system. They should test functionality and shouldn't break in case there are necessary internal changes made that don't affect the Apodini behavior. Please take this into consideration when reviewing the tests and point out possible wrong assumptions in the reviews.


tgymnich (approved):

Sources/Apodini/Properties/Request.swift
```diff
@@ -25,30 +25,6 @@ protocol Request: CustomStringConvertible, CustomDebugStringConvertible {
...
-    }
-    
-    
-    init() { }
-
-
-    mutating func inject(using request: Request) throws {
-        self.request = request
-    }
-}
```
awoc, line 49: I think it is still useful to access the `Request` in a Handler to get the current eventloop. My idea was it to use the current `Request` for scheduling of `Job`s in a Handler. Perhaps we can add `Request` to `dynamicValues` in order to access it with `@Environment` in a Handler.

PSchmiedmayer: @awoc I agree that having access to the `EventLoop` is important for various features. When working on the PR I already created an issue (#93) that documents this problem. As `Request` is an internal type and `_Request` was a left over from the very early days where we were using `Vapor.Request` and all the other functionality except the `eventLoop` (e.g. the `Endpoint`) shouldn't normally be accessible within a `Handler`.
I like the idea of accessing the `EventLoop` using the `Environment` in a non global manner as it is injected with every execution of a handler but would oppose injecting the whole `Request`.

awoc: My bad, I didn't see the open issue

PSchmiedmayer: No problem :)
Would an access to an EventLoop solve your problem or should we consider some sort of access to a Request type anyways?

awoc: Using only the EventLoop is totally sufficient 👍

PR #94, opened 2021-01-03
Adding back support for Handlers to return EventLoopFutures

bauer-andreas:
CC: @PSchmiedmayer, @nerdsupremacist, @hendesi, @lukaskollmer, @tgymnich, @moritzschuell 

# Adding back support for Handlers to return EventLoopFutures

Since moving from `ResponseEncodable` to `Encodable` (#61) we lost the ability to return `EventLoopFuture` in our `Handler`s.

## :bulb: Proposed solution

This PR resolves this by adding `Encodable` conformance to `EventLoopFuture`. 

### Problem that is solved

Handlers can now return any amount of nested `EventLoopFuture`s.

This PR is currently considered an intermediate solution, as the long term goal is probably to move to a custom built replacement for `Encodable`. See https://github.com/Apodini/Apodini/pull/73#issuecomment-749513032 and https://github.com/Apodini/Apodini/pull/73#pullrequestreview-560635838.

As the new `Action` enum was already merged, this PR turned out to be a bit bigger struggle than expected.
As `Action` as well as `EventLoopFuture` now conform to `Encodable` the user could in theory implement a `handle()` which returns any arbitrary nesting of those two types (e.g. `EventLoopFuture<Action<...>>`, `Action<EventLoopFuture<...>>` or `Action<Action<...>>`).

This PR enforces that an Action can only be on the "outer" level and prohibits the case of wrapping an Action into an Action, as this is basically nonsense. This is done via runtime checks. With our `Encodable` replacement I expect those restrictions to be incorporated into the type system, and thus leveraging those errors to compile time.

### Related PRs

This change was previously the scope of #73 (see https://github.com/Apodini/Apodini/pull/73#pullrequestreview-560635838)

### Testing

* The `CustomComponentTests`, which was disabled since we lost support for `EventLoopFuture`s, was reenabled.
* Added a few test cases to `ApodiniEncodableTests` to cover the newly added lines.

### Reviewer Nudging

Best entry point is probably looking at an example Handler (something like `FutureBasedHandler` in `ApodiniEncodableTests.swift`) and stepping through the request handler code, to review the implementation.


Sources/Apodini/Semantic Model Builder/InterfaceExporter/RequestHandler.swift
```diff
@@ -91,13 +99,64 @@ struct ActionVisitor: ApodiniEncodableVisitor {
...
+
+        var modifiers = modifiers
+        let next = modifiers.removeFirst()
+
+        return response.flatMap { encodable in
+            let transformed = request.enterRequestContext(with: next()) { responseTransformer in
+                responseTransformer.transform(response: encodable)
+            }
+            let responseFuture = eventLoop.wrapEncodableIntoFuture(encodable: transformed)
+            return transformNextResponse(responseFuture, modifiers)
```
moritzschuell: Why did you switch to a recursive approach for the response transformation? 🤔

Tests/ApodiniTests/RequestTests/ApodiniEncodableTests.swift
```diff
@@ -6,9 +6,10 @@
 //
 
 import XCTest
+import NIO
 @testable import Apodini
 
-final class ApodiniEncodableTests: ApodiniTests, ApodiniEncodableVisitor {
+final class ApodiniEncodableTests: ApodiniTests, EncodableContainerVisitor {
```
moritzschuell: `ApodiniEncodableTests` should probably be renamed.

moritzschuell (commented):
I think with the current state, a `ResponseTransformer` that can be applied to a `Handler` with response type `EventLoopFuture<String>` would have to have a signature for the `transform` method like so: `transform(response: EventLoopFuture<String>) -> EventLoopFuture<String>` (otherwise the compiler complains).
But the shared model unwraps futures and only hands the wrapped value (`String` in my example) to the transformers, right?

bauer-andreas:
> I think with the current state, a `ResponseTransformer` that can be applied to a `Handler` with response type `EventLoopFuture<String>` would have to have a signature for the `transform` method like so: `transform(response: EventLoopFuture<String>) -> EventLoopFuture<String>` (otherwise the compiler complains).
> But the shared model unwraps futures and only hands the wrapped value (`String` in my example) to the transformers, right?

@moritzschuell My vision/idea was that the ResponseTransformer would always get the "unwrapped" value, removing the need for the user to manually wait for the future to succeed. Thats the reason the above mentioned recursion was introduced, as a response transformer could again return a Future and the next hooked transformer would then also receive the unwrapped value and so on.
But quickly looking at the code I see that the current modifier methods are seemingly not build for that. Maybe I can add a additional case like that which was added for Action. Otherwise I think it is also just valid to move back to just passing the Future to the Transformers and to wait for the final `Encodable` replacement.

CC: @tgymnich

bauer-andreas:
As this PR is intended to be an intermediate step, I decided to revert the reduction/unpacking of EventLoopFutures between response transformer steps.

Tests/ApodiniTests/RequestTests/ApodiniEncodableTests.swift
bauer-andreas: Renamed to `CodableTests`

Sources/Apodini/Semantic Model Builder/InterfaceExporter/RequestHandler.swift
bauer-andreas: Removed this again, as explained in https://github.com/Apodini/Apodini/pull/94#issuecomment-753742262.
Removed the unwrapping of EventLoopFutures between transformer steps.

Tests/ApodiniTests/RequestTests/CodableTests.swift
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    struct FutureBasedHandler: Handler {
+        var eventLoop: EventLoop
+        var message: String
+
+        func handle() -> Action<EventLoopFuture<EventLoopFuture<EventLoopFuture<String>>>> {
+            // this tests if the `EventLoopFutureUnwrapper` properly unwraps multiple nested EventLoopFutures
+            // I hope no one would ever do that, but its possible, thus we need to handle it properly
+            .send(
```
PSchmiedmayer, line 28: I would consider this the other way around and to only allow that an `Action` carries an `Encodable` that is only a concrete type and not an `EventLoopFuture`. Looking over this example I am not 100% convinced that moving all the responsibility to `Encodable` is the perfect solution as It won't allow us to express this constraint on compile time as you have already mentioned in the PR description. 
Even if we replace `Encodable` with a custom solution I would consider this problem and the `Encodable` problem at two different layers. 

But to be honest I am not sure how to perfectly solve this at this point. I will give it a though in the morning as I have already experimented a bit with it while taking a look at the PR and will let you know if I have a solution that would be feasible in the near future to not block any other PRs/features relying on this.

Back to the `EventLoopFuture` nesting: I would consider `EventLoopFuture` the elements that can be nested on the outside and `Action`s can only contain `Encodable` types that are not `Action`s or `EventLoopFuture`s. 
This would also solve part of the `EventLoopFuture` problem mentioned in #97 as we could unwrap all `EventLoopFuture`s and allow `ResponseTransformers` to either only work on the `Encodable` type or `Action` where the `Element` is just an `Encodable` type and not an `Action` or `EventLoopFuture`.

If you e.g. consider a database access: The type of action is not known at the outer layer as it might depend on the elements retrieved from the database. If there are multiple elements it might be a `.send` while it could also be a `.finish` when there is only one element in the database and the `Handler` is trying to model a service side stream. In general the type and content of `Action` might be reliant on information only available in `EventLoopFutures` and therefore I would consider the possible signature the other way around: `EventLoopFuture<EventLoopFuture<EventLoopFuture<Action< String >>>>`

PSchmiedmayer (commented):
@Supereg Thanks for the fast creation of a PR to fix the problem that `Handler`s can't return `EventLoopFuturs`! 💪

I looked over your code and think the approach you have chosen is promising and seems to solve the current problem regarding `EventLoopFuturs`. Nevertheless I am not sure if pricing all the responsibility on `Encodable` is a good approach as it creates the runtime checks that need to enforce to e.g. not nest `Actions`. 

Regarding nesting types with `EventLoopFutures` where I have also added a comment in the source code at the test case: I would not consider that `Actions` should only be the outer layer but they should be only on the most inner layer and might include `Encodable` data carried in the `.finish` or `.send` cases. As shown in the example if e.g. the return value is dependent on a database access the decision if a `.send` or `.nothing` case is used can only be performed after a future as been evaluated. If the decision could be made before the decision could also be passed in the future chain so all `EventLoopFutures` wrap an `Encodable` or `Action` type.

During the review I have experimented a bit with the type constraints that could enforce this wrapping without runtime checks but I have not really been successful so far. I will take a look at this morning and will let you know if I can come up with an appropriate solution/idea.

Sources/Apodini/Modifier/ResponseModifer.swift
```diff
@@ -130,7 +130,7 @@ extension Handler {
     public func response<T: ResponseTransformer>(
         _ responseTransformer: @escaping @autoclosure () -> (T)
     ) -> ResponseModifier<Self, T> where Self.Response == T.Response {
-        if Self.Response.self is ApodiniEncodable.Type {
+        if Self.Response.self is EncodableContainer.Type {
             preconditionFailure("Actions cannot be transformed directly. Use a transformer on the type that is wrapped by the Action instead.")
```
lukaskollmer, line 134: There's several places in the code which operate on the assumption "if it conforms to `EncodableContainer` it must be an `Action`".
Is this the case, or am I misunderstanding something?
If yes, do we enforce this anywhere?
What about other types which might conform to `EncodableContainer`?

I think if `Action` really is the only type conforming to this protocol, and the protocol is being used to determine whether an object is an Action (and as a type-erased way to call the action's accept method), it might make more sense to rename it to `AnyAction`, `ActionProtocol` or something like that.

PSchmiedmayer:
@Supereg, @hendesi, @moritzschuell, @theMomax, @nerdsupremacist, @lukaskollmer 
I have experimented a bit after reviewing #94 and commenting on #97 and created a draft PR with very little description about the changes but some ideas in the code. As far as I can see this would solve the issues raised in the review of this PR and #97 by restructuring how the ResponseTransformer work. I think we should discuss how many elements from the draft and this PR should be taken into consideration when merging a solution into develop. I am interested in hearing your feedback and input! 🚀

bauer-andreas:
closing in favor of #98

Issue #97, opened 2021-01-04
Unwrap EventLoopFutures between response transformer steps


bauer-andreas:
Related: #94 
CC: @tgymnich (as you are working on the new `Encodable` replacement), @moritzschuell, @theMomax, @PSchmiedmayer, @nerdsupremacist 

As of now the Input type of `ResponseTransformers` must always match the Output type of the modified `Handler`.
E.g. having a handler with `func handle() -> String`, your transformed needs to be defined as `func transform(response: String) -> Output`.
There is one exception to this rule: `Action`s. For a `Handler` returning e.g. `Action<String>` you can defined a `ResponseTransformer` transforming the wrapped `String` value.

An additional edge case worth to be considered is a `Handler` returning `EventLoopFuture<...>`. Currently a ResponseTransformer must be defined to have an input type of `EventLoopFuture` if the `Handler` returns `EventLoopFuture`. I think it would be a great addition to the API if we support unwrapping `EventLoopFutures` between response transformer steps.
Below I supplied a list highlighting possible response transformer signatures for a given handle signature. 

* `func handle() -> String`
  * `func transform(response: String) -> Output`
* `func handle() -> Action<String>`
  * `func transform(response: String) -> Output`
  * `func transform(response: Action<String>) -> Output`
    This case is currently possible, to define (meaning no compile time error), but the current `RequestHandler` will currently always unwrap the Action, and will only call transformers for `.final` and `.send` (when a value is supplied).
  I think it would be a valid discussion if a response transformer should be allowed to transform the return action from a `Handler`. I would assume that it may be a problem when a `Handler` thinks it closed a server side stream with `.end`, but a `ResponseTransformer` modifies that. Otherwise I could also see benefits on allowing this.
* ~`func handle() -> Action<Action<String>>`~ **wrapping Actions is disallowed**
* `func handle() -> EvenLoopFuture<String>`
  * `func transform(response: String) -> Output`
  * `func transform(response: EventLoopFuture<String>) -> Output`
    I think allowing this is perfectly fine and just improves compatibility of possible building blocks.
* `func handle() -> Action<EventLoopFuture<String>>`
  * `func transform(response: String) -> Output`
  * `func transform(response: EventLoopFuture<String>) -> Output`
  * `func transform(response: Action<EventLoopFuture<String>>) -> Output`
    Depending on the discussion above, this case might be allowed.
* `func handle() -> EventLoopFuture<EventLoopFuture<String>>`
[...]

PR #98, opened 2021-01-04
Introduce the `Response` type

PSchmiedmayer:
# *Introduce the `Response` type*

We currently have the station that `Handler`s can't return `EventLoopFutures` and that using `Encodable` to express the requirement of a return type of `Handler` might complicate the implementation and handling of edge cases, e.g. nesting different `Encodable` types as seen in #94 and addresses the issues raised in #97

## :bulb: Proposed solution
This PR introduced the `ResponseTransformable` protocol that enables types to be transformed into a `Response` type. `Response` (previously `Action`) is used to express all possible returns from `Hanlder`'s `handle()` function.

Common types like `String`, `Int` (including variations like `UInt64`) , `Double`, `Float`,`Bool`, and `UUID`,  and the corresponding `EventLoopFuture`wrappers all confrom to `ResponseTransformable` out of the box.
Collection types (`Array`, `Dictionary`, `Set`) also conform to `ResponseTransformable` if the elements conforrm to `Encodable`.
Types that confrom to `Encodable` can confrom to `ResponseTransformable` without the need to implement any additional methods.

To make it easy to return custom types from `Handler`'s this PR introduces the `Content` type:
Types that can be returned from a `Handler`'s `handle` function should conform to `Content`.
`Content` includes the conformance to `Encodable`. If the types implement the `Encodable` requirements the type doesn't need to provide additional implementation steps to conform to `ResponseTransformable`.

In addition is simplifies the usage of `ResponseTransformer`: Can only be used to transform types conforming to `Encodable` to `Encodable` using a `ResponseTransformer`. `ResponseTransformer` can not transform a `Response` type to an other.

### Problem that is solved
Addresses the problems described in #94 and #97.

### Implications
Has no direct implications on the Exporters but heavily influences existing Apodini web services as the `ResponseTransformer` has changed to only support primate types that conform to `Encodable`.

The have renames the `Action` type `Response` and we now have a `ResponseTransfromable` protocol that incorporates the implementation that was previously located in the `Response` protocol including a different naming for the function to transform the conforming type to a `Response`.

To make conformance simpler for developers using Apodini we now have a `Content` protocol that includes the implementation of `Encodable` and `ResponseTransfromable` in one to make the conformance of external types easier.

#### ResponseTransformer
Example for a change in a `ResponseTransformer`:
```swift
struct EmojiTransformer: ResponseTransformer {
    private let emojis: String


    init(emojis: String = "✅") {
        self.emojis = emojis
    }


    func transform(response: String) -> String {
        "\(emojis) \(response) \(emojis)"
    }
}
```
changes to
```swift
struct EmojiTransformer: ResponseTransformer {
    private let emojis: String


    init(emojis: String = "✅") {
        self.emojis = emojis
    }


    func transform(content: String) -> String {
        "\(emojis) \(response) \(emojis)"
    }
}
```

#### Response
Types like `User`
```swift
struct User: Encodable  {
    let id: String
    let name: String
}
```
now have to conform to `Content`
```swift
struct User: Content  {
    let id: String
    let name: String
}
```

### Related PRs and issues
The PR is related to the discussion found in #94 and the issues raised in #97
In addition I have created #119 to document a potential need for AsyncResponseTransformer that could also return an EventLoopFuture. Allowing a user to return something that conforms to ResponseTransformable is not an option as this would introduce the possibility to transform a Response which would lead to the challenges described above.

### Testing
Included more tests to test the different `ResponseTransformer` and `EncodableResponseTransformer` mutations.

### Reviewer Nudging
Investigate the impact of the code changes on the existing exporters and how request are handled throughout Apodini.


Sources/Apodini/Modifier/ResponseModifier/EncodableResponseTransformer.swift
```diff
@@ -0,0 +1,78 @@
...
+    /// - Parameter response: The input as a type erasured `Action<AnyEncodable>`
+    /// - Parameter eventLoop: The `EventLoop` that should be used to retrieve the `Action` of the `Response`
+    /// - Returns: The output as a type erasured `EventLoopFuture<Action<AnyEncodable>>`
+    public func transform(response: Action<AnyEncodable>, on eventLoop: EventLoop) -> EventLoopFuture<Action<AnyEncodable>> {
+        guard let typedResponse = response.typed(Self.Response.self) else {
+            fatalError("Could not cast the `Action<AnyEncodable>` passed to the `ResponseTransformer` to the expected \(Action<Self.Response>.self) type")
+        }
+        
+        let transformed: Action<TransformedResponse.ResponseContent>
+        switch typedResponse {
```
moritzschuell: Why not use `typedResponse.map(transform)` here to replace the switch?

Sources/Apodini/Modifier/ResponseModifier/ResponseTransformer.swift
```diff
@@ -0,0 +1,67 @@
...
+//  
+//
+//  Created by Paul Schmiedmayer on 1/4/21.
+//
+
+import NIO
+
+
+/// A `ResponseTransformer` transforms a `response` of the type `Response` to a instance conforming to `TransformedResponse`
+public protocol SomeResponseTransformer: AnyResponseTransformer {
```
lukaskollmer: What is this being used for?

Sources/Apodini/Components/Action.swift
```diff
@@ -0,0 +1,120 @@
...
+//
+//  Created by Paul Schmiedmayer on 1/4/21.
+//
+
+import NIO
+
+
+/// Used to return from the `handle` method
+/// by `Components` that expose client-side streaming endpoints.
+public enum Action<Element: Encodable> {
```
lukaskollmer: Shouldn't the documentation here say service-side stream?

Sources/Apodini/Components/Response.swift
```diff
@@ -0,0 +1,90 @@
...
+    
+    public func action(on eventLoop: EventLoop) -> EventLoopFuture<Action<ResponseContent>> {
+        eventLoop.makeSucceededFuture(.automatic(self))
+    }
+}
+
+
+// MARK: Dictionary + Response
+extension Dictionary: Response where Key: Encodable, Value: Encodable {
+    public typealias ResponseContent = [Key: Value]
```
lukaskollmer: Why do the `Optional`, `Array` and `Dictionary` extensions have implementations whereas `String` et al below do not? I commented out the code and they still compiled and worked fine.

Sources/Apodini/Components/Response.swift
```diff
@@ -0,0 +1,90 @@
...
+}
+
+
+// MARK: - Primitive Types
+extension String: Response {}
+extension Int: Response {}
+extension Double: Response {}
+extension Float: Response {}
+extension Bool: Response {}
+extension UUID: Response {}
```
lukaskollmer: I'd just go all-in here and also add `{U}Int{8|16|32|64}`, as well as `Set` and `Data`
```suggestion
extension UUID: Response {}
extension Array: Response where Element: Encodable {}
extension Set: Response where Element: Encodable {}
extension Dictionary: Response where Key: Encodable, Value: Encodable {}
extension Optional: Response where Wrapped: Encodable {}
extension Data: Response {}
extension Int8: Response {}
extension Int16: Response {}
extension Int32: Response {}
extension Int64: Response {}
extension UInt8: Response {}
extension UInt16: Response {}
extension UInt32: Response {}
extension UInt64: Response {}
```

Sources/Apodini/Components/Action.swift
```diff
@@ -0,0 +1,120 @@
...
+            }
+        case .end:
+            return .end
+        }
+    }
+}
+
+
+// MARK: Action: Response
+extension Action: Response {
```
lukaskollmer: nitpick but I'd argue this conformance should be directly in the `struct Action` definition. if only to make it clearer to someone reading the code that every `Action` is an allowed `Response`

lukaskollmer (commented):
i really like this.

I think a big advantage of this approach of having `Response` as a protocol for "anything that can be returned from a handler" would be that it'd make it a lot easier to add other top-level response types in the future (apart from the primitive types, `EventLoopFuture`, and `Action`)

One more general change i'd recommend would be to maybe rename the `Action` type to something more specific. I think calling it `Action` is too broad/generic, since it only refers to the action to be taken based on a handler response.

As for the fact that returned types now also have to declare conformance to the `Response` protocol, I don't think this is a problem. Vapor already has the same limitation for types using their `Content` protocol.

An important thing I'd like to mention is that, in addition to "object → response" conversions, we also need to retain the ability to perform "response → object" conversions. This will be important for interactions between handlers, in order to access the value returned by another handler.
This currently is pretty simple, since we're using `Encodable`, and `Encodable` types usually are also `Decodable`. (I don't think this PR negatively affects this, but i think it might be good to keep in mind that in some cases we care not just about encoding response objects, but also about decoding them.)

Tests/ApodiniTests/ModifierTests/ResponseTransformerTests.swift
```diff
@@ -0,0 +1,279 @@
...
+                    Text("Hello Bernd")
+                        .response(EmojiEncodableResponseTransformer())
+                        .operation(.create)
+                }
+            }
+        }
+        
+        TestWebService.main(app: app)
+        
+        ResponseTransformerTests.emojiTransformerExpectation = self.expectation(description: "EmojiTransformer is exectured")
```
moritzschuell: Is `exectured` a typo?

moritzschuell (commented):
I really like the additions of this PR 👍

In my opinion we still need to think about whether we really want to allow `ResponseTransformer`s to change the case of an `Action` (this was also mentioned by @Supereg in #97). 
I guess, allowing such a transformation could cause some confusion / unexpected behavior, albeit most probably caused by the Apodini-User's fault.

Sources/Apodini/Components/Action.swift
theMomax: It is both, really: for client-side streaming, the `Handler` needs to be able to return `.nothing` for all but the last (which is `.final(...)`. For service-side streaming, you need the differentiation between `.send(...)` and `.final(...)`/`.end`.

Sources/Apodini/Components/Action.swift
```diff
@@ -0,0 +1,120 @@
...
+    /// Indicates that the given `Element` should be sent
+    /// to the client, and more elements might follow.
+    case send(_ element: Element)
+    /// Indicates that the given `Element` should be sent
+    /// as a final response to the client.
+    /// Will be the last message on the stream sent by the server.
+    case final(_ element: Element)
+    /// No specific action was defined by the `Handler` or `EncodableResponseTransformer`.
+    /// The eventual action will be decided by Apodini
+    case automatic(_ element: Element)
```
theMomax: I (and @Supereg) think that letting the exporter decide which `Action` is taken here is not something we should aim for. We see the danger of inconsistencies between exporters with the result that the control-flow logic of a `Handler` works for one exporter, but not another. If one exporter interprets `automatic` as `send` where all others interpret it as `final`, the `Handler` would be evaluated in a state it was never meant to be evaluated in when called from that one exporter. This is dangerous and the advantage of this approach over having `final` as the default is marginal.

Sources/Apodini/Components/Response.swift
```diff
@@ -0,0 +1,90 @@
...
+    /// The `Encodable` type that is returned from a `handle()` method in a `Handler`.
+    /// The type can be wrapped by an `EventLoopFuture` or an `Action`.
+    public typealias ResponseContent = Self
+    
+    /// Default Implementation for types conforming to `Response`
+    ///
+    /// Transforms an `Response` into an `Action` to be processed by the Apodini
+    /// - Parameter eventLoop: The `EventLoop` that should be used to transform the `Response` to an `Action` if needed
+    public func action(on eventLoop: EventLoop) -> EventLoopFuture<Action<Self>> {
+        eventLoop.makeSucceededFuture(.automatic(self))
```
theMomax: This must be changed to `.final` if we decide to remove `.automatic`. Same applies for all the default implementations below.

Sources/Apodini/Components/Response.swift
```diff
@@ -0,0 +1,90 @@
...
+
+
+// MARK: EventLoopFuture + Response
+extension EventLoopFuture: Response where Value: Response {
+    public typealias ResponseContent = Value.ResponseContent
+    
+    public func action(on eventLoop: EventLoop) -> EventLoopFuture<Action<ResponseContent>> {
+        flatMap { value in
+            value.action(on: eventLoop)
+        }
```
theMomax, line 47: Do we need to `.hop(to: )` the given `eventLoop` here? I am totally not sure, but might be something to think about in detail.

Tests/ApodiniTests/ComponentTests/ActionTests.swift
```diff
@@ -0,0 +1,162 @@
...
+            .final(message)
+        }
+    }
+    
+    struct FutureBasedHandler: Handler {
+        var eventLoop: EventLoop
+        var message: String
+
+        func handle() -> EventLoopFuture<EventLoopFuture<EventLoopFuture<Action<String>>>> {
+            // this tests if the `EventLoopFutureUnwrapper` properly unwraps multiple nested EventLoopFutures
```
theMomax: `EventLoopFutureUnwrapper` doesn't exist anymore.

theMomax (changes requested):
Looks pretty good! The most critical point @Supereg and I found (we did a pair-review, thus the weird "we" in all the comments) is the following:

We think that `Action.automatic` is not a desirable concept. It increases ambiguity on how a `Handler` should be exported/how it behaves. What is wrong with just using `.final` as the default? If that is not desirable in a certain case, the user can still opt for returning an `Action` on the `Handler`.

bauer-andreas (commented):
_There is a tldr at the bottom 🙃_

Maybe some nitpicking here, but we (me and @theMomax) think that maybe `Response` isn't the best name for the new protocol. And I think the refactoring done to `ResponseTransformer` shows best where we are coming from.

First of all I think it is to be discussed if we should allow `ResponseTransformer`s to alter the returned action. A Handler might free resources when he returns `.end` not expecting the returned `Action` to change. Thus such action transforming `ResponseTransformers` must be used with caution (my stance would be to even disallow it).

But should we go for allowing the transformation of `Action`s, we both agree that it is rather a corner case and not the "default" behavior. Like our idea would be that the (standard) `ResponseTransformer` is the one which isn't able to transform Actions. And we need to find a different name for the action transforming one.
However that's were we think `Response` is maybe a bit off as the protocol name. Because with that name the current naming of transformer makes sense. `ResponseTransformer` transform instances conforming to `Response` and `EncodableResponseTransformer` transforms values which are `Encodable` basically. Previously the "Response" in `ResponseTransformer` had no specific meaning, but saying "something returned as response to a request", but now its more or less coupled to the `Response` protocol. (Maybe a bit esoteric argumentation 🙃).
But our vision would be to have something like `ResponseTransformer` and `ActionResponseTransformer` (or whatever sound better than this thing `ActionTransformer`, idk), and having that in combination with `Response` as the protocol name is a bit odd, as the `ResponseTransformer` doesn't actually transform a instance of `Response`.

---

Your PR currently suggest a user facing API like the following:
```swift
struct User: Codable, Response {}
```

I think the API can be improved/made simpled, so the user doesn't have to implement two protocols.
And currently the type checking doesn't force the user **directly** to implement `Codable`. But I think there is probably no way around this.

Our first though was to come up with a protocol hierarchy depicted below.

```swift
public protocol BaseContent {
  // whatever we need
}

public protocol Content: BaseContent, Codable {}

protocol ContentContainer: BaseContent {}

extension EventLoopFuture: ContentContainer where Element: BaseContent {}
extension Action: ContentContainer where Element: Content {}
```

Basically `Response` becomes `BaseContent` and `ContentContainer` and the new `Content` protocol creates a union of
`Codable` and `BaseContent`.
Not sure if that is the pretties solution, as the `BaseContent` protocol (which should definitely be renamed) is declared as public.
However it probably better allows for future additions?

---

So to sum up, its maybe worth considering a different name for `Response`, we would like to see the "default"
`ResponseTransformer` to be the one without transforming Actions (if we allow it) and there should be a shortcut
to implement both `Response` and `Codable`.

So tldr: name nit picking & refinements in user facing API

bauer-andreas:
My #94 PR previously aimed to bring back `EventLoopFuture` support in a "quick and dirty" way (with a emphasize on quick), until we have the real deal as discussed in #73. Obviously this had some drawbacks where we didn't have proper type checking for stuff like nesting Actions in Actions.
However I think we can close #94 in favor of this PR, which already is kind of the real deal and should provide a good starting point to incorporate the ideas Mathias suggested(?).
Great work 👍

Sources/Apodini/Modifier/ResponseModifier/ResponseTransformer.swift
PSchmiedmayer: Thanks for noticing, this was a leftover of a previous attempt to restructure `ResponseTransformer`s, I have removed it 👍

Sources/Apodini/Components/Action.swift
PSchmiedmayer: I have updated the documentation on the renamed `Response` type

Sources/Apodini/Components/Response.swift
PSchmiedmayer: Thanks for the hint, I have updated the implementation to also support `Optional`, `Array` and `Dictionary` and different `Int` types.

Sources/Apodini/Modifier/ResponseModifier/EncodableResponseTransformer.swift
PSchmiedmayer: Great point, you are completely right, I have updated the implementation 👍

Tests/ApodiniTests/ModifierTests/ResponseTransformerTests.swift
PSchmiedmayer: Yes, this is a typo, I have fixed it 👍

Sources/Apodini/Components/Response.swift
PSchmiedmayer: Good point, thanks!
If we are already on the same event loop this is basically no overhead and a user would expect that the `EventLoopFuture` returned by this function is on the `EventLoop` passed in so I will add a `hop(to: )` call 👍

PSchmiedmayer:
@lukaskollmer @moritzschuell @Supereg @theMomax Thanks for your input! 
I have updated the PR based on your feedback.
I will convert it from a draft to a normal PR and we can close #94 👍

## Implementation Improvements
I have addressed the code level feedback from @lukaskollmer @theMomax, @Supereg and @moritzschuell, thanks for the input.  I have added comments under the original comments and resolved the comments that I have addressed. 

## Naming the Protocols 
From @lukaskollmer :
> One more general change i'd recommend would be to maybe rename the Action type to something more specific. I think calling it Action is too broad/generic, since it only refers to the action to be taken based on a handler response.

And @theMomax & @Supereg :
> So to sum up, its maybe worth considering a different name for Response, we would like to see the "default"

And the more detailed comments [in the comment of @Supereg](https://github.com/Apodini/Apodini/pull/98#pullrequestreview-563730223)

I have renamed the `Action` type `Response ` and we now have a `ResponseTransfromable` protocol that incorporates the implementation that was previously located in the `Response ` protocol including a different naming for the function to transform the conforming type to a `Response`.
To make conformance simpler for developers using Apodini we now have a `Content` protocol that includes the implementation of `Encodable` and `ResponseTransfromable` in one to make the conformance of external types easier. 
I have updated the PR description above the reflect these changes.

## ResponseTransformer
Thanks for the feedback from @moritzschuell, @theMomax and @Supereg on this. I agree that transforming an `Reponse` might pose some challenges for `Apodini` and might lead to unclear behavior from a developer's perspective.
`ResponseTransformer` now transform from an `Encodable` type to an other `Encodable` type and just use `Response.map`.
In addition I have created #119 to document a potential need for `AsyncResponseTransformer` that could also return an `EventLoopFuture`. Allowing a user to return something that conforms to `ResponseTransformable` is not an option as this would introduce the possibility to transform a `Response` which would lead to the challenges described above.

## `Response.automatic`
To get this PR into develop and unblock all elements of Apodini that rely on `EventLoopFuture` I have removed `Response.automatic` from this PR. I would nevertheless argue that it might be needed to reduce the possibility of some ambiguities that might arise from transforming all `Encodable` types to `.final`. I will create a new issue and will link it in the next team meeting and in this PR once I have created it.

Sources/Apodini/Components/Response.swift
PSchmiedmayer: As noted in [the comment below I have removed `Response.automatic`](https://github.com/Apodini/Apodini/pull/98#issuecomment-757932777) from the pull request. I will create an issue that documents the concern with `.automatic` in more detail and we can use this issue for a discussion around the topic, maybe there are other options to solve the issues I have in the back of my mind regarding automatically using `.final` for transforming `Encodable` types.

Sources/Apodini/Components/Action.swift
PSchmiedmayer: As noted in [the comment below I have removed `Response.automatic`](https://github.com/Apodini/Apodini/pull/98#issuecomment-757932777) from the pull request. I will create an issue that documents the concern with `.automatic` in more detail and we can use this issue for a discussion around the topic, maybe there are other options to solve the issues I have in the back of my mind regarding automatically using `.final` for transforming `Encodable` types.

hendesi (approved):
Sorry for the late review!

I really like the changes of this PR. From a database point of view, this finally resolves the `EventloopFuture` bug and re-enables database functionalities again. As I have nothing to add in terms of code changes and you have already addressed the issues that have been pointed out, I'd say we get this merged 👍

theMomax (approved):
All the remarks I made in my last review seem to be resolved, thank you!

bauer-andreas (commented):
A bit late to the party, but also a approval from my side. Great work 👍

PSchmiedmayer:
As described in #97 there is no need for an `.automatic` `Response` case as we don't allow for `Optional<Response<Encodable>>`. Therefore I won't create an issue [as mentioned in the comment above](https://github.com/Apodini/Apodini/pull/98#issuecomment-757932777).

PR #100, opened 2021-01-04
Use same service & method names for Protobuffer and gRPC exporter 

moritzschuell:
# Use same service & method names for Protobuffer and gRPC exporter

The Protobuffer exporter and gRPC exporter are using different naming strategies for services and their methods.
This means the generated `.proto` file does not actually describe what is exporter by the gRPC exporter.

## :bulb: Proposed solution
In a first step towards some shared logic between Protobuffer and gRPC exporter, this small PR adds an extension to `Endpoint` in `GRPCExporterUtility.swift`, that can be used to retrieve the service and method name for an `Endpoint`. We may extend this utility file to a larger shared logic in the future, if this is feasible in other situations as well.

### Problem that is solved
Inconsistent naming between Protobuffer & gRPC exporter.

### Implications
None, except for the two mentioned exporters.

none 🤖

### Related PRs
none

### Testing
Adds a new test case for the newly created default naming strategy.


Sources/Apodini/Semantic Model Builder/gRPC/GRPCExporterUtility.swift
```diff
@@ -0,0 +1,44 @@
+//
+//  GRPCExporterUtility.swift
```
nityanandaz: I prefer a different naming convention for extensions: `Endpoint+GRPC.swift`

Utility is very ambiguous and the former has a long history 😅 

> A category is usually declared in a separate header file and implemented in a separate source code file. In the case of XYZPerson, you might declare the category in a header file called XYZPerson+XYZPersonNameDisplayAdditions.h.

From: https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/CustomizingExistingClasses/CustomizingExistingClasses.html#//apple_ref/doc/uid/TP40011210-CH6-SW2

nityanandaz (changes requested):
Thanks 👍

nityanandaz (approved):

PR #102, opened 2021-01-06
Handle Connection in Shared Model + `Reducible`

theMomax:
# Handle Connection in Shared Model + `Reducible`

There was no implementation of handling `Connection` in the context of an exporter. In tests the `Connection` was updated using the `.withEnvironment` method. This method did not use `apply` and thus would not find `@Environment(\.connection)`s which are hidden inside e.g. a `DynamicProperty`.

Furthermore, until now the concept of a `Request` was rather "a message from the client" and not "an input for the `Handler`". At the same time a `Request` is absolutely necessary for evaluating a `Handler`. This becomes a problem once the evaluation of a `Handler` can be triggered by something else but an incoming `Request`, e.g. an `@ObservedObject`. 

## :bulb: Proposed solution
Updating the `Connection` is handled by the shared model, more specifically `InternalConnectionContext` and `InternalEndpointRequestHandler`. The `ConnectionContext`'s `handle` method was extended by a parameter `final: Bool = true`. This is the API provided to the exporter for controlling the `Connection`'s `state` property:
```swift
// Connection.state == .end
context.handle(request: latestInput, eventLoop: eventLoop)
context.handle(request: latestInput, eventLoop: eventLoop, final: true)

// Connection.state == .open
context.handle(request: latestInput, eventLoop: eventLoop, final: false)
```

Furthermore, a new protocol was added:
```swift
protocol Reducible {
    func reduce(to new: Self) -> Self
}
```
The `ExporterRequest` must conform to this protocol:
```swift
func createConnectionContext<I: InterfaceExporter>(for exporter: I) -> AnyConnectionContext<I> where I.ExporterRequest: Reducible
```
The default implementation of `Reducible` just returns the `new` `Self`, which would be a stateless approach. However, an exporter can use this to implement a stateful approach:
 E.g. the WebSocket exporter will use this mechanism to aggregate incoming messages:
```swift
// first client-message
// { 
//     ...
//     "parameters": {
//         "userId": 3,
//         "documentId": 5
//     }
// }
@Parameter var userId: Int // 3
@Parameter var documentId: Int // 5

// second client-message
// { 
//     ...
//     "parameters": {
//         "documentId": 12
//     }
// }
@Parameter var userId: Int // 3                     <- this didn't change and is not missing
@Parameter var documentId: Int // 12
```

Now, the `ConnectionContext` can use the `reduce` function to derive a `latestRequest: I.ExporterRequest?` from incoming messages:

```swift
// exporterRequest is the incoming request (argument of the `handle` method)
let newRequest = self.latestRequest?.reduce(to: exporterRequest) ?? exporterRequest
            
let validatedRequest = try validator.validate(newRequest, with: eventLoop)

self.latestRequest = newRequest

return self.requestHandler(on: Connection(state: final ? .end : .open, request: validatedRequest))
```
RequestHandler now operates on a `Connection`, which in turn carries the `request`. It uses `connection.enterConnectionContext` to inject both, `@Environment(\.connection)` and `@Parameter` based on the `Connection` object.

### Problem that is solved
* Handling of `Connection` was difficult to implement for exporters.
* `Request`s can be aggregated without the exporter having to manage state.

### Implications
The API for controlling the `Connection.state` is backwards compatible.
Everything previously conforming to `ExporterRequest` must now also conform to `Reducible`. Empty conformance is available.

I encountered a bug I couldn't really grasp while implementing `enterConnectionContext`. While `Connection` was successfully added to the `Environment.dynamicValues` dictionary, the map lookup inside `wrappedValue` could not find it.
Bildschirmfoto 2021-01-04 um 17 53 56

Removing the `ObjectIdentifier`-wrapper around the keypath from the dictionary's key fixed the bug. Still, I don't really get why that is and why the bug didn't appear for the old `withEnvironment` function.

### Related PRs
~~None I think.~~ #103 shows how the WebSocket exporter's uses the new API for `Connection` and `Reducible`.

### Testing
Tests that previously used `withEnvironment` to inject `Connection` were rewritten to use `enterConnectionContext`. `Reducible` is not tested as it contain any real logic yet. A follow-up PR containing the WebSocket exporter's implementation will test the described aggregation scenario.

### Reviewer Nudging
Start at the `InternalConnectionContext`. This is where the new logic lives. The rest is basically just API changes/additions to support this logic. From `InternalConnectionContext.handle` continue to `InternalExporterRequestHandler.callAsFunction` and then `Connection.enterConnectionContext` in `Traversable.swift`. Thats basically it.


Sources/Apodini/Semantic Model Builder/REST/RESTEndpointHandler.swift
```diff
@@ -6,7 +6,7 @@ import Foundation
 @_implementationOnly import Vapor
 import protocol FluentKit.Database
 
-extension Vapor.Request: ExporterRequest, WithEventLoop {}
+extension Vapor.Request: ExporterRequest, WithEventLoop, Reducible {}
```
bauer-andreas: If `Reducible` is a required protocol, can we make so that the Exporter doesn't need to specify the conformance explicitly? I think that would do a good job hiding this complexity to Exporter which don't use aggregated Requests.

theMomax: You are right, this was a bit ugly. I fixed it. (When I first tried, I had some problems with `Self` as a protocol requirement and then forgot about this convenience extension when I resolved that issue.)

moritzschuell (approved):
Looks great, thank you! 
I have nothing to comment 🙂

PR #103, opened 2021-01-06
WebSocket: Support for Action + Connection

theMomax:
# WebSocket: Support for Action + Connection

The WebSocket exporter did not support `Action.end` or `Action.nothing` and did not induce a final evaluation of the `Handler` with `Connection.state` set to `.end`.

The calls to `context.handle` were not synchronized, which would have caused trouble when used with async `Handler`s (i.e. ones returning an `EventLoopFuture`).

## :bulb: Proposed solution
The exporter function was adapted to handle all different cases of `Action` and communicates accordingly to the `Router`.

In order to support `ConnectionState.end`, the protocol run by the `VaporWSRouter` had to be adapted: The meaning of the `CloseContextMessage` was changed from "this context is now closed" to "I am not going to send another message on this context". Thus a context is now closed, once both participants have sent that message.

Furthermore, aggregation of client-messages is no longer done by the `InputParameter` implementation of the `WebSocketInfrastructure` target, but is implemented in `WebSocketInterfaceExporter` using the `Reducible` protocol.

When the client sends the `CloseContextMessage`, the exporter invokes the `ConnectionContext.handle` function with `final: true` and an empty input (which equals the latest input after aggregation). 

The calls to `context.handle` were synchronized by implementing a custom `map`-like `Publisher` that awaits futures returned by its transformer-closure. This `SyncMap` `Publisher` is non-blocking.

### Problem that is solved
* WebSocket exporter would crash on usage of `Apodini.end` or `Apodini.nothing` and could not control `Connection.state`.
* Code duplication for implementing input-aggregation was avoided by embracing shared-model approach.
* WebSocket exporter could not be used safely with async `Handler`s

### Implications
VaporWSRouter's protocol changed a bit, but there wasn't even a client-implementation before, so nobody cares I guess😁.

I also added a Client implementation. It allows for sending a series of client-messages to an endpoint and collecting all responses into an array.

### Related PRs
#102 provides basis for shared-model part. ~~thus #102 must be merged first.~~ #102 was merged.

### Testing
`WebSocketInterfaceExporterTests.testWebSocketConnectionClientStreamSchema` and other WebSocket exporter tests cover the exporter. There are extra tests for the `SyncMap` publisher called `CombineSyncMapTests`.

### Reviewer Nudging

I suggest you first read the docs of `.syncMap`. Then look at WebSocketInterfaceExporter.swift and WebSocketInterfaceExporterTests.swift and finally come back to the rest of the WebSocketInfrastructure target.

@nerdsupremacist could you please take a detailed look at the `syncMap` implementation and the related tests? It works find in my context, but since its rather complex and open source I don't want to deliver some half-baked code.


Sources/Apodini/Semantic Model Builder/REST/RESTEndpointHandler.swift
```diff
@@ -6,7 +6,7 @@ import Foundation
 @_implementationOnly import Vapor
 import protocol FluentKit.Database
 
-extension Vapor.Request: ExporterRequest, WithEventLoop {}
+extension Vapor.Request: ExporterRequest, WithEventLoop, Reducible {}
```
bauer-andreas: Is this up to date with the current develop?
If so I would leave pretty much the same comment as on https://github.com/Apodini/Apodini/pull/102#discussion_r552575194.
But it seems like it just got lost while rebasing(?)

theMomax: The automatic conformance was not lost, but the redundant conformances were added back in. Should be fixed now. Thanks!

awoc (approved):
Looks really good. Your comments helped a lot 👍

Issue #96, opened 2021-01-03
[BUG] Protobuffer generation: handle `Apodini.PathParameter`


nityanandaz:
**Describe the bug**
`TestWebSevice` produces an invalid `.proto` declaration:
```proto
syntax = "proto3";

service V1User:E1A4F8A7-9E95-4Ea3-Ae9F-50Cb99B19792Service {
  rpc handle (IntMessage) returns (UserMessage);
}
```

**To Reproduce**
1. Run target `TestWebService`
2. Open link http://0.0.0.0:8080/apodini/proto

**Expected behavior**
🤷 TBD

**Additional context**
- https://github.com/Apodini/Apodini/blob/bf59a8d120936c1735b2e580baeb4b62ca822604/Sources/TestWebService/main.swift#L98

bauer-andreas:
Some context:
I'm currently in the process of creating a dedicated path model in the shared semantic model builder to more easily parse PathComponents and PathParameters. Currently the raw `Parameter` instance is supplied to the `PathComponent` array, which is not really great to handle.
A standalone `@PathParameter` instance (or a `@Parameter` instance derived from a `@PathParameter`) doesn't yield any useful information. In fact, the only info we have is the `id`. As a `@PathParameter` can be defined in multiple `Handlers` we even don't have a distinct name/label for it.
So we really only have proper information for a `@PathParameter` in the context of a specific `Endpoint`/`Handler`. As this is given, it isn't really a problem. I just means that with the current approach it is rather ugly to handle path parameters.

As I need a proper PathParameter model as well when working on the implementation of the Relationship retrieval, first step will be to create a dedicated PR creating such a model, which is more powerful.

With that you could then use the parameter name in the service name, but not sure if that improves your situation, as having the parameter name isn't ideal as well (as having a service name `V1useruserId` isn't semantically valuable), but would at least remove the UUID string.

nityanandaz:
This is indeed resolved and no longer produces an invalid `.proto` file 🙂 

https://github.com/Apodini/Apodini/blob/0ac1944d5ebfc51315078fe0c70f1c985f9c785c/TestWebService/Sources/TestWebService/UserSubsystem/Components/UserComponent.swift#L12

becomes

```proto
service V1UseruseridService {
  rpc userhandler (UserHandlerMessage) returns (UserMessage);
}

message UserHandlerMessage {
  IntMessage userId = 1;
}

...
```

PR #104, opened 2021-01-06
Incorporated a path parameter model into the semantic model

bauer-andreas:
# Incorporated a path parameter model into the semantic model

Currently a `PathComponent` array is used as the path representation of a `Endpoint`. As of now Apodini has three types conforming to `PathComponent` (`_PathComponent` to be more specific): `String`, `Parameter<Element>` and `Version`.

Having the generically typed `Parameter`, it doesn't allow for little on the fly checks like "is this current path component a PathParameter". It is impossible without constructing a Visitor.

In addition, `Parameter` carries information which is represented in a unfavorable way, like for example having the raw `Parameter<Element>` doesn't allow to access the wrapped Type in the case where `Element` is an Optional (similar to how the `EndpointParameter<Type>` unwraps the optional type in its generic.
But more importantly, `Parameter` carries information which might not be valid: Referring to the below example of a `@PathParameter` definition, properties like `name`, `defaultValue`, `options` of `Parameter` are non existent for instances created from a `PathParameter`. Really the only information currently is the `id` of the `PathParameter`, which is used to match up `Parameter` definitions in the `Handler`.

The name of a `PathParameter` can really only be determined, if we know the `Handler` where the `PathParameter` is used. Additionally it creates a situation where the name of a `PathParameter` is ambiguous. 

Example `PathParameter` definition:

```swift
struct UserHandler: Handler {
  @Parameter var userId: String
}

@PathParameter var param: String
var content: some Component {
  Group($userId) {
    UserHandler(userId: $param)
    PostHandler(userId: $param)
  }
}
```

## :bulb: Proposed solution

This PR introduces a new `EndpointPath` model.
Essentially this is a `enum` with three cases:
```swift
enum EndpointPath {
    case root
    case string(_ path: String)
    case parameter(_ parameter: AnyEndpointPathParameter)
}
```
This makes quick checks as described above rather simple.

The `.parameter` holds an instance of `EndpointPathParameter<Type>` (The `Type` generic is identical to how it is done in the `EndpointParameter<Type>` model).
Such a path parameter instance can be scoped to a specific Endpoint. If such a scope is present, information like the `name` becomes available. This scoping is done automatically for every available API where the `Endpoint` is known. To be specific this is the case for **all** exporter facing APIs. So an Exporter will always deal with scoped `EndpointPathParameters` (thus having a valid name).
But because a `@PathParameter` might not be defined inside a `Handler` but still contained in the `path`, the name might still default to the string representation of the `id`. See below.

### Problem that is solved

As above explained, the new model addresses any complications a Exporter would currently encounter when dealing with PathParameters.

This PR additionally addresses the issue mentioned in #55. This Draft PR highlighted the edge case, that a user could pass a `Parameter` into the path, but not define a `Parameter` in the Handler defined under this path.

My https://github.com/Apodini/Apodini/pull/55#issuecomment-743771845 detailed that most exporters should be able to gracefully workaround such cases, with the only disadvantage that they can't provide a user friendly name for the given parameter (like e.g. the user facing documentation generated by the OpenAPI exporter).
For such scoped path parameter instances, the `name` property will default to the `pathId` (meaning the string representation of the parameter `id`).
Exporters can use the `scopedEndpointHasDefinedParameter` property to check for such scenarios, where no Parameter declaration is found in the `Handler`.

### Implications

Current Exporter implementations need to refactor any of their `PathBuilder`s to adapt to the new model and take advantage of newly added naming information.
I've done that for any exporter already in develop. As I hope that #35 will get merged soon, I will adapt that Exporter as well.

As the PR heavily refactors the PathParameter model, I decided to include a first version of the `PathComponentFunctionBuilder` mentioned in the `Relationship.md` proposal. For my restructured parser of `PathComponents` I needed to consider that the future implementation will need a way to parse described `PathComponentModifier`s. 

### Related PRs

* #55 Pointing out the possibility of unhandled illegal `@PathParameter` definitions

### Related Issues

* #96: Using this PR it creates a somewhat tolerable solution to create default service names

### Testing

Current test case already cover those newly introduced models as those are used in the different exporters.

### Reviewer Nudging

There are two separate things to consider, for one the restructured `PathComponent` with its modifiers, and secondly the `EndpointPath` model building on top of that.


Sources/Apodini/Semantic Model Builder/ParameterModelBuilder.swift
```diff
@@ -0,0 +1,114 @@
...
+                label: label,
+                necessity: .optional
+            )
+        } else {
+            var `default`: Element?
+            if let value = parameter.defaultValue {
+                `default` = value
+            }
+
+            endpointParameter = EndpointParameter<Element>(
```
awoc: I think you can remove `<Element>` here

Sources/Apodini/Semantic Model Builder/ParameterModelBuilder.swift
```diff
@@ -0,0 +1,114 @@
...
+        name: String,
+        label: String,
+        necessity: Necessity
+    ) -> AnyEndpointParameter {
+        var `default`: Element.Member?
+        if let value = self.defaultValue {
+            `default` = value.optionalInstance
+        }
+
+        return EndpointParameter<Element.Member>(
```
awoc: Same here

Sources/Apodini/Semantic Model Builder/REST/RESTInterfaceExporter.swift
```diff
@@ -60,29 +50,59 @@ extension Operation {
...
+
+        switch bindAddress {
+        case .hostname:
+            var uriPrefix = configuration.tlsConfiguration == nil ? "http://" : "https://"
+            uriPrefix += configuration.hostname
+
+            let port = configuration.port
+            if port != 80 {
+                uriPrefix += ":\(port)"
+            }
```
awoc: I guess you can also check for port 443 if tls is configured

Sources/Apodini/Components/Path/RelationshipNameModifier.swift
```diff
@@ -0,0 +1,36 @@
+//
+// Created by Andi on 05.01.21.
+//
+
+struct RelationshipNameContextKey: ContextKey {
+    // Move to Optional type once fixed https://github.com/Apodini/Apodini/issues/75
+    static var defaultValue: String = ""
```
awoc: Shouldn't this trigger a redundant type annotation swiftlint warning? 🤔

Sources/Apodini/Semantic Model Builder/Model/Endpoint.swift
```diff
@@ -123,102 +130,108 @@ struct Endpoint<H: Handler>: AnyEndpoint {
...
-                child!.addEndpoint(&endpoint, at: pathComponents)
-            } else {
-                fatalError("Encountered PathComponent which isn't a _PathComponent!")
+
+                child = EndpointsTreeNode(path: next, parent: self)
+                nodeChildren[next] = child
             }
+
+            // swiftlint:disable:next force_unwrapping
+            child!.addEndpoint(&endpoint, context: &context)
```
awoc, line 233: You could also use `if let` to remove force unwrapping. Then you also don't need to check `child` for `nil` in line 190.

Sources/Apodini/Semantic Model Builder/ParameterModelBuilder.swift
bauer-andreas: Ah yes, the original version back then, didn't have the type information which the `default` value delivers now. Can indeed be removed 👍

Sources/Apodini/Semantic Model Builder/REST/RESTInterfaceExporter.swift
bauer-andreas: Thanks, overlooked that case, good you pointed that out 👍

Sources/Apodini/Semantic Model Builder/ParameterModelBuilder.swift
bauer-andreas: same as above

Sources/Apodini/Semantic Model Builder/Model/Endpoint.swift
bauer-andreas: I did the requested refactoring as you suggested and rediscovered why I wrote it that way about a month ago (you may have noticed the failing build 🙃):
If you use if let you end up with the need of calling `addEndpoint` twice or the need to declare a additional variable (basically a if let where the body just sets the variable 🤔 I thought both were ugly and went for force unwrap. But I try to find a neat way of handling that maybe.

Sources/Apodini/Semantic Model Builder/ParameterModelBuilder.swift
bauer-andreas: Fixed in f02f2cf

Sources/Apodini/Semantic Model Builder/ParameterModelBuilder.swift
bauer-andreas: Fixed in f02f2cf

Sources/Apodini/Semantic Model Builder/REST/RESTInterfaceExporter.swift
bauer-andreas: Fixed in f02f2cf

Sources/Apodini/Components/Path/RelationshipNameModifier.swift
bauer-andreas: It actually doesn't trigger one 🙃 Maybe because its also used to define the associated value of the `ContextKey`
But have nonetheless fixed it f02f2cf

Sources/Apodini/Semantic Model Builder/Model/Endpoint.swift
bauer-andreas: I tried some stuff, but I think the approach using force_unwrap is honestly the nicest way of writing that function.
Using `if let`, I either have to define a kind of "temp variable", and the `if let` body pretty much consist of assigning that variable, or I would need to have the `addEndpoint` call twice, which isn't a improvement either.
Thus I think the force_unwrapping approach depicts the control flow best. But feel free to leave any change suggestions if you find a better way 👍

theMomax: What about optional chaining?

```suggestion
            child?.addEndpoint(&endpoint, context: &context)
```

Sources/Apodini/Semantic Model Builder/Model/Endpoint.swift
```diff
@@ -64,26 +64,27 @@ struct Endpoint<H: Handler>: AnyEndpoint {
...
+    var absolutePath: [EndpointPath] {
+        storedAbsolutePath
     }
+    private var storedAbsolutePath: [EndpointPath]! // swiftlint:disable:this implicitly_unwrapped_optional
+
     var relationships: [EndpointRelationship] {
-        treeNode.relationships
+        storedRelationship
     }
+    private var storedRelationship: [EndpointRelationship]! // swiftlint:disable:this implicitly_unwrapped_optional
```
awoc, line 78: I don't really see the need to basically declare the same attribute twice. I think this solution would also work and achieve the same thing:

```swift
private(set) var absoulePath: [EndpointPath] = []
private(set) var relationships: [EndpointRelationship] = []
```
This would also remove the need for force unwrapping

awoc (commented):
I just noted down another small improvement. Otherwise the PR looks good to me 👍

Sources/Apodini/Semantic Model Builder/Model/Endpoint.swift
bauer-andreas: The intention here is, that both properties are not considered valid until inserted into a `EndpointTreeNode`. If one would access those properties before it got inserted, you know something has gone fatally wrong.
I wanted to make that explicit in the code, and that's the reason I decided against doing an empty default value as you suggest it.
I could find a way to remove the implicitly unwrapped optional (by making the `stored...` properties Optional and doing an unwrap in the computed properties, creating a fatalError if they are indeed nil.
However I'm not sure how to achieve the same thing with removing the additional properties 🤔

Sources/Apodini/Semantic Model Builder/Model/Endpoint.swift
bauer-andreas: @theMomax I would argue against that. It is semantically incorrect (being a "execute this if it is the case that child is available" vs "the child is to be inserted here") and silently ignores any potential regressions made in the code above.

If the majority is for removing that force_unwrap and you guys can agree on a replacement for this I'm all for it 👍. Just stating my opinion here.

bauer-andreas: I could maybe see the following working, by (ab)using the `??` operator with a closure

```swift
let next = context.nextPath()

let child = nodeChildren[next] ?? {
    // as we create a new child node we need to check if there are colliding path parameters
    switch next {
    case .parameter:
        if childContainsPathParameter { // there are already some children with a path parameter on this level
            fatalError("When inserting endpoint \(endpoint.description) we encountered a path parameter collision on level n-\(context.pathCount): "
                + "You can't have multiple path parameters on the same level!")
        } else {
            childContainsPathParameter = true
        }
    default:
        break
    }

    let child = EndpointsTreeNode(path: next, parent: self)
    nodeChildren[next] = child
    return child
}()

child.addEndpoint(&endpoint, context: &context)
```

Not sure if this really an improvement to code readability though.

tgymnich: Here is my take on this:
```
let child: EndpointsTreeNode

if let nextChild = nodeChildren[next] {
   child = nextChild
} else {
   // as we create a new child node we need to check if there are colliding path parameters
   switch next {
   case .parameter:
      if childContainsPathParameter { // there are already some children with a path parameter on this level
         fatalError("When inserting endpoint \(endpoint.description) we encountered a path parameter collision on level n-\(context.pathCount): "
                       + "You can't have multiple path parameters on the same level!")
      } else {
         childContainsPathParameter = true
      }
      default:
         break
      }

   child = EndpointsTreeNode(path: next, parent: self)
   nodeChildren[next] = child
}

child.addEndpoint(&endpoint, context: &context)

bauer-andreas: Okay, who created a poll 😅😂

awoc: I'm also fine with the original version

awoc (approved):

bauer-andreas:
Just noted a small regression yesterday with the generation of Relationships. They need to be generated once the tree is in its final state. Moving away from the computed relationship property broke this.
Will fix this before merging. (rebased and fixed the issue in 12feeb6)

bauer-andreas:
Note: We can refactor the `RelationshipNameContextKey` once #115 is merged.

tgymnich (approved):

PR #105, opened 2021-01-06
Introduce StaticInterfaceExporter protocol

bauer-andreas:
# Introduce StaticInterfaceExporter protocol

Currently any `InterfaceExporter` is forced to define a `ExporterRequest` type and implement a `retrieveParameter` method.

## :bulb: Proposed solution

This PR adds a second exporter type, namely the `StaticInterfaceExporter` protocol, which removes the requirement for a `ExporterRequest`  type and the definition of a `retrieveParameter` function.

### Problem that is solved

The current situation is problematic for "static" interface exporters like OpenAPI which only server a representation of the WebService but not the WebService itself (see https://github.com/Apodini/Apodini/pull/35#discussion_r551764556).

### Implications

A `BaseInterfaceExporter` protocol was created to prevent code duplication of methods shared by those two exporter types.
The protocol should not be used directly, meaning it is considered to be a internal component of the Exporter API.
Direct usage is also prevented by the Exporter API.

### Related PRs

* #35 introduces the OpenAPI Exporter. Thus that PR would need to be adjusted if this PR gets merged first.
 My plan however is to give priority to #35 and merging this PR second (don't want to create another blocking change for #35). Once its merged I will adjust this PR to modify the OpenAPI exporter accordingly.

### Testing

The newly added parts are currently uncovered as no `StaticInterfaceExporter` exists in our codebase. However once #35 is merged (and this PR is adjusted) those parts will be covered automatically.

### Reviewer Nudging

I consider review of this PR simple.

### DIFF
--- a/Sources/Apodini/Semantic Model Builder/InterfaceExporter/InterfaceExporter.swift
+++ b/Sources/Apodini/Semantic Model Builder/InterfaceExporter/InterfaceExporter.swift
@@ -2,7 +2,7 @@
 // Created by Andi on 22.11.20.
 //
 
-import class Vapor.Application
+@_implementationOnly import class Vapor.Application
 import protocol NIO.EventLoop
 import protocol FluentKit.Database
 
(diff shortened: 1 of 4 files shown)


bauer-andreas:
As #35 is now merged, I made the OpenAPI Exporter conform to the new `StaticInterfaceExporter` protocol. The PR is now ready to merge from my side.

lschlesinger (approved):
Thanks for providing special treatment for `OpenAPIExporter` 💐

theMomax (approved):
LGTM👍🏻

PR #106, opened 2021-01-06
Adding parameter name collisions check

bauer-andreas:
# Adding parameter name collisions check

Currently parameter names aren't checked for collisions (something I already mentioned in https://github.com/Apodini/Apodini/pull/55#issuecomment-743771845).
e.g. one could define the following:

```swift
struct TestHandler: Handler {
  @Parameter("param")
  var param0: Int
  @Parameter("param")
  var param1: SomeStruct
}
```

This isn't a problem if the types match, but complicates things of types don't match.

## :bulb: Proposed solution

The PR introduces the ability for Exporters to define their expected namespaces for parameter naming for a given endpoint.

Exporter can defined an array of `ParameterNamespace`s on the Exporter Implementation. A `ParmaterNamespace` is either one standalone `ParameterType` (like `.lightweight`, defining that a parameter name must be unique for all `.lightweight` parameters) or an array of `ParameterType`s (defining that a parameter must be unique for all parameter types listed in the array).

In addition there are default definitions for the two most used namespaces: `.global` (default) and `.individual`.

### Problem that is solved

As parameter naming is not only derived via its property name, but also by supplying overrides using string constants, the user can introduce colliding parameter names and we can't rely solely on the swift property name uniqueness.

We discussed that Apodini should enforce `.global` (the default) namespace by default and add a configuration to allow more lenient parameter namespaces as defined by the exporters. 
This is currently sketched with the `allowLenientParameterNamespaces` property. We are awaiting the outcome of our `Configuration` DSL to properly make this option configurable.

### Related PRs
--

### Testing
Tests were added covering a variety of namespace definitions.


hendesi (approved):
Sorry for the late review! The code looks good to me. 👍 

Regarding the global name forcing, I don't really see any drawbacks beside the thing you already mentioned. Although this is something which, in my eyes, would not have a huge impact on developers using Apodini. I'm no REST expert, but I would imagine, that one would choose different parameter names for query and body anyway (?). Nevertheless it would be important to document this very well, so there is no confusion for developers.

I like the solution you have already implemented and I think it solves the issue for now very well, so I'm going to approve this PR. If you still want to discuss it in a larger round, you can leave it open until next week 👍

bauer-andreas:
Latest push rebased onto latest `develop` and added some missing documentation.
Additionally `.individual` will currently be the default namespace when the exporter doesn't specify any (or doesn't even call `exportParameter`).

I would like to use the meeting to make a decision on "using consistent collision checks" vs. "exporter specific collision checks". Even If we should decide for using consistent collisions checks, I think this PR is still a valuable addition of documentation 🙃 (besides the collision check implementation).

PSchmiedmayer:
@Supereg Thanks for the PR!
I think having exporter specific collision checks makes sense to me. In case a user would only need a REST Exporter the user might benefit from these exporter specific collision checks. Apart from this I would suggest that `.global` is the default option  if an exporter doesn't specify any further collision checks.
Independent of my opinion I would suggest to add it to the meeting agenda of the next meeting and discuss it there with the whole course.

Sources/Apodini/Properties/Parameter.swift
```diff
@@ -21,7 +21,13 @@ public struct Parameter<Element: Codable>: Property {
 
     
     var id = UUID()
-    var name: String?
+    var name: String? {
+        didSet {
+            if let name = name {
+                precondition(!name.isEmpty, "The name for Parameter cannot be empty!")
```
lukaskollmer: not that it would matter (or even necessarily be an improvement over the current code), but you could do
```swift
precondition(!(str ?? " ").isEmpty, "Parameter name cannot be empty. Use `nil` to un-define a parameter name.")
```
to get rid of the if let

bauer-andreas: Not sure if I like how the `if let` is replaced.
Also not sure if we should suggest to use `nil`, as the real solution is just to not specify a name?

bauer-andreas:
Alrighty, made the adjustment discussed today via d0250f4

First of all I moved the `ParameterNamespace` definition to a static property on the `InterfaceExporter`. This allows to have better control over ensuring that collisions checks are definitely called.
With this, I also based my `ParameterNamespace` definitions on the [OptionSet](https://developer.apple.com/documentation/swift/optionset) protocol (thanks for the heads up @PSchmiedmayer), makes it a bit cleaner 👍 

Lastly, I added my vision of a configuration flag of allowing more lenient namespace definitions made by the Exporters (e.g. by the Rest exporter):  `SharedSemanticModelBuilder.allowLenientParameterNamespaces`. It is currently set to `true` by default. Once configured I would envision having a default value of `false`, therefore always using the default `.global` namespace.

Sources/Apodini/Properties/Parameter.swift
```diff
@@ -50,16 +55,17 @@ public struct Parameter<Element: Codable>: Property {
     ///   - options: Options passed on to different interface exporters to clarify the functionality of this `@Parameter` for different API types
     public init(_ name: String, _ options: Option...) {
         self.defaultValue = nil
-        self.name = name
         self.options = PropertyOptionSet(options)
+        defer { // swiftlint:disable:this inert_defer
+            self.name = name
+        }
```
PSchmiedmayer: I am not sure why you use a defer statement here. As this would be the last operation it will be executed last anyways the defer would not be needed here.
```suggestion
        self.name = name
```

Sources/Apodini/Properties/Parameter.swift
```diff
@@ -70,8 +76,10 @@ public struct Parameter<Element: Codable>: Property {
     ///   - options: Options passed on to different interface exporters to clarify the functionality of this `@Parameter` for different API types
     public init(wrappedValue defaultValue: Element, _ name: String? = nil, _ options: Option...) {
         self.defaultValue = defaultValue
-        self.name = name
         self.options = PropertyOptionSet(options)
+        defer { // swiftlint:disable:this inert_defer
+            self.name = name
+        }
```
PSchmiedmayer: I am not sure why you use a defer statement here. As this would be the last operation it will be executed last anyways the defer would not be needed here.
```suggestion
        self.name = name
```

Sources/Apodini/Semantic Model Builder/Model/Endpoint.swift
```diff
@@ -34,13 +34,29 @@ protocol AnyEndpoint: CustomStringConvertible {
...
+    /// - Returns: Returns the `AnyEndpointParameter` if a parameter with the given `id` exists on that `Endpoint`. Otherwise nil.
     func findParameter(for id: UUID) -> AnyEndpointParameter?
-    
-    func exportParameters<I: InterfaceExporter>(on exporter: I) -> [I.ParameterExportOutput]
+
+    /// Internal method which is called to call the `InterfaceExporter.export(...)` method on the given `exporter`.
+    ///
+    /// - Parameter exporter: The `BaseInterfaceExporter` used to export the given `Endpoint`
+    /// - Returns: Whatever the export method of the `InterfaceExporter` returns (which equals to type `EndpointExporterOutput`) is returned here.
+    func exportEndpoint<I: BaseInterfaceExporter>(on exporter: I) -> I.EndpointExportOutput
```
PSchmiedmayer: Thanks for all the documentation 🚀

Sources/Apodini/Semantic Model Builder/InterfaceExporter/InterfaceExporter.swift
```diff
@@ -85,6 +90,13 @@ protocol InterfaceExporter: BaseInterfaceExporter {
     func retrieveParameter<Type: Decodable>(_ parameter: EndpointParameter<Type>, for request: ExporterRequest) throws -> Type??
 }
 
+extension BaseInterfaceExporter {
+    static var parameterNamespace: [ParameterNamespace] {
+        [] // default is to return nothing, to distinguish if exporter did actually set something
```
PSchmiedmayer: As far as I can see the empty array is replaced with a `[.global]` version in **ParameterNamespace.swift** in `nameCollisionCheck` and we don't check for an empty array anywhere else. It might be more clearer for someone who checks out `parameterNamespace` to find the default option that will eventually be used here instead in `nameCollisionCheck`.

Sources/Apodini/Semantic Model Builder/Model/ParameterNamespace.swift
```diff
@@ -0,0 +1,107 @@
...
+    /// Internal method which kicks off the parameter namespace collision checks
+    func parameterNameCollisionCheck(in namespaces: ParameterNamespace...) {
+        parameters.nameCollisionCheck(on: self, in: namespaces)
+    }
+}
+
+private extension Array where Element == AnyEndpointParameter {
+    func nameCollisionCheck(on endpoint: AnyEndpoint, in namespaces: [ParameterNamespace]) {
+        var namespaces = namespaces
+        if namespaces.isEmpty {
```
PSchmiedmayer, line 76: In case we change the default value to [.global] throwing an error there if an exporter does return an empty collection would be appropriate as this would somehow be a developer error/mistake as the developer would have overwritten the behavior and doesn't provide a strategy.

Sources/Apodini/Semantic Model Builder/SharedSemanticModelBuilder.swift
```diff
@@ -126,7 +131,11 @@ class SharedSemanticModelBuilder: SemanticModelBuilder, InterfaceExporterVisitor
 
     private func call<I: BaseInterfaceExporter>(exporter: I, for node: EndpointsTreeNode) {
         for (_, endpoint) in node.endpoints {
-            #warning("The result of export is currently unused. Could that be useful in the future?")
+            // before we run unnecessary export steps, we first verify that the Endpoint is indeed valid
+            // in the case of not allowing lenient namespace definitions we just pass a empty array
+            // which will result in the default namespace being used
+            endpoint.parameterNameCollisionCheck(in: allowLenientParameterNamespaces ? I.parameterNamespace : [])
```
PSchmiedmayer: I would suggest to explicitly pass `[.global]` as the strictest option here.

PSchmiedmayer (approved):
Looks really good to me, thanks for the implementation! I only had a few smaller comments that could be addressed before merging but don't think we would need a second review for them from my side 🚀

Sources/Apodini/Properties/Parameter.swift
bauer-andreas: The defer is intentional, as otherwise the `didSet` will not execute. `didSet` or `willSet` are not executed for values set within the initializer, probably makes sense in some cases. So my decision was to use defer instead of creating some method which is called in every initializer. This would also work for me if defer isn't considered a good solution. Or we move the empty name check into the `EndpointParameter` or something.

Sources/Apodini/Properties/Parameter.swift
bauer-andreas: Same as above. 👍

Sources/Apodini/Semantic Model Builder/Model/ParameterNamespace.swift
bauer-andreas: Sounds good to me. I will adjust those parts as described in your other comments as well 👍

Sources/Apodini/Properties/Parameter.swift
PSchmiedmayer: Oh, this is an interesting behavior of `defer` I was not aware of. If we decide to keep this I would recommend to add comment explaining the behavior.

But I might have a better solution for the problem.
I would suggest to add a private initializer that includes the precondition and make all properties except of `element` constants. All other initializers then call the private initializer. The 
```swift
    let id: UUID
    let name: String?
    let options: PropertyOptionSet<ParameterOptionNameSpace>
    let defaultValue: Element?
    
    private var element: Element?

    // ...

    public init() {
            // The only time where we need to pass in a parameter even though it is not an input of the initializer, otherwise we would have an endless loop.
        self.init(id: UUID())
    }

    // All other initializer call the private initializer too.
    // ...

    private init(id: UUID = UUID(), defaultValue: Element? = nil, name: String? = nil, options: PropertyOptionSet<ParameterOptionNameSpace> = PropertyOptionSet()) {
        if let name = name {
            precondition(!name.isEmpty, "The name for Parameter cannot be empty!")
        }
        
        self.id = id
        self.name = name
        self.options = options
    }
```
The private initializer then includes the precondition check and as all properties are constants we don't need to care about any possible updates to the `name` property.

bauer-andreas: yeah of course. I could have tried to think of simpler solutions like yours 😄
I guess I just wanted to try out `didSet` and make it work 🙃

I think the unified initializer is the best solution 👍 Thanks

bauer-andreas:
@PSchmiedmayer Addressed your two points with 162a73a (removing the usage of `[]` as a indication for default namespaces) and 4f89bb9 (cleaning up the `Parameter` initializer by moving all the logic into one unified and private initializer).

A quick comment or reaction if that solves your points would be great, would merge the PR then.

EDIT: Not sure if @nerdsupremacist still wants to leave a review? Should I wait?

Sources/Apodini/Properties/Parameter.swift
```diff
@@ -20,8 +20,8 @@ public struct Parameter<Element: Codable>: Property {
     public typealias Option = AnyPropertyOption<ParameterOptionNameSpace>
 
     
-    var id = UUID()
-    var name: String?
+    let id: UUID
+    let name: String?
     private var element: Element?
     internal var options: PropertyOptionSet<ParameterOptionNameSpace>
     internal var defaultValue: Element?
```
PSchmiedmayer: We could also make these constants too.

PSchmiedmayer (approved):
@Supereg The commit resolve the issues I had for the PR. The only very nitpicky comment would be that the other types of a `Parameter` could now also be constants but apart from this I would be happy with merging it.

I think it is fine to merge the PR without a review from @nerdsupremacist . He already has a lot to review and take a look at and it reflects what we have decided on in the team and work meeting 👍

Sources/Apodini/Properties/Parameter.swift
bauer-andreas: Done in 5b7b50a

bauer-andreas:
> I think it is fine to merge the PR without a review from @nerdsupremacist . He already has a lot to review and take a look at and it reflects what we have decided on in the team and work meeting 👍

Alrighty 👍 In any case, if he has any request, we can still tackle it in a follow up PR.

PR #108, opened 2021-01-07
Create targets for Jobs and Notifications

awoc:
# Create targets for Jobs and Notifications

At the moment, the `Notifications` and `Jobs` are part of the Apodini core target. Since both features are not necessary for exporters it is better to move them to separate targets in the project.

## :bulb: Proposed solution
Move `Notifications` and `Jobs` to different targets. This would mean that users need to import both the targets, besides `Apodini`, if they want to use the functionality.

### Implications
- Create a new package called XCTApodini to share testing code with other targets. This new target would include:
    - `XCTAssertRuntimeFailure`: Method to check for runtime failures.
    - `XCTApodini`: Most of the code from `ApodiniTests` is moved to this class. This would allow other `testTargets`to reuse code from `ApodiniTests`. `XCTApodini` also includes a method to add migrations to the testing database if needed.
- Create a new public struct called `EnvironmentValue` in the Apodini core target to allow other targets to add `EnvironmentValues`. This is needed since `EnvironmentValues` is restricted to be internal only and can therefore not be used by other targets.
- The modifer `addNotifications` from the `DatabaseConfiguration` was moved to the Notifications target. If all database settings will be moved to a different target then this target also needs to be imported to Notifications.

### Related PRs
- Based on #82

### Testing
The tests in the `NotificationCenter` and `Jobs` were refactored and new test cases were added.

### Reviewer Nudging
It would be good to check if no underlying assumptions were affected by this restructuring.


Sources/Jobs/Scheduler.swift
```diff
@@ -103,7 +106,7 @@ private extension Scheduler {
                                                        _ keyPath: KeyPath<K, T>) throws -> JobConfiguration {
         let identifier = ObjectIdentifier(keyPath)
         let jobConfiguration = try JobConfiguration(SwifCron(cronTrigger))
-        EnvironmentValues.shared.values[identifier] = job
+        _ = EnvironmentValue(keyPath, job)
```
tgymnich: would this work?
```suggestion
        EnvironmentValue(keyPath, job)
```

Sources/Notifications/DatabaseConfiguration+Notifications.swift
```diff
@@ -0,0 +1,11 @@
+import Apodini
+
+extension DatabaseConfiguration {
+    /// Adds a database migration which is used by the `NotificationCenter`.
+    /// 
+    /// This will add the models: `DeviceDatabaseModel`, `DeviceTopic`, and `Topic` to the database.
+    public func addNotifications() -> Self {
+        _ = self.addMigrations(DeviceMigration())
```
tgymnich: maybe we can add `@discarableResult` to `addMigrations(_:)` to avoid `_ =`.

tgymnich (commented):
Have you considered just leaving all the tests in the ApodiniTests Target?

Sources/Jobs/Scheduler.swift
awoc: I made EnvironmentValue `@discardableResult`

Sources/Notifications/DatabaseConfiguration+Notifications.swift
awoc: I agree with you. Just returning the result from `addMigrations(_:) ` also works

awoc:
> Have you considered just leaving all the tests in the ApodiniTests Target?

I thought it was better to also create different `testTargets` when separating `Notifications` and `Jobs` from the core Apodini target.

tgymnich (approved):

lschlesinger (approved):
I think it is a useful restructuring! Maybe it's worth mentioning the implication, i.e. tests classes that use shared functionality such as "XCTAssertRuntimeFailure" would have to import `XCTApodini`, in the meeting on Tuesday for everyone that did not see this PR and the related changes.

Issue #109, opened 2021-01-08
[INTERNAL CHANGE] Align data structure for node / tree relationship in shared model


lschlesinger:
**Is your change related to a problem? Please describe.**
Currently, we have two data structures modeling node types coexisting in the same target (`Model` in `Shared Model Builder`): 
- [`Tree.swift`](https://github.com/Apodini/Apodini/blob/develop/Sources/Apodini/Semantic%20Model%20Builder/Model/TypeReflection/Tree.swift) (used for recursive type reflection in `ProtobufferInterfaceExporter` and `OpenAPIInterfaceExporter`) and
- `EndpointsTreeNode` in [`Endpoint.swift`](https://github.com/Apodini/Apodini/blob/develop/Sources/Apodini/Semantic%20Model%20Builder/Model/Endpoint.swift) to particularly model nested endpoints with all the associated information associated to each `Endpoint`.

**Propose a solution**
As @nityanandaz  pointed out, it would be desirable to refactor `EndpointsTreeNode` to the more general representation of `Node`s in a `Tree` by using the available data structure to build `Node<AnyEndpoint>`s.

PR #110, opened 2021-01-08
Preparatory refactoring: Endpoint(sTreeNode|Parameter)

nityanandaz:
# Preparatory refactoring: Endpoint(sTreeNode|Parameter)

### `Endpoint` and `EndpointsTreeNode` are tightly coupled

However, one is a (composite) data type and the other merely a data structure.
I suggest to reuse/merge the different `Tree` implementations in a follow-up PR for https://github.com/Apodini/Apodini/issues/109

### `Sources/Apodini/Semantic Model Builder/Model/EndpointParameter.swift` is growing

I simply suggest to split up the current file and move Types and Extensions into separate files, that need not to be in the same file.
It should ease navigating the project.

## :bulb: Proposed solution

Decouple `Endpoint` and `EndpointsTreeNode`.

### Problem that is solved

Tight coupling.

### Implications

No regressions shall be introduced.

### Reviewer Nudging

It's most likely easier to observe the newest state instead of the changes.

### Additional Information

- https://en.wikipedia.org/wiki/Data_type
- https://martinfowler.com/articles/preparatory-refactoring-example.html


Sources/Apodini/Semantic Model Builder/Model/EncodeOptionalEndpointParameter.swift
```diff
@@ -0,0 +1,36 @@
...
+    func createParameterWithWrappedType(
+        name: String,
+        label: String,
+        necessity: EndpointParameterNecessity
+    ) -> AnyEndpointParameter
+}
+
+// MARK: - Parameter+EncodeOptionalEndpointParameter
+
+extension Parameter: EncodeOptionalEndpointParameter where Element: ApodiniOptional, Element.Member: Codable {
```
bauer-andreas, line 18: The `EncodeOptionalEndpointParameter` and its extension to `Parameter` are really only a hacky way to get to the Wrapped type, if the generic type of `Parameter` is a `Optional`. It is only used during Model Building and I don't consider that a "project internal API".

I would suggest to make it private, so it isn't visible to the other files in the project, which would require it being move to the ModeBuilder itself. Similar to the refactoring I have done in one of my latest PRs: https://github.com/Apodini/Apodini/blob/806528af4cc6d192e5ab4f527389d221d39bcc12/Sources/Apodini/Semantic%20Model%20Builder/ParameterModelBuilder.swift#L84

And the PathModelBuilder equivalent: https://github.com/Apodini/Apodini/blob/806528af4cc6d192e5ab4f527389d221d39bcc12/Sources/Apodini/Semantic%20Model%20Builder/PathModelBuilder.swift#L81

Sources/Apodini/Semantic Model Builder/Model/Endpoint.swift
```diff
@@ -65,12 +59,8 @@ struct Endpoint<H: Handler>: AnyEndpoint {
...
     var parameters: [AnyEndpointParameter]
     
-    var absolutePath: [_PathComponent] {
-        treeNode.absolutePath
-    }
-    var relationships: [EndpointRelationship] {
-        treeNode.relationships
-    }
+    fileprivate(set) var absolutePath: [_PathComponent] = []
+    fileprivate(set) var relationships: [EndpointRelationship] = []
```
bauer-andreas, line 63: Something discussed here: https://github.com/Apodini/Apodini/pull/104#discussion_r553823242

`absolutePath` and `relationships` isn't valid until after the Endpoint got placed into the tree. As illustrated in the linked comment I would love to make that behavior explicit to catch any invalid accesses (which should never happen) immediately.

In the linked PR I added a `fileprivate mutating func onInserted(at:)` to the Endpoint, which handles any necessary initialization once inserted into the tree. Maybe we can coordinate those PRs so something like this is introduced. Would remove the init steps from `addEndpoint` into the `Endpoint` itself, redoing clutter and giving more flexibility I think.

Sources/Apodini/Semantic Model Builder/Model/Endpoint.swift
```diff
@@ -137,19 +127,35 @@ class EndpointsTreeNode {
...
+        
+        return parent.absolutePath + [path]
     }()
     
     lazy var relationships: [EndpointRelationship] = {
-        var relationships: [EndpointRelationship] = []
-        
-        for (name, child) in nodeChildren {
-            child.collectRelationships(name: name, &relationships)
+        func collectRelationships(_ node: EndpointsTreeNode, name: String) -> [EndpointRelationship] {
```
bauer-andreas, line 138: Any reason for nesting the function into a lazy property? Feels a bit unclean to me.

Sources/Apodini/Semantic Model Builder/Model/Endpoint.swift
```diff
@@ -137,19 +127,35 @@ class EndpointsTreeNode {
...
+                // or just a regular path component. To be adapted.
+                let name = name + (child.path is RequestInjectable ? "" : "_" + childName)
+                relationships += collectRelationships(child, name: name)
+            }
+            
+            return relationships
         }
         
-        return relationships
+        return nodeChildren.reduce([]) { result, item in
```
bauer-andreas: Swiftlint warning here, wants to replace it with `reduce(:into)` which makes sense(?)

Sources/Apodini/Semantic Model Builder/Model/Endpoint.swift
```diff
@@ -158,104 +164,50 @@ class EndpointsTreeNode {
     }
     
     func addEndpoint<H: Handler>(_ endpoint: inout Endpoint<H>, at paths: [PathComponent]) {
-        if paths.isEmpty {
+        guard !paths.isEmpty else {
             // swiftlint:disable:next force_unwrapping
```
bauer-andreas: Misplaced SwiftLint comment, previously addressing the force_unwrap in the precondition error message

Sources/Apodini/Semantic Model Builder/Model/Endpoint.swift
```diff
@@ -269,21 +221,21 @@ private struct AnyHashableEndpoint: Hashable, Equatable {
     }
 }
 
-
 extension EndpointsTreeNode {
```
bauer-andreas, line 223: Would it make sense to move this extension (which is currently only used in Tests and which use case I'm not entirely sure of) into its own file as well?

Sources/Apodini/Semantic Model Builder/SharedSemanticModelBuilder.swift
```diff
@@ -97,15 +98,16 @@ class SharedSemanticModelBuilder: SemanticModelBuilder, InterfaceExporterVisitor
...
     }
     
     override func finishedRegistration() {
         super.finishedRegistration()
         
         webService.finishedParsing = true
         
-        webService.root.printTree() // currently only for debugging purposes
+        // currently only for debugging purposes
+        print(webService.root.debugDescription)
```
bauer-andreas: Don't forget to remove debugging `prints` before this PR gets merged 🙃

Sources/Apodini/Semantic Model Builder/Model/Tree.swift
```diff
@@ -134,3 +134,46 @@ extension Node {
...
+        reduce(Set()) { (partialResults, next) in
+            var set = Set(arrayLiteral: next)
+            for result in partialResults {
+                set.formUnion(result)
+            }
+            return set
+        }
+    }
+    
+    func collect2<U>() -> Set<U> where T == Array<U>, U: Hashable {
```
bauer-andreas: Can we maybe think of a better name than appending `2` as a suffix 🙃

bauer-andreas (changes requested):
Thanks for the PR. I think this change was way overdue 👍

Noticed incorrect header comment with "File.swift", I would either remove those altogether or adjust them to reflect the correct filename.
There are still some SwiftLint warnings in the code (which the Action runner seemingly ignored). Hope those get fixed soon 👍 

Did I understand you correctly, that the longterm plan is to base the `EndpointsTree` structure on the `Tree` and `Node` structures, right? Looking at the requirements of the `EndointsTreeNode` I would assume `Node` would need to be converted to a class to allow proper inheritance, right? 

We might want to coordinate merging of this PR and #104

Sources/Apodini/Semantic Model Builder/Model/Endpoint.swift
```diff
@@ -158,104 +164,50 @@ class EndpointsTreeNode {
...
-    func printTree(indent: Int = 0) {
-        let indentString = String(repeating: "  ", count: indent)
-        
-        print(indentString + path.description + "/ {")
+        var pathComponents = paths
         
-        for (operation, endpoint) in endpoints {
-            print("\(indentString)  - \(operation): \(endpoint.description) [\(endpoint.identifier.rawValue)]")
+        guard let first = pathComponents.removeFirst() as? _PathComponent  else {
+            fatalError("Encountered PathComponent which isn't a _PathComponent!")
```
lschlesinger, line 182: I tried to think about covering this with a test case, my feeling is, that if possible at all, this would be a very artificial test case - how should this happen?

Sources/Apodini/Semantic Model Builder/Model/Endpoint.swift
```diff
@@ -8,9 +8,6 @@ protocol AnyEndpoint: CustomStringConvertible {
     /// An identifier which uniquely identifies this endpoint (via its handler)
     /// across multiple compilations and executions of the web service.
     var identifier: AnyHandlerIdentifier { get }
-    
-    /// Description of the `Handler` this endpoint was generated for
-    var description: String { get }
```
lschlesinger, line 13: why is this removed here?

Sources/Apodini/Semantic Model Builder/Model/Tree.swift
nityanandaz: Yes, I renamed them to `Tree.collectValues`. The `2` would've indicated that we are dealing with a `Collection` inside the Tree "Collection".

Sources/Apodini/Semantic Model Builder/SharedSemanticModelBuilder.swift
nityanandaz: You are right 🙂

Sources/Apodini/Semantic Model Builder/Model/Endpoint.swift
nityanandaz: I somehow prefer the more strict immutable version, but I get it: it can be more efficient 😄

Sources/Apodini/Semantic Model Builder/Model/Endpoint.swift
nityanandaz: `AnyEndpoint.description` is required by the `CustomStringConvertible` inheritance. I don't see why we would need to repeat that.
If `description` is supposed to have a different meaning, then I would remove the inheritance to avoid differing expectations.

lschlesinger: yep...totally right, thanks for clarifying!

bauer-andreas:
Some status updates from my side. You said in the standup that this PR isn't a huge focus for you guys currently.
As I'm currently working on #147 the `Endpoint.swift` file gets bigger and bigger, and thus I will incorporate some of the refactoring into that said PR (like splitting up Endpoint and the tree itself, and some other things too).
I'll just wanted to mention that, so we are not both working on the same thing. Otherwise this would create way too much overhead resolving all the conflicts.

nityanandaz:
@Supereg Thanks for the heads up. I will gradually introduce some the ideas that I implemented here, when I see fit. 🚀

PR #112, opened 2021-01-08
[POC] EncoderConfiguration and visions for a potential InterfaceExporterConfiguration

bauer-andreas:
# Implement a draft version of the JSONEncoderConfiguration

Currently Exporters like the `RESTInterfaceExporter` aren't configurable, specifically in the way they encode the response.

## :bulb: Proposed solution

We definitely need some sort of `RESTConfiguration` (InterfaceExporter specific Configurations in general) sometime in the future. Especially to **change** the used encoder to e.g. something XML based.

However this Draft PR explores a solution where the user can configure any `JSONEncoder` instances used in the whole Apodini project, by using KeyPaths and providing a value for it.
As `JSONEncoder` is declared as `open`, it can be overwritten, thus the configuration also supports providing custom implementation of `JSONEncoder`.

```swift
var configuration: some Configuration {
    JSONEncoderConfiguration()
        .with(\.outputFormatting, value: [.prettyPrinted, .withoutEscapingSlashes])
}
```

I could maybe even imagine doing a generic `EncoderConfiguration` (and `DecoderConfiguration` accordingly) and adding a default initializer as it is right now. Just weren't able to do this right away as `JSONEncoder` seemingly confirms to `TopLevelEncoder` which is part of the `Combine` framework (? idk). Probably one of the reason why `Vapor` has its own `ContentEncoder` protocol.

In addition one could even write a completely abstracted version, for any kind of KeyPath Configurable objects (useful for at least a `DecoderConfiguration`).

### Problem that is solved

User can configure JSONEncoder instances used in the project.

### Implications

Consider this Draft as a Proof of Concept. It isn't ready to be used, as `Configuration` currently doesn't allow for storing such configurations in a `Exporter` accessible way. I assume we would need to wait for #82 to support that.

For the long term, where we have Exporter specific `Configuration`s I could imagine, that e.g. the user could define a `RESTConfiguration`. Within this `RESTConfiguration` the user can place a `EncoderConfiguration` (and `DecoderConfiguration`). Such a definition would then overwrite a global definition of a `EncoderConfiguration` instance.

So maybe its worth working on some infrastructure code to support `InterfaceExporterConfiguration`s (@hendesi)?

### DIFF
--- a/Sources/Apodini/Configurations/JSONEncoderConfiguration.swift
+++ b/Sources/Apodini/Configurations/JSONEncoderConfiguration.swift
@@ -1,0 +1,74 @@
+//
+// Created by Andi on 08.01.21.
+//
+
+import Foundation
+import class Vapor.Application
+
+/// This configuration provides a way to alter the configuration of `JSONEncoder` instances used in Apodini.
+/// As `JSONEncoder` is declared `open`, it also allows for supplying your own subclass.
+///
+/// A JSONEncoderConfiguration definition for the standard `JSONEncoder` could look like the following:
+/// ```swift
+/// @ConfigurationBuilder
+/// var configuration: some Configuration {
+///     JSONEncoderConfiguration()
+///         .with(\.outputFormatting, value: [.withoutEscapingSlashes, .prettyPrinted])
+///         .with(\.dateEncodingStrategy, value: .secondsSince1970)
+/// }
+/// ```
+public struct JSONEncoderConfiguration<Encoder: JSONEncoder>: Configuration {
+    private let initializer: () -> Encoder
+    private var configurations: [AnyKeyPathConfiguration<Encoder>] = []
+
+    /// Creates a new `JSONEncoderConfiguration` instance with a custom `JSONEncoder` implementation.
+    public init(_ initializer: @escaping @autoclosure () -> Encoder) {
+        self.initializer = initializer
+    }
+
+    public func with<Value>(_ property: WritableKeyPath<Encoder, Value>, value: Value) -> Self {
+        var configuration = self
+        configuration.configurations.append(KeyPathConfiguration(property, value: value))
+        return configuration
+    }
+
+    public func configure(_ app: Application) {
+        // TODO somehow store this configuration so Exporters can access it
+    }
+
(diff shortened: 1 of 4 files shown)


PSchmiedmayer (commented):
@Supereg This is a great idea!! 
This could showcase the benefits of the `Configuration` DSL and the approach described in your comment above about overriding e.g. an `EncoderConfiguration ` with a `RESTConfiguration` specific `EncoderConfiguration`.

I think that merging #82 is necessary to get some flexibility in the `Configuration` area. Getting input from @tgymnich,  @hendesi, and @awoc would be great. I think this use case would be an other reason to have a builder based configuration that creates an `Vapor.Application` and makes it available to `InterfaceExporters` [as described in the review in #82](https://github.com/Apodini/Apodini/pull/82#pullrequestreview-561858174).

hendesi (commented):
As already said in rocket chat, I really like the idea to offer the user some configuration for the exporter. 👍 This is definitely currently missing in Apodini. I could imagine that other exporters could also make use of that. 
As we are coming close to the end of the semester, I think we should try and go for the some infrastructure code to support `InterfaceExporterConfiguration`. This would provide a setup for the future work on Apodini and also allow us to tackle some general configuration, like port/ hostname. 

If you decide to go for it, I would be happy to help you on that. 👍

bauer-andreas:
> As already said in rocket chat, I really like the idea to offer the user some configuration for the exporter. 👍 This is definitely currently missing in Apodini. I could imagine that other exporters could also make use of that.
> As we are coming close to the end of the semester, I think we should try and go for the some infrastructure code to support `InterfaceExporterConfiguration`. This would provide a setup for the future work on Apodini and also allow us to tackle some general configuration, like port/ hostname.
> 
> If you decide to go for it, I would be happy to help you on that. 👍

@hendesi Thanks. My current aim would be to wait for #82 to get merged, as it is pretty much required for us to be able to store and deliver configurations to the exporters (to my understanding).
The feature isn't on my top priority list (or rather I have a bunch of stuff on my list I also want to tackle 😅), so we will have to see in what detail we can tackle this 👍

PR #113, opened 2021-01-08
Add client streaming handler to GRPC exporter

moritzschuell:
# Add client streaming handler to GRPC exporter

Currently, the GRPC exporter only supports the [Request-Response](https://github.com/Apodini/Apodini/blob/develop/Documentation/Communicational%20Patterns/3.%20Pattern%20Implementation/3.2.%20Request-Response.md) communicational pattern (or "unary" how it is called in GRPC terminology).
GRPC in general does support all 4 communicational patterns though.

## :bulb: Proposed solution
This PR adds support for the [client-side streaming](https://github.com/Apodini/Apodini/blob/develop/Documentation/Communicational%20Patterns/3.%20Pattern%20Implementation/3.3.%20Client-Side%20Stream.md) communicational pattern.

#### `GRPCMessage`
I had to switch from using the `Vapor.Request` as the `ExporterRequest` of the `GRPCInterfaceExporter` to a new request-type called `GRPCMessage`. 
Up until now (i.e. in the unary / request-response case), each HTTP request would only deliver one GRPC message to the service. 
However, with the introduction of the client-streaming handler this no longer holds true. One HTTP request can now deliver multiple HTTP frames, each potentially holding multiple GRPC messages. Further, one GRPC message can span multiple HTTP frames.

With this PR, the client-streaming handler can read multiple `GRPCMessage`s from a received HTTP frame. However, `GRPCMessage`s spanning multiple HTTP frames are not supported by this implementation yet. 
This is still an open point, see issue #127.

#### `ServiceTypeModifier`
Further, this PR also adds the `ServiceTypeModifier` for `Handlers`. It allows the user to explicitly set which type of communicational pattern should be exported for a handler:
```swift
TraditionalGreeter().serviceType(.clientStreaming)
```
The service-type was also added to the `Endpoint` protocol.
Default is always set to `.unary`.
The `ServiceType` enum offers one case for each communicational pattern:
```swift
public enum ServiceType {
    case unary
    case clientStreaming
    case serviceStreaming
    case bidirectional
}
```

For the GRPC exporter, currently there is no strategy implemented for automatically deriving the communicational pattern of a handler. Thus, it relies on the user to set the `.serviceType(..)` on each handler that should not be exported as a `.unary` endpoint.

### Problem that is solved
✅ The user can explicitly specify which communicational pattern should be exported for a `Handler`
✅ The GRPC exporter now also supports client-side streaming

### Implications
None really, for anything not related to GRPC.

As a result of the `GRPCInterfaceExporter` now using `GRPCMessage`, also the unary GRPC handler had to switch from `Vapor.Request` to `GRPCMessage`.

### Related PRs
No related PRs.

### Testing
The PR adds test cases for the two supported cases of mappings between GRPC messages and HTTP frames:
- one GRPC message in one HTTP frame
- multiple GRPC messages in one HTTP frame

### Reviewer Nudging
I think it is best to start with looking at the changes to the unary GRPC handler. Look at how it now uses the `GRPCMessage`, and how this message is parsed from the payload of the `Vapor.Request`. Then, the new client-streaming handler should be pretty easy to understand. It is essentially the same, just with support for multiple HTTP frames and multiple `GRPCMessages` in one frame.


Sources/Apodini/Semantic Model Builder/gRPC/GRPCMessage.swift
```diff
@@ -0,0 +1,40 @@
+//
+//  GRPCMessage.swift
+//  
+//
+//  Created by Moritz Schüll on 07.01.21.
+//
+
+import Foundation
+@_implementationOnly import Vapor
```
tgymnich: do you need vapor here?

Sources/Apodini/Semantic Model Builder/gRPC/GRPCService/GRPCService.swift
```diff
@@ -23,6 +24,77 @@ class GRPCService {
...
+    /// To be more specific, messages are layered on top of data frames.
+    /// A data frame may have many gRPC messages,
+    /// or if a gRPC message is quite large it might span multiple data frames."
+    ///
+    /// - Parameter data: The data to read the GRPC  messages from
+    internal func getMessages(from data: Data) -> [GRPCMessage] {
+        var data = data
+        var messages: [GRPCMessage] = []
+        var hasNext = false
+        repeat {
```
tgymnich: Is it maybe possible to express this loop in terms of `data.count`?

Sources/Apodini/Modifier/ServiceTypeModifier.swift
```diff
@@ -0,0 +1,53 @@
...
+        value = nextValue()
+    }
+}
+
+public struct ServiceTypeModifier<H: Handler>: HandlerModifier {
+    public let component: H
+    let serviceType: ServiceType
+
+    init(_ component: H, serviceType: ServiceType) {
+        self.component = component
```
lschlesinger, line 36: At some point we could switch to consistently using `handler` instead of `component` as variable name?

Sources/Apodini/Semantic Model Builder/Model/Endpoint.swift
```diff
@@ -20,6 +20,10 @@ protocol AnyEndpoint: CustomStringConvertible {
 
     var operation: Operation { get }
 
+    /// The communication pattern that is expressed
+    /// by this endpoint.
+    var serviceType: ServiceType { get }
```
lschlesinger, line 24: Do you think this should be part of the "shared" endpoint model? Are there use cases for other exporters or is this only related to gRPC? Btw this could also apply to `ServiceTypeModifier`, `GRPCMethodModifier` and `GRPCServiceModifier`: maybe I'm missing something, but it seems as if these could be moved into `Semantic Model Builder/gRPC/...`?

Sources/Apodini/Semantic Model Builder/SharedSemanticModelBuilder.swift
```diff
@@ -66,6 +66,7 @@ class SharedSemanticModelBuilder: SemanticModelBuilder, InterfaceExporterVisitor
         super.register(handler: handler, withContext: context)
         
         let operation = context.get(valueFor: OperationContextKey.self)
+        let serviceType = context.get(valueFor: ServiceTypeContextKey.self)
```
lschlesinger, line 69: see above, I myself also did not include the `customDescription` set via `DescriptionModifier` in `Endpoint` but rather accessed the context in a [function](https://github.com/Apodini/Apodini/pull/117/files#diff-b0f79c59dcf157375189003b00adde1270c0c73de3ec8906ecf8d2cb850c83f9R56) of the `OpenAPIPathBuilder` object, where I needed it.

lschlesinger: maybe we should think about moving `serviceType`, `serviceName` and `methodName` (in `Endpoint+GRPC`) also out of `Endpoint` as this - from my understanding - is only supposed reflect "shared" properties of endpoints.

Sources/Apodini/Semantic Model Builder/gRPC/GRPCInterfaceExporter.swift
```diff
@@ -36,105 +36,91 @@ class GRPCInterfaceExporter: InterfaceExporter {
...
+
+        do {
+            // we need to wrap the type into a struct to
+            // actually have a messsage.
+            let wrappedType = RequestWrapper<Type>.self
+            // set the fieldNumber to the one annotated at the
+            // parameter, or use default interference if none is
+            // annotated at the parameter.
+            FieldNumber.setFieldNumber(fieldTag)
+            let wrappedDecoded = try ProtoDecoder().decode(wrappedType, from: request.data)
```
lschlesinger: nit: when are you using `ProtoXYZ` and when `ProtobufferXYZ`?

Sources/Apodini/Semantic Model Builder/gRPC/GRPCMessage.swift
```diff
@@ -0,0 +1,40 @@
...
+    init(from data: Data, length: Int, compressed: Bool) {
+        self.data = data
+        self.length = length
+        self.compressed = compressed
+    }
+
+    /// TRUE if all fragments for this message have been collected.
+    /// FALSE othwise.
+    var isComplete: Bool {
+        if data.count < length {
```
lschlesinger: why not simply return `!(data.count < length)`

Sources/Apodini/Semantic Model Builder/gRPC/GRPCService/GRPCService.swift
```diff
@@ -23,6 +24,77 @@ class GRPCService {
         self.serviceName = name
         self.app = app.vapor.app
     }
+
+    internal func checkContentType(request: Vapor.Request) -> Bool {
+        let contentType = request.headers.first(name: .contentType)
+        switch contentType {
+        case "application/grpc", "application/grpc+proto":
```
lschlesinger: maybe we could rather create an enum for content types.

Sources/Apodini/Semantic Model Builder/gRPC/GRPCService/GRPCServiceClientStreaming.swift
```diff
@@ -0,0 +1,98 @@
...
+
+                    // retrieve all GRPC messages that were delivered in this request
+                    // (may be none, one or multiple)
+                    var messages = self.getMessages(from: data)
+                    // retain the last message, to run it through the handler
+                    // once the .end message was received.
+                    lastMessage = messages.popLast()
+
+                    messages
+                        // For now we only support messages delivered in one
```
lschlesinger: shouldn't it be: "we only support **one** message delivered in one frame, and multiple messages delivered in one frame."?

Sources/Apodini/Semantic Model Builder/gRPC/GRPCService/GRPCServiceClientStreaming.swift
```diff
@@ -0,0 +1,98 @@
...
+                        // frame, and multiple messages delivered in one frame.
+                        // One message delivered in multiple messages is not yet supported.
```
lschlesinger: and here: "One message delivered in multiple frames..."?

lschlesinger (changes requested):
Looks good :) Just had a few comments/questions...

Sources/Apodini/Semantic Model Builder/gRPC/GRPCService/GRPCService.swift
tgymnich: this is how vapor handles it: https://github.com/vapor/vapor/blob/master/Sources/Vapor/HTTP/Headers/HTTPMediaType.swift

Sources/Apodini/Semantic Model Builder/gRPC/GRPCMessage.swift
moritzschuell: Nope 😉 Good catch, thanks!

Sources/Apodini/Semantic Model Builder/gRPC/GRPCInterfaceExporter.swift
```diff
@@ -36,105 +36,91 @@ class GRPCInterfaceExporter: InterfaceExporter {
...
+
+        do {
+            if endpoint.serviceType == .unary {
+                try service.exposeUnaryEndpoint(name: methodName, context: context)
+                app.logger.info("Exported unary gRPC endpoint \(serviceName)/\(methodName)")
+            } else if endpoint.serviceType == .clientStreaming {
+                try service.exposeClientStreamingEndpoint(name: methodName, context: context)
+                app.logger.info("Exported client-streaming gRPC endpoint \(serviceName)/\(methodName)")
+            } else {
+                app.logger.warning("""
```
lschlesinger, line 62: "#warnings  should be moved to more descriptive and trackable GitHub issues that can be addressed step by step" 😄 this could be moved to an issue, couldn't it?

Sources/Apodini/Modifier/ServiceTypeModifier.swift
moritzschuell: Agree 👍 
But currently `HandlerModifier` requires the property to be called `component`.
I don't think this PR is the right place to refactor all existing modifiers.

lschlesinger: I see, let's keep this in mind and maybe create an issue for a refactoring PR.

Sources/Apodini/Semantic Model Builder/Model/Endpoint.swift
moritzschuell: I think the `serviceType` is relevant for any (also future) exporter that is able to export more than one communicational pattern. So, in my opinion this is worth to be placed in the shared endpoint model.

I agree that `GRPCMethodModifier` & `GRPCServiceModifier` should be moved to `Semantic Model Builder/gRPC/...`

Sources/Apodini/Semantic Model Builder/gRPC/GRPCInterfaceExporter.swift
moritzschuell: Thanks for pointing this out. @nityanandaz and I actually agreed to always use `Protobuffer` instead of abbreviations, so this has to be changed.

Sources/Apodini/Semantic Model Builder/gRPC/GRPCService/GRPCService.swift
moritzschuell: Thanks for the link @tgymnich 🙌 I created a `grpc` media type constant and replaced the string comparisons with it.

Sources/Apodini/Semantic Model Builder/gRPC/GRPCInterfaceExporter.swift
moritzschuell: You're right. I opened issue #142 and put a reference in the code.

However, I would still like to keep the warning-log for the user's information (as it's not a `#warning` but merely a log for the Apodini-user to understand what's happening).

Sources/Apodini/Semantic Model Builder/SharedSemanticModelBuilder.swift
moritzschuell: As said above, I think `serviceType` is actually an addition that could be of use for other exporters as well.
I agree for `serviceName` and `methodName`, and moved them out of the `Endpoint` extension.

nityanandaz: Good stuff 👍

Sources/Apodini/Semantic Model Builder/gRPC/GRPCInterfaceExporter.swift
nityanandaz: Thanks for creating the issue

Sources/Apodini/Semantic Model Builder/gRPC/GRPCMessage.swift
```diff
@@ -0,0 +1,35 @@
...
+    init(from data: Data, length: Int, compressed: Bool) {
+        self.data = data
+        self.length = length
+        self.compressed = compressed
+    }
+
+    /// TRUE if all fragments for this message have been collected.
+    /// FALSE othwise.
+    var isComplete: Bool {
+        !(data.count < length)
```
nityanandaz:
```suggestion
        data.count >= length
```

And I'd remove the comment or find a more specific name, e.g., `didCollectAllFragments`

Sources/Apodini/Semantic Model Builder/gRPC/GRPCService/GRPCServiceClientStreaming.swift
```diff
@@ -0,0 +1,99 @@
...
+                    // once the .end message was received.
+                    lastMessage = messages.popLast()
+
+                    messages
+                        // For now we only support
+                        // - one message delivered in one frame,
+                        // - multiple messages delivered in one frame.
+                        // One message delivered in multiple frames is not yet supported.
+                        // See `getMessages` internal comments for more details.
+                        .filter({ $0.isComplete })
```
nityanandaz:
```suggestion
                        .filter(\.isComplete)
```

Swift supports [keypath expressions as functions](https://github.com/apple/swift-evolution/blob/53c2e80bdede03b99aee683fe6399b6e4cf0bf95/proposals/0249-key-path-literal-function-expressions.md)

nityanandaz (commented):
Thanks for renaming many types to use _Protobuffer_

Sources/Apodini/Semantic Model Builder/gRPC/GRPCService/GRPCService.swift
moritzschuell: Yes it is 😉

Sources/Apodini/Modifier/ServiceTypeModifier.swift
```diff
@@ -0,0 +1,53 @@
...
+//  ServiceTypeModifier.swift
+//  
+//
+//  Created by Moritz Schüll on 07.01.21.
+//
+
+import Foundation
+
+/// Used to explicitly define the communicacational
+/// pattern thta is expressed by a `Handler`.
```
nityanandaz:
```suggestion
/// pattern that is expressed by a `Handler`.
```

Sources/Apodini/Modifier/ServiceTypeModifier.swift
```diff
@@ -0,0 +1,53 @@
+//
+//  ServiceTypeModifier.swift
+//  
+//
+//  Created by Moritz Schüll on 07.01.21.
+//
+
+import Foundation
+
+/// Used to explicitly define the communicacational
```
nityanandaz:
```suggestion
/// Used to explicitly define the communicational
```

Sources/Apodini/Semantic Model Builder/Model/Endpoint.swift
```diff
@@ -20,6 +20,10 @@ protocol AnyEndpoint: CustomStringConvertible {
 
     var operation: Operation { get }
 
+    /// The communication pattern that is expressed
```
nityanandaz: What is your rule of thumb for the length of comments? Personally, I have a guide at 100. So, ~50 (with indentation) seems narrow 😅😊

Sources/Apodini/Semantic Model Builder/gRPC/GRPCInterfaceExporter.swift
```diff
@@ -36,105 +36,94 @@ class GRPCInterfaceExporter: InterfaceExporter {
...
             }
 
-        let context = endpoint.createConnectionContext(for: self)
-
         // expose the new component via a GRPCService
         // currently unary enpoints are considered here
-        if let service = services[serviceName] {
-            service.exposeUnaryEndpoint(name: methodName, context: context)
+        let service: GRPCService
+        if let srvc = services[serviceName] {
```
nityanandaz: How about `existingService`?

Sources/Apodini/Semantic Model Builder/gRPC/GRPCInterfaceExporter.swift
```diff
@@ -36,105 +36,94 @@ class GRPCInterfaceExporter: InterfaceExporter {
...
+        }
 
-extension GRPCInterfaceExporter {
-    /// Returns the data contained in the body of the GRPC request.
-    private func bodyData(from request: Vapor.Request) -> Data? {
-        guard let byteBuffer = request.body.data else {
-            print("Cannot read body data.")
+        if request.data.isEmpty {
+            throw GRPCError.decodingError("""
+                No body data available to decode from.
```
nityanandaz:
```suggestion
                No body data available to decode.
```

Sources/Apodini/Semantic Model Builder/gRPC/GRPCMessage.swift
```diff
@@ -0,0 +1,33 @@
...
+//  GRPCMessage.swift
+//  
+//
+//  Created by Moritz Schüll on 07.01.21.
+//
+
+import Foundation
+
+class GRPCMessage: Apodini.ExporterRequest {
+    /// Default message  that can be used to call handlers in cases
```
nityanandaz:
```suggestion
    /// Default message that can be used to call handlers in cases
```

Sources/Apodini/Semantic Model Builder/gRPC/GRPCService/GRPCService.swift
```diff
@@ -23,14 +29,77 @@ class GRPCService {
...
+    internal func checkContentType(request: Vapor.Request) -> Bool {
+        // GRPC theoretically would also allow for other
+        // types of payload formats, e.g. JSON.
+        // We do only support proto payloads at the moment
+        if request.content.contentType == Self.grpc ||
+            request.content.contentType == Self.grpcproto {
+            return true
+        }
+        return false
+    }
```
nityanandaz:
```suggestion
    internal func contentTypeIsSupported(request: Vapor.Request) -> Bool {
        // GRPC theoretically would also allow for other
        // types of payload formats, e.g. JSON.
        // We only support proto payloads at the moment
        request.content.contentType == Self.grpc || request.content.contentType == Self.grpcproto
    }
```

`contentTypeIsSupported` should read better in an if-statement, no?

nityanandaz (approved):
Apply the proposed changes where you find them appropriate 👍

lschlesinger (approved):
👍

PR #115, opened 2021-01-09
Adding support for optional ContextKeys

### LINKED ISSUE #75: [BUG] SyntaxTreeVisitor context doesn't work well with optional values
**Describe the bug**
If a `ContextKey`'s `Value` is an optional type, node context added using this `ContextKey` won't be written to the node as expected.
The issue is that, within `ContextNode.addContext`, since `getNodeOnlyContextValue` and `getNodeContextValue` now return nested Optionals, we'll always enter the first branch of the if statement (and replace the new value with the perceived already existing one), whereas normally (for a non-optional `ContextKey.Value`) we'd enter the third branch (and replace the default value with the new one).

**To Reproduce**
See example in [lukaskollmer/visitor-context-bug](https://github.com/Apodini/Apodini/blob/lukaskollmer/visitor-context-bug/Sources/TestWebService/main.swift)

**Expected behavior**
It should be possible to use optional values as AST context.

**Screenshots**

bauer-andreas:
# Adding support for optional ContextKeys

Defining `ContextKeys` with an optional type doesn't work properly (see #75).

## :bulb: Proposed solution

This PR adds a new `OptionalContextKey` protocol, which can be used for `ContextKey`s without a default value.
To reflect this behavior, a specific `Context.get(...)` was added to return an optional, to support the case where no value was provided for a given value.

### Problem that is solved

#75 

Some of the gRPC modifiers already needed that behavior and were adjusted.

I provided a default implementation for the `reduce(...)` function, which is widely used.

### Implications

Any `ContextKey` MUST NOT be defined with the associated type being a `Optional<...>`. I added documentation stating this and additionally a runtime check, however I couldn't figure out a way to move this to a compiler warning (not sure if this is possible).

I have tried ways of adding optional `ContextKeys` by defining the `Value` associated type as `Optional<...>`, however this was a bit unclean approach, as one need to still define a `defaultValue` and the signature of the `reduce(...)` function wasn't really fitting.

### Related PRs
* #117 The `DescriptionModifier` can make use of the new `OptionalContextKey` protocol.

### Testing
Test cases checking `OptionalContextKey`s were added.

### Reviewer Nudging
--

### DIFF
--- a/Sources/Apodini/SyntaxTreeVisitor/ContextKey.swift
+++ b/Sources/Apodini/SyntaxTreeVisitor/ContextKey.swift
@@ -36,15 +36,6 @@ extension OptionalContextKey {
 /// A `ContextKey` is a `OptionalContextKey` with the addition of the definition of a default value.
 /// See implications of the reduction logic `OptionalContextKey.reduce(...)`.
 protocol ContextKey: OptionalContextKey {
-    /// The type of the value the `OptionalContextKey` identifies. The value MUST NOT be of type `Optional`.
-    /// The type is equal to `OptionalContextKey.Value`
-    associatedtype DefaultValue
-
     /// The default value this `ContextKey` provides.
-    static var defaultValue: DefaultValue { get }
-}
-
-extension ContextKey {
-    // I know the compiler creates a warning about this, but using it as a type constraint isn't actually the same thing
-    typealias Value = DefaultValue
+    static var defaultValue: Self.Value { get }
 }
--- a/Sources/Apodini/SyntaxTreeVisitor/ContextKey.swift
+++ b/Sources/Apodini/SyntaxTreeVisitor/ContextKey.swift
@@ -5,11 +5,34 @@
 //  Created by Paul Schmiedmayer on 6/26/20.
 //
 
-protocol ContextKey {
+/// A `OptionalContextKey` context key TODO document
+protocol OptionalContextKey {
     associatedtype Value
-    
-    static var defaultValue: Self.Value { get }
 
-    static func reduce(value: inout Self.Value,
-                       nextValue: () -> Self.Value)
+    static func reduce(value: inout Self.Value, nextValue: () -> Self.Value)
+}
+
+extension OptionalContextKey {
+    static func reduce(value: inout Self.Value, nextValue: () -> Self.Value) {
+        value = nextValue()
+    }
+}
+
+
+// TODO document
+protocol ContextKey: OptionalContextKey {
+    associatedtype DefaultValue
+
+    static var defaultValue: DefaultValue { get }
+}
+
+extension ContextKey {
+    // I know the compiler creates a warning about this, but using it as a type constraint isn't actually the same thing
+    typealias Value = DefaultValue
+}
+
+extension ContextKey {
+    static func reduce(value: inout Self.Value, nextValue: () -> Self.Value) {
+        value = nextValue()
+    }
 }
--- a/Sources/Apodini/SyntaxTreeVisitor/ContextKey.swift
+++ b/Sources/Apodini/SyntaxTreeVisitor/ContextKey.swift
@@ -5,10 +5,24 @@
 //  Created by Paul Schmiedmayer on 6/26/20.
 //
 
-/// A `OptionalContextKey` context key TODO document
+/// A `OptionalContextKey` serves as a key definition for a `ContextNode`.
+/// Optionally it can serve a reduction logic when inserting a new value into the `ContextNode`,
+/// see `OptionalContextKey.reduce(...)`.
+/// The `OptionalContextKey` is optional in the sense that it doesn't provide a default value, meaning
+/// it may not exists on the `Context` for a given `Handler`.
 protocol OptionalContextKey {
+    /// The type of the value the `OptionalContextKey` identifies. The value MUST NOT be of type `Optional`.
     associatedtype Value
 
+    /// This function can be optionally implemented to provide a reduction logic when inserting a new
+    /// value into a `ContextNode`. By default the previous value will just be overwritten.
+    /// As `OptionalContextKey` do not provide a default value, `reduce(...)` will not be called on first
+    /// insertion of a value for this `OptionalContextKey`.
+    ///
+    /// - Parameters:
+    ///   - value: The current value of the context key.
+    ///         The result of the reduction must be written into this inout parameter.
+    ///   - nextValue: The return value of the provided closure is the newly inserted value.
     static func reduce(value: inout Self.Value, nextValue: () -> Self.Value)
 }
 
@@ -19,20 +33,18 @@ extension OptionalContextKey {
 }
 
 
-// TODO document
+/// A `ContextKey` is a `OptionalContextKey` with the addition of the definition of a default value.
+/// See implications of the reduction logic `OptionalContextKey.reduce(...)`.
 protocol ContextKey: OptionalContextKey {
+    /// The type of the value the `OptionalContextKey` identifies. The value MUST NOT be of type `Optional`.
+    /// The type is equal to `OptionalContextKey.Value`
     associatedtype DefaultValue
 
+    /// The default value this `ContextKey` provides.
     static var defaultValue: DefaultValue { get }
 }
 
 extension ContextKey {
     // I know the compiler creates a warning about this, but using it as a type constraint isn't actually the same thing
     typealias Value = DefaultValue
-}
-
-extension ContextKey {
-    static func reduce(value: inout Self.Value, nextValue: () -> Self.Value) {
-        value = nextValue()
-    }
 }
--- a/Sources/Apodini/SyntaxTreeVisitor/ContextNode.swift
+++ b/Sources/Apodini/SyntaxTreeVisitor/ContextNode.swift
@@ -50,7 +50,7 @@ class ContextNode {
     func addContext<C: OptionalContextKey>(_ contextKey: C.Type = C.self, value: C.Value, scope: Scope) {
         let newValue: C.Value
 
-        if C.Value.self is ExpressibleByNilLiteral.Type {
+        if isOptional(C.Value.self) {
             fatalError("""
                        The `Value` type of a `ContextKey` or `OptionalContextKey` must not be a `Optional` type.
                        Found \(C.Value.self) as `Value` type for key \(C.self).
--- a/Sources/Apodini/SyntaxTreeVisitor/ContextNode.swift
+++ b/Sources/Apodini/SyntaxTreeVisitor/ContextNode.swift
@@ -28,21 +28,34 @@ class ContextNode {
             ?? getGlobalContextValue(for: contextKey)
             ?? C.defaultValue
     }
+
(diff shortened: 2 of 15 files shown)


lukaskollmer:
Hmmm not really a fan of the duplication tbh. I fear that this would make it, when changing anything about this code in the future, way too easy to miss that the change in fact needs to be applied to two places.

Also not sure about the "lets just wrap everything in an optional" approach, my guess would be that the vast majority of `ContextKey`s are non-optional, so this would introduce the optional-caused additional overhead for all context values, even though only very few actually need it.

Maybe we can do a protocol-based approach to detect if a `ContextKey` has an optional value? A conditional extension on the `ContextKey` type or something like that?

Apart from all this, I think having an `OptionalContextKey` would probably be the cleanest solution. (Or would that also require duplication?)

bauer-andreas:
> Hmmm not really a fan of the duplication tbh. I fear that this would make it, when changing anything about this code in the future, way too easy to miss that the change in fact needs to be applied to two places.

Feel the same. And obviously the end goal is to remove all this code duplication.
However, before doing unnecessary work I wanted to discuss how we should implement options `ContextKey`s from an API point of view.

This PR started with just using the same `ContextKey` protocol by just declaring the associated type as an Optional, as having just one unified protocol would be a great thing I though. However this may create unnecessary complications when implementing the `reduce` function.

But yes, you are seemingly feeling the same that a dedicated protocol might yield a cleaner solution.

> Maybe we can do a protocol-based approach to detect if a ContextKey has an optional value? A conditional extension on the ContextKey type or something like that?

Detecting if the ContextKey holds an optional isn't a problem. I think you talk about a similar approach I used for extracting the type information from optional `Parameter`s right? Was also my first attempt, but was a nightmare to handle the types. 

I will got back to the drawing table and see what I can do.

bauer-andreas:
Alright, so I have worked on an approach using a separate protocol and polished that one. I have updated the original message to reflect those changes, so maybe just have a look at it.

EDIT: also added some more regression test, so anyone in the future dealing with ContextNodes should be perfectly save making changes 👍

Sources/Apodini/SyntaxTreeVisitor/ContextKey.swift
```diff
@@ -5,11 +5,46 @@
...
+    /// The type of the value the `OptionalContextKey` identifies. The value MUST NOT be of type `Optional`.
+    /// The type is equal to `OptionalContextKey.Value`
+    associatedtype DefaultValue
+
+    /// The default value this `ContextKey` provides.
+    static var defaultValue: DefaultValue { get }
+}
+
+extension ContextKey {
+    // I know the compiler creates a warning about this, but using it as a type constraint isn't actually the same thing
```
moritzschuell: Why is that the case? 🙃

Sources/Apodini/SyntaxTreeVisitor/ContextNode.swift
```diff
@@ -28,31 +29,43 @@ class ContextNode {
...
+    private func getGlobalContextValue<C: OptionalContextKey>(for contextKey: C.Type = C.self) -> C.Value? {
         getNodeContextValue(for: contextKey) ?? parentContextNode?.getGlobalContextValue(for: contextKey)
     }
     
-    func addContext<C: ContextKey>(_ contextKey: C.Type = C.self, value: C.Value, scope: Scope) {
-        var newValue: C.Value
+    func addContext<C: OptionalContextKey>(_ contextKey: C.Type = C.self, value: C.Value, scope: Scope) {
+        let newValue: C.Value
+
+        if C.Value.self is ExpressibleByNilLiteral.Type {
```
moritzschuell: I am not sure whether this check is really the best way to determine if `Value` is an optional type.
There might be cases where custom types conform to `ExpressibleByNilLiteral`.

~~I search-engined a bit, and found that [`Mirror.DisplayStyle`](https://developer.apple.com/documentation/swift/mirror/displaystyle) has an `optional` case. 
Maybe that's a nicer approach, even though still only a runtime-check~~ (doesn't seem to work)

moritzschuell (commented):
Great stuff! Thanks for this PR.
I left a comment on the runtime-check for optional types, otherwise this looks good to me.

Sources/Apodini/SyntaxTreeVisitor/ContextKey.swift
bauer-andreas: I would categorize this as "swift compiler weirdness"

If `ContextKey` was just defined like the following ...

```swift
protocol ContextKey: OptionalContextKey {
    static var defaultValue: Value { get }
}
```

... one would assume a definition like this is valid:

```swift
struct ExampleContextKey: ContextKey {
  static let defaultValue: String = "asdf"
}
```

It isn't. Although default value is defined to be type `String`, the swift compiler can't infer the type of the `associatedtype Value`  located in the parent protocol `OptionalContextKey` (I assume because the definition of the associated type isn't in the same protocol?).

My workaround then was to define the associated type `DefaultValue` and just assign it to `Value`, like dedicated in the commented code.

Like the code is right now, the above `ExampleContextKey` compiles as expected. The compiler warning wishes to refactor `ContextKey` instead to ...

```swift
protocol ContextKey: OptionalContextKey where Value == DefaultValue {
    associatedtype DefaultValue

    static var defaultValue: DefaultValue { get }
}
```

... which is seemingly not the same. First of all it is not the same as to my understanding how the `where` clause works, and indeed the `ExampleContentKey` doesn't compile anymore without explicitly adding the `typealias Value = String`.

If anyone has a better workaround for this, feel free to suggest a change 🎉

Sources/Apodini/SyntaxTreeVisitor/ContextNode.swift
bauer-andreas: > There might be cases where custom types conform to ExpressibleByNilLiteral.

Our stance is, we don't allow `nil` literal, and thus I think the check for that protocol is actually valid(?). We do not just blacklist `Optional` type, but every type which allows `nil`(?).

In any case, this is internal API, so not sure if that is that big of a problem. But I assume adding a private protocol and conformance for that protocol to `Optional` would also be valid.

Sources/Apodini/SyntaxTreeVisitor/ContextKey.swift
PSchmiedmayer: @Supereg Thanks for the detailed explanation!

As you said an alternative to the `DefaultValue` and the extension below would be to just state 
```swift
    static var defaultValue: Self.Value { get }
```
As sated this would produce no compiler warning in this section as it is defined by the explicit declaration that `Value` is the associatedtype of `OptionalContextKey `.

As you stated the downside would be that you currently have to manually define an `typealias` like this to make the Swift compiler happy as it seems like it can't solve the type constraints across this protocol relationship at the moment:
```swift
struct OperationContextKey: ContextKey {
    typealias Value = Operation
    
    static var defaultValue: Operation = .automatic
}
```

II think this manual definition of a `typealias` would be an acceptable approach for me as this approach would reduce the number of `associatedtype`s by one and once the swift compiler gets smart enough this should be easier solvable then the current implementation with a second `associatedtype`. 
The two `associatedtype`s would point to the same type which might not be immediately visible when taking a look at the type definitions without taking a look at the externs below.

In the context of steadily reducing warnings on develop I would consider this a more likable solution as `ContextKey` and `OptionalContextKey` are both internal and not part of the public API, what do you think @Supereg @moritzschuell ?

Sources/Apodini/SyntaxTreeVisitor/ContextNode.swift
PSchmiedmayer: @Supereg @moritzschuell The `Runtime` framework offers a build in check to detect an optional type, we should be able to use this here. I have experimented a bit with this and created a test case to evaluate the function and it seems to work in my limits tests:
```swift
func testIsOptional() throws {
    func isOptional<T>(_ type: T.Type = T.self) throws -> Bool {
        let typeInfo = try Runtime.typeInfo(of: type)
        return typeInfo.kind == .optional
    }
    
    /// A custom type
    struct Test {}
    
    XCTAssertEqual(try isOptional(String.self), false)
    XCTAssertEqual(try isOptional(Int.self), false)
    XCTAssertEqual(try isOptional(Test.self), false)
    XCTAssertEqual(try isOptional(Optional<Test>.self), true)
    XCTAssertEqual(try isOptional(Optional<Test>.self), true)
    XCTAssertEqual(try isOptional(String?.self), true)
    XCTAssertEqual(try isOptional(String??.self), true)
    XCTAssertEqual(try isOptional(String???.self), true)
    XCTAssertEqual(try isOptional(Never.self), false)
}
```

Therefore we could introduce an internal method similar to `assertTypeIsStruct` that can look like this: 
```swift
func isOptional<T>(_ type: T.Type = T.self) throws -> Bool {
    let typeInfo = try Runtime.typeInfo(of: type)
    return typeInfo.kind == .optional
}
```
and could be used here:
```suggestion
        if isOptional(C.Value.self) {
```

Sources/Apodini/SyntaxTreeVisitor/ContextKey.swift
moritzschuell: @Supereg @PSchmiedmayer I also don't think that having to manually define the `typealias` would be a problem. And also, it somehow feels like the less "hacky" variant to me 🤓

bauer-andreas: Feedback incorporated with ff2e55e

Sources/Apodini/SyntaxTreeVisitor/ContextNode.swift
bauer-andreas: I incorporated your proposed change @PSchmiedmayer with 44b82fb.
I just modified the isOptional func to not be throwing (with documentation why this is possible).
Basically `typeInfo` only throws if the `kind` is something it does not support (or is not a "type"). If the kind is optional it will never throw, as it will generate a enum TypeInfo. Thus if typeInfo throws, we know that it can't be an optional.
I think that makes using the function a bit easier.

I also just added the test cases you provided 👍

bauer-andreas:
Depending on what gets merged first, #117 could make use of this new `OptionalContextKey` protocol.

moritzschuell (approved):
Looks good to me 🙂 Thanks for all your detailed explanations!

Sources/Apodini/SyntaxTreeVisitor/ContextNode.swift
PSchmiedmayer: Thats a good point, making it non throwing is a good addition 👍

PSchmiedmayer (approved):
Thanks for incorporating the changes! Looks like a great addition to Apodini 🎉

bauer-andreas:
Okay what do I doo with CodeCov falling with `77.64% (-0.00%) compared to 370e249` 😅 🙃

bauer-andreas:
administrator magic rules ✨

PSchmiedmayer:
The code coverage diff was so minor, I merged the branch into develop 👍

Issue #114, opened 2021-01-09
[EXPERIMENTAL CHANGE] Change the `descpription` of an `Endpoint<H: Handler>` to "shortname" of `Handler`


lschlesinger:
**Is your change related to a problem? Please describe.**
Currently, the `description` of an `Endpoint<H: Handler>` is set to [`String(describing: Handler)`](https://github.com/Apodini/Apodini/blob/develop/Sources/Apodini/Semantic%20Model%20Builder/Model/Endpoint.swift#L89)
To my understanding, as of now, the description is mainly used to print error messages. In case of more complex `Handler`s, (e.g., including several parameters... such as the `TraditionalGreeter` in the current main.swift), this can lead to unreadable error messages.
If one would try to add a `TraditionalGreeter` more than once with the same `operation` and at the same path, this would lead to an error message, that is hard to read:
![image](https://user-images.githubusercontent.com/41366522/104092982-bc691d00-5287-11eb-925a-547e41d1e249.png)

**Propose a solution**
I suggest to reduce the respective `description` to the name of the `Handler`, e.g. in the `TraditionalGreeter` in the given example.
This would lead to clean, readable outputs such as:
![image](https://user-images.githubusercontent.com/41366522/104093015-06ea9980-5288-11eb-847f-d4377c902a05.png)

Furthermore, I could think of other use cases where the `Handler`'s name can be harnessed. For example, the description of an `OpenAPI.Operation` (i.e. and endpoint, that is defined by a unique path and HTTP operation):
<img src="https://user-images.githubusercontent.com/41366522/104093147-e838d280-5288-11eb-8685-eece01fbe827.png" width="400" >

**QUESTION: Is  the endpoint's description, I want to refactor, used anywhere else except for error descriptions and tests or planned to be used for anything in this certain form in the future that would disallow for the proposed change?**

If not, refactoring would be an easy job ;)

bauer-andreas:
Maybe we can use `CustomDebugStringConvertible` and its `debugDescription` property to return that detailed description.

PR #117, opened 2021-01-11
`DescriptionModifier` as explicit, customizable endpoint description

lschlesinger:
# DescriptionModifier

As of now, the only textual "description" of an `Endpoint<H: Handler>` was available as [`String(describing: Handler)`](https://github.com/Apodini/Apodini/blob/develop/Sources/Apodini/Semantic%20Model%20Builder/Model/Endpoint.swift#L89). In the OAS, it is possible to include expressive descriptions on the endpoint (in OAS lingo: Operation Object) level, as described [here in the OAS documentation](https://swagger.io/specification/#operation-object).

## :bulb: Proposed solution
1. We provide the possibility of an expressive endpoint description that is either explicitly set by the developer using a newly introduced `DescriptionModifier` or takes the type name of the `Handler` as default.

Example:

2. As proposed in a related Issue #114 , we could use the current detailed (more technical description) as `debugDescription`

### Problem that is solved
We enable the developer to provide expressive description of an endpoint, which is useful in terms of documentation and can be used directly in OAS. 

### Implications
1. A new `description` field in the `OpenAPI.Operation` (which is built by the `OpenAPIPathsObjectBuilder` using the function `buildPathItemOperationObject<H: Handler>(from endpoint: Endpoint<H>)-> OpenAPI.Operation`) is introduced. 

  Exemplary OAS Output:
    

 Note, default would be `UserHandler`.

2. If needed, one would have to use the `debugDescription` to access the String representation of the respective `Handler` of an `Endpoint<H: Handler>`.
I therefore had to do some minor refactoring using the description in tests (`Tests/ApodiniTests/SyntaxTreeVisitorTests/EndpointIdentifierTests.swift`)

3. This description might also be integrated in the `.proto` file (as a comment for example?)

I have also integrated the use of the modifier in the `main.swift` to make it transparent to everyone, that this feature is available.

### Related PRs
No related PR I guess.

### Testing
I've added tests cases to test for both,  explicitly setting the `description` using the modifier and setting the `Handler` type name as default in: `ApodiniTests/ModifierTests/DescriptionModifierTests)`.

### Reviewer Nudging
The tests and screenshots clearly state the use case, changes are minor and easy to understand 😄 Used the random allocation mechanism for selecting the reviewers 😉


Tests/ApodiniTests/ModifierTests/DescriptionModifierTests.swift
```diff
@@ -0,0 +1,74 @@
...
+        let modelBuilder = SharedSemanticModelBuilder(app)
+        let visitor = SyntaxTreeVisitor(semanticModelBuilders: [modelBuilder])
+        let testComponent = TestComponentDescription()
+        Group {
+            testComponent.content
+        }.accept(visitor)
+        visitor.finishParsing()
+
+        // swiftlint:disable force_unwrapping
+        let treeNodeA: EndpointsTreeNode = modelBuilder.rootNode.children.first!.children.first!
```
moritzschuell: nit: Could you use `XCTUnwrap(...)` instead of force-unwrapping & disabling the SwiftLint rule?

moritzschuell (approved):
Looks great 🙌
Only left a small question.

Tests/ApodiniTests/ModifierTests/DescriptionModifierTests.swift
lschlesinger: Of course, thanks for the reminder, I was not aware of this ;)! solved by 5b6a526

Sources/Apodini/Semantic Model Builder/Model/Endpoint.swift
```diff
@@ -121,6 +118,18 @@ struct Endpoint<H: Handler>: AnyEndpoint {
     }
 }
 
+extension Endpoint {
+    var description: String {
+        let description = self.context.get(valueFor: DescriptionContextKey.self)
+        return description == DescriptionContextKey.defaultValue ? String(describing: type(of: self.handler)): description
```
bauer-andreas: Not sure if we should return the `DescriptionContextKey` here, which can be a rather detailed description?
As description is heavily used in the project in error or debug messages, I think just returning the `H.self` type name (like the struct name of the Handler) would be better? What do you think?
I would either just at a dedicated property for the DescriptionContextKey or just rely on the Context?

lschlesinger: I am not sure if I understand your comment... the approach was to return the `H.self` type name in case there was no description provided via `DescriptionModifier`.

bauer-andreas: So the idea is that `.description` always returns the struct name (like `"H.self"`).
And the description from the `DescrptionModifier` is stored in a different property, if that makes sense?

Sources/Apodini/Modifier/DescriptionModifier.swift
```diff
@@ -0,0 +1,37 @@
+//
+// Created by Lorena Schlesinger on 10.01.21.
+//
+
+struct DescriptionContextKey: ContextKey {
```
bauer-andreas: As #115 is now merged you could move to the `OptionaContextKey` protocol, instead of using an empty string as a default value 👍

Tests/ApodiniTests/OpenAPITests/OpenAPIDocumentBuilderTests.swift
```diff
@@ -36,6 +36,7 @@ final class OpenAPIDocumentBuilderTests: XCTestCase {
             paths: [
                 "test": .init(
                     get: .init(
+                        description: endpoint.description,
```
bauer-andreas, line 39: Might be some leftovers?

bauer-andreas: Nevermind, there are no custom modifiers, so this is expected, sorry.

bauer-andreas (approved):

PR #120, opened 2021-01-11
Fix DSL context

### LINKED ISSUE #111: [BUG] SyntaxTreeVisitor produces invalid node context values for Groups containing multiple Handlers
**Describe the bug**

For the following input web service:
```swift
    struct MyHandler: Handler {
        typealias Response = Never // make it crash if accessed
    }
    
    var content: some Component {
        Group("api") {
            Group("user") {
                MyHandler().operation(.read)
                MyHandler().operation(.create)
            }
        }
    }
}
```

the exported endpoints are:
```
/ {
  user/ {
    v1/ {
      api/ {
        - read: MyHandler() [0:0:0]
        - create: MyHandler() [0:0:1]
      }
    }
  }
}
```

This seems to be caused by putting multiple Handlers with different Operations into the same group.

It doesn't seem to be related to the nesting, this example also results in wrong paths:
```swift
var content: some Component {
    Group("api") {
        MyHandler().operation(.read)
        MyHandler().operation(.create)
    }
}
```

<br>

If, however, `WebService.content` is defined like this:
```swift
var content: some Component {
    Group("api") {
        Group("user") {
            MyHandler().operation(.read)
        }
    }
    Group("api") {
        Group("user") {
            MyHandler().operation(.create)
        }
    }
}
```
we're getting the expected output:
```
/ {
  v1/ {
    api/ {
      user/ {
        - create: MyHandler() [1:0:0]
        - read: MyHandler() [0:0:0]
      }
    }
  }
}
```

**To Reproduce**
Clone [`lukaskollmer/rest-path-bug`](https://github.com/Apodini/Apodini/tree/lukaskollmer/rest-path-bug) and run the TestWebService

lukaskollmer:
Fixes #111, by rolling back some of the changes introduced in #88.

We no longer try to ignore `TupleComponent`s when entering/exiting DSL content contexts.

Pro: we now get correct paths when generating endpoints.
Con: auto-generated endpoint/handler identifiers are a bit longer.

### DIFF
--- a/Sources/Apodini/SyntaxTreeVisitor/SyntaxTreeVisitor.swift
+++ b/Sources/Apodini/SyntaxTreeVisitor/SyntaxTreeVisitor.swift
@@ -112,7 +112,7 @@ class SyntaxTreeVisitor {
     
     private func formHandlerIndexPathForCurrentNode() -> HandlerIndexPath {
         let rawValue = currentNodeIndexPath
-            .map { String($0 - 1) } // We remove one from the current indexPath to have 0 as the first index
+            .map { String(max($0, 1) - 1) } // We remove one from the current indexPath to have 0 as the first index
             .joined(separator: ":")
         return HandlerIndexPath(rawValue: rawValue)
     }
--- a/Sources/Apodini/SyntaxTreeVisitor/SyntaxTreeVisitor.swift
+++ b/Sources/Apodini/SyntaxTreeVisitor/SyntaxTreeVisitor.swift
@@ -112,7 +112,7 @@ class SyntaxTreeVisitor {
     
     private func formHandlerIndexPathForCurrentNode() -> HandlerIndexPath {
         let rawValue = currentNodeIndexPath
-            .map { String(max($0, 1) - 1) } // We remove one from the current indexPath to have 0 as the first index
+            .map { String($0 - 1) } // We remove one from the current indexPath to have 0 as the first index
             .joined(separator: ":")
         return HandlerIndexPath(rawValue: rawValue)
     }
--- a/Sources/Apodini/Components/Group.swift
+++ b/Sources/Apodini/Components/Group.swift
@@ -24,19 +24,11 @@ public struct Group<Content: Component>: Component, SyntaxTreeVisitable {
     }
     
     func accept(_ visitor: SyntaxTreeVisitor) {
-        func forwardToContent() {
-            visitor.addContext(PathComponentContextKey.self, value: pathComponents, scope: .environment)
-            content.accept(visitor)
-        }
-        
-        if !String(describing: type(of: content)).hasPrefix("TupleComponent<") {
-            visitor.enterContent {
-                visitor.enterComponentContext {
-                    forwardToContent()
-                }
+        visitor.enterContent {
+            visitor.enterComponentContext {
+                visitor.addContext(PathComponentContextKey.self, value: pathComponents, scope: .environment)
+                content.accept(visitor)
             }
-        } else {
-            forwardToContent()
         }
     }
 }
--- a/Tests/ApodiniTests/RESTInterfaceExporterTests.swift
+++ b/Tests/ApodiniTests/RESTInterfaceExporterTests.swift
@@ -134,19 +134,16 @@ class RESTInterfaceExporterTests: ApodiniTests {
     
     func testEndpointPaths() throws {
         struct WebService: Apodini.WebService {
-            struct EmptyHandler: Apodini.Handler {
-                typealias Response = Never
-            }
             var content: some Component {
                 Group("api") {
                     Group("user") {
-                        EmptyHandler().operation(.read)
-                        EmptyHandler().operation(.create)
+                        Text("").operation(.read)
+                        Text("").operation(.create)
                     }
                 }
                 Group("api") {
                     Group("post") {
-                        EmptyHandler().operation(.read)
+                        Text("").operation(.read)
                     }
                 }
             }
--- a/Tests/ApodiniTests/RESTInterfaceExporterTests.swift
+++ b/Tests/ApodiniTests/RESTInterfaceExporterTests.swift
@@ -130,4 +130,62 @@ class RESTInterfaceExporterTests: ApodiniTests {
             XCTAssertEqual(container.data.name, name)
         }
     }
+    
+    
+    func testEndpointPaths() throws {
+        struct WebService: Apodini.WebService {
+            struct EmptyHandler: Apodini.Handler {
+                typealias Response = Never
+            }
+            var content: some Component {
+                Group("api") {
+                    Group("user") {
+                        EmptyHandler().operation(.read)
+                        EmptyHandler().operation(.create)
+                    }
+                }
+                Group("api") {
+                    Group("post") {
+                        EmptyHandler().operation(.read)
+                    }
+                }
+            }
+        }
+        
+        let builder = SharedSemanticModelBuilder(app)
+            .with(exporter: RESTInterfaceExporter.self)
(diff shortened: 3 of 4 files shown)


Sources/Apodini/SyntaxTreeVisitor/SyntaxTreeVisitor.swift
```diff
@@ -112,7 +112,7 @@ class SyntaxTreeVisitor {
     
     private func formHandlerIndexPathForCurrentNode() -> HandlerIndexPath {
         let rawValue = currentNodeIndexPath
-            .map { String($0 - 1) } // We remove one from the current indexPath to have 0 as the first index
+            .map { String(max($0, 1) - 1) } // We remove one from the current indexPath to have 0 as the first index
```
bauer-andreas: Would you mind explaining why this was changed? Or why the change in Group resulted in this change to be necessary 🤔
This max operation just seems a bit odd to me as it is not injective (meaning `0` and `1` would both map to the same result). But the test are seemingly passing so might just be my missing knowledge.

bauer-andreas (commented):
Looks good to me besides that one little comment I can't wrap my head around 🙃

bauer-andreas (approved):

Sources/Apodini/SyntaxTreeVisitor/SyntaxTreeVisitor.swift
lukaskollmer: There was an issue w/ one of the tests where the handler index path would contain -1, that's why i added the `max` call back in.

It is, in fact, not needed. (And now removed)

The issue seems to be the following: In some of the tests we generate the web service model the following way:
```swift
let builder = SharedSemanticModelBuilder(app).with(exporter: RESTInterfaceExporter.self)
let visitor = SyntaxTreeVisitor(semanticModelBuilders: [builder])
TestWebService().accept(visitor)
visitor.finishParsing()
```

Line 3 here gets resolved to [`Component.accept`](https://github.com/Apodini/Apodini/blob/b5b4be3c9a0f48748b85154d196257bdd76543d6/Sources/Apodini/Components/Component.swift#L61), which is not what we want (because, unlike [`WebService.visit`](https://github.com/Apodini/Apodini/blob/b5b4be3c9a0f48748b85154d196257bdd76543d6/Sources/Apodini/WebService.swift#L83), it doesn't insert the implicit root `Group`, and also doesn't set the version context).

We should be using [`WebService.register`](https://github.com/Apodini/Apodini/blob/b5b4be3c9a0f48748b85154d196257bdd76543d6/Sources/Apodini/WebService.swift#L77), since that properly sets up the root context.

Fun fact: this probably also means that some of the other tests which still use `someWebService.accept(visitor)` rather than `WebService.register` probably test against incorrectly generated endpoints (since they all are missing the root context).

(You can see this in [this commit](https://github.com/Apodini/Apodini/pull/120/commits/ad7dc19b78b43d6b4e5fefad43f5d5e842dd8971): the test passed w/ both the old and the new version, because the .accept version is missing the root level)

And the missing root level (alongside the then also missing implicit root Group) caused the index path indices to become negative (which the max call fixed).

Interestingly enough the issue w/ negative indices could only ever have occurred in our tests, since the `WebService.accept` function is private. Also means that the max call, when reached for a user-defined web service, always was effectively a no-op since `$0` always was >= 1.

bauer-andreas: Thanks for the detailed explanation 🚀 . So the real difference is the missing `Group` Component inserted by the `WebService` by default? Ah I see, if the `TestWebService` would just be a simple `Text` component for example, the `ContextNode` wouldn't be properly prepared, as there was not call to `enterContent` right?
Would it make sense to prepare the `SyntaxTreeVisitor` in such a way, such scenarios are handled as well? Not sure if one might remove the `Group` in the `WebService` in the future? If I remember correctly there was a time where `currentNodeIndexPath` hold a `0` by default right?

PSchmiedmayer (approved):
I checked out your implementation, it solves the problem and the test cases look great!

One suggestion I would have to reduce the number of elements in the `Handler` index.
We could remove the `Group` in the visit function of the `WebService`:
```swift
private func visit(_ visitor: SyntaxTreeVisitor) {
    visitor.addContext(APIVersionContextKey.self, value: version, scope: .environment)
    visitor.addContext(PathComponentContextKey.self, value: [version], scope: .environment)
    content.accept(visitor)
}
```
I have tested this with the current version of the branch. Nevertheless we should test if this as any other implications.
Apart from the comment above I approve the PR 🚀

lukaskollmer:
@PSchmiedmayer is it possible to somehow overrule the codecov thing here? There are, in fact, more tests now than we had before.
According to the [details page](https://github.com/Apodini/Apodini/pull/120/checks?check_run_id=1687329986) coverage for Group.swift is still at 100%, same as it was before.

PSchmiedmayer:
@lukaskollmer Yes, I can overrule the coverage report. It seems like the reduction in overall line numbers reduced in an overall reduction of code coverage even though the diff coverage is at 100%. I will merge the PR.
Independent of the PR getting merged, did you get the change to take a look [at the idea in the comment above](https://github.com/Apodini/Apodini/pull/120#pullrequestreview-565815252)? Would you consider this a feasible approach with no side effects that would bring the overall depth of the identifiers down by one level?

lukaskollmer:
@PSchmiedmayer ~~I don't see any issues with removing the implicit root `Group`. (tbh i don't really know we have it currently.)~~

Although I'd guess this would probably cause issues if the root component (ie, the component returned from `WebService.content`) is something different than `Group` or `TupleComponent`. For example, if the web service consists of only a `Handler`, it'd call this handler's accept function, which assumes it already is in a component context and therefore doesn't call either of the visitor.enter functions.

So if we remove the implicit Group, we'd need to wrap the `content.accept(visitor)` call in `WebService.visit` in both `visitor.enterContent` and `visitor.enterComponentContext` (at which point we've effectively re-implemented Group's visit behaviour)

Sources/Apodini/SyntaxTreeVisitor/SyntaxTreeVisitor.swift
lukaskollmer: Yeah pretty much. We used to have `currentNodeIndexPath = [0]` as the default value, but that was removed since it wasn't really needed.

We can move some of this into the visitor, but that'd mean we have to make sure that the visitor isn't reused (which would be a good idea anyway, since it shouldn't).
Also would mean that, while we'd then support WebServices w/ top-level non Group/Tuple components (see other comment below), other WebServices with such top-level components would have too many entries in the `currentNodeIndexPath` array (since we'd have both the visitor-added root context as well as the one introduced by the root component).

PSchmiedmayer:
@lukaskollmer Sounds reasonable, then we will leave it like this ✅

Sources/Apodini/SyntaxTreeVisitor/SyntaxTreeVisitor.swift
bauer-andreas: > [...] we have to make sure that the visitor isn't reused [...]

Think this would be a good addition. We are able to determine if parsing is complete afaik, we should be able to simple flip a bool and refuse any more calls to the SyntaxTreeVisitor 🤔 Or any other way which is maybe more "efficient".

> [...] other WebServices with such top-level components would have too many entries in the currentNodeIndexPath array (since we'd have both the visitor-added root context as well as the one introduced by the root component).

Is there a way to detect if we have such a root Group? (Or TupleComponenet, I think the same thin applies there right?)

This would allow removing the root Group in the WebService right? (If I followed the discussion below correctly).

PR #121, opened 2021-01-12
`EnvironmentObject` configuration

awoc:
# `EnvironmentObject` configuration

The property wrapper `@Environment` can be used to inject pre-defined values to a `Handler`. However, it is not possible to define custom key paths and use them with `@Environment`. This can be helpful when we want to share objects between different `Handler`s to for example share the state or change properties.

## :bulb: Proposed solution

The `EnvironmentObject` configuration takes as arguments a key path conforming a `ApodiniKeys` and the object. After the initialization of the web server, the object can be globally injected using `@Environment`. 

```swift
struct SomeHandler: Handler {
    @Environment(\Key.bird) var bird: Bird

    func handle() -> String {
        bird.name
    }
}

var content: some Component {
    SomeHandler()
}

    EnvironmentObject(Bird(name: "Rudi", age: 12), \Key.bird)
}

struct Key: Keychain {
    var bird: Bird
}

struct Bird {
    var name: String
    var age: Int
}
```

### Implications

- The documentation was edited to reflect the changes.
- ApodiniKeys was renamed to KeyChain

### Testing

The correct injection using `@Environment` was tested.

### Reviewer Nudging

This should be fairly easy.


hendesi (commented):
Sorry for the late review! Besides the one thing I mentioned, the code looks good to me 👍

Sources/Apodini/Configurations/EnvironmentObject.swift
```diff
@@ -0,0 +1,27 @@
...
+        self.value = value
+        self.keyPath = keyPath
+    }
+    
+    public func configure(_ app: Application) {
+        if let previousValue = EnvironmentValues.shared.values[ObjectIdentifier(keyPath)] {
+            print("""
+                Warning: A value associated with the key path \(type(of: keyPath)) is already stored.
+                The previous value \(String(describing: previousValue)) will be overwritten with \(String(describing: value)).
+                """)
```
hendesi: I would prefer to have a real logger warning here and not just a print statement. But this is something which is not consistent in Apodini anyway

hendesi (approved):

Sources/Apodini/Configurations/EnvironmentObject.swift
awoc: I agree this better than using `print`

awoc:
> Sorry for the late review! Besides the one thing I mentioned, the code looks good to me 👍

Well I just requested your review 19 hours ago. So, it was in fact a very quick review 😄

PR #124, opened 2021-01-12
Get rid of NSObject

tgymnich:
# *Get rid of NSObject*

We are currently using `NSObject` for `AnyPropertyOptionKey` to conform to `Hashable`.

## :bulb: Proposed solution
Instead of using the `Hashable` implementation of `NSObject` use `ObjectIdentifier` to give the object an identity and use that to implement `Hashable` and `Equatable`.

### Problem that is solved
Less legacy Objective-C code.

### Implications
None

### Related PRs
None


PSchmiedmayer (approved):
A good addition 👍

Issue #122, opened 2021-01-12
Default `class` `@Parameter`s allow information sharing.


theMomax:
**Describe the bug**
All `Handler`s use the same instance of a `@Parameter`'s default value. This is no problem for value-types. The `@Parameter`'s `set`ter is not available to the user, thus value-types cannot be altered and sharing the default value is fine.

However, reference-types (i.e. `class` objects) can be altered nevertheless. So, when using a `class` type on an `@Parameter` with a default value, `Handler`s share this default value **and** can alter it.

**To Reproduce**

Add the following handler to your code at ["v1", "greet"]:
```swift
class ClassString: Codable {
    var name: String
    
    init(name: String) {
        self.name = name
    }
}

struct Greeter: Handler {
        @Parameter var name: ClassString = ClassString(name: "Apodini")
        @Parameter var overwrite: Bool = false

        func handle() -> String {
            if overwrite {
                self.name.name = "Max"
            }
            return "Hi, \(name.name)!"
        }
    }
```
Open a WebSocket connection:
`websocat ws://127.0.0.1:8080/apodini/websocket`
Send the following messages:
```json
{"context":"a21f0915-825a-422e-b8dc-1876234cd08d","endpoint":"v1.greet"}
{"context":"a21f0915-825a-422e-b8dc-1876234cd08d","parameters": {"overwrite": true}}
```
It should answer with
```json
{"context":"A21F0915-825A-422E-B8DC-1876234CD08D","content":"Hi, Max!"}
{"context":"A21F0915-825A-422E-B8DC-1876234CD08D"}
```

Open a second WebSocket connection:
`websocat ws://127.0.0.1:8080/apodini/websocket`
Send the following messages (not that overwrite is `false`):
```json
{"context":"a21f0915-825a-422e-b8dc-1876234cd08e","endpoint":"v1.greet"}
{"context":"a21f0915-825a-422e-b8dc-1876234cd08e","parameters": {}}
```
It should answer with the same content.
```json
{"context":"A21F0915-825A-422E-B8DC-1876234CD08E","content":"Hi, Max!"}
{"context":"A21F0915-825A-422E-B8DC-1876234CD08E"}
```
Note that this behavior is not specific to the WebSocket exporter or its concept of context. 

**Expected behavior**
As the user of Apodini I imagine `Handler`s to be instantiated for each connection and thus `@Parameter`s to never be shared.

I see several ways to deal with this:

##### It's OK
[...]

PR #125, opened 2021-01-13
Avoid Sharing Parameter Default Values

theMomax:
# Avoid Sharing Parameter Default Values

All Handlers use the same instance of a @Parameter's default value. This is no problem for value-types. The @Parameter's setter is not available to the user, thus value-types cannot be altered and sharing the default value is fine.

However, reference-types (i.e. class objects) can be altered nevertheless. So, when using a class type on an @Parameter with a default value, Handlers share this default value and can alter it.

A detailed example is shown in #122.

## :bulb: Proposed solution
The `Parameter.defaultValue` is stored as a closure. This closure is called each time a new `ConnectionContext` is created. 

### Problem that is solved
See #122

### Implications
The closure is captured using the `@autoclosure` mechanism, thus there are no changes to the public API. Internal changes should not affect exporters.

The same pattern used used in this PR will most likely also be used for implementing the `@State` property wrapper.

### Related PRs
None.

### Testing
Previous tests should avoid regression regarding a `Parameter`'s `options`. The new `testMutationOnClassTypeDefaultParameterIsNotShared` tests if the default value is really not shared between `Handler`.

### Reviewer Nudging
Should be rather easy. The interesting part regarding parameter options (necessity, nullability, ...) happens in `EndpointParameter.swift`. The unwrapping of the closure happens in `Validator.swift`. Note that some I had to add another seemingly redundant initializer to `Parameter`. This was required because otherwise the swift compiler would crash.


Tests/ApodiniTests/EndpointsTreeTests.swift
```diff
@@ -48,7 +48,7 @@ final class EndpointsTreeTests: ApodiniTests {
             // swiftlint:disable:next force_unwrapping
             (0...times!)
                 .map { _ in
-                    "Hello \(name) born in \(birthdate.year)!"
+                    "Hello \(name) born in \("birthdate.year")!"
```
bauer-andreas: Those parentheses seem a bit odd to me? Is this intentional? Doesn't seem to impact anything but still.

Tests/ApodiniTests/ParameterMutabilityTests.swift
```diff
@@ -28,6 +28,29 @@ class ParameterMutabilityTests: ApodiniTests {
...
+        var string: String
+        
+        init(string: String) {
+            self.string = string
+        }
+    }
+    
+    struct TestHandlerUsingClassType: Handler {
+        @Parameter
+        var name = StringClass(string: "Apodini")
```
bauer-andreas: Would you mind adding a test case (or adding an assert to the test case) for a optional parameter with a default value.
As Parameters with optional Type are treated differently than parameters with a non optional type (and also the defaultValue closure is constructed differently) I would find a test case like described important.

bauer-andreas (approved):
Thanks for the quick fix of the bug. The solution looks good to me.
Just think we need a additional test case for Parameters with optional types, otherwise this is ready to merge from my side 👍

Tests/ApodiniTests/EndpointsTreeTests.swift
theMomax: No, it isn't...I had to "uncomment" quite some code in order to find out why the compiler crashed and didn't catch this one when adding stuff back in. Thanks for the hint!

Tests/ApodiniTests/ParameterMutabilityTests.swift
theMomax: Yes, seems reasonable. I will add one.

PSchmiedmayer (commented):
@theMomax I like the addition of capturing the default values using the auto closure 🚀
The only comment I would have would be to consider removing the `defaultClosurePresent` boolean flag and replace it with `defaultValue` being an optional itself (`(() -> Element)?`). 

I have experimented with this a bit while reviewing and testing your code and created a pull request to this branch to showcase the diff (#130).
What do you think about the idea? I think it would make the parameter implementation it bit clearer and would introduce a single source of truth for `defaultValue` and would enable us to put all the information into this single property. I reran the test suit and experimented a bit with the change and could not find any regressions.

theMomax:
> @theMomax I like the addition of capturing the default values using the auto closure 🚀
> The only comment I would have would be to consider removing the `defaultClosurePresent` boolean flag and replace it with `defaultValue` being an optional itself (`(() -> Element)?`).
> 
> I have experimented with this a bit while reviewing and testing your code and created a pull request to this branch to showcase the diff (#130).
> What do you think about the idea? I think it would make the parameter implementation it bit clearer and would introduce a single source of truth for `defaultValue` and would enable us to put all the information into this single property. I reran the test suit and experimented a bit with the change and could not find any regressions.

@PSchmiedmayer there is one core difference between our approaches: Your implementation requires default values of parameters to be constant while mine doesn't. I have no strong opinion on what approach is more desirable. See my detailed comment [here](https://github.com/Apodini/Apodini/pull/130#discussion_r557220113).

bauer-andreas:
> @PSchmiedmayer there is one core difference between our approaches: Your implementation requires default values of parameters to be constant while mine doesn't. I have no strong opinion on what approach is more desirable. See my detailed comment [here](https://github.com/Apodini/Apodini/pull/130#discussion_r557220113).

Not sure if I looked at the difference in implementation detailed enough, but wouldn't it be desirable that the default value does not change? In the case where a description of an WebService like OpenAPI creates it, include default values, the documentation and the actual behavior of the WebService might come out of sync?
In the end we can't fully check that and it's up to the user to not do any evil. With the current approach though, a user could return `nil`, which would be illegal for Parameters with non optional type and triggers `fatalError("Could not unwrap opiontal value for \(definition.description) even though it is not an optional. This should have been detected by 'checkNecessity'.")` in your approach, which doesn't quite describe the occurred issue.

theMomax:
> Not sure if I looked at the difference in implementation detailed enough, but wouldn't it be desirable that the default value does not change? In the case where a description of an WebService like OpenAPI creates it, include default values, the documentation and the actual behavior of the WebService might come out of sync?
> In the end we can't fully check that and it's up to the user to not do any evil. With the current approach though, a user could return `nil`, which would be illegal for Parameters with non optional type and triggers `fatalError("Could not unwrap opiontal value for \(definition.description) even though it is not an optional. This should have been detected by 'checkNecessity'.")` in your approach, which doesn't quite describe the occurred issue.

@Supereg I agree with your first point (if OpenAPI really needs a default value). However, I either don't get your last point or I think it is not valid.

The user can only return `nil`, if the `Parameter`'s `Element` is `ExpressibleByNilLiteral`. And then it is a valid value. Note that the `wrappedValue` taken by the `Parameter`'s `init`s is of type `() -> Element`. Only internally `() -> Element?` is used.

bauer-andreas:
> [...] However, I either don't get your last point or I think it is not valid.

> The user can only return nil, if the Parameter's Element is ExpressibleByNilLiteral. And then it is a valid value. Note that the wrappedValue taken by the Parameter's inits is of type () -> Element. Only internally () -> Element? is used.

Ah yeah, I mixed up stuff. Thanks for the clarification 👍 The optional came from the `Type` in `EndpointParameter` I see.

PSchmiedmayer (approved):
As said in the discussion in #130 I would also prefer non mutable default values. I have merged #130 in this branch. Thanks for the work @theMomax 👍

Sources/Apodini/Properties/Parameter.swift
```diff
@@ -23,8 +23,8 @@ public struct Parameter<Element: Codable>: Property {
     let id: UUID
     let name: String?
     private var element: Element?
-    internal let options: PropertyOptionSet<ParameterOptionNameSpace>
-    internal let defaultValue: Element?
+    internal var options: PropertyOptionSet<ParameterOptionNameSpace>
+    internal var defaultValue: (() -> Element)?
```
bauer-andreas: The merge commit seemingly reverted changes done in #106 (Moving to `let`)

theMomax:
I would be fine merging. But keep in mind that the following is still perfectly valid code:
```swift
@Parameter var number = Int.random()
```

PSchmiedmayer:
@theMomax That's a good point 😅
I think we will not be able to change that but at least this PR enable the functionality so you won't be able to easily change it once a `Handler` is initialized for a request and there would be no side effects in case you would (e.g. for a class).

PR #128, opened 2021-01-13
Replace ParticularType with general type info functions

PSchmiedmayer:
# *Replace ParticularType with general type info functions*

After reviewing #115 and merging it into develop I have realized that there is duplicate functionality between `ParticularType` and the global `isOptional` function.
 
## :bulb: Proposed solution
The removal of `ParticularType` reduces the number of allocations when evaluating type information and unifies the type information retrieval to the internal global functions.

### Problem that is solved
Reduces some code duplication in Apodini.

### Implications
Further usages of type information should use the global functions. 

### Related PRs
Is optional was introduced in #115 and [described in this comment thread](https://github.com/Apodini/Apodini/pull/115#discussion_r555314441).

### Testing
This PR introduces some tests for the global functions to get type information.

### Reviewer Nudging
It would be interesting if I missed some type information checking in other parts of Apodini that we could also unify as part of this PR.


awoc (approved):
I like the restructuring and putting it in a central place 👍

PR #129, opened 2021-01-13
Throwable Handler

PSchmiedmayer:
# *Throwable Handler*

Currently `Handler`s can't throw Swift `Error`s in the `Handler`'s `handle()` function.

## :bulb: Proposed solution
We allow `Handler`s to throw Swift `Error`s in the `Handler`'s `handle()` function by declaring it as `throws` and implementing propagating errors to an exporter.

### Problem that is solved
Swift `Error`s can now be thrown in the `Handler`'s `handle()` function.

### Implications
All exporters should consider the possibility that `EventLoopFutures` can fail. This was possible before if, e.g. the developer returned an `EventLoopFuture` in the `Handler`'s `handle()` function but now it is even easier to throw errors in a `Handler`'s `handle()` function.

### Testing
Introduces unit tests to test the error handling for the RESTExporter. Dosen't test the functionality for the other exporters. This will be discussed in the next team meeting.

### Reviewer Nudging
Feel free to point out possible issues that might arise in other exporters that I have not considered yet in the PR.


hendesi (approved):
Great job 👍 I really like this. It is also super beneficial for the database handler as we can now propagate the errors that happened in query or wherever back to the user. 
Maybe as we now have the infrastructure in the handlers and we more and more decouple vapor from the exposed Apodini, it might be time to think about an `ApodiniError` we can throw back to the user?

lukaskollmer:
this is a very good addition.

one question, a `handle() throws -> String` function throwing an error would be semantically equivalent to a `handle() -> EventLoopFuture<String>` function returning a failed future, right?

Sources/Apodini/Utilities/EventLoopFuture.swift
```diff
@@ -0,0 +1,34 @@
...
+    ///
+    /// If your callback function throws, the returned `EventLoopFuture` will error.
+    ///
+    /// - parameters:
+    ///     - callback: Function that will receive the value of this EventLoopFuture and return a new EventLoopFuture. The provided `callback` may optionally `throw`.
+    /// - returns: A future that will receive the eventual value.
+    @inlinable
+    public func tryFlatMap<NewValue>(file: StaticString = #file,
+                                     line: UInt = #line,
+                                     _ callback: @escaping (Value) throws -> EventLoopFuture<NewValue>) -> EventLoopFuture<NewValue> {
```
lukaskollmer: i'd call this `transform` iirc that's the terminology used in the swift stdlib

lschlesinger (approved):
Looks good to me! Thinking about how this would affect the `OpenAPIExporter`, I am wondering if this could be beneficial to provide more specific http (error) codes in the documentation. But I am not sure yet, if this change improves the information situation _before_ runtime. I have to investigate this topic ;)

PSchmiedmayer:
@hendesi Yes, adding some sort of error `Apodini` `HandlerError` that is more or less HTTP independent (`Vapor.Abort` is closely coupled to HTTP error codes) would be very beneficial. This error type could maybe also used in the description DSL that @Supereg @lukaskollmer and @lschlesinger wanted to investigate.

@lukaskollmer Yes, these two return types would be  semantically equivalent. 

@lschlesinger As this is only an addition to the runtime functionality this will probably not provide you any startup time information about errors. I think it will be near to impossible to detect what error might be thrown in a function so explicitly providing them on startup using a description DSL seems to be the only feasible approach for me at the moment. Sadly Swift don't support typed `throws` and therefore getting the different types of errors that could be thrown won't be possible using the type system. But maybe using an `HandlerError`/`ApodiniError` or similar in the description DSL might be a doable approach so any errors thrown in the function can be easily mapped to entries in the description DSL.

Sources/Apodini/Utilities/EventLoopFuture.swift
PSchmiedmayer: I wasn't sure about the naming too. As all `EventLoopFuture` functions are named `map` and `flatMap` and `flatMapThrowing` is weirdly a `map` that can throw. And there is no `mapThrowing` 🤷‍♂️

At fist I didn't want to also name the function `flatMapThrowing`. But looking at the naming problem now maybe it might be better to provide an overload to `flatMapThrowing` and add a `mapThrowing` function to clearly indicate the difference in code between the two?

hendesi: @PSchmiedmayer just out of curiosity what does `#file` / `#line` mean, it this the same as `#function`?

lukaskollmer: @hendesi yes. if it's the function parameter's default value it evaluates to the file name and line number of the caller.

lukaskollmer: @PSchmiedmayer i was talking about the `callback` parameter.

I'd say `tryFlatMap` is fine, if we want to clearly differentiate from `flatMapThrowing`. I'm not sure how i'd feel about overloading the function with different closure return types, although i guess it'd be fine since they both do the same thing, and in most cases the user doesn't actually care if its a map or a flatMap operation, they just want to somehoe process a future into a different future.

PSchmiedmayer: @hendesi There are a few of these literal expression in Swift. The Language reference provides a good overview over everything that is currently exposed using the `#...` literal expressions: https://docs.swift.org/swift-book/ReferenceManual/Expressions.html#ID389

PSchmiedmayer: Ah I see, I was missing that your comment referred to `callback`. I took `callback` as it is the same name used in `NIO` across the implementations of `flatMap` and so on.

I will rename `tryFlatMap` to `flatMapThrowing` so it is fully in line with the NIO naming. I agree with your opinion that is is probably not really important for the develop to differentiate between the overloads. I will also change the scope to internal so we don't expose the function for now.

lukaskollmer:
One more general issue i see with this (and this is something which I think would also already apply to handlers returning failed futures) is the risk of exposing internal errors to the user.
For example, if a handler uses some database component with a throwing api (eg `try db.save()`) and the error thrown by the database contains internal data, we'd currently simply return that to the client.
I'm not sure what would be the best way to handle this, but I think that by default throwing an error inside a handler should not expose any details about this error to the client.
Maybe this could be integrated with the above-mentioned `ApodiniError`, so that apodini performs a check whether an error thrown in a handler conforms to `ApodiniError`, and only returns the error to the client if it does. Otherwise, we'd return just the appropriate status code and some "an error was encountered" message.

PR #130, opened 2021-01-14
Remove defaultClosurePresent

PSchmiedmayer:
This PR showcases the idea presented in the [review](https://github.com/Apodini/Apodini/pull/125#pullrequestreview-567782890) of #125

### DIFF
--- a/Sources/Apodini/Semantic Model Builder/ParameterModelBuilder.swift
+++ b/Sources/Apodini/Semantic Model Builder/ParameterModelBuilder.swift
@@ -97,8 +97,8 @@ extension Parameter: EncodeOptionalEndpointParameter where Element: ApodiniOptio
                 guard let member = defaultValue().optionalInstance else {
                     fatalError(
                         """
-                        Encountered an internal Apodini error.
-                        Default values of `@Parameter`s are constants should not be changable by the developer using Apodini.
+                        Encountered an internal Apodini error: Default values of `@Parameter`s are constants.
+                        The developer using Apodini should make sure they do not change their value during runtime.
                         The orginal default value for the @Parameter was \(originalDefaultValue) and now it is nil.
                         """
                     )
--- a/Sources/Apodini/Semantic Model Builder/ParameterModelBuilder.swift
+++ b/Sources/Apodini/Semantic Model Builder/ParameterModelBuilder.swift
@@ -92,10 +92,16 @@ extension Parameter: EncodeOptionalEndpointParameter where Element: ApodiniOptio
         necessity: Necessity
     ) -> AnyEndpointParameter {
         var `default`: (() -> Element.Member)?
-        if let defaultValue = self.defaultValue, defaultValue().optionalInstance != nil {
+        if let defaultValue = self.defaultValue, let originalDefaultValue = defaultValue().optionalInstance {
             `default` = {
                 guard let member = defaultValue().optionalInstance else {
-                    fatalError("Could unwrap the value during startup. Default @Parameter values MUST NOT change.")
+                    fatalError(
+                        """
+                        Encountered an internal Apodini error.
+                        Default values of `@Parameter`s are constants should not be changable by the developer using Apodini.
+                        The orginal default value for the @Parameter was \(originalDefaultValue) and now it is nil.
+                        """
+                    )
                 }
                 return member
             }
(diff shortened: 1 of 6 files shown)


Sources/Apodini/Semantic Model Builder/ParameterModelBuilder.swift
```diff
@@ -92,9 +91,14 @@ extension Parameter: EncodeOptionalEndpointParameter where Element: ApodiniOptio
...
         necessity: Necessity
     ) -> AnyEndpointParameter {
-        var `default`: () -> Element.Member? = { nil }
-        if self.defaultClosurePresent {
-            `default` = { self.defaultValue()?.optionalInstance }
+        var `default`: (() -> Element.Member)?
+        if let defaultValue = self.defaultValue, defaultValue().optionalInstance != nil {
+            `default` = {
+                guard let member = defaultValue().optionalInstance else {
+                    fatalError("Could unwrap the value during startup. Default @Parameter values MUST NOT change.")
```
theMomax: This is basically exactly what I wanted to avoid with my approach. I see that this restriction indeed makes sense from an (idiomatic) REST/OpenAPI perspective. However, the compiler doesn't stop the developer from calling an arbitrary function to obtain the default value for a `Parameter`, so this fatal error would come rather unexpectedly I guess.

There are cases where I would consider a non-constant default value ok within a REST API, but I also think the cleaner (and standard) way of implementing this is to make the parameter a nullable optional with an annotation describing what is used if the parameter is null.
Consider the following example for the current time:
With variable default values one could just write the following:
```swift
struct Permission: Handler {
    @Parameter var user: User
    @Parameter var action: Action
    @Parameter var date = Date()

    func handle() -> Bool {
         PermissionService.isOk(for: user, toPerform: action, on: date)
    }
}
```
A more traditional way of doing it would probably be the following:
```swift
struct Permission: Handler {
    @Parameter var user: User
    @Parameter var action: Action
    // The current time is used if `date` is `nil`
    @Parameter var date: Date?

    func handle() -> Bool {
         let date: Date = self.date ?? Date()

         return PermissionService.isOk(for: user, toPerform: action, on: date)
    }
}
```

Note that in this example the version using a default value does not use an optional for `date`. This was just the first thing that came to my mind, but I am sure one could also come up with a scenario where the default value is an optional and changes from `nil` to `some` at certain events.

Basically I think that there are valid points for both strategies. I will leave this decision up to you. However, if you choose this PR, I would suggest one change:

Currently your fatal error states `Default @Parameter values MUST NOT change`, but it only enforces that the case of an `Optional` cannot change (my example from above using the default value for `date` would still work). I would suggest we either really enforce equality, or we change the message accordingly.

theMomax (changes requested):
I, too, experimented with this approach at first, but I didn't want to make the restriction for optional parameters you've made. I am fine with this approach. The only change request I have is that the message in the fatal error should match what really is enforced.

Sources/Apodini/Semantic Model Builder/ParameterModelBuilder.swift
```diff
@@ -92,10 +92,16 @@ extension Parameter: EncodeOptionalEndpointParameter where Element: ApodiniOptio
...
         var `default`: (() -> Element.Member)?
-        if let defaultValue = self.defaultValue, defaultValue().optionalInstance != nil {
+        if let defaultValue = self.defaultValue, let originalDefaultValue = defaultValue().optionalInstance {
             `default` = {
                 guard let member = defaultValue().optionalInstance else {
-                    fatalError("Could unwrap the value during startup. Default @Parameter values MUST NOT change.")
+                    fatalError(
+                        """
+                        Encountered an internal Apodini error.
+                        Default values of `@Parameter`s are constants should not be changable by the developer using Apodini.
```
theMomax:
```suggestion
                        Default values of `@Parameter`s are constants. The developer using Apodini should make sure they do not change their value during runtime.
```

theMomax (approved):
The sentence is still a bit weird, but everyone should get the gist😄

Sources/Apodini/Semantic Model Builder/ParameterModelBuilder.swift
PSchmiedmayer: @theMomax @Supereg Based on the discussion in #125 I would prefer to make default values non mutable.

I have updated the error description for the runtime failure in ParameterModelBuilder.swift. @theMomax Does this more closely reflect what you had in mind? As far as I can see at the current point it would be impossible for the developer to actually change the default value and therefore I have named it an internal Apodini error in the description.

PR #131, opened 2021-01-14
Protobuffer generation: handle many parameters

### LINKED ISSUE #62: Protobuffer generation: handle many Parameters
Methods of a gRPC service only accept one message argument:

```proto
service GreeterService {
  rpc handle (StringMessage) returns (StringMessage);
}
```

This is invalid:

```proto
service GreeterService {
  rpc handle (StringMessage, BoolMessage) returns (StringMessage);
}
```

However, we might want to declare many `Apodini.Paramter`s in a `Apodini.Component`:

```swift
struct Greeter: Handler {
    @Parameter
    var name: String
    
    @Parameter
    var isFormal: Bool
    
    func handle() -> String {...}
}
```

**Propose a solution**

_Alternative 1_: Interpret the Component itself as the top-level message type. 🥇

_Alternative 2_: Pre-generate _tuple messages_, but with named fields. Possibly with **gyb**.

_Alternative 3_: A top-level message with one repeated field of a _union type_ message using **oneof**.

**Additional context**
- https://github.com/Apodini/Apodini/pull/33#discussion_r541013183
- https://developers.google.com/protocol-buffers/docs/proto3#oneof
- https://en.wikipedia.org/wiki/Union_type
- https://en.wikipedia.org/wiki/Tagged_union

### LINKED ISSUE #140: Protobuffer: Support `Int` and `UInt`
It was previously unsupported, because we reasoned that we cannot default to `Int32` or `Int64`.

Assume a message type derived from:
```swift
struct Box {
  let value: Int
}
```

The default cannot be `Int32`: If `Box(value: Int.max)` is initialized on a 64-bit server, then it is truncated, when it is decoded on a 32-bit client. At least, the server can handle all incoming requests.
Vice versa, the default cannot be `Int64`.

**Propose a solution**
Read `Int.bitWidth` and further parameterize `ProtobufferBuilder`.

Alternatively, update the documentation: https://github.com/Apodini/Apodini/blob/develop/Sources/ProtobufferCoding/README.md

**Additional context**
https://github.com/Apodini/Apodini/blob/b126c36268c20cd8727dc8356473a5e359bf780a/Sources/Apodini/Utilities/TypeInfo.swift#L38

nityanandaz:
## Problem that is solved

- Closes #62
- Closes #140

## Implications

- Files in _SMB/gRPC/ProtobufferBuilder_ moved to _SMB/Protobuffer_
- _ProtobufferBuilderTests_ and _~ProtobufferBuilderInternalTests~_ are merged
- `EnrichedInfo.propertyInfo` and `EnrichedInfo.propertyOffset` are merged into `Apodini.PropertyInfo`. The information scope is narrower.

## Related PRs

The feedback of #150 will implemented there.

## Reviewer Nudging

I apologize that this PR got larger that anticipated, but I had the urge to move the implementation closer together.


moritzschuell (changes requested):
For some reason, this branch's `TestWebService` does not work for me 🤔
It fails in `EnrichedInfo.swift` line 58 with the `preconditionFailure(errorDescription)`:

`Thread 1: Fatal error: couldNotGetTypeInfo(type: () -> Swift.UInt, kind: Runtime.Kind.function)`

I tried clean-building and also replacing any `Int`s or `UInt`s in the `TestWebService` with their fixed-width counterparts, but it still keeps crashing.

Tests are passing like a charm, though.

Sources/Apodini/Semantic Model Builder/Protobuffer/Model+EnrichedInfo.swift
```diff
@@ -0,0 +1,89 @@
...
+                return try compatibleGenericName() + "Message"
+            case .tuple:
+                return try tupleName() + "Message"
+            default:
+                throw ProtobufferInterfaceExporter.Error(message: "Kind: \(kind) is not supported")
+            }
+        }
+    }
+    
+    func compatibleGenericName() throws -> String {
```
lschlesinger, line 75: I just today did the same task for OAS: handling type names including incompatible characters, such as `<`, `>`, and `,` 😄 ! I tried to not hide the information by what these characters are replaced, through defining an enum including some constants:
```swift
enum OpenAPISchemaConstants {
    static let replaceAngleBracket = "_"
    static let replaceCommaSeparation = "-comma-"
}
```
This also allowed by to use these properties in the tests and not getting out of sync in case I want to change something! Maybe you find that useful, too.
I came up with this enum solution for constants, as this is the only format swiftlint will not complain about.

lschlesinger: Additionally, I used regex, to catch cases where `X, Y` lacks the whitespace: `X,Y`. So I use:
```swift
.replacingOccurrences(of: "(\\,\\s|\\,)", with: OpenAPISchemaConstants.replaceCommaSeparation, options: .regularExpression)
```

Sources/Apodini/Utilities/Tree.swift
```diff
@@ -141,6 +141,28 @@ extension Node {
     }
 }
 
+extension Node {
+    func collectValues() -> Set<T> where T: Hashable {
```
lschlesinger, line 172: I find this very useful!! And as I already sneaked a peek at that PR before, I already incorporated this in my work by simply copying this to check if it solves my task - it worked. So I am happy to find this in develop, soon 👍

Sources/Apodini/Semantic Model Builder/Protobuffer/ProtobufferInterfaceExporter.swift
```diff
@@ -0,0 +1,197 @@
...
+
+// MARK: - ProtobufferInterfaceExporter.Builder Implementation
+
+extension ProtobufferInterfaceExporter.Builder {
+    static func buildMessage(_ type: Any.Type) throws -> Node<ProtobufferMessage> {
+        try buildCompositeMessage(type) ?? buildScalarMessage(type)
+    }
+    
+    static func buildCompositeMessage(_ type: Any.Type) throws -> Tree<ProtobufferMessage> {
+        try EnrichedInfo.node(type)
```
lschlesinger, line 111: I encountered that I need to do this "editing process" recursively, in order to catch the cases, where collection types are nested. The related PR #179 is not ready yet, but I think it is enough for you to identify if you might run in the same problems. The associated Issue #178 hopefully help to understand, what I try to reach in the PR.

lschlesinger (approved):
I find the `propertyInfo` refactoring reasonable 👍 Added comments regarding the string replacement in `compatibleType` and a link to a PR which I think is related to your `.edited()` procedure. If so, this can also be addressed in a separate PR - this one is good to go I would say 😉

Sources/Apodini/Semantic Model Builder/Protobuffer/Model+EnrichedInfo.swift
nityanandaz: Did you catch any cases were the space is actually missing? I expect this behavior to be stable 🙂

Sources/Apodini/Semantic Model Builder/Protobuffer/ProtobufferInterfaceExporter.swift
nityanandaz: I am also working/experimenting on a solution for this: https://github.com/Apodini/Apodini/blob/156cba7df7d1dbceb877cf95fd463078535770b1/Sources/Apodini/Semantic%20Model%20Builder/Model/TypeReflection/EnrichedInfo.swift#L148 on https://github.com/Apodini/Apodini/tree/refactor/EnrichedInfo.node

I will look into it 👍

Sources/Apodini/Semantic Model Builder/Protobuffer/Model+EnrichedInfo.swift
lschlesinger: I thought I had seen it somewhere in the debugger... but I just checked again, you're right!

moritzschuell (approved):
Looks good to me 👍 
This is a big step forward, thank you!

PR #135, opened 2021-01-14
`@State`

theMomax:
# `@State`

`Handler`s cannot maintain state across multiple requests of the same connection.

## :bulb: Proposed solution
The `@State` property wrapper as defined in our vision documents. It is supported for Handlers, Guards/SyncGuards and Transformers.

It uses `@autoclosure` to create separate instances of the contained value for each context. At first only this closure is stored. `State` has an internal `activate` method. This method must be called before the `wrappedValue` can be accessed. The `activate` method first evaluates the closure to get the initial value. Then it creates a private `Wrapper` class holding this value. Now `wrappedValue` can use this wrapper to update and retrieve the value. This `Wrapper` is a reference and is thus shared by all further copies of the `State` struct.

### Implications
There is no additional work for exporters relying on the shared model. Changes to the public API are purely additional, so any old code should still work.

@PSchmiedmayer I started out with a copy of [OpenSwiftUI's State implementation](https://github.com/Cosmo/OpenSwiftUI/blob/master/Sources/OpenSwiftUI/State.swift). However, I changed so much of it, that there is hardly anything left from that code. Thus I removed the license again. If you prefer to still include it, I can add it back in again.

### Testing
I wrote the `StateTests` checking that `@State` is not shared between different `ConnectionContext`s and that it indeed maintains state across different evaluations on the same context. Both scenarios are tested once with the `Element` being a reference- and once with it being a value-type.

### Reviewer Nudging
Start out with `State.swift`. Make sure you understand what is achieved by using the `Wrapper` and what `activate` does. Then take a look at the `EndpointInstance`. It instantiated with each new `ConnectionContext`. The last bigger change is the refactoring of `AnyGuard` which was necessary to implement its `activate` function. The rest basically doesn't contain any real logic.


moritzschuell (approved):
Looks good to me 👍 Thank's for moving forward with this!

Maybe a further test-case for two context's on the same exporter (instead of different exporters) with interleaved requests would be a nice addition.

theMomax:
> Looks good to me 👍 Thank's for moving forward with this!
> 
> Maybe a further test-case for two context's on the same exporter (instead of different exporters) with interleaved requests would be a nice addition.

You're right about the tests. Testing on the same exporter makes a lot more sense. I'll add one here and also adapt the tests for `@Parameter`.

PR #136, opened 2021-01-14
Exclude Mock Certificates from Test Target

tgymnich:
# Exclude Mock Certificates from Test Target

I forgot to exclude the certificates I added to test the http2 config from the test target

## :bulb: Proposed solution
Just exclude them

### Problem that is solved
Warning is fixed

### Implications
none

### Related PRs
#82

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -67,7 +67,9 @@ let package = Package(
                 "ConfigurationTests/mock_fcm.json",
                 "ConfigurationTests/mock_invalid_fcm.json",
                 "ConfigurationTests/mock.p8",
-                "ConfigurationTests/mock.pem"
+                "ConfigurationTests/mock.pem",
+                "ConfiguartionTests/cert.pem",
+                "ConfiguartionTests/key.pem"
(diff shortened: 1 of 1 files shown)


PSchmiedmayer (approved):

Package.swift
```diff
@@ -67,7 +67,9 @@ let package = Package(
                 "ConfigurationTests/mock_fcm.json",
                 "ConfigurationTests/mock_invalid_fcm.json",
                 "ConfigurationTests/mock.p8",
-                "ConfigurationTests/mock.pem"
+                "ConfigurationTests/mock.pem",
+                "ConfiguartionTests/cert.pem",
+                "ConfiguartionTests/key.pem"
```
awoc, line 72: Typo 🙃

PSchmiedmayer: Thanks, I fixed this with https://github.com/Apodini/Apodini/commit/beff00cfd7e821767afd6ef7ec21e42903a4fdb6

Issue #118, opened 2021-01-11
Upgrade OAS to correctly represent RESTful API with HATEOAS


lschlesinger:
It's more a shortcoming than a bug, as the described current behavior of the OpenAPIExporter was planned to be adapted iteratively to represent a RESTful application correctly. 

**Bug/problem description**
Currently, the `responses` section in the OpenAPI Specification does not correctly represent the response returned by a RESTful API. 
As an initial version, this section provides documentation for the return type of the `Handler`s `handle()` function.
Having a RESTful API always returning an object including a nested `data` object as well as a `_links` object, this section needs to be refactored. 

**To Reproduce**
Following examples base on the following example `TestWebService`

```swift
struct TestWebService: Apodini.WebService {

... // `Handler`s and other types

    @PathParameter var userId: Int

    var content: some Component {
        Group("swift") {
            Text("Hello Swift! 💻")
                .response(EmojiMediator())
            Group("5", "3") {
                Text("Hello Swift 5! 💻")
            }
        }
         Group("user", $userId) {
            UserHandler(userId: $userId)
        }
    }
}
```

#### Current behavior
Definition of **`handle()` function** in `Text.swift` returning a **String**:
```swift 
public func handle() -> String {
    text
}
```

A request to http://localhost:8080/v1/swift object would lead to the following **JSON** **response**:

```
{
    "data": "✅ Hello Swift! 💻 ✅",
    "_links": {
        "5_3": "http://127.0.0.1:8080/v1/swift/5/3",
        "self": "http://127.0.0.1:8080/v1/swift" 
    }
}
```
Whereas the current **OpenAPI Specification**  lists the response as a simple **String** that is returned:

``` 
"responses": {
    "200": {
        "content": {
            "text/plain": {
                "schema": {
                    "type": "string"
                }
            }
        }
    }
}
```
#### Possibly expected OAS

```
"responses": {
    "200": {
        "content": {
            "application/json": {
                "schema": {
                    "type": "object",
                    "properties": {
                        "data": {
                            "type": "string"
                        },
                        "_links": {
                            "type": "object",
                            "additionalProperties": {
                                "type": "string",
                                "format": "url"
                            }
                        }
                    }
                }
            }
        }
    }
    ...
}
```

#### Current behavior
Definition of **`UserHandler`** with respective **`handle()`**  function returning a **`User`** object:
```swift 
struct User: Codable {
    var id: Int
}

struct UserHandler: Handler {
    @Parameter var userId: Int

[...]

PR #86, opened 2020-12-30
Adding basic HATEOAS support to the REST exporter

bauer-andreas:
~Draft status: waiting for #85 to get merged, as features introduced there need to be incorporated~

# Adding basic HATEOAS support to the REST exporter

## :recycle: Current situation

In the current situation the REST exporter passes the Encodable instance returned from the `Handler` directly to the the Vapor Response encoder. HATEOAS information is not incorporated.

## :bulb: Proposed solution

This PR adds a very basic initial version of HATEOAS information to the responses, by adding a `_links` object to the json response.

### Problem that is solved

&ndash;

### Implications

&ndash;

### Related PRs

&ndash;

### Testing

Added a test case covering REST request handling. 

### Reviewer Nudging

&ndash;

PR #138, opened 2021-01-15
Dictionary Type Reflection & Prep. for HATEOAS

lschlesinger:
# Enable to handle response object returned from REST Exporter

The REST response will include a **dictionary** (i.e., `_links` in the built `ResponseContainer` holding the data itself and also HATEOAS information). So far, we can not properly reflect dictionaries in terms of providing a preprocessed structure of `Node<EnrichedInfo>` to further be used in the exporters, e.g. the `OpenAPIInterfaceExporter`.

Furthermore, we need to consider, that a response that is an object, i.e. a `ResponseContainer` would as of now always be handled as a "complex" type in the OAS and therefore be referenced in the response section and described in the components/schema section. As an example:
For a `Handler` like this:
```swift 
struct User: Codable {
    var id: Int
}

struct UserHandler: Handler {
    @Parameter var userId: Int

    func handle() -> User {
        User(id: userId)
    }
}
```
The expected response (wrapped in a type called [`ResponseContainer`](https://github.com/Apodini/Apodini/blob/develop/Sources/Apodini/Semantic%20Model%20Builder/REST/RESTEndpointHandler.swift#L11)) would be:
```json
{
    "data": {
        "id": 1
    },
    "_links": {
        "self": "http://127.0.0.1:8080/v1/user/:B285FABD-CA5A-4432-A456-8050864E4E05"
    }
}
```
What we currently get, if we treat `ResponseContainer` as every other complex type in the OAS is:
```json
"responses": {
    "200": {
        "content": {
            "application/json": {
                "schema": {
                  "$ref": "#/components/schemas/ResponseContainer"
                }
            }
        }
    }
}
```
where `ResponseContainer` would be further described in the `components/schema` section of the OAS.

What we would like to see in OAS is:
```json
"responses": {
    "200": {
        "content": {
            "application/json": {
                "schema": {
                    "type": "object",
                    "properties": {
                        "data": {
                            "$ref": "#/components/schemas/User"
                        },
                        "_links": {
                            "type": "object",
                            "additionalProperties": {
                                "type": "string",
                            }
                        }
                    }
                }
            }
        }
    }
}
```

## :bulb: Proposed solution

1. We provide the means to `handleDictionary`:
A dictionary of type `[String: User]` would be reflected as:
`EnrichedInfo` with `cardinality = .zeroToMany(.dictionary(key: EnrichedInfo(typeInfo: typeInfo(of: String.self), ...), value: EnrichedInfo(typeInfo: typeInfo(of: User.self),...))` 

As precondition for:

2. Within the `OpenAPIExporter` we are able to handle the case of a returned `ResponseContainer` as wrapper type.
As a first step, the `OpenAPIComponentsObjectBuilder` currently checks (by means of String comparison) if the reflected type is a `ResponseContainer` and will inline the type in OAS (see example above).
As soon as @Supereg and I have implemented the actual `ResponseContainer<T>` type as used in `RESTExporter`, this will be refactored accordingly. It is still part of this PR as this allows to already test all relevant logic.

### Problem that is solved

Prepare for #118; add infrastructure to properly reflect dictionaries.

### Implications

`ProtobufferBuilder` as well as `GraphQLESemanticModelBuilder` can include `.edited(handleDictionary)` in the chain of operation on a tree of `Node<EnrichedInfo>`s.

`EnrichedInfo`s   is extended by the enum `CollectionContext` to distinguish between arrays and dictionaries/maps. In case `cardinality` is `.zeroToMany`(indicating a collection), the `CollectionContext` has to be set accordingly. Potentially, `.set` might be another future `CollectionContext` for Swift `Set`.

As in `OpenAPIComponentsObjectBuilder` the `node`s (`Node<EnrichedInfo>`) `cardinality` is checked against, `EnrichedInfo` now conforms to `Equatable`.

### Related PRs

Builds on top of #33 and #35 (`EnrichedInfo` and OAS); uses ideas from #86 (HATEOAS).

### Testing

`OpenAPIComponentsObjectBuilderTests` tests if the OAS is built correctly for respective types, also the newly covered `let someDict = ["someKey": "someValue"]`. 

The "provisional" type `ResponseContainer<T>`, for the time being, mimics the response that will be sent by the REST application. The expected result is described in #118.

`EnrichedInfoTests` tests, if `EnrichedInfo` is built correctly and if enums `Cardinality` and `CollectionContext` are set correctly.   

### Reviewer Nudging

Start with `EnrichedInfo`, then `handleDictionary` and finally OAS related files. Hit me up, especially if the OAS related code is not clear to you.


Sources/Apodini/Utilities/TypeInfo.swift
```diff
@@ -36,6 +36,7 @@ func isOptional(_ type: Any.Type) -> Bool {
 
 // MARK: - Supported Scalar Types
 private let supportedScalarTypes: [ObjectIdentifier] = [
+    ObjectIdentifier(Int.self),
```
bauer-andreas, line 39: This may be related to #140.
To my knowledge `supportedScalarTypes` strongly resembles what the Protobuffer Exporters support in theirs of primitive types. In fact, In my opinion such exporter specific information should be moved into the exporter itself.
However that's currently not the case and thus might break the current gRPC exporter(?). If I'm not wrong (help @nityanandaz)

bauer-andreas (approved):
First of all, sorry for the delay.
Looks good to me (aside from the one comment).
However I'm not an export in the `EnrichedInfo` API, so you might want to wait for @nityanandaz review as well 👍

Sources/Apodini/Utilities/TypeInfo.swift
lschlesinger: @nityanandaz whats your opinion on this? If you manage to support `Int` (as planned with the mentioned issue #140), would it be reasonable to include `Int` in supportedScalarTypes?

Sources/Apodini/Semantic Model Builder/Model/TypeReflection/EnrichedInfo.swift
```diff
@@ -21,6 +23,50 @@ struct EnrichedInfo {
...
+            return (lhsKey, lhsValue) == (rhsKey, rhsValue)
+        default:
+            return false
+        }
+    }
+}
+
+extension EnrichedInfo: Equatable {
+    public static func == (lhs: EnrichedInfo, rhs: EnrichedInfo) -> Bool {
+        if lhs.typeInfo.mangledName != rhs.typeInfo.mangledName {
```
nityanandaz:
```suggestion
        lhs.typeInfo.type == rhs.typeInfo.type
            && lhs.propertyInfo?.name == rhs.propertyInfo?.name
            && lhs.propertiesOffset == rhs.propertiesOffset
            && lhs.cardinality == rhs.cardinality
```

Plus, I exchanged `typeInfo.mangledName` for `typeInfo.type`.

And I'd put `EnrichedInfo: Equatable` above the other two extensions, preserving the hierarchy.
Futhermore, I'd put all three of them at the end of the file, maybe below a a `MARK` comment:
`// MARK: - EnrichedInfo: Equatable`.

Sources/Apodini/Semantic Model Builder/Model/TypeReflection/EnrichedInfo.swift
```diff
@@ -11,7 +8,12 @@ struct EnrichedInfo {
     enum Cardinality {
         case zeroToOne
         case exactlyOne
-        case zeroToMany
+        case zeroToMany(CollectionContext)
+    }
+
+    enum CollectionContext {
+        case array
+        indirect case dictionary(key: EnrichedInfo, value: EnrichedInfo)
```
nityanandaz, line 16: @moritzschuell does `ProtobufferCoding` support dictionaries? For Protobuffer a wrapper type should be feasible:

```swift
struct DictionaryItem<K, V> {
  let key: K
  let value: V
}
```
```proto
message DictionaryOfKAndVMessage {
  repeated DictionaryItemOfKAndV values = 1;
}
```

Sources/Apodini/Utilities/TypeInfo.swift
nityanandaz: I agree that this is covered with the mentioned issue. I'm already started working on it, it should not block your PR.

Tests/ApodiniTests/SemanticModel/EnrichedInfoTests.swift
```diff
@@ -0,0 +1,160 @@
...
+//  Created by Lorena Schlesinger on 1/15/21.
+//
+
+import XCTest
+@_implementationOnly import Runtime
+
+@testable import Apodini
+
+class EnrichedInfoTests: ApodiniTests {
+    func testCardinalityIsEquatable() throws {
```
nityanandaz, line 11: This is not hard evidence, but I feel these tests are very fine-grained

Tests/ApodiniTests/SemanticModel/EnrichedInfoTests.swift
```diff
@@ -0,0 +1,160 @@
...
+        XCTAssertNotEqual(
+            EnrichedInfo.CollectionContext.dictionary(key: keyType, value: valueType),
+            EnrichedInfo.CollectionContext.array
+        )
+        XCTAssertNotEqual(
+            EnrichedInfo.CollectionContext.dictionary(key: keyType, value: valueType),
+            EnrichedInfo.CollectionContext.dictionary(key: keyType, value: valueType1)
+        )
+    }
+
```
nityanandaz, line 108: I prototyped a helper function. Maybe you find it useful:
```swift
func testCardinality(_ lhs: Any.Type, _ rhs: Any.Type, isEqual: Bool) {
    let keyPath = \Node<EnrichedInfo>.value.cardinality
    
    if isEqual {
        XCTAssertEqual(
            EnrichedInfo.node(lhs)[keyPath: keyPath],
            EnrichedInfo.node(rhs)[keyPath: keyPath]
        )
    } else {
        XCTAssertNotEqual(
            EnrichedInfo.node(lhs)[keyPath: keyPath],
            EnrichedInfo.node(rhs)[keyPath: keyPath]
        )
    }
}
```

Tests/ApodiniTests/SemanticModel/EnrichedInfoTests.swift
lschlesinger: That's right, I got a little little over-motivated: https://github.com/Apodini/Apodini/pull/143#issuecomment-761889533 😄

Sources/Apodini/Semantic Model Builder/Model/TypeReflection/EnrichedInfo.swift
lschlesinger: thanks, very good suggestions, solved via 978a99b

Tests/ApodiniTests/SemanticModel/EnrichedInfoTests.swift
lschlesinger: Thanks! This would entail to call `.edited(handleArray)` in between, or am I missing something? Otherwise it would always compare two default cardinality of `.exactlyOne`

nityanandaz: No, it wouldn't call `edited`, even though it was my intention, kind of. 😄
Anyways, my first thought was to avoid initializing `EnrichedInfo` manually.

However, now I am thinking of refactoring/removing this whole `edited` stuff and doing it in `EnrichedInfo.node`. Compare it to now, where we alter the `Tree` afterwards 😅

lschlesinger: I just refactored the code in e73f561 😄

Sources/Apodini/Semantic Model Builder/Model/TypeReflection/EnrichedInfo.swift
lschlesinger: Depending on whether you want to support proto3, I guess `maps` are the equivalent in protobuf? https://developers.google.com/protocol-buffers/docs/proto3#maps"

moritzschuell: @nityanandaz Sorry for the late reaction 🙃
No, as of now `ProtobufferCoding` does not support maps.

PR #141, opened 2021-01-17
Add (locally dispatched) invocable handlers

lukaskollmer:
A handler cannot easily access another handler's functionality

## :bulb: Proposed solution
This is the implementation of the first part of the API described in the [Inter-Component Communication vision document](https://github.com/Apodini/Apodini/blob/develop/Documentation/Components/Inter-Component%20Communication.md).

### Problem that is solved
We now have a simple API for invoking a handler from within another handler.

I'm not really happy with the naming here, feel free to suggest better typenames if you feel the same way.

### Testing
I have added unit tests for the new functionalities (handler invocation, array-based and custom-type-based parameter handling, default param value handling)

### Reviewer Nudging
The TestWebService contains two new handlers (another greeter and a random number generator) which call each other using the new API. This should be a good starting point


Sources/Apodini/RHI/RHIInterfaceExporter.swift
```diff
@@ -0,0 +1,100 @@
...
+}
+
+
+extension RHIInterfaceExporter {
+    // Used by the tests to get a new object for every test case.
+    // Ideally this function would be wrapped in some `#if TEST` condition, but that doesn't seem to be a thing
+    internal static func resetSingleton() {
+        Self.shared = nil
+    }
+}
```
awoc, line 90: I think you can also create this extension in your test file if it is only used for testing purposes

Sources/Apodini/Properties/Parameter.swift
```diff
@@ -37,6 +42,16 @@ public struct Parameter<Element: Codable>: Property {
         
         return element
     }
+    
+    
+    // This is here because, other than a `@Parameter` property itself (ie the storage), the projected value
+    // isn't private, meaning that we can use this to specify a parameter when remote-invoking a handler.
+    // We could alwo make the projectedValue return self, although it seems like a better idea to limit ourselves
```
awoc: Typo

Sources/Apodini/RHI/RHIInterfaceExporter.swift
lukaskollmer: The extension needs to be in the same file as the class definition because `shared`'s setter is private. If we move the extension somewhere else, the setter would need to be internal. The reason why this function exists (instead of just making the setter internal) is so that the singleton can, from outside the class, only be reset to `nil` but never set to a nonnil value (which would be possible if the setter isn't private anymore).

awoc: Thanks for clearing this up 👍

awoc (approved):

PSchmiedmayer:
@lukaskollmer A bit of spring / winter cleanup: I am closing this PR in favor of #246 which also includes this functionality. Feel free to remove the branch if it no longer needed 👍

PR #143, opened 2021-01-17
Remove Operation.automatic

bauer-andreas:
# Remove Operation.automatic

As of now, `Operation` defines a dedicated default case, namely `.automatic`.

## :bulb: Proposed solution

This PR removes the `.automatic` operation and moves the `OperationContextKey` to conform to the newly introduced `OptionalContextKey`.

We introduced the `.automatic` case with #17 in order to be able to distinguish between a explicitly set `.read` or `.read` as a default value. 

### Problem that is solved

Having `.automatic` being persisted in the model, and having the exporter decide on to what operation `.automatic` maps to, introduces conflict potential. For one you create cases where a DSL is valid with one exporter and invalid for the other (with that also the need for exporters to check for collisions is introduced).
But also, as I'm currently working on Relationship modeling, it creates complications where `.automatic` and `.read` are treated differently but in fact map to the same HTTP method (With latest changes on my local branch, a Relationship model not only points to a single Endpoint, but can point to multiple endpoint located under the same path but with different operations. For that situation we have the issue, that the Model itself would need to know to which operation `.automatic` maps to, but in fact doesn't).

If I remember correctly, we once talked about doing some sort of inference, when encountering `.automatic` (or rather a not explicitly set Operation). For example, a http exporter could map a Handler to `.create` if it contains `.content` parameters as `GET` typically doesn't have a body. However such an inference makes only sense in the scope of a particular exporter.

This PR makes such inference approaches currently impossible. If such inference is still wanted I would suggest to create an Issue for it, and then change the Model in such a way, that (1) the exporter can decide between `.read` and unset Operation again and (2) the Exporter can return the result of such an inference back to the model (This will be interesting on how to handle competing Exporters). 

### Related PRs
* #17 Introducing the `Operation` enum
* #xx _Once the Relationship draft PR becomes available_

### Testing
--

### Reviewer Nudging

Requesting review from @PSchmiedmayer as this is a more visionary decision about the DSL itself.


bauer-andreas:
As the total lines of numbers was reduced in the `RESTInterfaceExporter` the total coverage was reduced (by `-0.00%`).
So if we should merge that in the current state, we may need admin powers 🙃

lschlesinger:
> As the total lines of numbers was reduced in the `RESTInterfaceExporter` the total coverage was reduced (by `-0.00%`).
> So if we should merge that in the current state, we may need admin powers 🙃

I thought we tamed the codecov bot, to prevent something like this? IMO it does not make too much sense having coverage as the primary criterion for test suite effectiveness (i.e., as testing quality gate)... I personally caught myself optimizing for coverage the other day 😄

PSchmiedmayer:
@Supereg I would be fine with merging this PR and removing `.automatic` in favor of an optional value for now. Creating an issue sounds good to me. 

I would suggest to follow an approach where the semantic model tries to evaluate the possible `Operation` if no one has been passed from the syntax tree and all exporters follow this logic. This can either be done by 
1. using a shared logic inside the shared semantic model or
2. delegating this work to the exporters that have to reach a consensus on the operation type. 
 
I think going for the for the approach (1) in a first step would seem reasonable to me as this logic should be sharable between the exporters we currently have (e.g. based on Parameters and so on ...).

PSchmiedmayer (approved):
@Supereg I would be fine with merging this PR and removing `.automatic` in favor of an optional value for now. Creating an issue sounds good to me. 

I would suggest to follow an approach where the semantic model tries to evaluate the possible `Operation` if no one has been passed from the syntax tree and all exporters follow this logic. This can either be done by 
1. using a shared logic inside the shared semantic model or
2. delegating this work to the exporters that have to reach a consensus on the operation type. 
 
I think going for the for the approach (1) in a first step would seem reasonable to me as this logic should be sharable between the exporters we currently have (e.g. based on Parameters and so on ...).

PSchmiedmayer:
@lschlesinger @Supereg I will add a threshold for the overall code coverage while increasing the threshold for the diff code coverage which is more what we need. E.g. removing some lines in the project might decrease the code coverage while making sure that additions/changes are all covered is desirable.

PR #147, opened 2021-01-18
Incorporate Relationship Retrieval

bauer-andreas:
haha lol [meme](https://twitter.com/codewisdom/status/1353087614103023622)

# Incorporate Relationship Retrieval

At the moment we only have a very basic Relationship model, only retrieved by looking at the Component tree structure.
Additionally, with #104 I already incorporated the proposed PathComponent modifiers, although not yet used when creating Relationships.

## :bulb: Proposed solution

This PR tries to provide the full relationship retrieval process as proposed in #41.

### Problem that is solved

Provide the proposed mechanism for Relationship retrieval.

#### Refactorings

* The `WebServiceModel` was decoupled from the `SharedSemanticModeBuilder` and now has its own file inside the `Model` directory
* The `Endpoint` model and `EndpointsTreeNode` were fully decoupled and the `EndpointsTree` was moved into its own file in the `Semantic Model Builder` folder (This was kinda overdue, see #110, and as this PR added quite some code, SwiftLint force it to be split up with its max lines rule)
* Everything ExporterRequest (`ExporterRequest` protocol, `Reducible` protocol) got its own file inside `IntefaceExporter` directory.
* Instead of returning `AnyEncodable`, the `RequestHandler` returns now `HandledRequest` (wanted to name it Response but that collides with our Response enum, couldn't find something better), which is used to store context about injected parameters (for the given request) and store the actual response (like the literal AnyEncodable). This is needed to format e.g. the links section in the REST Exporter, where we need those informations to fill in path parameters which are known in that request context.
`HandledRequest` shares the same interface with `AnyEncodable`, making it fully transparent to other exporters not using Relationships (apart from type name).
* `ConntectionContext` was refactored to be a class. This was originally done to carry the information as stated above, which wasn't possible with the previous type erasure way. As I know found a different way to provide said information I wouldn't need that anymore. However I think this refactoring made using the `ConnectionContext` a bit simpler. Additionally this file was moved to `InterfaceExporter`, signifying that it is part of the Exporter API. (Seems like #183 move the ConnectionContext to a class as well).

### Mapping of the Proposal

Overview of the Relationship proposal and what is implemented or to be implemented (by chapter).
1. **Using the DSL "structure"**: ✅
   Collecting estrutural relationships is all done inside the `EndpointsTreeNode.constructStructuralRelationships()` and `constructStrutrualSelfRelationship()` functions. The `EndpointRelationship` model was modified to include information about the Operation of the relationship destination.
Additionally parsing of PathComponentModifiers (as described in chapter 3.) is done within those methods.
2. **Using Type Information**: ✅
    1. **Indexing Endpoints by their `Handler` return type**: ✅
    The return type of all available `Handler`s is captured within the `TypeIndexBuilder` (Note we only capture the ultimate response type, like the type of the last response transformer, or the Handler if none exist).
    With this information we can build entries in the `TypeIndex` for types which are unique, or can be considered a destination, and also collect candidates for "auto inheritance" as described in chapter 2.3 (implicit).
     2. **Using `.defaultRelationship` to resolve ambiguous return type information**: ✅
       Handling of this modifier is done inside the `TypeIndexBuilder`. Edge cases are properly handled.
     3. **Creating Relationships by referencing type information**: ✅
     The Relationship DSL is currently implemented using the `WithRelationships` protocol (not yet with the Metadata DSL).
     Everything regarding the Relationship DSL is placed within the `./Relationships` directory.
     Data gained through the Relationship DSL is collected in the `RelationshipsBuilder` (located in `SemanticModelBuilder/Relationship`).
     This information is then parsed using the fully assembled `TypeIndex`.
     Evaluating of this data is happening in `SharedSemanticModelBuilder.finishedRegistration()`.
3. **Manually defining Relationship Information**: ✅
    1. **Overriding the relationship name**: ✅
     The `.relationship(name:)` modifier is implemented and working.
      Parsing is done within the `PathModelBuilder` and utilized in `EndpointsTreeNode.constructStructuralRelationships()`.
    2. **Adding new relationships**: ✅
        1. **Providing type hints**: ✅
          This is parsed similarly to the RelationshipDSL
        2. **Creating `Relationship` instances**: ✅
           The usage of `.relationship(to:)` and `.destination(of:)` modifier with `Relationship` instances is fully supported.
           This information is collected in the `RelationshipBuilder`.
     3. **Hiding Relationships**: ✅
       The `.hideLink()` modifier is implemented and working. Parsing and handling same as for 3.1.
       As operations matter now for Relationships, and additional parameter was added to only hide relationships to destinations
       with a given Operation type.

### Additions to the Relationship DSL

* Some arguments and parameters were reorder or renamed. Look at the proposal diff.
* There is an added `RelationshipSource` (or type alias `Relationship` within the `WithRelationships` protocol) to add similar behavior like the `.relationship(name:to:)` modifier.
* We added a `identifying:` parameter to `PathParameter`, which is required when resolving path parameters with information gained through `Content` properties
* To not only support one KeyPath resolving a path parameter, the `RelationshipIdentification` with its `RelationshipIdentificationBuilder` was added. This function builder can be used for `Inherits`, `References` and `Relationship` definitions. See the example below:
 ```swift
struct SomeData: Content, WithRelationships {
  var someContent0: String
  var someContent1: String

  static var relationships: some Relationships {
    References<SomeContent>(as: "someContent") {
      Identifies<Param1>(at: \.someContent0)
      Identifies<Param2>(at: \.someContent1)
    }
  }
}

@PathParameter(identifies: Param1.self)
var param1: String
@PathParameter(identifies: Param2.self)
var param2: String

var content: some Component {
  Group($param1, $param2) {
    SomeHandler() // returning SomeContent
  }
  
}
```

### Related PRs

* #41: Introduced the original Relationship Retrieval PR
* #104: Introduced the first version of `PathComponentModifier`s

### Testing
_Extensive testing was added_

### Reviewer Nudging

I know that this PR kinda escalated it size. I thought about splitting it up, but not sure if that makes sense as everything kind of depends on each other in some or another way. When developing I couldn't really focus on one part of the proposal without to be required to think how this influences other parts.

If it helps we could schedule a meeting were we can try to step through this PR together.

Otherwise the code contains very detailed documentation. Additionally with the mapping above, describing which parts of the proposal map to which part in the code, I hope to help you a bit with the review.

### CODE REVIEW

**Comment by bauer-andreas** [comment]
The current PR has still an issue with name shadowing of Relationships.
To give some insights, our approach to name collision is the following:
explicit definition (`Relationship` instances, DSL) -> hides structural relationships -> hides inherited relationships.

The current PR has an issue where explicit relationships definitions are hidden by inherited relationships, if the inherited relationship is also a structural relationship of the Endpoint.

E.g. the following web service

```swift

struct TestData: Content, WithRelationships {
    static var relationships: Relationships {
        Inherits<Number>()
    }
}

var content: some Component {
    Group("dut") {
        TestDataHandler()
            .relationship(to: aRelationship)
        Number(int: 3)
            .operation(.update)
        Group("a") {
            Text("TextA")
        }
    }
    Group("other") {
        Text("TextOther")
            .destination(of: aRelationship)
    }
}
```

According to our rules the `"a"` relationship of `TestDataHandler` should point towards `/other`. Instead it points to `/dut/a`.

Issue is to be fixed.

bauer-andreas:
Addressed some code smells highlighted by SonarCloud.

* SonarCloud enforces max init parameter count of 7. Endpoint is an exception to this. To me it doesn't make sense to refactor this. [See](https://sonarcloud.io/project/issues?id=Apodini_Apodini&issues=AXcyv54RrWTuIHEk_c7v&open=AXcyv54RrWTuIHEk_c7v&pullRequest=147)
* Particularly this suggested [refactoring](https://sonarcloud.io/project/issues?id=Apodini_Apodini&issues=AXcyv55drWTuIHEk_c76&open=AXcyv55drWTuIHEk_c76&pullRequest=147) from switch to if statement seems impossible to me 🤔 Or at least wouldn't be possible with one if clause. 
* Additionally, SonarCloud seems to use the configuration of the master branch. Because the `EndpointIdentificationBuilder` function builder is not ignored. 

Other than that, everything smells fine.

Sources/Apodini/Modifier/OperationModifier.swift
```diff
@@ -15,6 +15,19 @@ public enum Operation {
...
+        case .create:
+            return "create"
+        case .read:
+            return "read"
+        case .update:
+            return "update"
+        case .delete:
+            return "delete"
+        }
+    }
```
nityanandaz:
```suggestion
    public var description: String {
        rawValue
    }
```

How about `String` for the `Operation.RawValue`? E.g. `Operation: String, ...`.

Sources/Apodini/Semantic Model Builder/REST/Operation+HTTPMethod.swift
```diff
@@ -0,0 +1,20 @@
+//
+// Created by Andreas Bauer on 22.01.21.
+//
+
+@_implementationOnly import Vapor
+
+extension Operation {
+    var httpMethod: Vapor.HTTPMethod {
```
nityanandaz: I would prefer this solution, again: https://github.com/Apodini/Apodini/pull/35#discussion_r553812093
If @lschlesinger hadn't changed their implementation, there would be now two `.httpMethod` extensions on Operation.

Sources/Apodini/Semantic Model Builder/Relationship/RelationshipBuilder.swift
```diff
@@ -0,0 +1,135 @@
...
+
+// MARK: RelationshipInstanceBuilder
+extension RelationshipInstance {
+    func build() {
+        validateDestinations()
+
+        precondition(!sources.isEmpty, """
+                                            Sources for the `RelationshipInstance` '\(name)' was empty! \
+                                            Please define at least one destination using the `relationship(to:)` modifier.
+                                            """)
```
nityanandaz: Indentation error

Sources/Apodini/Semantic Model Builder/Relationship/RelationshipBuilder.swift
```diff
@@ -0,0 +1,135 @@
...
+                let destinationEndpoint = destination.resolve()
+                relationship.addEndpoint(destinationEndpoint, name: name)
+            }
+
+            node.addRelationship(at: source, relationship)
+        }
+    }
+
+    private func validateDestinations() {
+        _ = destinations.reduce(into: Set<[EndpointPath]>()) { result, destination in
```
nityanandaz: You are discarding the result. A for-loop should be sufficient then 🙂

Sources/Apodini/Semantic Model Builder/Relationship/RelationshipCandidate.swift
```diff
@@ -0,0 +1,201 @@
...
+            return "link"
+        }
+    }
+
+    var debugDescription: String {
+        switch self {
+        case .inheritance:
+            return "inheritance"
+        case let .reference(name):
+            return "reference(name: \"\(name)\")"
```
nityanandaz:
```suggestion
            return #"reference(name: "\(name)")"#
```
They are _Extended String Delimiters_, if you didn't already know 🙂

Sources/Apodini/Semantic Model Builder/Relationship/TypeIndexBuilder.swift
```diff
@@ -0,0 +1,304 @@
...
+    init(_ reference: EndpointReference, type: Any.Type, markedDefault: Bool, pathParameters: [AnyEndpointPathParameter]) {
+        self.reference = reference
+        self.type = type
+        self.markedDefault = markedDefault
+        self.pathParameters = pathParameters
+    }
+}
+
+
+// <------------ Everything above has to to with capturing the type data. Everything below is actually building something with the data ------------>
```
nityanandaz: I highly recommend `// MARK: - ...`  comments. They appear in the Xcode Jump Bar and in the Minimap 🙂

Sources/Apodini/Semantic Model Builder/SemanticModelBuilder.swift
```diff
@@ -5,11 +5,15 @@
 //  Created by Paul Schmiedmayer on 11/3/20.
 //
 
+import Logging
+
 class SemanticModelBuilder {
     private(set) var app: Application
+    private(set) var logger: Logger
```
nityanandaz: I don't see a benefit and I could build the target `Apodini-Package` after removing the additions.

Sources/Apodini/Utilities/AnyEncodable.swift
```diff
@@ -24,9 +24,13 @@ public struct AnyEncodable: Encodable {
...
-        guard let anyEncoableWrappedValue = wrappedValue as? AnyEncodable else {
+        guard let anyEncodableWrappedValue = wrappedValue as? AnyEncodable else {
             return wrappedValue as? T
         }
-        return anyEncoableWrappedValue.typed(T.self)
+        return anyEncodableWrappedValue.typed(T.self)
+    }
+
+    internal func any() -> Any {
+        wrappedValue
```
nityanandaz: I think loosening the access level of `AnyEncodable.wrappedValue` would be a more honest approach. Or do we hide any meaningful information behind `Any`?

nityanandaz (dismissed):
meme👏review 😉

Sources/Apodini/Utilities/AnyEncodable.swift
bauer-andreas: Well I think I'm facing the same issue as Sadik, discussed in some previous meeting (https://github.com/Apodini/Apodini/pull/184#discussion_r564677759).

But maybe I can just used `typed(as: Encodable.self)` and then just throw away the type in the `HandledRequest` then (or keep, should also be fine).

EDIT: Nevermind `Value of protocol type 'Encodable' cannot conform to 'Encodable'; only struct/enum/class types can conform to protocols`

Tests/ApodiniTests/RelationshipTests/NameShadowingTests.swift
```diff
@@ -0,0 +1,62 @@
...
+            }
+        }
+        Group("other") {
+            Text("TextOther")
+                .destination(of: aRelationship)
+        }
+    }
+
+    func testInheritedRelationshipShadowing() {
+        return
```
awoc: Is this intentional?

Sources/Apodini/Properties/Options/PropertyOption.swift
```diff
@@ -3,7 +3,7 @@ import Foundation
 
 /// A type erasure of a `PropertyOption`
 public struct AnyPropertyOption<PropertyNameSpace> {
-    /// A type erasued `PropertyOptionKey` identfying the `AnyPropertyOption`
+    /// A type erasued `PropertyOptionKey` identifying the `AnyPropertyOption`
```
awoc: While you're at it 😄 
```suggestion
    /// A type erased `PropertyOptionKey` identifying the `AnyPropertyOption`
```

Sources/Apodini/Semantic Model Builder/InterfaceExporter/ConnectionContext.swift
```diff
@@ -0,0 +1,70 @@
+//
+//  ConnectionContext.swift
+//  
+//
+//  Created by Max Obermeier on 31.12.20.
+//
+
+import Foundation
+import Fluent
```
awoc: I think it is better to replace this import with SwiftNIO. You can probably also remove Foundation

awoc (commented):
I just left some smaller comments. Otherwise it looks really good to me!

Tests/ApodiniTests/RelationshipTests/NameShadowingTests.swift
bauer-andreas: Well yes, but actually no.
This is a regression test added for the mentioned issue https://github.com/Apodini/Apodini/pull/147#issuecomment-767087509.
The issue isn't yet fixed, and the file was committed by accident 😅

sonar-project.properties
```diff
@@ -17,6 +17,7 @@ sonar.links.homepage=https://github.com/Apodini/Apodini
 # Path is relative to the sonar-project.properties file.
 sonar.sources=Sources
 sonar.exclusions=Sources/Apodini/Components/ComponentBuilder.swift, \
+                 Sources/Apodini/RelationshipIdentificationBuilder.swift, \
```
awoc: I think you used the wrong file path

```suggestion
                 Sources/Apodini/Relationships/RelationshipIdentificationBuilder.swift, \
```

bauer-andreas:
The latest commit refactors the relationship building process a bit.
It moves lots of stuff of the building process out of the `Endpoint.swift` and `EndpointsTreeNode.swift` into the `RelationshipBuilder.swift`, creating better abstractions.
Amongst other this was to rewrite out the issue described in https://github.com/Apodini/Apodini/pull/147#issuecomment-767087509 which is now fixed (and added regression tests in `NameShadowingTests.swift`).

Sources/Apodini/Semantic Model Builder/Relationship/RelationshipBuilder.swift
bauer-andreas: Fixed within de45e10

Sources/Apodini/Semantic Model Builder/Relationship/RelationshipBuilder.swift
bauer-andreas: Fixed within de45e10

Sources/Apodini/Semantic Model Builder/Relationship/RelationshipCandidate.swift
bauer-andreas: Your example does not work as intended, as you seemingly also need to reescape the string Interpolation.
`#"reference(name: "\#(name)")"#`. Not sure what reads better.

Sources/Apodini/Semantic Model Builder/InterfaceExporter/ConnectionContext.swift
bauer-andreas: Seemingly some left overs. Good caught!

sonar-project.properties
bauer-andreas: Good caught. Resolved via 6c2b9bf

Sources/Apodini/Semantic Model Builder/InterfaceExporter/ConnectionContext.swift
bauer-andreas: Resolved via 6c2b9bf

Sources/Apodini/Properties/Options/PropertyOption.swift
bauer-andreas: Resolved via 6c2b9bf

Tests/ApodiniTests/RelationshipTests/NameShadowingTests.swift
bauer-andreas: Resolved via de45e10
See https://github.com/Apodini/Apodini/pull/147#issuecomment-768733721

Sources/Apodini/Utilities/AnyEncodable.swift
bauer-andreas: Resolved via 6c2b9bf

Sources/Apodini/Semantic Model Builder/SemanticModelBuilder.swift
bauer-andreas: Resolved via 6c2b9bf

Sources/Apodini/Semantic Model Builder/Relationship/TypeIndexBuilder.swift
bauer-andreas: Resolved via 6c2b9bf

Sources/Apodini/Semantic Model Builder/Relationship/RelationshipCandidate.swift
bauer-andreas: Resolved via 6c2b9bf

Sources/Apodini/Semantic Model Builder/REST/Operation+HTTPMethod.swift
bauer-andreas: Resolved via 6c2b9bf

Sources/Apodini/Modifier/OperationModifier.swift
bauer-andreas: Resolved via 6c2b9bf

bauer-andreas:
I have two conceptual/behavioral questions which might want to tackle:

1. Currently if `.hideLink()` is used, the `hideLink` property is set to true for the given `EndpointDestination`. As we first check for name collisions and then filter for hidden destinations (in the case of REST), we don't show the destination which is shadowed by the hidden destination (as they have the same name). Would it make sense for REST (as it is the only one respecting `hideLink`) that it stills shows any destinations which would normally be shadowed by another destination if `hideLink` would be ignored?
2. Do we want to inherit the `hideLink` property. Currently we do. If a endpoint A inherits relationships from another endpoint B, and those destinations are marked as hidden there, they will be marked hidden for A as well. Might there be a use case to not have them hidden?

CC: @PSchmiedmayer @nerdsupremacist

awoc (approved):
Looks good 👍

bauer-andreas:
> Looks good 👍

Thanks for working through the PR 👏

Documentation/Relationship/Relationship.md
```diff
@@ -275,7 +275,7 @@ struct Article: Identifiable, WithRelationships {
   var writtenBy: String
 
   static var relationships: some RelationshipDefinition {
-    References<User>(at: \.writtenBy, as: "author")
+    References<User>(as: "author", at: \.writtenBy)
```
PSchmiedmayer: When reading this I was interpreting the KeyPath as "identifying" the referenced `User` and therefore would consider:
```suggestion
    References<User>(as: "author", identifiedBy: \.writtenBy)
```
I think this somehow comes down to preference and I think both options are valid as the current description goes more in the direction of describing where in the response the relationship is listed (`at` the `writtenBy` property).

Documentation/Relationship/Relationship.md
```diff
@@ -463,7 +463,7 @@ struct TestService: WebService {
     Group {
       "user"
       $userId
-        .relationship(to: Greeting.self, name: "greeter")
+        .relationship(name: "greeter", to: Greeting.self)
```
PSchmiedmayer, line 466: I like the reordering of the parameters 🚀
While looking at the documentation here I was looking at the example below and noticed that I would have expected that the example output would have also appended the `userId` to the path in the links. As it is the same Id it could have also been replaced with 123, similar to how the `self` link is prefilled:
```json
{
  "id": "123",
  "name": "Rudi",
  "lastname": "Raser",
  "_links": {
    "self": "https://example.api/user/123",
    "greeter": "https://example.api/greeting/123"
  }
}
```
Would this be possible with the current model in the implementation? If not, I would consider it an future enhancement that we could document in the documentation and create an issue for it.

Sources/Apodini/Components/Component/BaseComponents/Group.swift
```diff
@@ -31,7 +31,9 @@ public struct Group<Content: Component>: Component, SyntaxTreeVisitable {
     func accept(_ visitor: SyntaxTreeVisitor) {
         visitor.enterContent {
             visitor.enterComponentContext {
-                visitor.addContext(PathComponentContextKey.self, value: pathComponents, scope: .environment)
+                var component = pathComponents
+                component.markEnd()
```
PSchmiedmayer, line 35: Nice idea of using the type information by wrapping path components 👍

Sources/Apodini/Configurations/Application/Application.swift
```diff
@@ -118,6 +118,9 @@ public final class Application {
         self.locks = .init()
         self.didShutdown = false
         self.logger = .init(label: "org.apodini.application")
+        #if DEBUG
+        self.logger.logLevel = .debug
```
PSchmiedmayer: Good addition 👍

Sources/Apodini/Modifier/DefaultRelationshipModifier.swift
```diff
@@ -0,0 +1,34 @@
+//
+// Created by Andreas Bauer on 16.01.21.
+//
+
+struct DefaultRelationshipContextKey: OptionalContextKey {
+    // We really only need to check if the context key exists on a Endpoint
+    // So the type is actually irrelevant, but Bool probably fits best?
+    typealias Value = Bool
```
PSchmiedmayer: I would propose to use `Void` as the `Value` as this indicates even more that there is no additional context carried with this `OptionalContextKey`:
```suggestion
    // This OptionalContextKey doesn't carry any additional context. 
    // We only need to check if the context key exists on a Endpoint
    typealias Value = Void
```
An other Apodini developer wound not be able to make an error and creating a `DefaultRelationshipContextKey` with the `value` being `false`, leading to an undefined behavior.
Checking if it is present is also a bit more descriptive:
```swift
let markedDefault = endpoint.context.get(valueFor: DefaultRelationshipContextKey.self) != nil
```

Sources/Apodini/Components/Path/HideLinkModifier.swift
```diff
@@ -2,33 +2,48 @@
...
-    }
+struct HideLinkContextKey: OptionalContextKey {
+    typealias Value = [Operation]
 }
 
 public struct HideLinkModifier: PathComponentModifier {
     let pathComponent: _PathComponent
+    // defines if only certain endpoints with certain Operations should be hidden
+    // empty array signals to hide everything
+    let operations: [Operation]
```
PSchmiedmayer, line 12: I would consider making usage of a Swift [`OptionSet`](https://developer.apple.com/documentation/swift/optionset) here. This might be a bit more descriptive instead of an empty array. It might also make your life a bit easier when checking it and there could be convince options like `mutating` (`create`, `update`, and `delete`) or something like this.

Sources/Apodini/Modifier/ResponseModifier/ResponseModifier.swift
```diff
@@ -41,3 +41,12 @@ extension ResponseModifier: SyntaxTreeVisitable {
         component.accept(visitor)
     }
 }
+
+extension Array where Element == LazyAnyResponseTransformer {
```
PSchmiedmayer, line 45: Good addition 👍

Sources/Apodini/Modifier/TypedRelationshipDestinationModifier.swift
```diff
@@ -0,0 +1,39 @@
+//
+// Created by Andreas Bauer on 20.01.21.
+//
+
+public struct TypedRelationshipDestinationModifier<H: Handler, To>: HandlerModifier {
```
PSchmiedmayer, line 5: Even though it is not needed here I would suggest to constraint `To` to `ResponseTransformable`:
```suggestion
public struct TypedRelationshipDestinationModifier<H: Handler, To: ResponseTransformable>: HandlerModifier {
```
Every type that is returned by a `Handler` must conform to the protocol and therefore it might give a developer an additional hint if the developer passes in a type that can not be returned by a `Handler`.

Sources/Apodini/Properties/Options/IdentifiyingType.swift
```diff
@@ -0,0 +1,30 @@
+//
```
PSchmiedmayer: Typo in the name of the Swift file: `IdentifiyingType.swift` -> `IdentifyingType.swift`

Sources/Apodini/Properties/Options/IdentifiyingType.swift
```diff
@@ -0,0 +1,30 @@
...
+public struct IdentifyingType: PropertyOption, Equatable {
+    /// The `Identifiable` type.
+    public let type: Any.Type
+    /// The `Identifiable.ID` type.
+    public let idType: Any.Type
+
+    public static func == (lhs: IdentifyingType, rhs: IdentifyingType) -> Bool {
+        lhs.type == rhs.type && lhs.idType == rhs.idType
+    }
+}
```
PSchmiedmayer: I would suggest to create an internal initializer that can only be created with an `Identifiable` type. Otherwise it would be hard to enforce that the `type` and `idType` match.
In addition I think that comparing the `type` should be sufficient for the `Equatable` implementation if we make sure that `idType` always matches the initial `Identifiable.ID` type.
```suggestion
public struct IdentifyingType: PropertyOption, Equatable {
    /// The `Identifiable` type.
    public let type: Any.Type
    /// The `Identifiable.ID` type.
    public let idType: Any.Type

    public static func == (lhs: IdentifyingType, rhs: IdentifyingType) -> Bool {
        lhs.type == rhs.type
    }
    
    init<Type: Identifiable>(identifying type: Type.Type = Type.self) {
        self.type = type
        self.idType = type.ID
    }
}
```

Sources/Apodini/Utilities/Optional.swift
```diff
@@ -26,3 +27,33 @@ extension Optional where Wrapped: ExpressibleByNilLiteral {
...
+    associatedtype Input = ApodiniOptional
+    associatedtype Output
+
+    func callAsFunction<T: ApodiniOptional>(_ value: T) -> Output
+}
+
+extension ApodiniOptionalVisitor {
+    @inline(never)
+    @_optimize(none)
+    func _test() { // swiftlint:disable:this identifier_name
```
PSchmiedmayer: You can make the function `fileprivate` as it is only used here and then you don't need the `// swiftlint:disable:this identifier_name` comment here.

TestWebService/Sources/TestWebService/UserSubsystem/Models/Post.swift
```diff
@@ -0,0 +1,11 @@
+//
+// Created by Andreas Bauer on 22.01.21.
+//
+
+import Foundation
+import Apodini
+
+struct Post: Content, Identifiable {
+    var id: UUID
```
PSchmiedmayer, line 9: It seems like an UUID triggers the error:
```
error org.apodini.application : Error(message: "Tuple: (UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8) is not supported")
```
When running the `TestWebService`. It seems to be related to the gRPC exporter group. @moritzschuell @nityanandaz and @Supereg do you have an idea where this might come from?

Tests/ApodiniTests/Helper/Miscellaneous/Response+forceUnwrap.swift
```diff
@@ -0,0 +1,17 @@
+//
```
PSchmiedmayer: This might be movable to XCTestApodini 👍 ... XCTest has an `XCTUnwrap` expression. We could overload it with our specific implementation for a parameter with the `Response` type.

Tests/ApodiniTests/Helper/Miscellaneous/TestingRelationshipFormatter.swift
```diff
@@ -0,0 +1,21 @@
+//
+// Created by Andreas Bauer on 23.01.21.
+//
+
+@testable import Apodini
+
+struct TestingRelationshipFormatter: RelationshipFormatter {
```
PSchmiedmayer: Similar to the other helpers this could also move to XCTApodini

Sources/Apodini/Relationships/RelationshipReference.swift
```diff
@@ -0,0 +1,59 @@
+//
+// Created by Andreas Bauer on 18.01.21.
+//
+
+/// A `RelationshipReference` can be used to create a referencing relationship for the annotated `Content` type.
+/// A relationship reference uses the value of properties holding the identifier of the target type
+/// too resolve a relationship with the given values.
```
PSchmiedmayer:
```suggestion
/// to resolve a relationship with the given values.
```

Sources/Apodini/Relationships/RelationshipIdentification.swift
```diff
@@ -0,0 +1,44 @@
...
+        resolver = identification.resolver()
+    }
+}
+
+/// A `RelationshipIdentification` provides additional information to resolve path parameter
+/// values to the destination of a relationship, e.g. defined by `RelationshipInheritance`, `RelationshipReference`
+/// or `RelationshipSource`.
+/// The path parameter which should be resolved must be annotated
+/// with a identifying type using `PathParameter.init(identifying:)`.
+public struct RelationshipIdentification<From, To: Identifiable> where To.ID: LosslessStringConvertible {
```
PSchmiedmayer, line 19: To be honest I was struggling with understanding the reason and purpose and had to go back to the documentation and PR description multiple times to understand `RelationshipIdentification`, `RelationshipInheritance`, `RelationshipReference` and `RelationshipSource`.
I think it would be great if the documentation added above the types would include small explanatory example when the type would be encountered and when/if a developer developing an Apodini web service would need to use the types.

E.g. it was sometimes hard figure out the different type aliases. E.g. for `RelationshipReference` a following example stating that `References` is a `typealias` of `RelationshipReference` and then explaining the example would be a great help.
```swift
struct User: Content, Identifiable, WithRelationships {
    var id: Int
    var writtenId = UUID()

    static var relationships: Relationships {
        References<Post>(as: "written", at: \.writtenId)
    }
}
```

Tests/ApodiniTests/RelationshipTests/RelationshipDSLTests.swift
```diff
@@ -0,0 +1,295 @@
...
+            Relationship(name: "TestC", of: TestC.self, parameter: \.cId)
+        }
+    }
+
+    struct AuthenticatedUser: Content, WithRelationships, Identifiable {
+        var id: Int
+        var secretName: String
+
+        static var relationships: Relationships {
+            Inherits<User>()
```
PSchmiedmayer, line 28: I like the concept of `Inherits` here, good example! I would suggest adding it to the documentation of the `RelationshipInheritance`. I found that reading through the test first was good to get a feeling of what must be possible. Adding this to the documentation as an example also helps to justify why this type is needed and why it is structured in this way.

Tests/ApodiniTests/RelationshipTests/NameShadowingTests.swift
```diff
@@ -0,0 +1,61 @@
...
+// Created by Andreas Bauer on 25.01.21.
+//
+
+import XCTest
+import XCTApodini
+@testable import Apodini
+
+class NameShadowingTests: ApodiniTests {
+    var aRelationship = Relationship(name: "a")
+    var bRelationship = Relationship(name: "b")
```
PSchmiedmayer: `bRelationship` doesn't seem to be used here. 
```suggestion
```

Tests/ApodiniTests/RelationshipTests/NameShadowingTests.swift
```diff
@@ -0,0 +1,61 @@
...
+                Text("TextA")
+            }
+        }
+        Group("other") {
+            Text("TextOther")
+                .destination(of: aRelationship)
+        }
+    }
+
+    func testInheritedRelationshipShadowing() {
```
PSchmiedmayer, line 49: I am not fully sure if I understand this test case. As there are no inherited relationships (`Int` doesn't seem to have any) there is nothing that could be shadowed here or am I missing something?

Sources/Apodini/Relationships/RelationshipInheritance.swift
```diff
@@ -0,0 +1,63 @@
...
+    }
+}
+
+extension RelationshipInheritance where To: Identifiable, To.ID: LosslessStringConvertible {
+    /// Creates a new `RelationshipInheritance`, inheriting from the specified type using the specified resolver.
+    /// Additionally it adds a specified resolver for a path parameter in the path of the destination.
+    /// - Parameters:
+    ///   - type: The type to inherit from.
+    ///   - keyPath: A resolver for a path parameter of the destination.
+    public init(from type: To.Type = To.self, at keyPath: KeyPath<From, To.ID>) {
```
PSchmiedmayer: Similar to the other initializers I would suggest to take a simple example from the tests and add it as a documentation here

Sources/Apodini/Relationships/RelationshipsVisitor.swift
```diff
@@ -0,0 +1,17 @@
+//
+// Created by Andreas Bauer on 18.01.21.
+//
+
+/// Ideally we would write a visitor using the
+/// AssociatedTypeRequirementsKit. As this has some issues right now
+/// this is a workaround for the workaround. Sadly required to be public.
+public protocol RelationshipVisitor {
```
PSchmiedmayer: @nerdsupremacist Do you have any input on this? 
@Supereg I would suggest to create an GitHub issue for this, we might be able to investigate and address this in a separate PR 👍

Sources/Apodini/Semantic Model Builder/InterfaceExporter/ConnectionContext.swift
```diff
@@ -0,0 +1,144 @@
...
+    ///   - exporterRequest: The exporter defined request to be handled.
+    ///   - final: True if this request is the last for the given Connection.
+    /// - Returns: The response for the given Request.
+    func handle(request: Exporter.ExporterRequest, final: Bool = true) -> EventLoopFuture<Response<HandledRequest>> {
+        handle(request: request, eventLoop: request.eventLoop, final: final)
+    }
+}
+
+
+class InternalConnectionContext<I: InterfaceExporter, H: Handler>: ConnectionContext<I> {
```
PSchmiedmayer: Very minor: The `InternalConnectionContext` has been around for quite some time but taking a look at its functionality where maybe naming it `HandlerSpecificConnnectionContext` would be a more appropriate name?

Sources/Apodini/Semantic Model Builder/InterfaceExporter/HandledRequest.swift
```diff
@@ -0,0 +1,106 @@
+//
+// Created by Andreas Bauer on 22.01.21.
+//
+
+import Foundation
+
+/// A HandledRequest describes the outcome of a `ConnectionContext.handle(...)`.
+public struct HandledRequest: Encodable {
```
PSchmiedmayer: If I understand this object correctly I would consider it more like a `ResponseFormatter` or some similar name. At first I was a bit confused about the `HandledRequest` name as I thought it was related to a request.

Sources/Apodini/Semantic Model Builder/InterfaceExporter/HandledRequest.swift
```diff
@@ -0,0 +1,106 @@
...
+            result = endpoint.selfRelationships().formatRelationships(into: result, with: formatter, context: context)
+        }
+
+        return result
+    }
+
+    /// Used to apply a `RelationshipFormatter` to THE self relationships of the `Endpoint`
+    /// with the context of this `HandledRequest` (path parameter values and property values of the response).
+    /// It is different to `formatSelfRelationships(into:with:for:)` that it used the one and only self
+    /// `RelationshipDestination` of the Endpoint (Using the Operation of the Endpoint).
```
PSchmiedmayer: I don't fully understand the statement you make in this sentence, maybe I am missing something there or can you clarify what you mean with this comment? In general I think it refers to the fact that `formatSelfRelationships(into:with:for:)` is used for everything except the self relationships (e.g. as GraphQL wouldn't need this step?).

Sources/Apodini/Semantic Model Builder/InterfaceExporter/RequestHandler.swift
```diff
@@ -32,16 +33,33 @@ class InternalEndpointRequestHandler<I: InterfaceExporter, H: Handler> {
...
                 }
             }
-            .flatMap { typedAction -> EventLoopFuture<Response<AnyEncodable>> in
-                self.transformResponse(typedAction.typeErasured, using: connection, on: request.eventLoop, using: self.endpoint.responseTransformers)
+            .flatMap { typedAction -> EventLoopFuture<Response<HandledRequest>> in
+                let transformed = self.transformResponse(
+                    typedAction.typeErasured,
+                    using: connection,
+                    on: request.eventLoop,
+                    using: self.instance.responseTransformers)
```
PSchmiedmayer: Minor: From a code formatting perspective I would move the parenthesis to the next line.
```suggestion
                let transformed = self.transformResponse(
                    typedAction.typeErasured,
                    using: connection,
                    on: request.eventLoop,
                    using: self.instance.responseTransformers
                )
```

Sources/Apodini/Semantic Model Builder/InterfaceExporter/RequestHandler.swift
```diff
@@ -32,16 +33,33 @@ class InternalEndpointRequestHandler<I: InterfaceExporter, H: Handler> {
...
+                }
             }
     }
 
+    private func mapToHandledRequest(_ response: Response<AnyEncodable>, validatedRequest: ValidatedRequest<I, H>) -> Response<HandledRequest> {
+        response.map { anyEncodable in
+            HandledRequest(
+                for: instance.endpoint,
+                response: anyEncodable,
+                parameters: validatedRequest.validatedParameterValues)
```
PSchmiedmayer:
```suggestion
                parameters: validatedRequest.validatedParameterValues
            )
```

Sources/Apodini/Semantic Model Builder/Model/Endpoint.swift
```diff
@@ -12,10 +12,8 @@ public protocol AnyEndpoint: CustomStringConvertible {
     /// Description of the `Handler` this endpoint was generated for
```
PSchmiedmayer, line 12: I like the refactoring into the `EndpointTreeNode` 🚀

Sources/Apodini/Semantic Model Builder/Model/EndpointPath.swift
```diff
@@ -98,8 +101,24 @@ public protocol AnyEndpointPathParameter: CustomStringConvertible {
...
     var description: String { get }
+    /// Defines the property type of the `PathParameter` declaration in a statically accessible way.
+    /// Be aware that for optional `PathParameter` this property holds the wrapped type of the `Optional`.
+    /// Use `PathBuilder` to access the `EndpointPathParameter` in a generic way.
+    ///
+    /// For the following declaration
+    /// ```
+    /// @PathParameter var name: String?
+    /// ```
+    /// this property holds `String.Type` and not `Optional<String>.self`.
```
PSchmiedmayer: As far as I understand from the refactoring made to `PathParameter` an optional value is no longer possible:
```swift
    /// Creates a new `@PathParameter`
    public init() {
        precondition(!isOptional(Element.self), "A `PathParameter` cannot annotate a property with Optional type!")
    }
```
This would eliminate the need for the comment here or am I missing something?

Sources/Apodini/Properties/PathParameter.swift
```diff
@@ -19,10 +22,21 @@ public struct PathParameter<Element: Codable & LosslessStringConvertible> {
...
+        precondition(!isOptional(Element.self), "A `PathParameter` cannot annotate a property with Optional type!")
+    }
+
+    /// Creates a new `@PathParameter` specifically stating the type it identifies.
+    /// The identified type must conform to `Identifiable` and the property type of the `PathParameter`
+    /// must match the type of the `id` property of the  identified type.
+    ///
+    /// - Parameter type: The type the PathParameter value identifies.
+    public init<Type: Encodable & Identifiable>(identifying type: Type.Type = Type.self) where Element == Type.ID {
+        self.identifyingType = IdentifyingType(type: type, idType: type.ID.self)
```
PSchmiedmayer: Should we also add a precondition check for the `Type` similar to `init()`? As far as I can see an optional type annotation also wouldn't make any sense here and would simplify the handling in the shared semantic modle.

Sources/Apodini/Semantic Model Builder/Model/WebServiceModel.swift
```diff
@@ -0,0 +1,70 @@
...
+
+    /// Retrieve the `Endpoint` located under the `EndpointPath.root`.
+    /// - Parameter operation: The `Operation` to retrieve the `Endpoint` for.
+    /// - Returns: The `AnyEndpoint` or nil if it doesn't exist.
+    public func getEndpoint(for operation: Operation) -> AnyEndpoint? {
+        rootEndpoints[operation]
+    }
+
+    /// Shortcut for calling `relationships(endpoint: .read, for: for)`
+    public func relationships(for operation: Operation) -> Set<RelationshipDestination> {
```
PSchmiedmayer: If this is called with a`.create` `Operation` this is not a shortcut for `relationships(endpoint: .read, for: for)`.

Sources/Apodini/Semantic Model Builder/Model/WebServiceModel.swift
```diff
@@ -0,0 +1,70 @@
...
+
+    /// Creates a `Set<RelationshipDestination` which ensures that relationship names
+    /// are unique (for all collected destination for: a given `Operation`).
+    /// The relationships are retrieved for a certain `Endpoint` defined by the given `Operation`.
+    ///
+    /// - Parameters:
+    ///   - endpoint: The `Operation` of the `Endpoint` to retrieve the Relationships from.
+    ///   - operation: The `Operation` of the Relationship destination to create a unique set for.
+    /// - Returns: The set of uniquely named relationship destinations.
+    public func relationships(endpoint: Operation, for operation: Operation) -> Set<RelationshipDestination> {
```
PSchmiedmayer: In general I am not really sure why this function takes two parameters. Why do go down a single level if there exists an endpoint for a single operation?
E.g. If we have an `Endpoint` for a `create` `Operation` we seem to only check in this subtree. Or am I missing something here?

Sources/Apodini/Semantic Model Builder/REST/RESTEndpointHandler.swift
```diff
@@ -2,79 +2,42 @@
...
+        let context = contextCreator()
 
-        var links = ["self": configuration.uriPrefix + endpoint.absolutePath.asPathString()]
-        for relationship in endpoint.relationships {
-            links[relationship.name] = configuration.uriPrefix + relationship.destinationPath.asPathString()
-        }
+        let responseFuture = context.handle(request: request)
 
-        return response.map { encodableAction in
+        return responseFuture.flatMap { (encodableAction: Response<HandledRequest>) in
```
PSchmiedmayer: I am not sure if I missed this in the code and it might also be out of scope for this PR: I don't see that the REST exporters handles `ApodiniError`s. We might want to create an issue for this.

Sources/Apodini/SyntaxTreeVisitor/SyntaxTreeVisitor.swift
```diff
@@ -91,6 +91,17 @@ class SyntaxTreeVisitor {
...
+        let responseTransformers = currentNode.getContextValue(for: ResponseTransformerContextKey.self)
+        let responseType = responseTransformers.responseType(for: H.self)
+
+        // Intermediate solution to parse `Content` types conforming to `WithRelationships`
+        // until the Metadata DSL creates a unified solution for such metadata.
+        if let relationships = responseType as? RelationshipVisitable.Type {
+            relationships.accept(self)
+        }
+        // let relationshipVisitor = StandardRelationshipsVisitor(visitor: self)
+        // relationshipVisitor(responseType)
```
PSchmiedmayer: Pease remove commented out code

Sources/Apodini/Semantic Model Builder/Relationship/PathParameterResolver.swift
```diff
@@ -0,0 +1,148 @@
...
+
+    init(destination type: Destination.Type = Destination.self, at keyPath: PartialKeyPath<Element>) {
+        self.identifyingType = IdentifyingType(type: type, idType: Destination.ID.self)
+        self.keyPath = keyPath
+    }
+
+    func resolves(parameter: AnyEndpointPathParameter) -> Bool {
+        // This line basically creates the restriction that we require the `identifyingType`
+        // to be supplied with the `@PathParameter` definition.
+        // As the lhs isn't Optional below check will always evaluate to false.
```
PSchmiedmayer: This line doesn't make sense to me, I am not sure what you want to express with this comment.

bauer-andreas:
thanks to the help of @nerdsupremacist with https://github.com/nerdsupremacist/AssociatedTypeRequirementsKit/commit/d861e6dd41ad0c1096a9a730611fa6a8cc307b0c, release TestWebService builds are now passing 🎉

PSchmiedmayer (approved):
@Supereg Thanks for adding the relationship information into Apodini! I am impressed by the amount of work and the attention to detail that went into the PR. 
Sorry that it took me so long to test everything and read the source code in its entirety. 
I had a few comments here and there while reviewing the PR and also had some questions to elements that I did not fully understand. But this shouldn't stop this PR from being merged and therefore you have my approval 🚀

Regarding the bigger review questions you have posed:
> Currently if .hideLink() is used, the hideLink property is set to true for the given EndpointDestination. As we first check for name collisions and then filter for hidden destinations (in the case of REST), we don't show the destination which is shadowed by the hidden destination (as they have the same name). Would it make sense for REST (as it is the only one respecting hideLink) that it stills shows any destinations which would normally be shadowed by another destination if hideLink would be ignored?

I would argue the shadowed destination should be shown when a `.hideLink()` is applied to a `Handler` that would shadow the relationship. It might be interesting to add a `ShowNoRelationshipFor(name: String)` or similar relationship information in the Relationship DSL / Metadata DSL in the future (as far as I can see we don't have anything like this for now). This could then be an "ultimate" hide indication for the Relationship algorithm.

> Do we want to inherit the hideLink property. Currently we do. If a endpoint A inherits relationships from another endpoint B, and those destinations are marked as hidden there, they will be marked hidden for A as well. Might there be a use case to not have them hidden?

I think it is fine that they are inherited. As far as I can see and understand endpoint B could manually add them again anyways couldn't it?

I hope I didn't miss any other open questions. If so let me know!

Sources/Apodini/Properties/Options/IdentifiyingType.swift
bauer-andreas: The amount of times I misspelled `Identifying` while writing this PR is uncountable 😅

Sources/Apodini/Utilities/Optional.swift
bauer-andreas: I think there was a problem, as the visitor protocol itself was not private. But moving that to private should to the trick 👍

Documentation/Relationship/Relationship.md
bauer-andreas: Agree, `identifiedBy:` is probably a bit more straightforward that the simple `at:`.

Documentation/Relationship/Relationship.md
bauer-andreas: I will revise this use case in more detailed again, but this should be supported out of the box.
The relationship modifier would create a `PartialRelationshipCandidate`. When it is later resolved (after the inheritance step; calling `ensureResolved` in the `TypeIndex`), resolvers for all path parameters (or property resolvers leading to the inherited `"self"`) are added for any explicit relationship definition.

The only thing what is not possible with `.relationship(name:to:)` is adding additional resolvers for additional path parameters. I had this in a prototype, but removed it again, since you could add such a modifier anywhere. This would be a problem, if you would add a resolver for a property (like what `References` does), but the property is then later eliminated by a response transformer.

TestWebService/Sources/TestWebService/UserSubsystem/Models/Post.swift
bauer-andreas: This error is already there for quite some time. afaik UUID is mapped to the said UInt8 tuple internally, which isn't supported.
Happy to move this out of my PR @nityanandaz  🙃

PSchmiedmayer: Oh, then I must have missed it before. Sure, then let's not address this in this PR. @nityanandaz @moritzschuell @Supereg can you create an issue for this or is this already addressed in some other PR that I am missing?

nityanandaz: I know this error. We will need to handle this separately

Documentation/Relationship/Relationship.md
PSchmiedmayer: > I will revise this use case in more detailed again, but this should be supported out of the box. 

Sounds great 🥳

> The only thing what is not possible with .relationship(name:to:) is adding additional resolvers for additional path parameters. I had this in a prototype, but removed it again, since you could add such a modifier anywhere. This would be a problem, if you would add a resolver for a property (like what References does), but the property is then later eliminated by a response transformer.

I am not 100% sure if I understood this correctly. You are saying that if the greeter would be located behind an additional path parameter it would be impossible to reference it would be impossible to reference it? Or am I on the wrong track here?

bauer-andreas: It is still possible to reference/link to it. We simply won't have a value for it when formatting. So we would just put `../{pathParamName}/..` into the `_links` section instead of a dedicated value (and the requester has to provide a value then if they want to create a request against that linked endpoint).
Does this bring clarification. Otherwise I can just state an extensive example.

nityanandaz (commented):
I reviewed the changes since I last viewed this PR, but I sadly don't feel comfortable making any comments that go beyond syntax (without investing considerably more time to understand the problem statement).

I think @PSchmiedmayer made good comments 👍

Documentation/Relationship/Relationship.md
bauer-andreas: To give some more context. 

`RelationshipSource` (or `Relationship` when in the `WithRelationships` namespace) is basically the DSL equivalent for the `.relationship(name:to:)` modifier. The `RelationshipSource` does indeed support additional PathParameterResolvers (e.g. using values from propreties). But the `RelationshipSource` also doesn't require that all path parameters to the destination are resolvable (=> meaning they appear as described above without a value in the `_links` section).

`RelationshipReference` (or `References`) on the other hand requires that all path parameters are resolved to the destination.

_Note: `RelationshipSource` is something not present in the original Relationship proposal, but though, exactly of those reasons presented, that it might be a good a idea to add a DSL equivalent for the `.relationship(...)` modifier._

Sources/Apodini/Modifier/TypedRelationshipDestinationModifier.swift
bauer-andreas: Not sure about that. As this would explicitly state that some type of Response<String> would be okay to pass. Even if this would be currently allowed, it is still semantically somewhat weird in my opinion.

Sources/Apodini/Properties/Options/IdentifiyingType.swift
bauer-andreas: I would aim to keep `idType` in the equality check. After all, this is what caught the issue with the AssociatedTypeRequirementsKit and its type visitor. But yes, the initializer is a good idea.

Tests/ApodiniTests/RelationshipTests/NameShadowingTests.swift
bauer-andreas: This is basically a regression test for the initial issue mentioned here https://github.com/Apodini/Apodini/pull/147#issuecomment-767087509.

`TestDataHandler` defines that it inherits from a Handler returning `Int` (as defined in `TestData` type). The `Number` handler returns `Int`. So `TestDataHandler` inherits from `Number` handler.
Additionally the `TestDataHandler` has a explicit `Relationship` to `/other` named `"a"`.

So: `TestDataHandler` has a structural relationship named `"a"` to `/dut/a`, it inherits the structural relationship named `"a"` pointing to `/dut/a` from `Number` (they share the same structural relationships) and has the manually defined `Relationship` to `/other` named `"a"`.

This Test case ensure that the explicit `Relationship` definitions is what is exposed, and isn't shadowed by the inherited `"a"` relationship, a very edge case scenario, which was previously an issue before the refactoring mentioned in the last work meeting.

Sources/Apodini/Semantic Model Builder/InterfaceExporter/HandledRequest.swift
bauer-andreas: This is essential a `AnyEncodable` (wrapping the Encodable response) with containing additional context of the handled request (specifically path parameters values and resolvers for relationships).
As its similarly structured to `AnyEncodable`, it can be transparently used by all exporters even if they don't care about relationships.
My initial thought was to name it like `Response` which obviously collided with our existing `Response` enum.
I'm not sure about `ResponseFormatter` as its not only about formatting, but the thing also holds the response itself 🤔

Sources/Apodini/Semantic Model Builder/InterfaceExporter/HandledRequest.swift
bauer-andreas: It tries to explain that it uses the self relationship which points to the Endpoint itself. So the operation of the relationship destination is the operation of the Endpoint. This relationship destination is guaranteed to always exists.
But of course if there are other handlers, there might be self relationships destinations for the other operations.
I try to clarify that a bit.

Sources/Apodini/Semantic Model Builder/Model/WebServiceModel.swift
bauer-andreas: Well you might want to select all `.read` Relationships of the root endpoint with the `.update` operation.

> E.g. If we have an Endpoint for a create Operation we seem to only check in this subtree. Or am I missing something here?

And yes you are right. Only if the Endpoint doesn't exists, we generate the default structural relationships, which are the same for all Endpoints under the same path.

Sources/Apodini/Semantic Model Builder/REST/RESTEndpointHandler.swift
bauer-andreas: I didn't quite follow the ApodiniError PR that closely. Is there additionally work required?
In any case it will be forwarded to Vapor creating some sort of error formatting, but not sure if this is the expected one.

Sources/Apodini/Semantic Model Builder/Relationship/PathParameterResolver.swift
bauer-andreas: The rest of the sentence is missing 😅 
It should continue as follows `As the lhs isn't Optional below check will always evaluate to false, should rhs be nil`.

The rhs might be nil when the user doesn't set the identifying type for a PathParameter. So this makes it the implicit requirement for `IdentifyingType` to exists on PathParameter which are expected to be resolved by properties.

bauer-andreas: The next sentence basically reiterates that, so I probably just remove that line.

Documentation/Relationship/Relationship.md
bauer-andreas: Solved via c25637b615c3d002f80f51d95d832e588483b6a3

Sources/Apodini/Modifier/DefaultRelationshipModifier.swift
bauer-andreas: Solved via 5ac7d7df3c2c1de6c3fd59031fa9f758d6822218

Sources/Apodini/Properties/Options/IdentifiyingType.swift
bauer-andreas: Solved via 5ac7d7df3c2c1de6c3fd59031fa9f758d6822218

Sources/Apodini/Properties/Options/IdentifiyingType.swift
bauer-andreas: Solved via 5ac7d7df3c2c1de6c3fd59031fa9f758d6822218

Sources/Apodini/Utilities/Optional.swift
bauer-andreas: Solved via 5ac7d7df3c2c1de6c3fd59031fa9f758d6822218

Sources/Apodini/Relationships/RelationshipReference.swift
bauer-andreas: Solved via 5ac7d7df3c2c1de6c3fd59031fa9f758d6822218

Tests/ApodiniTests/RelationshipTests/NameShadowingTests.swift
bauer-andreas: Solved via 5ac7d7df3c2c1de6c3fd59031fa9f758d6822218

Sources/Apodini/Semantic Model Builder/InterfaceExporter/ConnectionContext.swift
bauer-andreas: I went for `EndpointSpecificConnectionContext`, as the scope doesn't only exist of the `Handler`, but also guards and stuff (see `EndpointInstance`).

Solved via 5ac7d7df3c2c1de6c3fd59031fa9f758d6822218

Sources/Apodini/Semantic Model Builder/InterfaceExporter/HandledRequest.swift
bauer-andreas: Adjusted via 5ac7d7df3c2c1de6c3fd59031fa9f758d6822218

Sources/Apodini/Semantic Model Builder/InterfaceExporter/RequestHandler.swift
bauer-andreas: Solved via 5ac7d7df3c2c1de6c3fd59031fa9f758d6822218

Sources/Apodini/Semantic Model Builder/InterfaceExporter/RequestHandler.swift
bauer-andreas: Solved via 5ac7d7df3c2c1de6c3fd59031fa9f758d6822218

Sources/Apodini/Semantic Model Builder/Model/WebServiceModel.swift
bauer-andreas: Clarified the situation a bit with extended docs 👍 
Solved via 5ac7d7df3c2c1de6c3fd59031fa9f758d6822218

Sources/Apodini/SyntaxTreeVisitor/SyntaxTreeVisitor.swift
bauer-andreas: Solved via 5ac7d7df3c2c1de6c3fd59031fa9f758d6822218

Sources/Apodini/Semantic Model Builder/Relationship/PathParameterResolver.swift
bauer-andreas: Solved via 5ac7d7df3c2c1de6c3fd59031fa9f758d6822218

Sources/Apodini/Semantic Model Builder/REST/RESTEndpointHandler.swift
PSchmiedmayer: As far as there might be some small work that needs to be done to handle the error in the exporters. The exporter must conform to `StandardErrorCompliantExporter`. 

But this is definitely out of scope of the PR here. The WebSocket exporters showcases how a more custom error handling is implemented there 👍

bauer-andreas: Thanks for the clarification.
Yeah, would definitely put that into a separate PR 😅 
Maybe I can squeeze something in, next week. Otherwise it will be an Issue.

Sources/Apodini/Modifier/TypedRelationshipDestinationModifier.swift
nerdsupremacist: nit: Under what circumstances would we mark a relationship to an object that is not returned by a handler?
I think this is a fair constraint

Sources/Apodini/Properties/Parameter.swift
```diff
@@ -53,6 +53,10 @@ public struct Parameter<Element: Codable>: Property {
         self.defaultValue = defaultValue
         self.name = name
         self.options = PropertyOptionSet(options)
+
+        if option(for: PropertyOptionKey.http) == .path {
```
nerdsupremacist, line 57: great catch!

Sources/Apodini/Relationships/RelationshipIdentificationBuilder.swift.gyb
```diff
@@ -0,0 +1,26 @@
+// swiftlint:disable all
```
nerdsupremacist, line 1: gyb all the things!

Sources/Apodini/Relationships/RelationshipInheritance.swift
```diff
@@ -0,0 +1,63 @@
+//
+// Created by Andreas Bauer on 18.01.21.
+//
+
+/// A `RelationshipInheritance` can be used to indicate that the annotated `Content` type
+/// inherits Relationships from the specified target type.
+public struct RelationshipInheritance<From, To>: RelationshipDefinition {
```
nerdsupremacist, line 7: nit: Shouldn't `To` here also be constrained to `ResponseTransformable`?

Sources/Apodini/Relationships/RelationshipsVisitor.swift
nerdsupremacist: @Supereg I'm not sure I see the issue right now. Can you elaborate?

Sources/Apodini/Semantic Model Builder/InterfaceExporter/HandledRequest.swift
nerdsupremacist: nit: I'm also not super happy about the name here. I understand where it comes from though

Sources/Apodini/Semantic Model Builder/Model/EndpointPath.swift
```diff
@@ -108,6 +127,11 @@ public protocol AnyEndpointPathParameter: CustomStringConvertible {
     /// The name for the parameter. As PathParameter don't allow for custom name definitions
     /// this will always be the property name as defined in the `Handler`.
     var name: String { get }
+
+    /// The value for the given path parameter in the context of a specific request.
+    /// Nil if the path parameter does not have a resolved value.
+    /// See `resolved(value:)` for restrictions.
+    var erasuredResolvedValue: Any? { get }
```
nerdsupremacist: nit: `erasedResolvedValue`

Sources/Apodini/Semantic Model Builder/REST/RESTDefaultRootHandler.swift
```diff
@@ -0,0 +1,22 @@
...
+    let configuration: RESTConfiguration
+    let relationships: Set<RelationshipDestination>
+
+    // Registers a GET handler on root path
+    func register(on app: Vapor.Application) {
+        app.get(use: self.handleRequest)
+    }
+
+    func handleRequest(_: Vapor.Request) -> ResponseContainer {
+        let links = relationships.formatRelationships(into: [:], with: LinksFormatter(configuration: configuration))
```
nerdsupremacist, line 19: nit: Could we somehow configure the relationships formatter from outside? That would be really cool if people could inject their own RelationshipFormatter's

nerdsupremacist (approved):
overall lgtm

have a couple of nits but in my opinion they can be solved after merging

Sources/Apodini/Relationships/RelationshipsVisitor.swift
bauer-andreas: @nerdsupremacist The previous version was built around `AssociatedTypeRequirementsKit`.
See [here](https://github.com/Apodini/Apodini/blob/9e4ac526e1d69cab2dfc5c9e43825b8439e61773/Sources/Apodini/Relationships/RelationshipsVisitor.swift)

For me this was still resulting in a fatalError. They one about "function not found" (somewhat like that). It seemed to me that the _test workaround might just not work for type visitors, but after all it works for the already packaged type visitors. So not sure what happened here.

bauer-andreas: Just comparing to the AssocKit sources, why are there double `.self` 😅 
https://github.com/nerdsupremacist/AssociatedTypeRequirementsKit/blob/d861e6dd41ad0c1096a9a730611fa6a8cc307b0c/Sources/AssociatedTypeRequirementsVisitor/visitors.swift#L837

Sources/Apodini/Semantic Model Builder/InterfaceExporter/HandledRequest.swift
bauer-andreas: Any other suggestions maybe?

EDIT: Like I'm okay with `ResponseFormatter`, just think it might not be the most transparent one for other exporters.

Sources/Apodini/Components/Path/HideLinkModifier.swift
bauer-andreas: the comment is actually outdated.
Empty array will lead to nothing being hidden.
So I don't think I need `OptionSet` here.

Sources/Apodini/Semantic Model Builder/InterfaceExporter/HandledRequest.swift
PSchmiedmayer: As discussed in the meeting: `EnrichedContent` 🚀

Tests/ApodiniTests/RelationshipTests/NameShadowingTests.swift
PSchmiedmayer: Thanks for the explanation, makes sense!

Sources/Apodini/Modifier/TypedRelationshipDestinationModifier.swift
bauer-andreas: As discussed in the work meeting: We will leave it as is.

Sources/Apodini/Semantic Model Builder/REST/RESTDefaultRootHandler.swift
bauer-andreas: Definitively. But would call this out of scope of this PR.
I will create an issue for it?

Sources/Apodini/Semantic Model Builder/InterfaceExporter/HandledRequest.swift
bauer-andreas: As of the work meeting, we decided for `EnrichedResponse`

Sources/Apodini/Relationships/RelationshipInheritance.swift
bauer-andreas: Same as in https://github.com/Apodini/Apodini/pull/147#discussion_r568601289

Documentation/Relationship/Relationship.md
PSchmiedmayer: Thanks for the detailed explanation, now I understand 👍

Sources/Apodini/Modifier/TypedRelationshipDestinationModifier.swift
PSchmiedmayer: As discussed in the meeting we will keep the `To` generic type unconstrained.

Sources/Apodini/Semantic Model Builder/Model/WebServiceModel.swift
PSchmiedmayer: As discussed in the meeting: I have understood the functionality of the function and @Supereg will update the documentation based on my feedback.

Sources/Apodini/Properties/PathParameter.swift
bauer-andreas: Added a call to `self.init()` running the same check for all initializers.
Anyways this would also have been caught in the `Parameter` precondition check.
But fixed with upcoming commit.

Sources/Apodini/Relationships/RelationshipsVisitor.swift
bauer-andreas: solved by @nerdsupremacist in 9e38a34

Sources/Apodini/Semantic Model Builder/Model/EndpointPath.swift
bauer-andreas: Seems like some leftovers.

Sources/Apodini/Relationships/RelationshipsVisitor.swift
nerdsupremacist: One `.self` wasn't enough

Sources/Apodini/Semantic Model Builder/Model/EndpointPath.swift
bauer-andreas: Resolved in f82004a

Tests/ApodiniTests/Helper/Miscellaneous/Response+forceUnwrap.swift
bauer-andreas: Resolved in f82004a

Tests/ApodiniTests/Helper/Miscellaneous/TestingRelationshipFormatter.swift
bauer-andreas: Resolved in f82004a

Sources/Apodini/Properties/PathParameter.swift
bauer-andreas: Resolved in f82004a

Sources/Apodini/Semantic Model Builder/Model/WebServiceModel.swift
bauer-andreas: Resolved in f82004a

Sources/Apodini/Semantic Model Builder/Model/EndpointPath.swift
bauer-andreas: Resolved via 908d44c

Sources/Apodini/Relationships/RelationshipIdentification.swift
bauer-andreas: Examples were added with 799f331. Hope that's as expected.

Sources/Apodini/Relationships/RelationshipInheritance.swift
bauer-andreas: Resolved via 799f331

PR #148, opened 2021-01-18
Add properties of Application to `@Environment`

awoc:
# Add properties of `Application` to `@Environment`

Idea from @nerdsupremacist 

Many properties of `Application` should also be accessible in a `Handler`, e.g., the current database, apns and many more. Currently every property of `@Environment` needs to implement an `EnvironmentKey` and extend `EnvironmentValues` to be used with `@Environment`. This PR would add every property of `Application` to `@Environment` and remove the need to define special `EnvironmentValues` and `EnvironmentKey`s.

## :bulb: Proposed solution

Adapt the implementation `@Environment` to also accept key paths of `Application` properties. 

### Implications

- Shouldn't have any implication on the current implementation and usage of `@Environment`.
- Restricts the access level of `apns` and `fcm` in the `NotificationCenter` which can now be injected directly with `@Environment`.

### Testing

- Includes simple test for injection

### Reviewer Nudging

It would be good to check if no important properties are exposed by these changes.


nerdsupremacist (approved):
LGTM

Sources/Apodini/Environment/Environment.swift
```diff
@@ -54,6 +62,11 @@ public struct Environment<K: KeyChain, Value>: Property {
         self.keyPath = keyPath
     }
     
+    /// Initializer of `Environment` specifically for `Application` for less verbose syntax.
+    public init(_ keyPath: KeyPath<K, Value>) where K == Application {
```
nerdsupremacist, line 66: nit: For a future version of this I think we could make Application work identically to Environment. Because if I understand this PR correctly the key paths are different, right?

nerdsupremacist (approved):

Issue #146, opened 2021-01-18
[MINOR FEATURE] OAS support for enum types


lschlesinger:
**Is your feature request related to a problem? Please describe.**
As of now, enum types are not reflected properly in the OAS. 

**Propose a solution**
In order to make OAS describe enums properly, the supporting data structure of reflecting a response type recursively, i.e. `Node<EnrichedInfo>` has to be adapted likewise. This issue should be tackled by a PR, that:

1. Provides a function: `handleEnum()` to enable proper reflection and preprocessing of enum type in a `Node<EnrichedInfo>` as part of the data structure provided in `Shared Semantic Model/Model/TypeReflection`

2. Provides the identification of enum types within the `OpenAPIComponentsObjectBuilder` to properly build the respective JSON Schema: 

The result of a handler like this:
```swift
struct Status: Handler {
    enum Color: String, Codable, ResponseTransformable {
        case delivered, ordered, ready
    }
    func handle() -> Color {
        .ordered
    }
}
```
should be:
```json
...
"responses": {
  "200": {
    "content": {
      "application/json": {
        "schema": {
          "type": "string",
          "enum": [
            "delivered",
            "ordered",
            "ready"
            ]
          }
        }
      }
    }
  }
...
```

**Additional context**
Find more information on enumerations on sagger in the [documentation](https://swagger.io/docs/specification/data-models/enums/)

This issue is related to #64, i.e., providing `handleEnum()` for   `Node<EnrichedInfo>` will also be a prerequisite for this issue.

lschlesinger:
closed by #149

PR #149, opened 2021-01-18
OAS enumeration support

lschlesinger:
# OAS enumeration support

The current implementation of `OpenAPIExporter` was not able to include enums properly in the schema. 

## :bulb: Proposed solution
Implement logic to handle `Node<EnrichedInfo>` resulting from reflecting an enum type in the `OpenAPIComponentsObjectBuilder`.

### Problem that is solved
Enums are represented properly as expected by the Open API Specification.
This PR addresses #146.

The result of a handler such as:
```swift
struct Status: Handler {
    enum Color: String, Codable, ResponseTransformable {
        case delivered, ordered, ready
    }
    func handle() -> Color {
        .ordered
    }
}
```
is now:
```json
...
"responses": {
  "200": {
    "content": {
      "text/plain": {
        "schema": {
          "type": "string",
          "enum": [
            "delivered",
            "ordered",
            "ready"
            ]
          }
        }
      }
    }
  }
...
```

### Implications

- New type check (`isEnum`) based on reflected type in `Utilities/TypeInfo`

The representation as enum types in the OAS described [here](https://swagger.io/docs/specification/data-models/enums/) in the official documentation.

### Related PRs
none.

### Testing
Added a test in `TypeInfoTest`.
Checked for the correct schema creation in `OpenAPIComponentsObjectBuilderTests`

### Reviewer Nudging
Not much changed, easy to review 🦑


Tests/ApodiniTests/Utilitites/TypeInfoTests.swift
```diff
@@ -26,6 +25,24 @@ class TypeInfoTests: ApodiniTests {
...
+    func testIsEnum() {
+        /// A custom type
+        enum Test {
+            case unit
+            case integration
+            case system
+        }
+
+        XCTAssertEqual(isEnum(Test.self), true)
+        XCTAssertEqual(isEnum(Int.self), false)
```
awoc, line 38: You could also use the abbreviations `XCTAssertTrue` or `XCTAssertFalse`

awoc (approved):
Looks good 👍

Tests/ApodiniTests/Utilitites/TypeInfoTests.swift
lschlesinger: thanks! I sticked to the pattern inside this test file, as it appears to sometimes be more readable without the short-hand syntax. :)

PR #150, opened 2021-01-18
Protobuffer generation: support Int and UInt

### LINKED ISSUE #140: Protobuffer: Support `Int` and `UInt`

nityanandaz:
## Problem that is solved

- Closes #140
- Create `IntegerWidthConfiguration` and `IntegerWidthCodingStrategy` to set a width preference for `Int` and `UInt`.

### Related PRs

- [x] Merge #131 first, because I restructured the tests there.
- [x] Merge #159, then configure `ProtobufferDecoder` and `ProtobufferEncoder`

### Testing

- [x] I will add tests that depend on the `Int.bitWidth`

### Reviewer Nudging

- [x] I am open for a better name for `isAmbiguousSupportedFixedWidthInteger`


Sources/Apodini/Utilities/TypeInfo.swift
```diff
@@ -50,3 +52,14 @@ func isSupportedScalarType(_ type: Any.Type) -> Bool {
     supportedScalarTypes
         .contains(ObjectIdentifier(type))
 }
+
+func isAmbiguousSupportedFixedWidthInteger(_ type: Any.Type) -> Bool {
```
PSchmiedmayer: Maybe `isSupportedVariableWidthInteger` would be a more descriptive name?

Sources/Apodini/Semantic Model Builder/gRPC/ProtobufferBuilder/Extensions/Model+EnrichedInfo.swift
```diff
@@ -35,20 +54,40 @@ extension ProtobufferMessage.Property {
...
-            return nil
+fileprivate extension TypeInfo {
+    func compatibleName() throws -> String {
+        if isSupportedScalarType(type) {
+            let infix = isAmbiguousSupportedFixedWidthInteger(type)
+                ? String(describing: Int.bitWidth)
+                : ""
+            let suffix = isSupportedScalarType(type)
+                ? ""
+                : "Message"
```
PSchmiedmayer: I would suggest adding if statements here to make the code a bit easier to read.

PSchmiedmayer (commented):
From a first glance those changes look good to me, I only had two small comments.

In addition to use `.bitWidth` to generate the correct protobuf description I would suggest to add some sort of configuration to the exporter (or Apodini in general) that would allow a developer to explicitly specify what type of `bitWidth` the developer would suspect to have when running the web service. 
If the `bitWidth` is different from what would be expected by the developer we should either fail at startup or at least log a logging error that the expected `bitWidth` of an `Int`/`UInt` differs from the expected platform as defined by the configuration.

PSchmiedmayer (approved):
Thanks for incorporating the feedback from the last review 🚀

Sources/Apodini/Configurations/VariableWidthIntegerConfiguration.swift
```diff
@@ -0,0 +1,37 @@
+//
+//  Created by Nityananda on 28.01.21.
+//
+
+public struct VariableWidthIntegerConfiguration: Configuration {
+    // MARK: Nested Types
+    public enum Option: Int {
```
moritzschuell: nit: Maybe a name more precise than `Option` would be better? Something like `IntWidthOption` 🤔

Sources/Apodini/Semantic Model Builder/Protobuffer/ProtobufferInterfaceExporter.swift
```diff
@@ -10,7 +10,9 @@ class ProtobufferInterfaceExporter: StaticInterfaceExporter {
         let message: String
     }
 
-    internal enum Builder {}
+    enum Builder {
+        static var variableWidthIntegerPreference: Int?
```
moritzschuell: Could we reuse `Option` (or `IntWidthOption` 😁) instead of the `Int` type? To enforce that only valid values (32 and 64) can be set?

Sources/Apodini/Semantic Model Builder/gRPC/GRPCService/GRPCService.swift
```diff
@@ -95,7 +99,17 @@ extension GRPCService {
     /// Encodes the given encodable value
     /// to  `Data` using Protobuffer encoding
     private func encode(_ value: Encodable) throws -> Data {
-        let message = try ProtobufferEncoder().encode(AnyEncodable(value))
+        let encoder = ProtobufferEncoder()
+        if let configuration = app.storage[VariableWidthIntegerConfiguration.Key.self] {
+            let variableWidthIntegerStragegy: VariableWidthIntegerStrategy?
```
moritzschuell:
```suggestion
            let variableWidthIntegerStrategy: VariableWidthIntegerStrategy?
```

Sources/ProtobufferCoding/Encoder/ProtoEncodingContainer.swift
```diff
@@ -108,6 +108,16 @@ internal class InternalProtoEncodingContainer {
...
+                try encodeInt32(Int32(value), tag: tag)
+            case .sixtyFour:
+                try encodeInt64(Int64(value), tag: tag)
+            }
+            return
+        }
+        
         if MemoryLayout<Int>.size == 4 {
             try encodeInt32(Int32(value), tag: tag)
         } else if MemoryLayout<Int>.size == 8 {
```
moritzschuell: Maybe instead of `encoder.variableWidthIntegerStrategy` being optional and keeping those `MemoryLayout.size` checks, we could make `encoder.variableWidthIntegerStrategy` non-optional and set it to the system's architecture-width by default in the constructor of the encoder?

moritzschuell (commented):
Great work @nityanandaz 🙌
Thanks for also adding the configuration functionality to the Proto en-/decoder!

I only left some small comments, once they are addressed I think this can be merged.

Sources/Apodini/Configurations/VariableWidthIntegerConfiguration.swift
```diff
@@ -0,0 +1,38 @@
...
+    
+    // MARK: Nested Types
+    enum Key: StorageKey {
+        typealias Value = VariableWidthIntegerConfiguration
+    }
+    
+    // MARK: Methods
+    public func configure(_ app: Application) {
+        guard rawValue <= Int.bitWidth else {
+            preconditionFailure(
```
lschlesinger: This [helper for XCTAssertRuntimeFailure](https://github.com/Apodini/Apodini/blob/develop/Sources/XCTApodini/XCTAssertRuntimeFailure.swift) should help testing this precondition...

lschlesinger (approved):
I could not find any potential for code improvements. Just a hint for testing the precondition, if you weren't already aware of this. Regarding the coverage in general, I would guess it is not too hard to add tests cases (similar to `testVariableWidthIntegerConfiguration()`) with varying configurations in combination with `Int`, `UInt`, `[Int]`etc.? Maybe I am missing the difficulties in doing so 😄 .
 One comment regarding the variable naming, I think I would prefer `IntWidthConfiguration` instead of `VariableWidthIntegerConfiguration` this would then also be consistent with @moritzschuell 's [suggestion](https://github.com/Apodini/Apodini/pull/150/#discussion_r568047310) to rename `Option` to `IntWidthOption`, wouldn't it? I think it's easier to read as it is used very often and still captures the essence.

Sources/Apodini/Configurations/VariableWidthIntegerConfiguration.swift
nityanandaz: Thanks for pointing out the function. I can use it, but it sadly wouldn't be of any use in the near future, if I haven't missed something:

1. `XCTAssertRuntimeFailure` depends on `CwlPreconditionTesting`
2. `CwlPreconditionTesting` depends on `platforms: [.macOS]` in `Package.swift`
3. Macs will most likely never be 32-bit again.

Therefore, we cannot trigger the failure in a test case 😅

lschlesinger: Haha 😆  I didn't really consider what is being checked here, my comment was only triggered by the uncovered precondition 🤦🏽‍♀️

Sources/ProtobufferCoding/Encoder/ProtoEncodingContainer.swift
nityanandaz: I do that now: 40da4b12a554d66823168ba28484d7f3c668ce7f

Sources/Apodini/Configurations/VariableWidthIntegerConfiguration.swift
nityanandaz: The final names are `IntegerWidthConfiguration` and `IntegerWidthCodingStrategy` 👍

moritzschuell (approved):
Thank you for implementing the suggested changes.
I have no more comments 😁

PR #153, opened 2021-01-18
Add HTTP Configuration

### LINKED ISSUE #137: [BUG] Command-line args cause app to crash
**Describe the bug**
Previously one could provide command-line arguments to the app.
E.g. for the certificate used by the `HTTP2Configuration`: `--cert="/some/path/cert.pem"`.

When doing this on the current `develop`, this results in:
```
Fatal error: Error raised at top level: .unknownCommand("--cert=/some/path/cert.pem", available: ["boot", "serve", "routes"]): file Swift/ErrorType.swift, line 200
```

I think this changed behavior was introduced with the decoupling of `Vapor.Application`.

**To Reproduce**
Start the app with an additional command-line argument, e.g. `--cert="/some/path/cert.pem"`.

**Expected behavior**
Custom command-line arguments should be forwarded, as previously.

**Possible solution**
As of now, I think only the `HTTP2Configuration` is actually offering the possibility to use command-line arguments to configure it (with certificates in this case).
So I see two possibilities:
- Remove the possibility to use command-line args from `HTTP2Configuration`, since it also offers the possibility to configure the certificates in code. However, this would make it necessary to recompile the server each time the user might want to change certifcate 🙁
- Restore previous behavior by adding command-line arg support to our `Apodini.Application`.

tgymnich:
# Add HTTP Configuration

Currently it is not possible to set the port and host of the http server.

## :bulb: Proposed solution
This PR adds `HTTPConfiguration`, which is able to set the port and host using the DSL and command line arguments.

### Problem that is solved
Serverless runtimes can configure Apodini from the "outside".

### Implications
For the port and host command line options to apply the HTTPConfiguration needs to be added to `Webservices.configuration`.

### Related
fixes #137 

### Reviewer Nudging
its best to start with `HTTPConfiguration.swift`


Tests/ApodiniTests/ConfigurationTests/DatabaseConfigurationTests.swift
```diff
@@ -1,10 +1,3 @@
-//
-//  DatabaseConfigurationTests.swift
-//  
-//
-//  Created by Tim Gymnich on 13.1.21.
-//
-
```
theMomax, line 7: Why would you delete those?

Sources/Apodini/Configurations/HTTPConfiguration.swift
```diff
@@ -0,0 +1,81 @@
...
+
+/// A `Configuration` for HTTP.
+/// The configuration can be done in two ways, either via the
+/// command line arguments --hostname, --port and --bind or via the
+/// function `address`
+public class HTTPConfiguration: Configuration {
+    private var address: BindAddress?
+
+    enum HTTPConfigurationError: Error {
+        case incompatibleFlags
```
theMomax, line 20: Please add an extension so that throwing this error results in a nice error message.

Sources/Apodini/Configurations/Application/Application+Vapor.swift
```diff
@@ -13,15 +13,21 @@ extension Vapor.Application {
         var app: Vapor.Application
 
         func didBoot(_ application: Application) throws {
-            try app.start()
+            if let address = application.http.address {
+                try app.server.start(address: Vapor.BindAddress(from: address))
+            } else {
+                try app.server.start()
+            }
+            try app.boot()
```
theMomax, line 21: Are you sure this is correct? Executing `app.boot()` in a function called `didBoot` just smells like there is something wrong. I haven't looked into the details of Vapor boot-up processes, though.

Tests/ApodiniTests/ConfigurationTests/HTTPConfigurationTests.swift
```diff
@@ -0,0 +1,46 @@
+@testable import Apodini
+import XCTest
+
+final class HTTPConfigurationTests: ApodiniTests {
```
theMomax, line 4: I think it would make sense to have a tests that checks if command line arguments can overwrite in-code configurations.

theMomax (commented):
Looks good! I would just like to see one more test case where you explicitly test overwriting in-code parameters with command line arguments.

Sources/Apodini/Configurations/Application/Application+Vapor.swift
tgymnich: Did boot relates to the Apodini application. So we boot up the vapor application after the Apodini application did boot.

Tests/ApodiniTests/ConfigurationTests/DatabaseConfigurationTests.swift
tgymnich: for consistency with the other tests

Sources/Apodini/Configurations/HTTPConfiguration.swift
tgymnich: what do you mean with create an extension. Is there a specific protocol I should implement?

Tests/ApodiniTests/ConfigurationTests/HTTPConfigurationTests.swift
tgymnich: good point! done.

Sources/Apodini/Configurations/Application/Application+Vapor.swift
theMomax: Ah, I see...thanks for the explanation.

Sources/Apodini/Configurations/HTTPConfiguration.swift
theMomax: I think the best one would be LocalizedError.

tgymnich: done

Sources/Apodini/Configurations/HTTP2Configuration.swift
```diff
@@ -27,20 +28,37 @@ public class HTTP2Configuration: Configuration {
...
+            var keyPath: String?
+        }
+
+        let signature = try Signature(from: &commandInput)
+
+        if let certPath = signature.certPath, let keyPath = signature.keyPath {
+            certData = try Data(contentsOf: URL(fileURLWithPath: certPath))
+            keyData = try Data(contentsOf: URL(fileURLWithPath: keyPath))
+        } else {
+            print("Cannot read certificate / key file provided via command line")
```
theMomax: I found another little one: Please use the logger instead of print (might be more cases than just this one).

tgymnich: At this point the logger / Apodini app is not yet available. But I found some other instances of print and fixed those. Using the Environment at this point feels wrong.

tgymnich: I made the errors fatalErrors.

theMomax (approved):
LGTM!

PR #155, opened 2021-01-18
Treat Warnings as Errors

tgymnich:
just testing for now...


tgymnich:
We cannot get rid of all warnings at the moment.

PR #156, opened 2021-01-19
ObservedObjects for `Job`s

awoc:
# ObservedObjects for `Job`s

The implementation for `@ObservedObject` and `@Published` is missing.

## :bulb: Proposed solution

`@Published` will execute a closure to signal changes of the object. By collecting every `@Published` property of `ObservableObject`s Apodini can subscribe to changes and trigger the evaluation of a `Job`. `@ObservedObject` includes a property called `changed` which can be used to check if the evaluation was triggered by an `ObservableObject`.

Example web service:

```swift
struct SomeJob: Job {
        @ObservedObject(\Keys.job2) var job2: AnotherJob
        
        func run() {
            // `_` used to access to changed property of `@ObservedObject`
            if _job2.changed {
                print("Someone else triggered me \(job2.num).")
            } else {
                print("I was normally scheduled.")
            }
        }
    }

    class AnotherJob: Job {
        @Published var num = 0
        
        func run() {
            num += 1
        }
    }

    struct Keys: KeyChain {
        var job1: SomeJob
        var job2: AnotherJob
    }

        // Executed once at an even minute
        Schedule(SomeJob(), on: "*/2 * * * *", runs: 1, \Keys.job1)
        // Executed every minute
        Schedule(AnotherJob(), on: "* * * * *", \Keys.job2)
    }
```

### Implications

~~- Exposes `OpenCombine` to the Apodini web service~~
-> Uses closure approach now, removes the need for OpenCombine

- `Handler` includes a method to collect every `ObservedObject` declared in it
- The insertion of the Job EnvironmentValue was moved to the initializer
- The property wrapper check in `Scheduler.checkPropertyWrappers` is not working as intended and was changed-> Will be addressed in new PR

### Testing

- Includes basic tests for `ObservedObject`in Apodini
- Includes tests for uses cases of `ObservedObject` with `Job`s

### Reviewer Nudging

A good starting point is to take a look at the property wrappers `@Published` and `@ObservedObject`. After that the `ObservedObjectTests` class in the `JobTests` target offers examples how to use the property wrappers.


moritzschuell (approved):
Looks great 👍 Thanks for implementing this.

Just for my understanding as to how exporters can use this: 
A `ConnectionContext` could now register a callback for `valueDidChange` in each `ObservedObject` of the `Handler`, to be able to get notified for changes on these objects?

awoc:
> Looks great 👍 Thanks for implementing this.
> 
> Just for my understanding as to how exporters can use this:
> A `ConnectionContext` could now register a callback for `valueDidChange` in each `ObservedObject` of the `Handler`, to be able to get notified for changes on these objects?

Yes, that is exactly the idea.

Documentation/AdditionalFunctionality/Jobs.md
```diff
@@ -117,7 +117,7 @@ Besides emitting events, `Job`s can also listen to changes of other [ObservedObj
...
+The _RegisterHandler_ is responsible to count every access to its path and store this information in _VisitorObservedObject_. Using `Environment(value, keyPath)` in the `configuration` stored property, which takes as arguments a key path and the corresponding value, we can declare the _VisitorObservedObject_ as a Singleton. This allows us to use the `ObservableObject` in the web service with the property wrappers `@Environment` and `@ObservedObject` and the key path. The `Job` will send a weekly push notification based on the total number of visitors and also when someone accessed the _RegisterHandler_ by listening to changes to the _VisitorObservedObject_. To differentiate between these two cases we use the property `changed` of `@Published`  to check if the `Job` was triggered by being scheduled or by the _VisitorObservedObject_.
```
theMomax: I think this should be "the property `changed` of `ObservedObject`" to be consistent with your implementation, shouldn't it?

Sources/Apodini/Properties/ObservedObject.swift
```diff
@@ -0,0 +1,101 @@
...
+            fatalError("The object \(String(describing: self)) cannot be found in the environment.")
+        }
+        set {
+            element = newValue
+        }
+    }
+    
+    /// Property to check if the evaluation of the `Handler` or `Job` was triggered by this `ObservableObject`.
+    /// Read only property
+    public var changed: Bool {
```
theMomax, line 33: Just a suggestion...

```suggestion
public enum ObservedObjectState: Bool {
    case changed = true
    case unchanged = false
}

    public var projectedValue: ObservedObjectState {
```
One could then write just `$someObservedObject` instead of `_someObservedObject.changed` with the returned type still making clear what the value means.

I'm not sure if you could copy this code 1:1, but I think you get the gist. 

This is just some syntactic sugar and it totally depends on taste, so you can also just leave it as is :).

Sources/Apodini/Properties/ObservableObject.swift
```diff
@@ -0,0 +1,6 @@
+/// Objects that notify subscribers when `@Published` properties change.
+///
+/// Refer to the documentation of
+/// [ObservedObject](https://github.com/Apodini/Apodini/blob/develop/Documentation/Communicational%20Patterns/2.%20Tooling/2.4.%20ObservedObject.md)
```
theMomax: Please add more docs here in code. I don't think we should cross reference our textual documentation like this as such references will break most likely in the future.

Sources/Apodini/Properties/ObservedObject.swift
```diff
@@ -0,0 +1,101 @@
+import Foundation
+
+/// Property wrapper that subscribes to an `ObservableObject` and evaluates a `Handler` or `Job` on changes.
+///
+/// Refer to the documentation of
+/// [ObservedObject](https://github.com/Apodini/Apodini/blob/develop/Documentation/Communicational%20Patterns/2.%20Tooling/2.4.%20ObservedObject.md)
```
theMomax: Same as for the other doc-reference. Please add more information to the code at least if you still want to include the link.

Documentation/AdditionalFunctionality/Jobs.md
awoc: Good catch. Thank you for pointing this out!

Sources/Apodini/Properties/ObservedObject.swift
awoc: I like the use `projectedValue` and think that it is more swift-like than using `_`. However, I think when just using `$someObservedObject` the return type might be a bit confusing because it is not directly object related and returns something else. For this reason I think `_someObservedObject.changed` better serves the purpose.

Sources/Apodini/Properties/ObservableObject.swift
awoc: Changed the documentation

Sources/Apodini/Properties/ObservedObject.swift
awoc: Changed the documentation

Issue #152, opened 2021-01-18
[BUG] Test Coverage Varies Between Builds


PSchmiedmayer:
**Describe the bug**
The test coverage of [`Sources/Apodini/Semantic Model Builder/Model/EndpointPath.swift`](https://github.com/Apodini/Apodini/blob/develop/Sources/Apodini/Semantic%20Model%20Builder/Model/EndpointPath.swift#L72) seems to fluctuate between builds. Sometimes line 72 (`return false`) of the `EndpointPath: Equatable` extension is not covered and executed when running the test suite.
```swift
extension EndpointPath: Equatable {
    static func == (lhs: EndpointPath, rhs: EndpointPath) -> Bool {
        switch (lhs, rhs) {
        case (.root, .root):
            return true
        case let (.string(lhsPath), .string(rhsPath)):
            return lhsPath == rhsPath
        case let (.parameter(lhsParameter), .parameter(rhsParameter)):
            return lhsParameter.id == rhsParameter.id
        default:
            return false
        }
    }
}
```
The default case seems to be executed sometimes while sometimes the test coverage and a breakpoint at the line seems to indicate that the line is not executed on some test runs.

**To Reproduce**

1. Set a breakpoint on the default case of the `EndpointPath: Equatable` extension.
2. Run the test suite multiple times.
3. Sometimes the line is executed, sometimes it is not executed.

**Expected behavior**
The test coverage shouldn't change between executions of the test suite and the behavior of the Apodini code should be deterministic.

**Environment**

 - Swift Version: 5.3
 - Xcode Version: 12.3 (12C33)

lschlesinger:
The test with non-deterministic behavior seems to be [`SharedSemanticModelBuilderTests.testEndpointsTreeNodes`](https://github.com/Apodini/Apodini/blob/develop/Tests/ApodiniTests/SemanticModel/SharedSemanticModelBuilderTests.swift#L106).
I think the problem is in [this line](https://github.com/Apodini/Apodini/blob/develop/Sources/Apodini/Semantic%20Model%20Builder/Model/Endpoint.swift#L235).

My guess would be that this is due to the dictionary (which is inherently an unordered collection) that we use for [`nodeChildren: [EndpointPath: EndpointsTreeNode]` in `EndpointsTreeNode`](https://github.com/Apodini/Apodini/blob/develop/Sources/Apodini/Semantic%20Model%20Builder/Model/Endpoint.swift#L165).
[...]

PR #157, opened 2021-01-19
[BUG FIX] Add test case to stop flaky test coverage

lschlesinger:
# Add test case to avoid flaky test coverage

We had a bug producing flaky test coverage results.

## :bulb: Proposed solution
As proposed in https://github.com/Apodini/Apodini/issues/152#issuecomment-762516823, we would add a test case that ensures to cover all cases in [`EndpointPath: Equatable`](https://github.com/Apodini/Apodini/blob/62dbacbb64be9f3782637d94cab7840a943d9aea/Sources/Apodini/Semantic%20Model%20Builder/Model/EndpointPath.swift#L62) as done by:

`testEndpointPathEquatable()` in `EndpointsTreeTests`

### Problem that is solved
#152 

### Implications
We can expect stable test coverage results.

### Reviewer Nudging
Questions would be:

1. Is this the right test class to add the test?
2. Is is too exhaustive to test all cases, even though we know, that only the default case was not deterministically executed?

### DIFF
--- a/Tests/ApodiniTests/SemanticModel/EndpointsTreeTests.swift
+++ b/Tests/ApodiniTests/SemanticModel/EndpointsTreeTests.swift
@@ -7,6 +7,7 @@
 
 @testable import Apodini
 import XCTest
+import Foundation
 
 
 final class EndpointsTreeTests: ApodiniTests {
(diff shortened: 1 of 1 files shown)


bauer-andreas (approved):

PSchmiedmayer (approved):
Looks good to me, thanks a lot!

Issue #95, opened 2021-01-03
[BUG] Protobuffer generation: type hierarchy Array<Optional<T>> throws error


nityanandaz:
**Describe the bug**
`Optional` is handled before `Array` and never after.

https://github.com/Apodini/Apodini/blob/bf59a8d120936c1735b2e580baeb4b62ca822604/Sources/ProtobufferBuilder/ProtobufferBuilder.swift#L72

**To Reproduce**
```swift
func testOptionalArray() throws {
    struct Message {
        let value: [Bool]?
    }
    
    XCTAssertNoThrow(try buildMessage(Message.self))
}

func testArrayOfOptional() throws {
    struct Message {
        let value: [Bool?]
    }
    
    XCTAssertThrowsError(try buildMessage(Message.self))
}
```

**Expected behavior**
🤔 Protobuffer does not naturally support "Optional".

**Additional context**
- https://github.com/Apodini/Apodini/pull/33#discussion_r550999658

nityanandaz:
`Optional<Array<T>>` is problematic, too. @lschlesinger 

Ref.: https://github.com/Apodini/Apodini/blob/6f05dd21cbee1721139edbbd90ccd12aeba9e71a/Sources/Apodini/Semantic%20Model%20Builder/OpenAPI/Document/OpenAPIComponentsObjectBuilder.swift#L58

nityanandaz:
@lschlesinger regarding your comment: I disagree that `[Cardinality]` would be suitable with the current cases. E.g., `[.exactlyOne, .zeroToMany]` cannot be valid.
This would rule out an [OptionSet](https://developer.apple.com/documentation/swift/optionset) as well.

nityanandaz:
@moritzschuell how would `ProtobufferCoding` handle `Array<Optional<T>>` or `Optional<Array<T>>`, respectively?

`Array<Optional<T>>`: I would assume that nil values are omitted in the payload.
`Optional<Array<T>>`: I would drop the optional. Thus, `Optional<Array<T>>.none` and an empty array (`[]`) will be the same.

moritzschuell:
@nityanandaz Sorry for the late reaction. This got lost among all the notifications 🙁

As you maybe noticed, the approach in #159 is exactly as you suggested:
`Array<Optional<T>>` only encodes the non-nil values in the array
`Optional<T>` (including `Optional<Array<T>>`) only encodes something if the optional is non-nil

PR #159, opened 2021-01-19
Add support for Int, UInt, and Array<Optional<T>> to Protobuffer encoding/decoding

moritzschuell:
# Add support for `Int`, `UInt`, and `Array<Optional<T>>` for Protobuffer encoding/decoding

Addresses these issues from the Protobuffer encoding / decoding perspective: 
#95: Support for `Optional<T>` and `Array<Optional<T>>`
#140: Support for `Int` and `UInt` (this PR is the Protobuffer encoding / decoding counterpart for the PR #150 by @nityanandaz )

## :bulb: Proposed solution
This PR adds support for two more types and allows to encode and decode optionals properly.

### Optionals
Optionals are only encoded if they hold a value, `nil` will not be encoded.
`Array<Optional<T>>` behaves similarly, and only encodes the non-nil elements in the array.
E.g. `[1, 2, nil, 4, 5]` yields the same encoded data as `[1, 2, 4, 5]`.

### `Int` and `UInt`
Similarly to #150, the encoder / decoder will behave differently depending on the systems architecture.
The bit-width of the data type (`Int` and `UInt`) is checked, and either the 32-bit or 64-bit fixed-width type is used internally by the encoder / decoder.

### Implications
The addition of `Int` and `UInt` causes the server to behave differently on different architectures. As a consequence, when moving the Apodini app from a 32-bit to a 64-bit machine (and vice-versa), Apodini will decode `Int` and `UInt` differently (and also generate different `proto`-files as of #150).
However, the user can still "force" consistent behavior across architectures by simply only using fixed-width `Int`s in her code.

### Related Issues and PRs
#95: Support for `Optional<T>` and `Array<Optional<T>>`
#140: Support for `Int` and `UInt`
#150: Proto-IDL generation support for `Int` and `UInt`

### Testing
Tests are added for encoding / decoding 
- positive & negative `Int`s
- `UInt`s
- optional single values
- arrays of optional values
- 
(optionals are only tested for a few types, not for every single available type)

### Reviewer Nudging
The PR contains only incremental changes, adding a few more types and encode / decode methods.
Should be pretty straight-forward to review, as no (complex) logic is changed or added.


moritzschuell:
@PSchmiedmayer I added quite a few test cases to this, however coverage is still far off the target. 
Would this be OK for your anyways, or do you have a suggestion on further useful test-cases that I can add?

PSchmiedmayer:
@moritzschuell I took a look at the current test coverage and the unit tests you have added and would be fine with the coverage. 
It would be great if the reviewers would take a closer look at edge cases that we might want to add but I don't see a reason to add more redundant tests as most of the uncovered code follows a similar semantic as the code that is already tested.

Once you have a review from members of the @Apodini/apodini team I can merge the PR using the "admin powers" I have even if it might not hit the diff or the overall target 👍

theMomax (commented):
The code looks pretty reasonable from my point of view.

I have a remark about your encoding strategy though (just know, I don't have any experience with protobuf, so maybe what I'm saying doesn't make any sense 😄):

* About size of Uint/Int: I think just using the server's system default is not ideal (but also not a big deal-breaker as most servers are 64bit anyways). Could you imagine providing a configuration that allows for fixing this to one or the other value for all systems?
* About encoding of `Array<T?>`: With your solution: can the client even know how long the array was originally? And if so: can the client know what elements were `nil`? E.g. if I wanted to send the measurements for each day of the month, but I assume measurements can be missing, I could derive from what month the data is for, how many days there should be. However, if one measurement is missing, I would never know what day this missing measurement belongs to, right?

nityanandaz:
@theMomax Thanks for the suggestion to provide configuration options. For reference, @PSchmiedmayer made this comment on my PR, too: https://github.com/Apodini/Apodini/pull/150#pullrequestreview-570818892

Sources/ProtobufferCoding/Decoder/KeyedProtoDecodingContainer.swift
```diff
@@ -7,6 +7,7 @@
 
 import Foundation
 
+// swiftlint:disable discouraged_optional_boolean
```
nityanandaz: Could you place the comment closer to the lines that it affects?

Sources/ProtobufferCoding/Decoder/KeyedProtoDecodingContainer.swift
```diff
@@ -93,7 +94,15 @@ class KeyedProtoDecodingContainer<Key: CodingKey>: InternalProtoDecodingContaine
     }
 
     func decode(_ type: Int.Type, forKey key: Key) throws -> Int {
-        throw ProtoError.decodingError("Int not supported, use Int32 or Int64")
+        let keyValue = try extractIntValue(from: key)
+        if let value = data[keyValue]?.last {
+            if MemoryLayout<Int>.size == 4 {
```
nityanandaz: I guess this works as well. I think that `Int.bitWidth == 32` shows just a little bit more clearly why we test the condition 🙂

Sources/ProtobufferCoding/Decoder/KeyedProtoDecodingContainer.swift
nityanandaz: Same for the other `...Container`s

Sources/ProtobufferCoding/TypeInfo.swift
```diff
@@ -0,0 +1,88 @@
...
+    [UInt?].self,
+    [UInt32?].self,
+    [UInt64?].self,
+    [Double?].self,
+    [Float?].self,
+    [Data?].self
+]
+
+internal func isPrimitiveSupported(_ type: Any.Type) -> Bool {
+    primitiveSupportedTypes.contains(where: { $0 == type })
```
nityanandaz, line 60: This looks very much like `Apodini/Utilities/TypeInfo.swift`. Sadly, an import would create cyclic dependencies 😕

nityanandaz (changes requested):

Sources/ProtobufferCoding/Decoder/KeyedProtoDecodingContainer.swift
moritzschuell: I would prefer not to use `Int.bitWidth`.
`if Int.bitWidth == 32 { ... }` produces a compiler warning `Will never be executed` on 64-bit machines (and vice versa).
Doing it via `MemoryLayout<Int>` does not create this warning.

Sources/ProtobufferCoding/TypeInfo.swift
moritzschuell: Yes, you're right. Parts of the code is even simply copied from `Apodini/Utilities/TypeInfo.swift`. But as you said, using `TypeInfo` from Apodini target does not work.

Sources/ProtobufferCoding/Decoder/KeyedProtoDecodingContainer.swift
nityanandaz: I didn't know that 👍

Sources/ProtobufferCoding/Decoder/KeyedProtoDecodingContainer.swift
moritzschuell: Good point, thanks 👍 I moved the comments closer to the affected lines.

Tests/ProtobufferCodingTests/Encoding/ComplexEncodingTests.swift
```diff
@@ -0,0 +1,97 @@
...
+    let expectedComplexMessage = Data([
+                            8, 199, 159, 255, 255, 255, 255, 255, 255, 255,
+                            1, 16, 185, 96, 32, 1, 40, 2, 65, 88, 168, 53,
+                            205, 143, 28, 200, 64, 74, 11, 72, 101, 108, 108,
+                            111, 32, 87, 111, 114, 108, 100, 82, 6, 1, 2, 3,
+                            253, 254, 255, 90, 36, 10, 34, 72, 97, 108, 108, 111,
+                            44, 32, 100, 97, 115, 32, 105, 115, 116, 32, 101,
+                            105, 110, 101, 32, 83, 117, 98, 45, 78, 97, 99, 104,
+                            114, 105, 99, 104, 116, 46, 117, 126, 228, 64, 70
+    ])
```
nityanandaz:
```suggestion
    let expectedComplexMessage = Data([
        8, 199, 159, 255, 255, 255, 255, 255, 255, 255,
        1, 16, 185, 96, 32, 1, 40, 2, 65, 88, 168, 53,
        205, 143, 28, 200, 64, 74, 11, 72, 101, 108, 108,
        111, 32, 87, 111, 114, 108, 100, 82, 6, 1, 2, 3,
        253, 254, 255, 90, 36, 10, 34, 72, 97, 108, 108, 111,
        44, 32, 100, 97, 115, 32, 105, 115, 116, 32, 101,
        105, 110, 101, 32, 83, 117, 98, 45, 78, 97, 99, 104,
        114, 105, 99, 104, 116, 46, 117, 126, 228, 64, 70
    ])
```

Tests/ProtobufferCodingTests/Encoding/OptionalEncodingTests.swift
```diff
@@ -0,0 +1,124 @@
...
+import XCTest
+@testable import ProtobufferCoding
+
+class OptionalEncodingTests: XCTestCase {
+    func testEncodeSinglePositiveOptionalInt32() throws {
+        let expected = Data([8, 185, 96])
+        let number: Int32? = 12345
+
+        let encoded = try ProtobufferEncoder().encode(number)
+        XCTAssertEqual(encoded, expected, "testEncodeSinglePositiveOptionalInt32")
```
nityanandaz: Are these messages beneficial? 
- `swift test`'s output is already grouped by test case and test method,
- and you would always need to update them if you change the method name 🙂

Tests/ProtobufferCodingTests/Encoding/OptionalEncodingTests.swift
```diff
@@ -0,0 +1,124 @@
...
+        let encoded = try ProtobufferEncoder().encode(message)
+        XCTAssertEqual(encoded, expected, "testEncodeRepeatedOptionalInt32Message")
+    }
+
+    func testEncodeRepeatedOptionalFloatMessage() throws {
+        let expected = Data([
+                                10, 16, 250, 62, 246, 66, 207, 119,
+                                246, 66, 121, 233,
+                                246, 66, 78, 34, 247, 66
+        ])
```
nityanandaz:
```suggestion
        let expected = Data([
            10, 16, 250, 62, 246, 66, 207, 119,
            246, 66, 121, 233,
            246, 66, 78, 34, 247, 66
        ])
```

Tests/ProtobufferCodingTests/Encoding/OptionalEncodingTests.swift
```diff
@@ -0,0 +1,124 @@
...
+        let encoded = try ProtobufferEncoder().encode(message)
+        XCTAssertEqual(encoded, expected, "testEncodeRepeatedOptionalFloatMessage")
+    }
+
+    func testEncodeRepeatedOptionalDoubleMessage() throws {
+        let expected = Data([
+                                10, 16, 117, 107, 126, 84, 52, 111,
+                                157, 65, 219, 209, 228, 84, 52, 111,
+                                157, 65
+        ])
```
nityanandaz:
```suggestion
        let expected = Data([
            10, 16, 117, 107, 126, 84, 52, 111,
            157, 65, 219, 209, 228, 84, 52, 111,
            157, 65
        ])
```

nityanandaz (approved):

moritzschuell:
@theMomax I agree with you that allowing the user to configure the bit-size to use is a good idea. As @nityanandaz suggested in #150, we can probably use the configuration of the Protobuffer IDL exporter.

Regarding the second point, I am not 100% sure. To my current understanding, `repeated` optional fields are actually not supported by the current Protobuffer version. Further, `repeated` fields are usually encoded with "implicit presence", meaning they are simply not in the encoded byte-stream rather than being explicitly encoded as "not present".

In my opinion the chosen approach is the most suitable way to bring together the Swift and Proto worlds, even though it's not an ideal solution.
Would you be OK with merging this version?

theMomax:
> Would you be OK with merging this version?
@moritzschuell yes! As I mentioned, I have no experience with protobuf at all, so I just pointed out things that did not fit in with my intuition. If you think this is the best way I'm fine with merging. I just wanted to make these points weren't treated carelessly.

Sources/ProtobufferCoding/Decoder/KeyedProtoDecodingContainer.swift
```diff
@@ -93,7 +93,15 @@ class KeyedProtoDecodingContainer<Key: CodingKey>: InternalProtoDecodingContaine
     }
 
     func decode(_ type: Int.Type, forKey key: Key) throws -> Int {
-        throw ProtoError.decodingError("Int not supported, use Int32 or Int64")
+        let keyValue = try extractIntValue(from: key)
```
PSchmiedmayer: I know that I am a bit late to the party here but given the topic of the PR I was a bit confused about the name of this function. First I was expecting that it actually extracting an int value but it extracts a key that is an integer (as defined by the protobuf spec to uniquely identify fields in a message). Maybe renaming the function and clarifying the functionality in the comment of the function would be beneficial for a first time reader of the code. E.g. `convertKeyToProtobufIntIndex`, `extractProtocolbufferIntKey`,  `extractProtobufIntIndex` or something similar?

Sources/ProtobufferCoding/Decoder/KeyedProtoDecodingContainer.swift
```diff
@@ -93,7 +93,15 @@ class KeyedProtoDecodingContainer<Key: CodingKey>: InternalProtoDecodingContaine
...
+        if let value = data[keyValue]?.last {
+            if MemoryLayout<Int>.size == 4 {
+                return try Int(decodeInt32(value))
+            } else if MemoryLayout<Int>.size == 8 {
+                return try Int(decodeInt64(value))
+            }
+        }
+        throw ProtoError.decodingError("No data for given key")
```
PSchmiedmayer: This code snipped is used quite often as part of the PR. If possible, I would suggest to move it to a throwable initializer of the type `Int` and reuse that initializer throughout the PR? The same applies to the same `if else` constructs that are used throughout the PR for other `Int`ish types.

Sources/ProtobufferCoding/Decoder/UnkeyedProtoDecodingContainer.swift
```diff
@@ -81,7 +81,14 @@ class UnkeyedProtoDecodingContainer: InternalProtoDecodingContainer, UnkeyedDeco
     }
 
     func decode(_ type: Int.Type) throws -> Int {
-        throw ProtoError.decodingError("Int not supported, use Int32 or Int64")
+        if let value = popNext().last {
+            if MemoryLayout<Int>.size == 4 {
+                return try Int(decodeInt32(value))
+            } else if MemoryLayout<Int>.size == 8 {
+                return try Int(decodeInt64(value))
+            }
```
PSchmiedmayer: Similar to the comment above I think that this file could also benefit from an extension of the `Int` type. The sea applies to other `Int`ish types 👍

PSchmiedmayer (approved):
@moritzschuell I looked over the PR, great to see support for the platform specific `Int` and `UInt` types, thanks for all the work.
I just had a few code level comments, they are mostly related to reducing the code duplication introduced in the PR by moving some of the functionality to initializers defined in extensions of the `Int`, `UInt` and `Arrays` where the `Element` is one of the Int types to reduce the code duplication. Apart from this the PR looks good to me and can be merged after potentially reducing some of the code duplications.

Tests/ProtobufferCodingTests/Encoding/OptionalEncodingTests.swift
PSchmiedmayer: Agree 👍

Sources/ProtobufferCoding/Decoder/KeyedProtoDecodingContainer.swift
moritzschuell: I agree with you that `extractIntValue` is definitely not the best naming for this function 😁
Changed it to `convertToProtobufferFieldNumber`.

Sources/ProtobufferCoding/Decoder/KeyedProtoDecodingContainer.swift
moritzschuell: I moved the width-dependent decoding into extensions of the respective types, and the encoding parts into the shared internal en-/decoder containers. This reduces the code duplication across the several coding containers.

Tanks for the suggestion of using type extensions for this. Didn't occur to me.

PSchmiedmayer:
@moritzschuell Let me know when I should merge the PR 👍

moritzschuell:
@PSchmiedmayer I think I addressed all comments and we can merge it now 👍 Thank you!

PR #160, opened 2021-01-19
Refactor EndpointsTreeNode.printTree

nityanandaz:
We sometimes use print statements and sometimes use a logger

## :bulb: Proposed solution
Use `app.logger.info` for `EndpointsTreeNode`

## Reviewer Nudging
I can replace more `print`s, if pointed out

**Before**
```
2021-01-21T17:07:42+0100 warning codes.vapor.directory-config : No custom working directory set for this scheme, using /Users/nitai/Library/Developer/Xcode/DerivedData/Apodini-ecnwqazjaoyymagxspghqildiwwa/Build/Products/Debug
/ {
  v1/ {
    - read: Text [0:0]
    greet/ {
      - read: TraditionalGreeter [0:2:-1:0]
    }
    swift/ {
      - read: Text [0:4:-1:0:0]
      5/ {
        3/ {
          - read: Text [0:4:-1:0:1:0]
        }
      }
    }
    user/ {
      :D608F60E-961B-4D24-B176-49C135EF21E4/ {
        - read: UserHandler [0:5:-1:0]
      }
    }
    rand/ {
      - read: Random [0:3:-1:0]
    }
    auction/ {
      - read: Auction [0:1:-1:0]
    }
  }
}
...
```

**After**
```
2021-01-21T17:12:09+0100 warning codes.vapor.directory-config : No custom working directory set for this scheme, using /Users/nitai/Library/Developer/Xcode/DerivedData/Apodini-ecnwqazjaoyymagxspghqildiwwa/Build/Products/Debug
2021-01-21T17:12:09+0100 info org.apodini.application : /
→ v1/
  - read: Text [0:0]
  → auction/
    - read: Auction [0:1:-1:0]
  → swift/
    - read: Text [0:4:-1:0:0]
    → 5/
      → 3/
        - read: Text [0:4:-1:0:1:0]
  → rand/
    - read: Random [0:3:-1:0]
  → user/
    → :A24D4351-C3F1-4F8A-958C-6494A54F0CB1/
      - read: UserHandler [0:5:-1:0]
  → greet/
    - read: TraditionalGreeter [0:2:-1:0]
...
```


Sources/Apodini/Semantic Model Builder/Model/EndpointsTreeNode+CustomDebugStringConvertible.swift
```diff
@@ -0,0 +1,29 @@
...
+
+extension EndpointsTreeNode: CustomDebugStringConvertible {
+    var debugDescription: String {
+        if let parent = parent {
+            return parent.debugDescription
+        }
+        
+        let node = Node(root: self) { root in
+            Array(root.children)
+        }
```
bauer-andreas, line 13: Would you mind explaining the usage of Node here? Aren't those currently still two separate things?

nityanandaz: We can still leverage the advantage that `Node` is generic. Therefore, we wouldn't need to implement `.description` for another Tree again.

I don't mean to bind `Node` to TypeReflection. How about I move the file to `Apodini/Utilities`?

lschlesinger: I think this refactoring is a good idea :) also moving `Node` to `Apodini/Utilities` seems more suitable!

bauer-andreas (commented):
Looks good to me. 👍 

Just for the overview, would you mind posting new vs. old output. Would be great for a better overview imo

Sources/Apodini/Semantic Model Builder/SharedSemanticModelBuilder.swift
```diff
@@ -94,10 +94,10 @@ class SharedSemanticModelBuilder: SemanticModelBuilder, InterfaceExporterVisitor
         
         webService.finish()
         
-        webService.root.printTree() // currently only for debugging purposes
+        app.logger.info(.init(stringLiteral: webService.root.debugDescription))
```
bauer-andreas, line 97: Is the init needed? Shouldn't LogMessage be able to be implicitly constructed from a String instance?

bauer-andreas (approved):
Sorry for the delay from my side. Thanks for tackling this 👍

Sources/Apodini/Semantic Model Builder/SharedSemanticModelBuilder.swift
nityanandaz: Yes, it's needed 😕. As you said, it's a String instance, not a String literal (a value that appears directly in your source code)

PR #162, opened 2021-01-19
Restructure the TestWebService

PSchmiedmayer:
# *Restructure the TestWebService and Add Products*

The `TestWebService` was a huge monolith that grew over time

## :bulb: Proposed solution
Restructures the `TestWebService` into smaller files that each have a dedicated functionality.

## ❓ Implications
No implications on Apodini itself.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,5 @@
 // swift-tools-version:5.3
+
 import PackageDescription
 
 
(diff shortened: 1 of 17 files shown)


theMomax (changes requested):
I would suggest renaming `Sources/TestWebService/Mediators` to `Sources/TestWebService/Transformers` as that is the term Apodini uses internally and exposes externally (e.g. `ResponseTransformer`).

PSchmiedmayer:
@theMomax Thanks, a good catch! I have updated the PR to rename `Mediator` to `Transformer`.

theMomax (approved):
LGTM👍

PR #163, opened 2021-01-19
Export Additional Products

PSchmiedmayer:
# *Export Additional Products*

Apodini web services defined outside of the Apodini project can not use the `ApodiniDatabase`, `Jobs`, and `Notifications` targets as they are only defined internally and not exported as `products` in Package.swift.

## :bulb: Proposed solution
This PR exposes `ApodiniDatabase`, `Jobs`, and `Notifications` as`products` in Package.swift.

### Problem that is solved
Other web service implementations can now use the `ApodiniDatabase`, `Jobs`, and `Notifications` implementations.

### Implications
No implications on Apodini itself. Contributors of `ApodiniDatabase`, `Jobs`, and `Notifications` have to be aware that their public interface can now used by external dependencies.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -75,7 +75,8 @@ let package = Package(
             name: "ApodiniTests",
             dependencies: [
                 .product(name: "XCTVapor", package: "vapor"),
-                .target(name: "XCTApodini")
+                .target(name: "XCTApodini"),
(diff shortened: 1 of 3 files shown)


moritzschuell (approved):
Looks good to me 👍

PR #166, opened 2021-01-21
`ApodiniError` and `@Throws`

theMomax:
# `ApodiniError` and `@Throws`

The developer is able to return `Error`s from a `Handler`'s `handle()` function either directly or via an `EventLoopFuture`. What was missing is a standard that enables the developer to instruct exporters on how to deal with this error.

E.g. the developer cannot define the HTTP status code or if an associated WebSocket connection can remain open or must be closed.

Furthermore, static exporters (e.g. the OpenAPI exporter) were not able to see if and what a `Handler` throws.

## :bulb: Proposed solution

The core element is the `ApodiniError` as it carries all the information:
```swift
/// An error that can be thrown from `Handler`s and receives special treatment from compliant interface exporters.
public struct ApodiniError: Error {
    // ...
    private let `type`: ErrorType
    
    private let reason: String?
    
    private let description: String?
    
    private let options: PropertyOptionSet<ErrorOptionNameSpace>
    // ...
}
```

The simplest properties are `reason` and `description`. The difference is that the `reason` is always exposed, while the `description` is only exposed in `DEBUG` mode.

The `type` can be one of six common error types (e.g. `badInput`, `forbidden`, `serverError`) or the generic `other`.

`options` holds the information for the exporters. Each exporter can define its own options. The Apodini core does not define one. However, if an exporter defines an option, it has to provide a default value for each `ErrorType`. This requirement is expressed as the `StandardErrorCompliantOption`:
```swift
protocol StandardErrorCompliantOption: PropertyOption {
    static func `default`(for type: ErrorType) -> Self
}
```

E.g. for the `WebSocketStatusCode`:
```swift
extension WebSocketErrorCode: StandardErrorCompliantOption {
    static func `default`(for type: ErrorType) -> Self {
        switch type {
        case .badInput:
            return .dataInconsistentWithMessage
        case .serverError:
            return .unexpectedServerError
        default:
            return .normalClosure
        }
    }
}
```

### Exporter API

The `ApodiniError` is a `StandardError` and a `StandardErrorContext`:
```swift
protocol StandardErrorContext {
    func option<Option: StandardErrorCompliantOption>(for key: PropertyOptionKey<ErrorOptionNameSpace, Option>) -> Option
}

protocol StandardError: Error, StandardErrorContext {
    func message<E: StandardErrorCompliantExporter>(for exporter: E.Type) -> String
}
```

Any exporter that conforms to `StandardErrorCompliantExporter` can not only retrieve its `option`s from an `ApodiniError`, but also a `message`. This message depends on the exporter:
```swift
protocol StandardErrorCompliantExporter: InterfaceExporter {
    associatedtype ErrorMessagePrefixStrategy: Apodini.ErrorMessagePrefixStrategy = CustomErrorMessagePrefix
    static func messagePrefix(for context: StandardErrorContext) -> String?
}
```
There are default implementations for `messagePrefix` that can be enabled using `typealias ErrorMessagePrefixStrategy = StandardErrorMessagePrefix` or `typealias ErrorMessagePrefixStrategy = NoErrorMessagePrefix`.

This prefix can depend on any option (e.g. the HTTP status code would be reasonable for the REST exporter). The option doesn't even have to be public, e.g. for the `StandardErrorMessagePrefix` the `ErrorType` is used.
```swift
switch error.option(for: .errorType) {
    case .badInput:
        return "Bad Input"
    case .notFound:
        return "Resource Not Found"
    case .unauthenticated:
        return "Unauthenticated"
    case .forbidden:
        return "Forbidden"
    case .serverError:
        return "Unexpected Server Error"
    case .notAvailable:
        return "Resource Not Available"
    case .other:
        return "Error"
}
```

Exporters only have to deal with `ApodiniError`s. Any `Error` can be transformed to an `ApodiniError` using its `.apodiniError` property.

### Public Developer API

The most notable thing about the public API is that `ApodiniError` has no public initializers. The only way to obtain an `ApodiniError` is via the `@Throws` property wrapper's `wrappedValue`. It's initializer basically takes the same arguments as `ApodiniError`'s internal initializers and the wrapped `ApodiniError` is created from those arguments. The reason why I chose this approach is that I want the developer to declare the thrown errors as properties on the `Handler`. This way they can be accessed at startup and taken into account by static exporters.
```swift
struct Auction: Handler {
    @Throws(.badInput, reason: "The placed bid was too low") var bidTooLowError: ApodiniError
    
    @Parameter var bid: UInt

    @State var highestBid: UInt = 0
    
    // ...

    func handle() throws -> Response<String> {
           if bid > highestBid {
                highestBid = bid
                return .send("bid accepted")
            } else {
                throw bidTooLowError
            }
        // ...
    }
}
```

In order to make the construct more flexible, `ApodiniError` provides a public `callAsFunction` method that allows for modifying both `reason` and `description` (but not `type` or `options`!). Thus we could also write
```swift
throw bidTooLowError(description: "highest: \(highestBid); received: \(bid)")
```

Of course, the developer could still initialize the `Throws` inline and access its `wrappedValue` directly, but I think the way the API is written makes it clear that this is not how it is intended.

Options can be passed in similarly to `@Parameter`, e.g. if I wanted to penalize fiddling around with illegal values:
```swift
@Throws(.badInput, reason: "The placed bid is negative", .wsConnectionConsequence(.closeConnection)) var bidIsNegative: ApodiniError
```

### Problem that is solved
* There is a standard way for exporters to expose error-options, e.g. status codes or connection impact.
* There is a standard way for exporters to retrieve a meaningful error message from an error that automatically hides debug information in production
* The errors that are thrown by a `Handler` are known at startup time (if the developer sticks to the style-guide)

### Implications
The public API is purely additional. Every exporter that wants to make usage of the `ApodiniError`'s message has to conform to `StandardErrorCompliantExporter` and thus either implement the `messagePrefix` function or select one of the default implementations using a typealias. Each exporter can create as little or many options as they like. 

Retrieving certain options on startup for static exporters has not been implemented yet, but would be easily doable once the wanted option exists using the `execute` function of `Traversable.swift` and `@Throws`' conformance to `StandardErrorContext`.

### Testing
There are no tests yet. I will start writing tests once the proposed approach has general approval.

### Reviewer Nudging
Start out with `ApodiniError.swift` and get an idea how all the protocols play together. Then take a look at the WebSocketExporter's usage of `ApodiniError` in `WebSocketErrorOptions.swift` and `WebSocketInterfaceExporter.swift`. Afterwards you can inspect `Throws.swift` and see how it is used on a `Handler` in `Auction.swift`. The rest is just changes to the WebSocket infrastructure that came along with implementing the WebSocket exporter's options.


PSchmiedmayer:
@theMomax I really like the concept of an `ApodiniError`. It closely reflects the ideas we had in the last team meeting and even goes beyond with the `@Throws` property wrapper which is a nice addition!

The only conceptual feedback I would have is that the `static func messagePrefix(for error: StandardErrorContext) -> String` function of `StandardErrorCompliantExporter` should have a default implementation that reflects the naming of the different error types (... and maybe that the debug descriptions or the web socket exporter include some nicer language 🙃).
In addition I think it would be good to have a standard way of accessing the `ErrorType` from a `StandardErrorContext` without the need to define your own `WSErrorType` or similar type alias for each exporter as I would suspect that most exporter's `messagePrefix` will first and foremost really on the `ErrorType` with e.g. REST being an exception that might want to use an HTTP Status Code option.

Apart from this I like the type based approach and reusing the experience we have from the `@Parameter`s options for this feature! Looking forward to seeing this in the develop branch soon! I think this will also be helpful for everyone's example projects :)

Sources/Apodini/Properties/Throws.swift
```diff
@@ -0,0 +1,49 @@
+//
+//  Throws.swift
+//  
+//
+//  Created by Max Obermeier on 21.01.21.
+//
+
+import Foundation
+
+/// A property wrapper that can be used on `Handler`s to obtain `ApodiniError`s.
```
lschlesinger, line 10: nit: it is rather _in_ `Handler`s isn't it?

Sources/Apodini/Semantic Model Builder/WebSocket/WebSocketErrorOptions.swift
```diff
@@ -0,0 +1,99 @@
...
+
+extension AnyPropertyOption where PropertyNameSpace == ErrorOptionNameSpace {
+    /// An option that sepcifies the consequence an associated error has on the WebSocket connection and/or context.
+    public static func wsConnectionConsequence(_ consequence: WSConnectionConsequence) -> AnyPropertyOption<ErrorOptionNameSpace> {
+        AnyPropertyOption(key: .wsConnectionConsequence, value: consequence)
+    }
+}
+
+// MARK: WSErrorType
+typealias WSErrorType = ErrorType
```
lschlesinger: Maybe you want to name types consistently, e.g. `WebSocketErrorCode` vs. `WSErrorType`, or did I miss the underlying convention here?

Sources/Apodini/Semantic Model Builder/WebSocket/WebSocketInterfaceExporter.swift
```diff
@@ -9,48 +9,18 @@ import Fluent
 @_implementationOnly import WebSocketInfrastructure
 @_implementationOnly import OpenCombine
 @_implementationOnly import Runtime
```
lschlesinger: Why do you need Runtime here?

theMomax:
@PSchmiedmayer thanks for your feedback!

> In addition I think it would be good to have a standard way of accessing the `ErrorType` from a `StandardErrorContext`

I made the `ErrorType` an internal, exporter-agnostic option defined in the Apodini core. That way every exporter can access it from a `StandardErrorContext`.

> and maybe that the debug descriptions or the web socket exporter include some nicer language 🙃

Sure 😁

> the static func messagePrefix(for error: StandardErrorContext) -> String function of StandardErrorCompliantExporter should have a default implementation that reflects the naming of the different error types

That was something I explicitly avoided. The author of the extension should have to think about the naming and if it makes sense for the protocol in use. I understand that this might be unnecessary in some cases. I came up with the following solution: The exporter has to define a strategy, where the default strategy is providing a custom implementation, but it only takes one `typealias` to switch the strategy to one with a default implementation.

```swift
protocol ErrorMessagePrefixStrategy {}

struct StandardErrorMessagePrefix: ErrorMessagePrefixStrategy {}

struct NoErrorMessagePrefix: ErrorMessagePrefixStrategy {}

struct CustomErrorMessagePrefix: ErrorMessagePrefixStrategy {}

protocol StandardErrorCompliantExporter: InterfaceExporter {
    associatedtype ErrorMessagePrefixStrategy: Apodini.ErrorMessagePrefixStrategy = CustomErrorMessagePrefix
    
    static func messagePrefix(for context: StandardErrorContext) -> String?
}

extension StandardErrorCompliantExporter where ErrorMessagePrefixStrategy == StandardErrorMessagePrefix {
    static func messagePrefix(for error: StandardErrorContext) -> String? {
        switch error.option(for: .errorType) {
        // ...
        }
    }
}

extension StandardErrorCompliantExporter where ErrorMessagePrefixStrategy == NoErrorMessagePrefix {
    static func messagePrefix(for error: StandardErrorContext) -> String? {
        return nil
    }
}
```

This way the author at least notices they can customize this and must opt in to one of the standard implementations.

PSchmiedmayer:
@theMomax Thanks, I like the solution of having the `ErrorType` as an internal, exporter-agnostic option. The `StandardErrorMessagePrefix`, `NoErrorMessagePrefix`, and `CustomErrorMessagePrefix` protocols are a nice way to configure the `StandardErrorCompliantExporter `, I think this is a good approach, thanks for incorporating the feedback 👍

lschlesinger (approved):
Cool new feature! I only have some very minor comments.... Good that @PSchmiedmayer already gave helpful conceptual feedback ;) Maybe I'll ask for some support when integrating this in the OpenAPIExporter!

Sources/Apodini/Properties/Throws.swift
theMomax: I would argue properties are declared **on** structures🤔

Sources/Apodini/Semantic Model Builder/WebSocket/WebSocketErrorOptions.swift
theMomax: Good point👍🏻

Sources/Apodini/Semantic Model Builder/WebSocket/WebSocketInterfaceExporter.swift
theMomax: An artifact from a long time ago...thanks for the hint!

Sources/Apodini/Properties/Throws.swift
lschlesinger: Ah well, I am not too used to the swift lingo, so go with _on_, it's fine!

PR #168, opened 2021-01-22
Introduce Upload & Download Handlers

hendesi:
# Introduce Upload & Download Handlers

Currently there is no possibility in Apodini to upload images or files to the server and download them if needed. As this is something a web service should offer, I added the functionality to up- and download files to the directory structure of Apodini.

## :bulb: Proposed solution
To get this to work, I introduced a `Directory` property to the Apodini Application, which basically mirrors the behaviour of the Vapor directory to enable the saving in different sub folders. The Configurations use a `Directories` property which is an enum to specify the different directories Apodini uses.
This PR adds a `Uploader` handler, which takes an `UploadConfiguration` and a file that comes with the request. The `UploadConfiguration` allows the user to define the directory he wants the file to be saved in as well as an optional subPath in that folder.
```
Uploader(UploadConfiguration(.default, subPath: "Misc/"))
```

It also adds a `SingleDownloader` and `MultipleDownloader`, with which a user can either download single or multiple files under the given file name. The download handlers can also be configured using a `DownloadConfiguration`. Here can the user specify where the `Downloader` should look for the file.

```
MultipleDownloader(DownloadConfiguration(.default))

SingleDownloader(DownloadConfiguration(.default))
``` 

### Problem that is solved
Apodini now also the up- and download of files and images to its webservice.

### Implications
No real implications as its mostly new code.

### Related PRs
The default max. body size is currently hardcoded in the `WebService.swift`. Once #153 is merged, the max body size can added to the http configuration and this line can be removed.

### Testing
Tests for Downloads and Uploads and their corresponding configurations are included.

### Reviewer Nudging
One could have a look at the `Downloaders` and `Uploader` to see how the handlers work and then look at the provided test cases to see how they can be used and what request input they take.


Sources/Apodini/Semantic Model Builder/Model/TypeReflection/EnrichedInfo.swift
```diff
@@ -49,7 +49,8 @@ extension EnrichedInfo {
                         guard !errorDescription.contains(keyword) else {
                             return nil
                         }
-                        preconditionFailure(errorDescription)
+                        return nil
+//                        preconditionFailure(errorDescription)
```
hendesi, line 53: Without removing this line, I would get a runtime error shortly after startup. 
@nityanandaz Do you know by any chance why that's the case or is it fine to leave it like this. I did some research but couldn't find the cause

nityanandaz: I address this issue here https://github.com/Apodini/Apodini/pull/131, https://github.com/Apodini/Apodini/pull/131#pullrequestreview-571524716.

The problem is that `Parameter.defaultValue` is a function type since recently.

hendesi: Thanks! So I'll leave it like that for now

Sources/Apodini/Configurations/Application/Application+Directory.swift
```diff
@@ -0,0 +1,81 @@
+#if os(Linux)
+import Glibc
+#else
+import Darwin.C
+#endif
+import Foundation
+
+/// `Directory` represents a specified directory.
+/// It can also be used to derive a working directory automatically.
+public struct Directory {
```
awoc, line 11: If possible, I think it would be better to move this file to `ApodiniDatabase`

Sources/Apodini/WebService.swift
```diff
@@ -51,6 +51,8 @@ extension WebService {
         // Adds the created application instance to `EnvironmentValues`.
         // Can be used `@Environment` to access properties.
         EnvironmentValues.shared.values[ObjectIdentifier(Application.Type.self)] = app
+
+        app.vapor.app.routes.defaultMaxBodySize = "1mb"
```
awoc, line 55: For a future PR: Once we have the target `ApodiniVapor` (#169),  I think we can also move this to `ApodiniDatabase`.

Sources/ApodiniDatabase/FileHandlers/DownloadConfiguration.swift
```diff
@@ -0,0 +1,95 @@
+import Foundation
+import NIO
```
awoc: I think you can remove this import

Sources/ApodiniDatabase/FileHandlers/MultipleDownloader.swift
```diff
@@ -0,0 +1,51 @@
...
+/// If you want to download a single files, please refer to `SingleDownloader`
+public struct MultipleDownloader: Handler {
+    @Environment(\.directory)
+    private var directory: Directory
+    
+    @Environment(\.fileio)
+    private var fileio: NonBlockingFileIO
+    
+    @Environment(\.eventLoopGroup)
+    private var eventLoopGroup: EventLoopGroup
```
awoc, line 17: In the future I think it will be better to use the eventLoop of the request

Sources/ApodiniDatabase/FileHandlers/UploadConfiguration.swift
```diff
@@ -0,0 +1,44 @@
...
+            var url = URL(fileURLWithPath: mainPath)
+            for pathComponent in subPath.pathComponents {
+                url.appendPathComponent(pathComponent.description)
+                if !fileManager.fileExists(atPath: url.relativePath) {
+                    try fileManager.createDirectory(at: url, withIntermediateDirectories: false, attributes: nil)
+                }
+            }
+            return url.relativePath.appending("/").appending(fileName)
+        } catch {
+            print(error.localizedDescription)
```
awoc: Maybe you can use the logger here and in all other places instead of the print statement.

Sources/ApodiniDatabase/FileHandlers/MultipleDownloader.swift
hendesi: I agree! I think for now this is the only way to access an event loop from the handler.

Sources/Apodini/Configurations/Application/Application+Directory.swift
hendesi: I think this is not possible since the directory gets initialised in the Core and it is currently not supported to access this from outside Apodini.

hendesi:
@awoc I addressed all issue that were possible to fix. Could you please take another look?

@PSchmiedmayer I would like to ignore the sonarCloud analysis for this PR, because there a bunch of issue which can/should not be fixed (e.g. `Directories` naming of `public`/ `default` which should just be named like this. )
Is it possible to disable the checks for single lines? Like in swift line

PSchmiedmayer:
@hendesi I think it is fine to ignore most of the SonarCloud warnings for this PR. I would argue that the following issue should be resolved: https://sonarcloud.io/project/issues?id=Apodini_Apodini&issues=AXcqhsk_UB5eMyZTOMtc&open=AXcqhsk_UB5eMyZTOMtc&pullRequest=168
The nesting of the function could be reduced by a bit 👍

awoc (approved):
Looks good 👍

PR #171, opened 2021-01-22
Make Apodini Work When Compiling Using a Release Configuration

### LINKED ISSUE #167: [BUG] Target TestWebService does not Build with Configuration `Release`
**Describe the bug**
When setting "Build Configuration" to "Release" in Xcode on target "TestWebService", building fails with three errors I could not easily fix:

image

**To Reproduce**
Select "Release" setting:

image

Press `Cmd+B`

**Expected behavior**
Apodini should compile with release settings.

 - Swift Version: 5.3.2 (swiftlang-1200.0.45 clang-1200.0.32.28)
 - Xcode Version: 12.3 (12C33)

PSchmiedmayer:
# *Make Apodini Work When Compiling Using a Release Configuration*

As documented in #167 Apodini doesn't compile when compiling in a release configuration.
In addition there is an issue that `AssociatedTypeRequirementsKit` fails at runtime when starting an Apodini `WebService` after building it using a release configuration.

## :bulb: Proposed solution
Fixes compile errors that caused that the release configuration could not build.
Move the `TestWebService` to a separate package and add Github Actions and test cases to test that a `WebService` compiles and does successfully startup. 
@nerdsupremacist I have not yet found a solution for the runtime failure due to `AssociatedTypeRequirementsKit`: You can inspect the error when running the tests of the `TestWebService` in release mode. `AssociatedTypeRequirementsKit` seems to call a method that is no longer available and runs into an fatalError: https://github.com/Apodini/Apodini/runs/1750026279#step:6:1585 

### Problem that is solved
The `WebService` now compiles on a release configuration and we added some regression tests.
Closes #167

### Implications
The `TestWebService` is now stored in a different directory.
This PR adds new GitHub Actions that test if the `TestWebService` can start up. The required GitHub Actions will be updated after this PR.

### Testing
Added regression tests.


lukaskollmer:
Question: Do release builds contain the same run-time type information as debug builds? Eg, do things like `Mirror` or the `typeInfo` function work the same way regardless of whether it's a debug or a release build, or would that somehow be affected? Are there optimisations applied to release builds which would interfere w/ some of the runtime stuff we're doing?

PSchmiedmayer:
@lukaskollmer Yes, I suspect that this might be the case why we have the fatal error here with `AssociatedTypeRequirementsKit `: https://github.com/Apodini/Apodini/runs/1750026279#step:6:1585 ... but I think that `Mirror` seems to work fine on release builds.

Sources/Apodini/Components/Component/Component.swift
```diff
@@ -54,6 +54,14 @@ private protocol HandlerVisitorHelperImplBase: AssociatedTypeRequirementsVisitor
     func callAsFunction<H: Handler>(_ value: H) -> Output
 }
 
+extension HandlerVisitorHelperImplBase {
+    @inline(never)
+    @_optimize(none)
+    func _test() {
+        _ = self(Text(""))
```
nerdsupremacist, line 63: # Context on what's going on here
This method is never actually called.
But since we call the implementation of `callAsFunction` manually by checking the witness table ourselves, the compiler will doesn't see any uses of our function and will delete the implementation for release builds. Since we don't want the users of Apodini to have to disable optimizations, we needed another way to trick the compiler not to optimize this away.

So the way this works is that `AssociatedTypeRequirementsVisitor` requires a `_test` function where you're supposed to call `callAsFunction`. 
And since `callAsFunction` is being called from a protocol extension, all implementations of `callAsFunction` for this protocol, will be marked as if it's possible that they will be called at some point

Sources/Apodini/SyntaxTreeVisitor/ContextNode.swift
```diff
@@ -88,13 +88,12 @@ class ContextNode {
             // we need to check if `OptionalContextKey` is type of `ContextKey`
             // aka if the `ContextKey provides a default value. Because if it provides a defaultValue
             // we need to call `reduce(...)` with it before inserting.
-            let visitor = StandardContextKeyTypeVisitor()
-            let defaultValue = visitor(contextKey)
 
             // if defaultValue is nil the contextKey didn't conform to `ContextKey` => doesn't have a default value
-            if let defaultValue = defaultValue {
+            if let contextKey = contextKey as? HasDefaultValue.Type {
```
nerdsupremacist, line 93: # Context
`AssociatedTypeRequirementsTypeVisitor` is still broken for release builds. So this is just moving away from it, since we can

PSchmiedmayer:
@nerdsupremacist It would be great if you can give the PR a second look. I will need to merge it with admin privileges as we have renamed the GitHub Actions and we don't have a high enough diff test coverage in this PR.
After this PR is merged I will update the required checks to reflect the new GitHub Actions.

bauer-andreas (approved):
Looks good to me.
Neat workaround for the `AssociatedTypeRequirementsKit` issue 👍 Thanks for the detailed explanation.

TestWebService is basically its own SPM project now, right? Will it share the `.build` directory with the "parent" Apodini project? Or will it act on its own, downloading all dependencies again? In this case it might make sense to add similar `.build` directory caching using `action/cache`?

Additionally I noticed that for the `TestWebService` we test both, debug and release configuration. Would it make sense to add that to the "regular" build and test Action as well?
Prevent the situation we are in right now?

PSchmiedmayer:
@Supereg Good points. Yes, the TestWebService will use a separate `.build` directory. I will add cache actions. I will also add a release build action for the main target.

One thing we should discuss in general: One thing I noticed with this PR is that by having the Package.resolved file in the repo we always used fixed versions for the development. A Package.resolved file is ignored by anything that imports a library and therefore it is ignored by the `TestWebService` when it resolves the `Apodini` target. 
Therefore we currently have a build error for any web service using Apodini at its current state on develop: Apodini doesn't build as our Package.resolved file uses `0.2.2` of `AssociatedTypeRequirementsKit` while its current version is `0.3.0` which actually includes source breaking changes. If we wouldn't have a Package.resolved file and would not cache the dependencies we would have caught this error a few days ago using some CI builds that ran for the different PRs.

But as far as I can see there is no easy workaround for this. Maybe a `swift package update` before the build would resolve the issue? But then caching the dependencies doesn't really make too much sense anymore. As our "cache" guru: @Supereg, what's your opinion about this?

bauer-andreas:
@PSchmiedmayer In my opinion it might actually be a good thing, that we have **reproducible** dependency versions for our CI builds (by checking in `Package.resolved`). I could envision that it would be really frustrating, if the code runs locally but fails in the CI build of your PR. Especially as CI builds aren't really great to debug.

And should we implement a way where PR CI builds run with the latest dependencies, the problem of untested dependency versions getting into our release build, still exists. I don't think we can forever guarantee a consistent stream of PRs checking compatibility?

Instead I would propose to directly address the problem of users ending up with newer, untested dependency versions of Apodini. I would propose two actions:

1. The Swift Package Manager is strongly built around [Semantic Versioning](https://semver.org). Thus we can expect, that our dependencies follow those semantic versioning rules as well. Therefore I would suggest, that we don't allow automatic version upgrades which contain breaking changes.
In the `AssociatedTypeRequriementsKit` example: The project still has a major version number of `0`. Thus any version increment (even patch level increments) is allowed to contain breaking changes (API wise and functionality wise; after all its an early development version) [[4.](https://semver.org/#spec-item-4)] (_I wonder why spm is automatically upgrading such development versions_). I would suggest to use fixed version numbers (using `.exact` requirement) or only allowing patch level increments (using `.upToNextMinor` requirement; would allow for automatically pulling in bug fixes).
By default SPM uses `.upToNextMajor` for any dependency. Depending on our trust in a certain dependency we might want to consider changing that to `.upToNextMinor` in some cases. After all, minor version increments allow for internal changes [[7.](https://semver.org/#spec-item-7)], which are often combined with the introduction of new bugs.
2. As a second step I would setup weekly (or whatever time frame makes sense) automated builds of the `release` (and `develop`) branch running always on the latest pulled dependencies ("latest" according to our rules we built in step 1.). This would address the possibility that even patch level increments (or minor level increments, where allowed) could alter the build output in an unfavorable way. This would allow us to keep build time optimizations for our PR builds. After all, those happen pretty often and sometimes with very little changes, keeping frustration low by shorter build times.

One problem still unsolved, is the maintenance overhead of manually updating dependencies (and checking in those changes via `Package.resolved`). afaik spm doesn't have a command to list all updatable packages. And it seems like there isn't good tolling available for spm which informs you about a dependency updates in an automated way like e.g. GitHubs dependabot (other than watching all GitHub repos for new releases). Though this problem already exists today, as we also have to manually update major releases. But I think it's a good thing to force dependency updates to be a manual/active step in some cases, as it directly highlights where a potential error is coming from.

So I think we have to find a middle ground of which dependencies require a manual Apodini package update a which don't. And complement that with scheduled CI. Imo this makes most sense to combat this issue.

PSchmiedmayer:
@Supereg Thanks for the input!
1. I agree with constraining `AssociatedTypeRequriementsKit` to the `upToNextMinor` version as it is already done in this PR. This should in fact resolve the issue you have discussed. I have also done the same for `OpenCombine` to also constraint it `upToNextMinor`.
2. I think this is a good idea! I looked up the syntax for GitHub Actions and it should be fairly simple to construct such an action. I think this Action could also address the problem you have described below. The action could run a `swift package update`, check if there is any change in the `Package.resolved` file and if so test everything and create a PR with the updated Package.resolved file. I will create an issue and might tackle this next week 🚀

PSchmiedmayer:
@Supereg I have incorporated your feedback, the GitHub Actions for the `TestWebService` now use a cache for the dependencies and the main build and test action also builds the web service using the release configuration.

.github/workflows/test-webservice.yml
```diff
@@ -16,37 +16,43 @@ defaults:
...
         configuration: [debug, release]
     steps:
     - uses: actions/checkout@v2
     - uses: maxim-lobanov/setup-xcode@v1.1
       with:
         xcode-version: latest
+    - uses: actions/cache@v2
+      with:
+        path: .build
+        key: ${{ runner.os }}-spm-cachev2-${{ hashFiles('**/Package.resolved') }}
```
bauer-andreas: The TestWebService Action uses the same caching key pattern as the "regular" action. If the hash function would evaluate to the same result (they don't) they would store conflicting .build directories.

The hash function don't evaluate to the same value though as of the changed working directory, thus the `**/Package.resolved` will only find `TestWebService/Package.resolved` (I assume).
Though I would argue to make that difference explicit using a different key pattern.

Another note. As `**/Package.resolved` is still used in the "default" action, the hash will consider both the root `Package.resolved` and the `TestWebService/Package.resolved`, which is a bit weird. Maybe its worth removing the `**` matching. I can't find a case where it makes sense, wonder why they put that into the example over at `action/cache`.

The test-webservice action never touches the root `.build` folder right?

nerdsupremacist (approved):

.github/workflows/test-webservice.yml
PSchmiedmayer: That's a good point, I will update the PR to only include the correct `Package.resolved` file 👍

PR #176, opened 2021-01-24
Single Read Handler

hendesi:
# Single Read Handler

As already mentioned in this [review](https://github.com/Apodini/Apodini/pull/70#pullrequestreview-571105829), there is currently no possibility to retrieve just one item from the database by its id. As this is something every web service should offer, this PR adds the corresponding functionality.

## :bulb: Proposed solution
Adding a handler that returns the object found by the id in the database. Nothing too fancy.

### Problem that is solved
Now single object queries (by id) are possible.

### Testing
Yes, one test covering the functionality of the new handler.

### Reviewer Nudging
These are just minor additions. Should be self-explanatory.


nityanandaz (approved):
I have only a name suggestion, `ReadOne` instead of `SingleRead`.

I think it would go well with `ReadAll` (I saw that this is just `Read`). And it reads more like a verb, like Create, Update and Delete, instead of a noun.

PR #179, opened 2021-01-25
Handle nested collection types by introducing `recursiveEdit`

### LINKED ISSUE #178: [BUG] OAS: Nested collection types are not reflected properly
**Describe the bug**
Nested collection types are not reflected properly, meaning that the transformation of the tree with `Node<EnrichedInfo>` did not produce expected results after handling _nested_ collection types (array and dictionary). As a result, the schema was not created as expected.

**To Reproduce**
Create a struct that incorporate nested collection types, such as:
```swift
struct Game: Content {
    /// Associated players, i.e. "the team", key is player ID.
    var players: [String: Player]
}

struct Player: Content {
    var name: String?
    var hand: [Card]
}
```
After creating the tree, we trigger the transformation as conducted in [`OpenAPIComponentsObjectBuilder`](https://github.com/Apodini/Apodini/blob/0415b881ea0b5917d8633510163aedfd4f4f4c45/Sources/Apodini/Semantic%20Model%20Builder/OpenAPI/Document/OpenAPIComponentsObjectBuilder.swift#L131): 
```swift
static func node(_ type: Any.Type) throws -> Node<EnrichedInfo>? {
        let node = try EnrichedInfo.node(type)
            .edited(handleOptional)?
            .edited(handleArray)?
            .edited(handleDictionary)?
            .edited(handlePrimitiveType)?
            .edited(handleUUID)
        return node
    }
```
What I observe, is that the correct representation of a nested collection type directly depends on the order of calling             `.edited(handleArray)?` and `.edited(handleDictionary)?`. In the presented case, after we handle dictionary, we should handle the nested array type `hand`, i.e. a property of `Player` stored in this dictionary. However, you can construct example, where it is exactly the other way around.

A wrong representation would include the schema for type Array in the components section such as:
```json
"Array": {
	"type": "object",
	"properties": {
	  "_buffer": {
	    "$ref": "#/components/schemas/_ArrayBuffer"
	  }
	},
	"required": [
	  "_buffer"
	]
}
```
Indicating that the tree was not transformed/pruned correctly for array types (nested in another collection type, here the `players` dictionary).

**Expected behavior**
We would expect, that the transformation of the `Node<EnrichedInfo>` tree for `Game.self` would resolve the dictionary stored in `players` as well as the array stored in a `Player`s `hand`.
```json
  "components": {
    "schemas": {
      "Game": {
        "type": "object",
        "properties": {
          "players": {
            "type": "object",
            "additionalProperties": {
              "$ref": "#/components/schemas/Player"
            }
          }
        },
        "required": [
          "players",
        ]
      },
      "Player": {
        "type": "object",
        "properties": {
          "name": {
            "type": "string"
          },
          "hand": {
            "type": "array",
            "items": {
              "$ref": "#/components/schemas/Card"
            }
          }
        },
        "required": [
          "hand",
        ]
      },
      "Card": {
        "type": "object",
        "properties": {
          "number": {
            "type": "integer"
          }
        },
        "required": [
          "number"
        ]
      }
    }
  }
```
AND **NOT** including a schema for Array in the components section.

**Additional context**
This heavily rely on how the different handle functions passed into `.edited` are written and how the tree of `Node<EnrichedInfo>` is constructed.

lschlesinger:
# Handle nested collection types by introducing `recursiveEdit`

Nested collection types are not described properly in OAS. This is due to the static order of handling arrays and dictionaries [here in `OpenAPIComponentsObjectBuilder`](https://github.com/Apodini/Apodini/blob/0415b881ea0b5917d8633510163aedfd4f4f4c45/Sources/Apodini/Semantic%20Model%20Builder/OpenAPI/Document/OpenAPIComponentsObjectBuilder.swift#L131). The situation with example is also described by the respective issue #178.

## :bulb: Proposed solution
```swift
extension OpenAPIComponentsObjectBuilder {
    static func node(_ type: Any.Type) throws -> Node<EnrichedInfo>? {
        let node = try EnrichedInfo.node(type)
        var counter = 0
        return try recursiveEdit(node: node, counter: &counter)
    }
    
    private static func recursiveEdit(node: Node<EnrichedInfo>, counter: inout Int) throws -> Node<EnrichedInfo> {
        if counter > OpenAPISchemaConstants.allowedRecursionDepth {
            fatalError("Error occurred during transforming tree of nodes with type \(node.value.typeInfo.name). The recursion depth has exceeded the critical value of \(OpenAPISchemaConstants.allowedRecursionDepth)")
        }
        counter += 1
        let before = node.collectValues()
        guard let newNode = try node
            .edited(handleOptional)?
            .edited(handleArray)?
            .edited(handleDictionary)?
            .edited(handlePrimitiveType)?
            .edited(handleUUID) else {
            fatalError("Error occurred during transforming tree of nodes with type \(node.value.typeInfo.name).") }
        let after = newNode.collectValues()
        return after != before ? try recursiveEdit(node: newNode, counter: &counter) : node
    }
}
```

### Problem that is solved
#178 

### Implications
`EnrichedInfo: Hashable` -> since the condition for recursively calling `recursiveEdit` is that the recently handled node, i.e. after calling `.edited(...)`, has changed compared to its state when entering this function. This comparison relies on the `collectValues()` returning `Set<T> where T: Hashable`.

Refactoring of `schemaName` as the previously taken `TypeInfo.mangledName` provided by `Runtime` was not suitable for collection types. The function `createSchemaName` will now produce `ArrayofCard` instead of `Array` for example.

### Related PRs
@nityanandaz is facing the same problem and works on a solution for
https://github.com/Apodini/Apodini/pull/131#discussion_r563973311

### Testing
`testCreateEnrichedInfoTree()` is newly introduced.

### Reviewer Nudging
Please look specifically at `recursiveEdit` function.
Random reviewers picked.


theMomax (approved):
Everything looks good and reasonable👍🏻

PR #180, opened 2021-01-25
Move scheduler to application and add tests

awoc:
# Move `Scheduler` to `Application` and add tests

Currently, the Scheduler is part of `EnvironmentValues` and created as a Singleton. By moving it to `Application` the handling of the `Scheduler` is simplified and the `Scheduler` can access the `EventLoopGroup` directly. Also, the property wrapper check of `Job`s didn't include a check for `@Parameter`.

## :bulb: Proposed solution

Refactor the implementation of the `Scheduler` to work with this approach. 

### Implications

- No implication on the usage
- Now also supports all `Activatable`s properties
- Exposes `Activatable` and some function of `Traversable`

### Testing

Tests for the correct scheduling and handling in time of `Job`s were added.

### Reviewer Nudging

Should be easy.


Sources/Jobs/Scheduler.swift
```diff
@@ -82,6 +101,8 @@ private extension Scheduler {
     
     func schedule<T: Job>(_ job: T, with config: JobConfiguration, _ runs: Int, on eventLoop: EventLoop) {
         guard runs > 0, let nextDate = try? config.cron.next() else {
+            print("CANCEL \(runs)" )
```
theMomax: Please use the logger.

Sources/Jobs/Scheduler.swift
```diff
@@ -97,7 +118,7 @@ private extension Scheduler {
     func checkPropertyWrappers<T: Job>(_ job: T) throws {
         for property in Mirror(reflecting: job).children {
             switch property.value {
-            case is Environment<EnvironmentValues, Connection>:
+            case is Environment<EnvironmentValues, Connection>, is RequestBasedPropertyWrapper:
```
theMomax: Please always use execute of Traversable.swift instead of mirror. Otherwise DynamicProperty and Properties are not supported.

theMomax (changes requested):
The only problem is the usage of Mirror instead of Traversable (which was there before). I think now would be a good time to fix it.

hendesi (approved):
Please have a look at the changes @theMomax mentioned. Aside from that, It looks good to me 👍

Sources/Jobs/Scheduler.swift
awoc: Deleted this statement

Sources/Jobs/Scheduler.swift
awoc: I use the methods `apply` and `execute` of `Traversable` now. Since `Job`s have their own target I made both methods public. Furthermore, I declared `Activatable` public to work with `Job`s. This means that `@State` can now be used with `Job`s and is activated only once at creation and stays activated. 

Also, `apply` and `Activatable` need to be public to function with the new approach of `@Environment` #181 which will also conform to `Activatable`.

Sources/Apodini/Properties/State.swift
```diff
@@ -56,12 +56,13 @@ extension State where Element: ExpressibleByNilLiteral {
 /// An `Activatable` element may allocate resources when `activate` is called. These
 /// resources may share information with any copies made from this element after `activate`
 /// was called.
-protocol Activatable {
+public protocol Activatable {
+    /// Activates the given element.
     mutating func activate()
```
theMomax, line 61: `Activatable` is public but luckily cannot be called by the user on property wrappers, because `handle()` is not mutating.

Sources/Jobs/Job+Traversable.swift
```diff
@@ -0,0 +1,38 @@
...
+
+        apply({ (activatable: inout Activatable) in
+            activatable.activate()
+        }, to: &selfCopy)
+
+        return selfCopy
+    }
+    
+    /// Checks if only valid property wrappers are used with `Job`s.
+    func checkPropertyWrapper() throws {
```
theMomax: Could be implemented inside main target as generic global instead of extension.

Sources/Jobs/Job+Traversable.swift
```diff
@@ -0,0 +1,38 @@
...
+        }, on: self)
+        
+        try execute({ (_: RequestBasedPropertyWrapper) in
+            throw JobErrors.requestPropertyWrapper
+        }, on: self)
+    }
+    
+    /// Subscribes to all `ObservedObject`s.
+    func subscribe(configuration: JobConfiguration) {
+        execute({ (observedObject: AnyObservedObject) in
```
theMomax: Since this function depends on your target-internal `JobConfiguration` it cannot be transferred to the main target as is. I'd suggest to make a public `subscribe<T>(on target: T, using callback: (AnyObservedObject) -> Void)` global in the main target. You can then use this function to implement your extension here.

Sources/Jobs/Job+Traversable.swift
```diff
@@ -0,0 +1,38 @@
...
+            observedObject.valueDidChange = {
+                observedObject.setChanged(to: true)
+                // Executes the `Job` on its own event loop
+                _ = configuration.scheduled?.futureResult.hop(to: configuration.eventLoop)
+                observedObject.setChanged(to: false)
```
theMomax: I don't think this logic works as intended. The `observedObject.setChanged` should be called **after** the future has completed, shouldn't it?

theMomax (commented):
Besides one logic-issue (not sure about that one, though) I have one concern about this PR. It exposes quite a lot of internal, low-level functionality to the user of Apodini. I added some suggestions on how you could avoid making `execute` and `apply` public. @PSchmiedmayer I would be interested in your opinion on this topic. Maybe it would be better **not** to separate Jobs into their own target.

PSchmiedmayer:
@theMomax @awoc I agree that we expose some more public information by having Jobs in a separate target but as `Activatable` can not be used on a `@State` property wrapper and accessing the `@State` or `@Environment` property wrapper itself isn't normally necessary I would be fine with this. I would suggest to incorporate the suggestions of @theMomax to keep `execute` and `apply` private as this would be a bigger concern to me.

Sources/Jobs/Job+Traversable.swift
awoc: Yes, you're right this makes it safer

Sources/Jobs/Job+Traversable.swift
awoc: Good idea!

Sources/Jobs/Job+Traversable.swift
awoc: Fixed with `c3c4523`

Sources/Jobs/Job+Traversable.swift
awoc: Fixed with `c3c4523`

Sources/Jobs/Job+Traversable.swift
awoc: Instead of "hopping" I'm executing it now directly on the eventloop

theMomax (approved):
LGTM!

Sources/Apodini/Properties/State.swift
```diff
@@ -56,12 +56,13 @@ extension State where Element: ExpressibleByNilLiteral {
 /// An `Activatable` element may allocate resources when `activate` is called. These
 /// resources may share information with any copies made from this element after `activate`
 /// was called.
-protocol Activatable {
+public protocol Activatable {
```
PSchmiedmayer: Based on the changes in the last commits I would suggest that we can make `Activatable` and the `activate` function internal again:
```suggestion
protocol Activatable {
```

Sources/Apodini/Properties/State.swift
```diff
@@ -56,12 +56,13 @@ extension State where Element: ExpressibleByNilLiteral {
...
+    /// Activates the given element.
     mutating func activate()
 }
 
 extension State: Activatable {
-    mutating func activate() {
+    public mutating func activate() {
```
PSchmiedmayer: Based on the changes in the last commits I would suggest that we can make `Activatable` and the `activate` function internal again:
```suggestion
   mutating func activate() {
```

PSchmiedmayer (commented):
@awoc Thanks for incorporating the feedback by @theMomax . I only had a small comment when looking at the PR. In addition I have created #190 to document that we might want structure the global functions a bit but I would not include this in this PR and therefore created the #190 issue.

PR #181, opened 2021-01-25
Combine `EnvironmentValues` and `Storage`

### LINKED ISSUE #164: [INTERNAL CHANGE] Combine the functionality of `EnvironmentValues` and `Storage`
**Is your change related to a problem? Please describe.**

PR #82 introduced a `Storage` struct that offers a dictionary to store properties used for the configuration of `Application`. The functionality of `Storage` and the existing `EnvironmentValues` used by `@Environment` is almost identical. The concept of `EnvironmentValues` is, by using extensions, to declare stored properties which can then be accessed using `@Environment` and a key path. Storage works the same way. Therefore, it would be helpful to only have one central and convenient place for storage related settings in Apodini. This would also reduce code duplication in the project.

**Propose a solution**

Adapt the functionality of `Storage` to work with `@Environment`. This would mean overloading the subscripts of `Storage` and changing stored properties of `EnvironmentValues` using `EnvironmentKey` to work with `StorageKey`.

awoc:
# Combine `EnvironmentValues` and `Storage`

At the moment `@Environment` uses `EnvironmentValues` and `Application` to store values which can be injected. The  `Storage` struct a similar interface for `Application`. Moving to logic of `EnvironmentValues` to `Storage` would reduce code duplication and offer a central space for storage in Apodini.

## :bulb: Proposed solution

Instead of defining extensions for `EnvironmentValues`, `Application` is used. This also works with other targets. 

`@Environment` requires an `Application` instance to access the `Storage`. This is achieved by conforming to `ApplicationInjectable` which injects the instance once at server start up. Like `@State`, `@Environment` now also conforms to `Activatable` which allows to restrict the access of the wrapped value to be only used in a request. This means that `@Environment` always needs to be activated before accessing the wrapped value.

### Problem that is solved

#164 

### Implications

- Renames the property `db` of `Application` to `database`
~~- Removes `@Environment(\.database)`. The same functionality is also possible with `@Environment(\.db)`~~
- The `Application` instance needs to be passed to `Environment` at server startup and also for testing
- `@ObservedObject` also conforms to `ApplicationInjectable` since it needs access to the storage.
- The global `inject` function in Traversable was declared public since it is required by the `Scheduler` to work with `@Environment` and `@ObservedObject`
- Renames `KeyChain` to `EnvironmentAccessible`

### Related PRs

- #82 introduced `Application` and `Storage`

### Testing

No new tests are introduced. The existing ones were refactored to work with this implementation. 

The helper function `mockQuery` helps with the creation of requests with `ApplicationInjectable`s. 

### Reviewer Nudging

One could start with the changes to `@Environment` and then have a look at `Storage`.


Sources/Apodini/Application/Application+Connection.swift
```diff
@@ -1,21 +1,18 @@
...
-}
-
-extension EnvironmentValues {
+extension Application {
     /// A Property identifying the `Connection` which provides an abstract view on the underlying protocol's state.
+    /// - Note: The setter should never be directly called and only serves as a placeholder.
     public var connection: Connection {
-        get { self[ConnectionEnvironmentKey.self] }
-        set { self[ConnectionEnvironmentKey.self] = newValue }
+        get { Connection() }
```
PSchmiedmayer: I would even suggest to add a `fatalError` for accessing a `Connection` on the `Application` level should not possible as this is stored on a `Environment` property wrapper level. 

Thinking about this now we could reconsider adding back the `EnvironmentKeys` as indicators that they are only stored in the `@Environment` property wrapper's key value storage and accessing them does not even take a look at the `Application` storage while accessing an environment value using an `Application` key path first checks the storage of the single `@Environment` property wrapper and then the `Application` storage?

Sources/ApodiniDatabase/Handlers/Create.swift
```diff
@@ -4,7 +4,7 @@ import Apodini
 /// A Handler that creates, if possible, an object in the database that conforms to `DatabaseModel` out of the body of the request.
 /// It uses the database that has been specified in the `DatabaseConfiguration`.
 public struct Create<Model: DatabaseModel>: Handler {
-    @Apodini.Environment(\.database)
+    @Apodini.Environment(\.db)
```
PSchmiedmayer: I would suggest to keep the name `database` instead of `db`. I would suggest renaming the property name of the `Application`

Sources/Apodini/Application/Application+Connection.swift
awoc: I think the idea was to have a default `Connection` with the state `.end`. But I think the default case is also handled by all exporters, so changing this the getter to a `fatalError` shouldn't change anything. Maybe @moritzschuell can have a look at this.

I don't think it is necessary to re-introduce `EnvironmentKeys`. The behavior of `@Environment` is to always check `dynamicProperties` first and only if it is missing the `Application` storage is accessed.

PSchmiedmayer: Sounds good, I think that handling this on the exporter level or ideally on the `SemanticModelLevel` sounds good to me.
Ok, makes sense 👍

PSchmiedmayer:
@awoc Please make sure that the comments make in #180 are also reflected in this PR so we don't expose too much internal functionality in the public interface 🚀

Apart from this: It seems like the reporting of the code coverage failed. Let me know if this continues to happen or if we need to manually merge the PR once @nityanandaz and @sadikekin reviewed the PR.

Sources/Notifications/Application+NotificationCenter.swift
```diff
@@ -0,0 +1,7 @@
+import Apodini
+
+extension Application {
+    var notificationCenter: NotificationCenter {
+        .init(app: self)
```
nityanandaz: If I understand correctly, NotificationCenter is more transient now. Therefore, I suggest to make NotificationCenter a value type:
```swift
public struct NotificationCenter { ... }
```
I believe this would be a little bit more efficient than initializing and discarding a class instance many times, but I am not an expert on what semantics are expected for this type  🙂

Sources/Apodini/Application/Application+Databases.swift
```diff
@@ -44,14 +44,13 @@ extension Application {
         }
     }
 
-    // swiftlint:disable identifier_name
     /// default database
-    public var db: Database {
-        self.db(nil)
+    public var database: Database {
```
nityanandaz, line 48: I prefer this over the abbreviation 👍

Sources/Apodini/Application/Application+Connection.swift
```diff
@@ -1,21 +1,18 @@
...
 import Foundation
 
-enum ConnectionEnvironmentKey: EnvironmentKey {
-    static var defaultValue = Connection()
-}
-
-extension EnvironmentValues {
+extension Application {
     /// A Property identifying the `Connection` which provides an abstract view on the underlying protocol's state.
+    /// - Note: This property only serves as a placeholder and should never be called.
```
nityanandaz:
```suggestion
    ///
    /// This property only serves as a placeholder and should never be called.
```

If you make a break (leave one line blank), Xcode will already display this section below a heading _Discussion_

Sources/Notifications/Application+NotificationCenter.swift
awoc: Good idea! I changed it to a struct now. 

Also, I noticed that a warning is logged every time `fcm` from the `Application` is recreated. That's why I retrieve it from the storage instead of recreating it.

Sources/Apodini/Application/Application+Connection.swift
moritzschuell: @awoc Yes, you're right. The `ConnectionContext` will make sure to always supply a `Connection` to the handler, with state `.end` by default. So, an application level `Connection` access should never happen (and also doesn't really make sense semantically).

PSchmiedmayer:
@awoc I just checked out the open PRs and also saw this one. I remember adding some comments a while ago but I lost a bit track of the PR. Do you need an additional review by anyone from the @Apodini/apodini team or myself?

awoc:
> @awoc I just checked out the open PRs and also saw this one. I remember adding some comments a while ago but I lost a bit track of the PR. Do you need an additional review by anyone from the @Apodini/apodini team or myself?

@PSchmiedmayer I would be happy about a review 😄 This PR should be ready to merge

Sources/Apodini/Application/Application+FCM.swift
```diff
@@ -0,0 +1,32 @@
+//
+//  Application+FCM.swift
+//  
+//
+//  Created by Tim Gymnich on 30.12.20.
+//
+
+import FCM
+
+extension Application {
```
PSchmiedmayer, line 10: I am not sure if this is already addressed by an other PR (might very well be out of scope for this PR) but we should think about moving this file and the FCM dependencies to the Notifications target.

Sources/Apodini/Properties/Environment.swift
```diff
@@ -0,0 +1,70 @@
...
+}
+
+/// Properties that need an `Application` instance.
+protocol ApplicationInjectable {
+    /// injects an `Application` instance
+    mutating func inject(app: Application)
+}
+
+/// A protocol to define key paths that can be used with `@Environment` to retrieve pre-defined objects.
+public protocol KeyChain { }
```
PSchmiedmayer: I think the name `KeyChain` originates from the proposal written at the beginning of the semester. I am not really sure it is a great name as I always start thinking about the iOS/macOS keychain functionality when seeing this protocol. Maybe `EnvironmentAccessible` or `EnvironmentKeyPath` or something similar might be better fit?

Sources/Apodini/Semantic Model Builder/SemanticModelBuilder.swift
```diff
@@ -89,7 +92,7 @@ class SemanticModelBuilder: InterfaceExporterVisitor {
                     return AnyHandlerIdentifier(handlerIndexPath.rawValue)
                 }
             }(),
-            handler: handler,
+            handler: handler.inject(app: app),
```
PSchmiedmayer: I would move this to the creation of the `guards` and `responseTransformers`, it seems a bit hidden in the initializer here and I almost missed it.

nityanandaz (approved):
Looks good to me 👍

Sources/Notifications/Application+NotificationCenter.swift
PSchmiedmayer: As far as I can see the `NotificationCenter` instance is recreated every time. This is to avoid side effects? Could it also be stored in the storage or am I missing something here?

PSchmiedmayer (commented):
I looked over the diff since my last comment and only had a few smaller comments from my side. I have also re-requested a review from the @Apodini/apodini team 👍

Sources/Apodini/Application/Application+FCM.swift
awoc: Yes, this also addressed by another branch. However, it requires to have a separate `VaporTarget` first before moving it to the `Notifications` target

PSchmiedmayer: Sounds good 👍

Sources/Notifications/Application+NotificationCenter.swift
awoc: I will also store in storage

moritzschuell (approved):
Nice work 🙌 Looks good to me.

PR #183, opened 2021-01-26
Allow exporters to listen for changes of observed objects via ConnectionContext

moritzschuell:
# Allow exporters to listen for changes of observed objects via `ConnectionContext`

`Handler`s have the `collectObservedObjects()` function, however exporters work with the `ConnectionContext`. Exporters do not have direct access to the handler of the context, and as a consequence cannot listen to changes of observed objects in the handler.

## :bulb: Proposed solution
This PR adds a small protocol `ObservedListener` and modifies the `ConnectionContext` protocol, such that exporters can register a listener to a context to get informed about changes in the context's handler.

### Problem that is solved
The ability of exporters to listen for changes in observed objects is needed for the implementation of the service-side communicational pattern.

### Implications
No implications for exporters.
Only adds the option for exporters to register a listener at a `ConnectionContext`.

### Testing
Tests are included that test the registration of an `ObservedListener` at a `ConnectionContext`, as well as whether the `onObservedDidChange` callback gets called properly. 

### Reviewer Nudging
First take a look at the new `ObservedListener` protocol and the changes to `ConnectionContext`. Then, the test case should explain how the listener can be used by an exporter. That's it 🤓


theMomax:
I am pretty sure this won't work as implemented right now...

Currently all the `Wrapper`s are initialized with their parents (the property wrappers `@Published` and `@ObservedObject`). Thus all these wrappers are shared across all copies of the property wrappers and will overwrite each other's `valueDidChange` callbacks. Same applies for the `changed` property. I suggest we take a look at it together.

theMomax:
I think this is ready for review now. Even though the gRPC exporter does not incorporate the new API yet, the API is complete and functional. The WebSocket exporter was adapted to work with `ObservedObject`, so there is a working PoC. There is also a simple end-to-end test for the WebSocket exporter demonstrating that `ObservedObject`s indeed emit events that result in messages to the client.

Sources/Apodini/Properties/Published.swift
```diff
@@ -1,40 +1,42 @@
 /// Property wrapper that can be used to annotate properties inside of `ObservableObject`s.
 /// The `ObservableObject` will notify its subscribers if a `Published` property changes.
 @propertyWrapper
-public struct Published<Element>: Property {
-    private var wrapper: Wrapper<Element>
-    private var wrappedValueDidChange: Wrapper<(() -> Void)?>
+public class Published<Element>: Property {
```
awoc: Doesn't changing this property wrapper to a class violate the condition that all objects conforming to `Property` have to be a struct? However, I agree that it makes sense to declare it as a class.

Sources/Apodini/Properties/ObservedObject.swift
```diff
@@ -7,17 +7,16 @@ import Foundation
...
     public var wrappedValue: Element {
         get {
             if let element = element {
                 return element
             }
-            if let objectIdentifer = objectIdentifer,
+            if let objectIdentifer = objectIdentifier,
                let element = EnvironmentValues.shared.values[objectIdentifer] as? Element {
                 return element
             }
```
awoc, line 22: Does the bug with the missing `ObjectIdentifier` in in the environment still occur or is there an update?

Sources/Jobs/Schedule.swift
```diff
@@ -17,6 +17,10 @@ public class Schedule<K: KeyChain, T: Job>: Configuration {
     ///     - on: Crontab as a String.
     ///     - keyPath: Associates a `Job` for later retrieval.
     public init(_ job: T, on cronTrigger: String, _ keyPath: KeyPath<K, T>) {
+        // Activate any `ObservedObject`s on the job.
+        var job = job
+        activate(&job)
```
awoc, line 22: I think it is better to remove these lines an keep only one activation step in the `Scheduler`. 

I'm not really sure why I already add the job to the `Environment` in the initializer. I will fix this in #180

Sources/Jobs/Schedule.swift
```diff
@@ -33,6 +37,10 @@ public class Schedule<K: KeyChain, T: Job>: Configuration {
     ///     - runs: Number of times a `Job` should run.
     ///     - keyPath: Associates a `Job` for later retrieval.
     public init(_ job: T, on cronTrigger: String, runs: Int, _ keyPath: KeyPath<K, T>) {
+        // Activate any `ObservedObject`s on the job.
+        var job = job
+        activate(&job)
```
awoc, line 42: Same here

Sources/Jobs/Scheduler.swift
```diff
@@ -108,11 +113,11 @@ private extension Scheduler {
...
     func subscribe(job: Job, to observedObject: AnyObservedObject) {
-        observedObject.valueDidChange = {
-            observedObject.setChanged(to: true)
+        self.observations.append(observedObject.register {
+            observedObject.changed = true
             job.run()
-            observedObject.setChanged(to: false)
-        }
+            observedObject.changed = false
+        })
```
awoc, line 120: I don't think that it is necessary to store all observations in the `Scheduler`

awoc (commented):
I like the changes to `@ObservedObject` and `@Published` and think that they're really helpful. I just left some smaller comments

Sources/Apodini/Properties/Published.swift
theMomax: You're right, `Published` shouldn't even be a `Property` as it isn't handled using the functionality of `Traversable.swift`.

Sources/Apodini/Properties/ObservedObject.swift
theMomax: It is gone. We actually forgot to add the according element to the environment🙈. It seemed so buggy because it was added asynchronously afterwards and thus was present at the time one would start debugging.

Sources/Jobs/Schedule.swift
theMomax: Don't think so...if I remove it, tests fail.

Sources/Jobs/Scheduler.swift
theMomax: Yes, it is necessary. Otherwise the `Observation` isn't kept in memory and the registered callback is lost. You can observe this by running the tests and placing a breakpoint on `job.run()`. It won't be executed when the `observations.append` was removed. For some reason your tests still pass if removing, so the tests aren't strict enough.

Sources/Apodini/Properties/Published.swift
theMomax: fixed with 4ef4e85

Sources/Jobs/Schedule.swift
theMomax: same here😉

Sources/Jobs/Scheduler.swift
theMomax: I modified on job-test so it actually checks if the job is run in 8631127.

Sources/Jobs/Schedule.swift
awoc: I will handle this on my branch 😄

awoc (approved):

theMomax:
I agreed with @moritzschuell it would be best to move the gRPC integration to another PR as it is not that small of a task. Thus I am merging now.

Issue #177, opened 2021-01-24
[MINOR FEATURE] Allow handle() to return Void


hendesi:
**Is your feature request related to a problem? Please describe.**
As found in [thread](https://github.com/Apodini/Apodini/pull/70/files/00759a1d3bf1f57b10ed377218010dd67d9848e0#r559817626), there is currently no possibility to return an `EventloopFuture<Void>` from a `handle()` function. This is due to `Void` being unable to conform to `ResponseTransformable`. Adding the conformance or a workaround would be beneficial for handlers that offer functionality which does not need to return something (e.g. `Delete` handler in the database target) 

**Propose a solution**
We could think of transforming `Void` to `.final(Void())` to be able to return it as a `Response`. Another option for the handlers, not really recommended though, would be to return `.nothing`. This is not really recommended as this does not create any response at all to the client which can be misleading. 

3. Option: Define a struct that conforms to `Encodable` and use/return it as a type alias for `Void`. This would allow us to have a clear and unified way, we and developers can use to indicate that a response is empty.

nerdsupremacist:
We need to think about how void would be represented in GraphQL, since GraphQL requires that every field returns a value. So void doesn't really exist. Perhaps we can represent it as an enum with a single case?

hendesi:
@PSchmiedmayer @Supereg 
Over the weekend I experimented a lot with refactoring the `Response` in order to support void. Also with looking at the [refactoring](https://gist.github.com/PSchmiedmayer/b26e268f3eee5092d6cb419abf68de98) proposed by Paul. 
While refactoring of `Response` was quite doable, I seem to get stuck on `ResponseTransformable`. If we want to be able to return `EventloopFuture<Void>` from handle(), we need to move away from constraining `ResponseTransformable.Content` to `Encodable`. As the internal structure relies heavily on expecting a encodable type, this would mean to refactor basically every exporter and context to manually check if content is conforming to encodable. This would still mean that at some point we lose the `Void` type and convert into something encodable. 

Although I spent quite a few hours on this...if you have any comments or tip or I missed something there completely, please let met know :D 

Since for now the only real use case that would profit from this refactoring is the `Delete` handler and there is not much time left in this project, I would propose this solution (similar to 3. option above): 
```
public struct NoContent: Encodable {
    var status: HTTPResponseStatus
    
    public init(_ status: HTTPResponseStatus) {
        self.status = status
    }
    
    public static var ok: NoContent {
        NoContent(.ok)
    }
}
[...]

PR #165, opened 2021-01-20
OpenAPI Specification Output Endpoint and Swagger-UI

lschlesinger:
# Configurable OpenAPIExporter Endpoint(s)

## :recycle: Current situation

Currently the OAS is served on a dedicated (vapor) endpoint `/openapi` in json format. 
As of now, only json export of OAS is supported.

## :bulb: Proposed solution
The objective is to provide options:

- [x] if the result should be served as json or yaml (both supported by OAS 3.0) -> select **output format**
- [x] if and where the output file should be served -> specify **URL for plain document**
- [x] if and where swagger-ui should be served -> specify **URL for swagger-ui**

Therefore, `OpenAPIConfiguration` was extended by these properties with default values:

```swift
var outputFormat: OpenAPIOutputFormat = .JSON
var outputEndpoint: String = "openapi"
var swaggerUiEndpoint: String = "openapi-ui"
```

### Question
**Ideas for how/where to set the  from outside are welcome :blush:**

### Problem that is solved
The PR tackles issue #107.

### Implications
1. Developer can specify OpenAPI specific parameters via theses properties.
2. One dedicated endpoint (e.g. `/openapi`) serves the specification in the respective format `.json` or `.yaml`
2.1 I had to include the package [YAMS]("https://github.com/jpsim/Yams.git") as dependency in the Apodini target.
3. One dedicated endpoint (e.g. `/openapi-ui`) serves the specification in `swagger-ui` allowing the developer to inspect and explore the API as one would expect when creating an OAS:

4. As suggested by @lukaskollmer I added `OpenAPIStorageKey` to make `OpenAPI.Document` accessible throughout the Apodini application.

5. Since 053f38a, `OpenAPIConfiguration: Configuration` enables the developer to define OpenAPI specific parameters. This can be achieved as shown below:

```swift
struct TestWebService: Apodini.WebService {
    @PathParameter var userId: Int
    
    var content: some Component {
        // Hello World! 👋
        Text("Hello World! 👋")
            .response(EmojiTransformer(emojis: "🎉"))
    }
    
    var configuration: Configuration {
        OpenAPIConfiguration(
            outputFormat: .json,
            outputEndpoint: "oas",
            swaggerUiEndpoint: "oas-ui",
            title: "The great TestWebService - presented by Apodini")
    }
}

try TestWebService.main()
```

## :heavy_plus_sign: Additional Information

What should be addressed in a future PR:
- Possible alignment with other exported (documentation) files such as the file created by `ProtobufferBuilder`?

[...]

PR #184, opened 2021-01-26
Graphql Interface Exporter

sadikekin:
# GraphQL Interface Exporter

Currently, nothing exports graphQL code.

## :bulb: Proposed solution
The tree is generated from the path and used for the GraphQL Schema creation

### Problem that is solved
Prior to this PR, people couldn't generate GraphQL code. By this branch, I am aiming to export GraphQL responses that answer users' queries.

### Implications
-

-

### Related PRs
-

### Testing
-

### Reviewer Nudging
GraphQLInterfaceExporter.swift is a good point to start. After, GraphQLSchemaBuilder.swift is suggested


Sources/Apodini/Semantic Model Builder/GraphQL/GraphQLInterfaceExporter.swift
```diff
@@ -0,0 +1,150 @@
...
+    case invalidResponse
+    case noBody
+}
+
+
+class GraphQLInterfaceExporter: InterfaceExporter {
+
+
+    // GraphQL Schema
+    private var schema = try! GraphQLSchema(
```
nerdsupremacist: Do we use this initial version? You seem to always override it in finish exporting. I'd suggest making it optional and throwing an error if the schema is never finished exporting

Package.swift
```diff
@@ -4,154 +4,158 @@ import PackageDescription
...
-            ],
-            exclude: [
-                "Helper/mock_fcm.json",
-                "Helper/mock_invalid_fcm.json",
-                "Helper/mock.p8",
-                "Helper/mock.pem"
-            ]
-        )
-    ]
+        name: "Apodini",
```
nerdsupremacist: What's the change here? indentation level?

.idea/Apodini.iml
```diff
@@ -0,0 +1,2 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
nerdsupremacist: Let's add all the .idea stuff to the repo's gitignore

Sources/Apodini/Semantic Model Builder/GraphQL/GraphQLInterfaceExporter.swift
```diff
@@ -0,0 +1,150 @@
...
+
+        // Create Swift dict
+        var query = "{}"
+        let data = body.data(using: .utf8)!
+        do {
+            if let jsonArray = try JSONSerialization.jsonObject(with: data, options: .allowFragments) as? [String: AnyObject] {
+                if let q = jsonArray["query"] as? String {
+                    query = q
+                } else {
+                    print("jsonArray doesn't have query parameter")
```
nerdsupremacist: please restrain from prints in the exporter code

@PSchmiedmayer do we already have a common logger in place that can be accessed from the exporter?

nerdsupremacist: If the query isn't present we should throw an error

Sources/Apodini/Semantic Model Builder/GraphQL/GraphQLInterfaceExporter.swift
```diff
@@ -0,0 +1,150 @@
...
+    private func graphqlServer(_ req: Vapor.Request) throws -> EventLoopFuture<String> {
+        guard let body = req.body.string else {
+            throw NetworkError.noBody
+        }
+
+        // Create Swift dict
+        var query = "{}"
+        let data = body.data(using: .utf8)!
+        do {
+            if let jsonArray = try JSONSerialization.jsonObject(with: data, options: .allowFragments) as? [String: AnyObject] {
```
nerdsupremacist: Please use `Decodable` with the `JSONDecoder` to retrieve the query

Sources/Apodini/Semantic Model Builder/GraphQL/GraphQLInterfaceExporter.swift
```diff
@@ -0,0 +1,150 @@
...
+        return Vapor.Response(status: .ok, headers: ["Content-Type": "text/html"], body: html)
+    }
+
+    private func graphqlServer(_ req: Vapor.Request) throws -> EventLoopFuture<String> {
+        guard let body = req.body.string else {
+            throw NetworkError.noBody
+        }
+
+        // Create Swift dict
+        var query = "{}"
```
nerdsupremacist: I'm pretty sure you can leave this unintialized, since there's shouldn't be a case where we run a query without there being a query in the request

Sources/Apodini/Semantic Model Builder/GraphQL/GraphQLInterfaceExporter.swift
```diff
@@ -0,0 +1,150 @@
...
+        } catch let error as NSError {
+            print(error)
+            throw NetworkError.noBody
+        }
+
+        return try graphql(schema: self.schema,
+                request: query,
+                context: req,
+                eventLoopGroup: req.eventLoop).map { result -> String in
+            result.description
```
nerdsupremacist: We should return JSON here. I understand that `.description` is the string representation of JSON, but Vapor can set things like the content-type better if we give it a `ResponseEncodable` type and use JSON to encode the data

Sources/Apodini/Semantic Model Builder/GraphQL/GraphQLInterfaceExporter.swift
```diff
@@ -0,0 +1,150 @@
...
+            } else {
+                print("bad json")
+                throw NetworkError.noBody
+            }
+        } catch let error as NSError {
+            print(error)
+            throw NetworkError.noBody
+        }
+
+        return try graphql(schema: self.schema,
```
nerdsupremacist: you don't need `self`. It's considered a good practice to not write `self` unless absolutely necessary

Sources/Apodini/Semantic Model Builder/GraphQL/GraphQLInterfaceExporter.swift
```diff
@@ -0,0 +1,150 @@
...
+
+    func exportParameter<Type: Codable>(_ parameter: EndpointParameter<Type>) -> String {
+        parameter.name
+    }
+
+    func finishedExporting(_ webService: WebServiceModel) {
+        self.schema = self.graphQLPath.generate()
+    }
+
+    func retrieveParameter<Type: Decodable>(_ parameter: EndpointParameter<Type>, for request: GraphQLRequest) throws -> Type?? {
```
nerdsupremacist: Why double optional? Does `.none` here mean something else than `.some(.none)`

Sources/Apodini/Semantic Model Builder/GraphQL/GraphQLInterfaceExporter.swift
```diff
@@ -0,0 +1,150 @@
...
+        if let queryArgDict = request.args.dictionary,
+           let val = queryArgDict[parameter.name] {
+
+            if (val.isNumber) {
+                return val.int as? Type
```
nerdsupremacist, line 89: Ok now I see what you meant in the work meeting. 
Here's where we'll have to do some work to support: `UInt`, as well as `Enums` and `InputObjects`

Sources/Apodini/Semantic Model Builder/GraphQL/GraphQLSchemaBuilder.swift
```diff
@@ -0,0 +1,248 @@
...
+    if (type == String.self) {
+        return GraphQLString
+    } else if (type == Int.self || type == UInt.self) {
+        return GraphQLInt
+    } else if (type == Float.self) {
+        return GraphQLFloat
+    } else if (type == Bool.self) {
+        return GraphQLBoolean
+    }
+    return GraphQLString
```
nerdsupremacist: I'm not so sure about this? I feel like we could end up marking a lot of types as Strings and not even notice until they crash.
It's always best to fail as early as you can tell somethings wrong

Sources/Apodini/Semantic Model Builder/GraphQL/GraphQLSchemaBuilder.swift
```diff
@@ -0,0 +1,248 @@
+//
+// Created by Sadik Ekin Ozbay on 06.01.21.
+//
+
+@_implementationOnly import Vapor
+import GraphQL
```
nerdsupremacist: Please use @_implementationOnly for GraphQL too

Sources/Apodini/Semantic Model Builder/GraphQL/GraphQLSchemaBuilder.swift
```diff
@@ -0,0 +1,248 @@
+//
```
nerdsupremacist, line 1: The spacing in this file is a bit inconsistent. Please only one empty line between functions or methods (unless you have a Mark comment)

Sources/Apodini/Semantic Model Builder/GraphQL/GraphQLSchemaBuilder.swift
```diff
@@ -0,0 +1,248 @@
...
+        for c in responseTypeHead.children {
+            if let propertyInfo = c.value.propertyInfo {
+                currentFields[propertyInfo.name] = GraphQLField(type: responseTypeHandler(for: c), resolve: { source, args, context, info in
+                    return try propertyInfo.get(from: source)
+                })
+            }
+        }
+
+        let typeName = responseTypeHead.value.typeInfo.name
+        return try! GraphQLObjectType(name: typeName, fields: currentFields)
```
nerdsupremacist: Don't force try. It's better to throw the error. Then someone can attempt to catch it gracefully. 
For example this statement will crash if someone uses an Empty struct

Sources/Apodini/Semantic Model Builder/GraphQL/GraphQLSchemaBuilder.swift
```diff
@@ -0,0 +1,248 @@
...
+            }
+            hasIncomingEdge.insert(child)
+            indx -= 1
+        }
+
+
+    }
+
+    private func nameExtractor(for node: String) -> String {
+        node.components(separatedBy: "_").filter({ $0 != "" }).last ?? "None"
```
nerdsupremacist: Again, not sure if that "None" default is a good idea

Sources/Apodini/Semantic Model Builder/GraphQL/GraphQLSchemaBuilder.swift
```diff
@@ -0,0 +1,248 @@
...
+            if let responseType = self.responseTypeTree[nodeName], let responseContext = self.leafContext[nodeName] { // It has handler
+                let fieldName = self.graphQLRegexCheck(for: responseType.value.typeInfo.name.lowercased())
+                currentFields[fieldName] = self.graphQLFieldCreator(for: responseType,
+                        responseContext,
+                        self.args[nodeName] ?? [:])
+            }
+
+            for child in childrenList {
+                let childName = self.graphQLRegexCheck(for: self.nameExtractor(for: child)) // child.components(separatedBy: "_").filter({ $0 != "" }).last ?? "None"
+                // if let childrenList = self.tree[node] {
```
nerdsupremacist: always remove commented out code

Sources/Apodini/Semantic Model Builder/GraphQL/GraphQLSchemaBuilder.swift
```diff
@@ -0,0 +1,248 @@
...
+                currentFields[childName] = generateSchemaFromTreeHelper(child)
+            }
+
+            let checkedNodeName = self.graphQLRegexCheck(for: nodeName)
+            self.types[checkedNodeName] = try! GraphQLObjectType(name: checkedNodeName, fields: currentFields)
+            return GraphQLField(type: self.types[checkedNodeName]!, resolve: { _, _, _, _ in "Emtpy" })
```
nerdsupremacist: I'm a bit lost. When does this happen? And why are you returning "Empty"?

nerdsupremacist (changes requested):

Sources/Apodini/Utilities/AnyEncodable.swift
```diff
@@ -9,7 +9,7 @@ import Foundation
 
 
 public struct AnyEncodable: Encodable {
-    private let wrappedValue: Encodable
+    public let wrappedValue: Encodable
```
PSchmiedmayer, line 12: As discussed in the last meeting there is no need to make the wrappedValue public. If you know the desired type you can access the `typed` function you find below.

Sources/Apodini/WebService.swift
```diff
@@ -39,23 +40,23 @@ extension WebService {
...
 
         webService.register(
-            SharedSemanticModelBuilder(app)
-                .with(exporter: RESTInterfaceExporter.self)
-                .with(exporter: WebSocketInterfaceExporter.self)
-                .with(exporter: OpenAPIInterfaceExporter.self)
-                .with(exporter: GRPCInterfaceExporter.self)
-                .with(exporter: ProtobufferInterfaceExporter.self),
-            GraphQLSemanticModelBuilder(app)
+                SharedSemanticModelBuilder(app)
```
PSchmiedmayer: Please remove the intention diff here and on the other files that were affected by your formatting in the IDE.

Sources/TestWebService/RandomSubsystem/Components/RandomComponent.swift
```diff
@@ -8,7 +8,7 @@
 import Apodini
 
 
-struct RamdomComponent: Component {
+struct RandomComponent: Component {
```
PSchmiedmayer: 👍

Sources/Apodini/Semantic Model Builder/GraphQL/GraphQLInterfaceExporter.swift
```diff
@@ -0,0 +1,150 @@
...
+                                                  .then(response => response.json())
+                                                  .catch(() => response.text());
+                                              ReactDOM.render(
+                                                React.createElement(GraphiQL, { fetcher: graphQLFetcher }),
+                                                document.getElementById('graphiql'),
+                                              );
+                                            </script>
+                                          </body>
+                                        </html>
+                                        """
```
PSchmiedmayer: Please use resources to store .html files or other content. You can take a look at #165 and how the .html file is stored as a resource there.

Sources/Apodini/Semantic Model Builder/GraphQL/GraphQLInterfaceExporter.swift
```diff
@@ -0,0 +1,150 @@
...
+    let app: Vapor.Application
+    let graphQLPath: GraphQLSchemaBuilder
+
+    required init(_ app: Application) {
+        self.graphQLPath = GraphQLSchemaBuilder()
+        self.app = app.vapor.app
+
+        // For Query
+        self.app.post("graphql", use: self.graphqlServer)
+        self.app.get("graphql", use: self.graphQLIDE)
```
PSchmiedmayer: I would suggest to move this function to the `finishedExporting` method. Even though it doesn't really make a difference I would suggest to register the routes once all the functionality of the `interfaceExporter`.

Sources/Apodini/Semantic Model Builder/GraphQL/GraphQLInterfaceExporter.swift
```diff
@@ -0,0 +1,150 @@
...
+@_implementationOnly import Vapor
+import GraphQL
+
+enum NetworkError: Error {
+    case unauthorised
+    case timeout
+    case serverError
+    case invalidResponse
+    case noBody
+}
```
PSchmiedmayer: I would suggest to use the `ApodiniError` as much as possible. You you need to provide any additional input you can define options for an `ApodiniError` as the WebSocket exporter does. As far as I can see the errors thrown here could all be mapped to `ApodiniError` cases that are already part of Apodini. This also reduced the error handling need you will have to take a look at because these errors could be the same ones as thrown by the Apodini developer.

Sources/Apodini/Semantic Model Builder/GraphQL/GraphQLInterfaceExporter.swift
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    private func graphqlServer(_ req: Vapor.Request) throws -> EventLoopFuture<String> {
+        guard let body = req.body.string else {
+            throw NetworkError.noBody
+        }
+
+        // Create Swift dict
+        var query = "{}"
+        let data = body.data(using: .utf8)!
```
PSchmiedmayer: Please use guard statements with appropriate errors so you don't produce runtime crashes when handling requests.

Sources/Apodini/Semantic Model Builder/GraphQL/GraphQLInterfaceExporter.swift
PSchmiedmayer: If you use the JSONDecoder you can also reduce the different error catches you currently have in the code. You will probably need to create a `Query` struct that can be decoded here 👍

Sources/Apodini/Semantic Model Builder/GraphQL/GraphQLInterfaceExporter.swift
```diff
@@ -0,0 +1,150 @@
...
+            )
+    )
+
+
+    let app: Vapor.Application
+    let graphQLPath: GraphQLSchemaBuilder
+
+    required init(_ app: Application) {
+        self.graphQLPath = GraphQLSchemaBuilder()
+        self.app = app.vapor.app
```
PSchmiedmayer: I would suggest to save the app instance itself instead of only saving the `Vapor` app instance here. The `Apodini` app instance will enable you to also access the `Logger` and so on.

Sources/Apodini/Semantic Model Builder/GraphQL/GraphQLInterfaceExporter.swift
PSchmiedmayer: Yes, as mentioned in my comment in the initializer I would suggest to save the `Apodini.Application` instance instead of `Vapor.Application`. The `Apodini.Application` allows you to directly access our global `Logger` with ease.

Sources/Apodini/Semantic Model Builder/GraphQL/GraphQLInterfaceExporter.swift
```diff
@@ -0,0 +1,150 @@
...
+            print(error)
+            throw NetworkError.noBody
+        }
+
+        return try graphql(schema: self.schema,
+                request: query,
+                context: req,
+                eventLoopGroup: req.eventLoop).map { result -> String in
+            result.description
+        }
```
PSchmiedmayer: Please try to match the formatting we use in Apodini:
```suggestion
        try graphql(
            schema: self.schema,
            request: query,
            context: req,
            eventLoopGroup: req.eventLoop
        )
            .map { result -> String in
                result.description
            }
```
or alternatively
```suggestion
        try graphql(schema: self.schema,
                    request: query,
                    context: req,
                    eventLoopGroup: req.eventLoop)
            .map { result -> String in
                result.description
            }
```

Sources/Apodini/Semantic Model Builder/GraphQL/GraphQLInterfaceExporter.swift
PSchmiedmayer: I think that the current implementation is missing a lot of different edge cases that must be covered here. E.g. complex structs or elements can not be decoded in the moment. I think it would be interesting how the decoding for more complex types would work.
I think we will need to collect what parameters are possible for each endpoint or need to extend the `InterfaceExporter` to get all Parameters at once and return a value for each one step by step. E.g. it make a difference if there is only one parameter (e.g. one `Int`) and therefore we can use the top level value or of we have two parameters (`name: String`, `age: Int`) and therefore need to search for a dictionary that contains String keys and different ambiguous values that can be decoded for the corresponding keys. As far as I can see this is currently not done by any other exporter but #131 is approaching this for the Protocol Buffer generator.

Sources/Apodini/Semantic Model Builder/GraphQL/GraphQLInterfaceExporter.swift
```diff
@@ -0,0 +1,150 @@
+//
+// Created by Sadik Ekin Ozbay on 03.01.21.
+//
+
+@_implementationOnly import Vapor
+import GraphQL
```
PSchmiedmayer: Please use `@_implementationOnly` to import any new dependencies.

Sources/Apodini/Semantic Model Builder/GraphQL/GraphQLSchemaBuilder.swift
```diff
@@ -0,0 +1,248 @@
...
+            return GraphQLInt
+        }
+        if (type == Bool.self) {
+            return GraphQLBoolean
+        }
+        if (type == Float.self) {
+            return GraphQLFloat
+        }
+        return nil
+    }
```
PSchmiedmayer: Seems to me that is function is a duplicate of the function above but includes a slights variation on the type that can not be matched.

Sources/Apodini/Semantic Model Builder/GraphQL/GraphQLSchemaBuilder.swift
```diff
@@ -0,0 +1,248 @@
...
+
+    private func responseTypeHandler(for responseTypeHead: Node<EnrichedInfo>) -> GraphQLOutputType {
+        let typeValTemp = self.typeToGraphQL(type: responseTypeHead.value.typeInfo.type)
+
+        if let typeVal = typeValTemp {
+            return typeVal as! GraphQLOutputType
+        }
```
PSchmiedmayer: I think you can remove the force cast here:
```suggestion
        if let typeVal = typeValTemp as? GraphQLOutputType {
            return typeVal
        }
```

Sources/Apodini/Semantic Model Builder/GraphQL/GraphQLSchemaBuilder.swift
```diff
@@ -0,0 +1,248 @@
...
+
+
+        // Array / Optional
+        if let newResponseHead = try! responseTypeHead.edited(handleArray)?.edited(handleOptional) {
```
PSchmiedmayer: I would suggest to make the overall function throwing to reduce the need to use `try!` in multiple places.

Sources/Apodini/Semantic Model Builder/GraphQL/GraphQLSchemaBuilder.swift
```diff
@@ -0,0 +1,248 @@
...
+            let vaporRequest = gContext as! Vapor.Request
+            let response: EventLoopFuture<Response<AnyEncodable>> = mutableContext.handle(request: request, eventLoop: vaporRequest.eventLoop.next())
+
+            return response.flatMapThrowing { encodableAction -> Any? in
+                switch encodableAction {
+                case let .send(element),
+                     let .final(element):
+                    return element.wrappedValue
+                case .nothing, .end:
+                    return ".nothing, .end"
```
PSchmiedmayer, line 106: We had a discussion today about handling `.nothing` or `.end`. @hendesi will approach this based on the issue #177

Sources/Apodini/Semantic Model Builder/GraphQL/GraphQLSchemaBuilder.swift
```diff
@@ -0,0 +1,248 @@
...
+        var currentPath = endpoint.absolutePath.map {
+            $0.description.lowercased()
+        }.filter {
+            $0.first != ":"
+        }
+
+        // Remove `root`
+        currentPath.removeFirst()
+        // Remove `v1`
+        currentPath.removeFirst()
```
PSchmiedmayer: I think that we should make this a bit more generic in the long term then just removing the first two path components as this might change in the future.

Sources/Apodini/Semantic Model Builder/GraphQL/GraphQLSchemaBuilder.swift
```diff
@@ -0,0 +1,248 @@
...
+
+        print("The current path is ->", currentPath)
```
PSchmiedmayer: Please use a logger throughout the exporter.

Sources/Apodini/Semantic Model Builder/GraphQL/GraphQLSchemaBuilder.swift
```diff
@@ -0,0 +1,248 @@
...
+        if (currentPath.count > 1) {
+            for ix in 0..<currentPath.count {
+                currentSum.append(currentPath[ix])
+                currentSum.append("_")
+                currentPath[ix] = currentSum
+            }
+        }
+
+        // Get leaf name
+        let leafName = currentPath.last ?? "None"
```
PSchmiedmayer: I would suggest to move the name creation into a separate function to make it clearer that this functionality es encapsulated. It helps to reduce the length and complexity of this function.

Sources/Apodini/Semantic Model Builder/GraphQL/GraphQLSchemaBuilder.swift
```diff
@@ -0,0 +1,248 @@
...
+        while (indx >= 1) {
+            let child = currentPath[indx], parent = currentPath[indx - 1]
+            if (self.tree.keys.contains(parent)) {
+                self.tree[parent]!.insert(child)
+            } else {
+                self.tree[parent] = [child]
+            }
+            hasIncomingEdge.insert(child)
+            indx -= 1
+        }
```
PSchmiedmayer: To be honest I am a bit lost about the functionality at the end of the function here. It would be great to split the function out into smaller functions or types. Maybe some comments would also help a bit.

Sources/Apodini/Semantic Model Builder/GraphQL/GraphQLSchemaBuilder.swift
```diff
@@ -0,0 +1,248 @@
...
+
+
+    }
+
+    private func nameExtractor(for node: String) -> String {
+        node.components(separatedBy: "_").filter({ $0 != "" }).last ?? "None"
+    }
+
+    // To handle `Names must match /^[_a-zA-Z][_a-zA-Z0-9]*$/`
+    private func graphQLRegexCheck(for str: String) -> String {
```
PSchmiedmayer, line 180: I would suggest to move this to a `String` extension.

PSchmiedmayer (changes requested):
@sadikekin Great to see the first bigger pull request for the GraphQL exporter! I took a look at the different sections of the code and added some suggestions for some improvements. I would suggest that you make a first iteration of the PR based on the feedback from @nerdsupremacist and myself. After this I would suggest to get a review from two random members of the `Apodini` team to get some additional feedback from your peers in a second iteration 👍

Looking forward to see the GraphQL functionality in Apodini soon!!🚀

Sources/Apodini/Semantic Model Builder/GraphQL/GraphQLInterfaceExporter.swift
bauer-andreas: The double optional is a requirement of the Exporter API.
See https://github.com/Apodini/Apodini/pull/81#discussion_r548919891.
We decided after a long call on this in order to distinguish between "value does not exist" (`nil`) and "nil is the value" `Optional.some(nil)` (for there later there exist a `.null` shortcut).

Package.swift
sadikekin: I think, also i added some libraries (graphQL and swiftJSON)

Sources/Apodini/Semantic Model Builder/GraphQL/GraphQLSchemaBuilder.swift
sadikekin: If they are in the middle of the tree

hendesi:
@sadikekin In case you didn't know, you can check for SwiftLint errors by running `swiftlint` in the terminal on your local machine. Then you don't have to commit to check if there are any errors. You can also run `swiftlint autocorrect` to automatically correct some of the violations 👍

PSchmiedmayer:
@sadikekin As I see that you are currently struggling with SwiftLint: You can also [install and execute SwiftLint locally and run it from the command line](https://github.com/realm/SwiftLint#using-homebrew). SwiftLint also has a `autocorrect` command that might also be helpful 👍
(edit: the exact same as @hendesi is saying 😄)

PR #187, opened 2021-01-27
fix default handler index paths

lukaskollmer:
# fix default handler index paths

Fixes a bug where the default-generated handler identifiers (ie their index paths) could contain negative elements.
The issue was that we missed a `visitor.enterComponentContext` call for non-custom components.

Example:
```swift
struct WrappedComponent<T: Component>: Component {
    let wrapped: T
    var content: T { wrapped }
}

    var content: some Component {
        WrappedComponent(WrappedComponent(WrappedComponent(Text())))
    }
}
```

Here, all `WrappedComponent`s would use the default visit implementation, which only calls `enterContent` but not `enterComponentContext`. If we replace the `WrappedComponent`s with pathless `Group`s or manually constructed `TupleComponent`s (all three options would be semantically equivalent) we'd get the correct behaviour since Group's and TupleComponent's accept implementations call both functions (`enterContent` and `enterComponentContext`).

### Testing
added a test to catch this in the future

### Reviewer Nudging
look at the code, run the example

### DIFF
--- a/Sources/Apodini/Components/Component/Component.swift
+++ b/Sources/Apodini/Components/Component/Component.swift
@@ -39,7 +39,9 @@ extension Component {
             HandlerVisitorHelperImpl(visitor: visitor)(self)
             if Self.Content.self != Never.self {
                 visitor.enterContent {
-                    content.accept(visitor)
+                    visitor.enterComponentContext {
+                        content.accept(visitor)
+                    }
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


theMomax (approved):
LGTM...it didn‘t break anything, so I guess it’s fine. I could imagine it was implemented like that because back then we thought any custom components would be what we now call Handler.

lukaskollmer:
@theMomax iirc the index paths were added after the Handlers, but yeah I don't think I ever really considered the use case of declaring custom `Component`s.

PSchmiedmayer:
@lukaskollmer Thanks 🚀

PR #192, opened 2021-01-30
Make the Exporter API public

bauer-andreas:
# Make the Exporter API public

Currently all InterfaceExporters live in the same core `Apodini` target. Thus it was fine to have the Exporter API to be internal.
With latest aims to move all exporters into their own targets, the Exporter API needs to part of the public interface.

## :bulb: Proposed solution
Make all "public" Exporter API actually public.

### Problem that is solved
This is the first step to allow exporters to live in their own target.
This should also @tgymnich to continue with moving Vapor Engine into its own target.

### Implications

* This PR also "flattens" the SemanticModelBuilder, so it properly resembles that we only have the one and only Semantic Model Builder.
* The `TypeReflection` API used by OpenAPI and Protobuffer Exporters was not yet touched. Would be great if one of you (@nityanandaz @lschlesinger) could do that, as you are more comfortable with the code base, and know better what is indeed for public use.

### Related PRs
--

### Testing
--

### Reviewer Nudging
--


bauer-andreas:
As of the additional overhead created by the AssociatedTypeRequirementsKit workaround, the code coverage was reduced slightly.

And I don't really know why the checks don't run 🤔

tgymnich (approved):
LGTM

bauer-andreas:
> LGTM

now I just need to figure out how to not crash amazonlinux and get the `-0.06` coverage back somehow 🙃

PSchmiedmayer:
@Supereg The Amazon Linux build has run successfully. I can merge the PR using my admin privileges if you want to 👍

bauer-andreas:
For me the PR is in a weird state where the checks seemingly run successfully, but show as unreported in the PR.
But yes if you liked to use your admin powers it would be great.

PSchmiedmayer:
I have update the codecov configuration and triggered all the builds again, let's see if everything passes now.

PR #193, opened 2021-01-31
Improve the GitHub Actions

PSchmiedmayer:
# *Name of the PR*

1. The amazonlinux TestWebService tests seemed to have failed once. From the build logs it looked like the linux instance the action was running on was taking longer then expected to start up the web service.
2. Some runs of the CI did not use the cache to speed up the build process
3. The Codecov action did not report the test results reliably 

## :bulb: Proposed solution
This PRs address the issues noted above:
1. This PR increased the test timeout
2. This PR improves the caching mechanism
3. This PR addresses some error logs found in the Codecov action


.github/workflows/test-webservice.yml
```diff
@@ -28,8 +29,8 @@ jobs:
         xcode-version: latest
     - uses: actions/cache@v2
       with:
-        path: .build
-        key: ${{ runner.os }}-spm-cachev2-${{ hashFiles('Package.resolved') }}
+        path: TestWebService/.build
```
bauer-andreas, line 32: Is the cache action not impacted by the `working-directory` definition?

bauer-andreas (approved):
looks good 👍 

Out of curiosity, which unit are those deadlines? minutes or seconds?

.github/workflows/test-webservice.yml
PSchmiedmayer: Yes, it seems like it. The previous version didn't work and the cache only worked once a used absolute paths undefended of the workspace path.

Regarding the timeout: It is defined seconds, I think about 5 seconds should be reasonable to start up the web service 🚀

bauer-andreas: Weird behavior 😅 

Would it make sense to increase that even further? Should boot in 5 seconds, but we also have seen GitHub Actions being really slow 🙃 But lets just see how it goes.

PSchmiedmayer: Yes, I could also imaging extending the timeout even more if this is needed. On the other hand this gives us some hint about changes that dramatically increase the startup time of the web service so I would suggest to leave it at the current timeout for now and increase it once we have the need for it or we notice that builds start failing due to a timeout.

.github/workflows/build-and-test.yml
```diff
@@ -15,17 +15,20 @@ jobs:
     name: macOS ${{ matrix.configuration }} test build
     runs-on: macos-latest
     strategy:
+      fail-fast: false
       matrix:
         configuration: [debug, release]
     steps:
     - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
```
tgymnich, line 24: @PSchmiedmayer why did you add this line? Is there a reason we need to fetch all history for all branches and tags?

PSchmiedmayer: @tgymnich There was a log error in the job that uploaded the codecov report that the action was requiring a fetch-depth of 0 or bigger then 1 and we had the issue that the report was not always reported in the PRs and after removing the log error the codecov reports continued to work again.

tgymnich: @PSchmiedmayer thanks for clarifying this!

PR #194, opened 2021-02-01
Database improvements

hendesi:
# Improvement Of Database Handlers

While implementing the web service for my sample project, I noticed that some convenience features are currently missing in Apodini. The things I noticed were the creation of multiple items and the reading of all items.

## :bulb: Proposed solution
Added a `CreateAll` handler which enables the creation of an array of objects. Furthermore I modified `ReadAll` to return all items in the db by default when no parameters are provided

### Problem that is solved
A User can now create multiple items with a single request and read all items when providing no parameters

### Related PRs
Just a small addition, so no related prs.

### Testing
Adjusted existing tests for the new functionality

### Reviewer Nudging
Should be fairly easy.

### DIFF
--- a/Sources/ApodiniDatabase/Handlers/Create/Create.swift
+++ b/Sources/ApodiniDatabase/Handlers/Create/Create.swift
@@ -13,4 +13,6 @@ public struct Create<Model: DatabaseModel>: Handler {
     public func handle() -> EventLoopFuture<Model> {
         object.save(on: database).transform(to: object)
     }
+    
+    public init() {}
 }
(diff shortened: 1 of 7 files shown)


Sources/ApodiniDatabase/Handlers/Create/Create.swift
```diff
@@ -13,4 +13,6 @@ public struct Create<Model: DatabaseModel>: Handler {
     public func handle() -> EventLoopFuture<Model> {
         object.save(on: database).transform(to: object)
     }
+    
+    public init() {}
```
awoc, line 17: I think the other Handlers like `Delete` and `ReadOne` would also benefit from an initializer

hendesi: Good point! I've added the missing initializers

awoc (approved):
Looks good 👍

PR #199, opened 2021-02-01
Replace Vapor.Abort with ApodiniError in NotificationCenter

awoc:
# Replace Vapor.Abort with ApodiniError in NotificationCenter

~~The update of the `APNS` and `APNSwift` package currently leads to an infinite loop when sending push notifications as described in this issue: https://github.com/kylebrowning/APNSwift/issues/100.~~ -> Fixed by #181 

Also, the `NotificationCenter` still uses `Abort` of Vapor when not finding objects in the database. The same behavior can now be expressed directly with Apodini by using `ApodiniError`.

## :bulb: Proposed solution

~~- By simpling adding a new attribute to `send` the problem is fixed. Described in: https://github.com/vapor/apns/pull/26~~
- Change `Abort(.notFound)` to `notFound` of `ApodiniError`

### Implications

Shouldn't have any implications.

### Testing

- Tests in the `NotificationCenter` were extended to capture `ApodiniError`.


hendesi (approved):
Looks good! 👍

PR #200, opened 2021-02-02
Protobuffer generation: handle UUID

nityanandaz:

Traced back to /Users/nitai/Developer/github/Apodini/Sources/Apodini/Semantic Model Builder/Protobuffer/Model+EnrichedInfo.swift:86

The concerned handler/service will not appear in the `.proto` file.
Every other handler/service is not affected.

## :bulb: Proposed solution

Represent `UUID` with a string.

### Problem that is solved

The handler/service is not excluded from the `.proto` file.

### Implications

None, really.

### Related PRs

#147; Ref.: https://github.com/Apodini/Apodini/pull/147#discussion_r567745845


bauer-andreas:
> Represent UUID with a string.

Are there any solutions other than encoding it as a string? Sounds like a lot of wasted space to me.
Is there any way to just store it as ~two~ plain bytes?

nityanandaz:
@Supereg I think that should be possible. With 16 bytes at least 🙂

bauer-andreas:
> 16 bytes at least

yeah, totally messed up the math 😂

nityanandaz:
@Supereg Actually, I would prefer to keep encoding with a string. For the sake of simplicity.

UUID would request other encoders to encode itself as a string as well. Making an exception would be _unnecessarily_ complex.

moritzschuell (approved):
Looks great, thank you! 
Since UUID also encodes itself as String, I think using String is the right choice. 👍

bauer-andreas (approved):

PR #204, opened 2021-02-02
Refactor NotificationCenter

awoc:
# Refactor NotificationCenter

This PR refactors some of the functionalities of the `NotificationCenter`. 

At the moment it is not checked if the push notification service (APNS or FCM) is configured before sending out a notification to it. This leads to an error when sending out a push notification to a `topic`. This method tries to send messages to all devices subscribed to a specific `topic`,  independent of the service. If only one service is used it would throw an error because the configuration for the other one is missing.

## :bulb: Proposed solution

Fix the described problem by adding computed properties that check if APNS or FCM was configured.

### Implications

- Before sending out push notifications, Apodini checks if the service was configured.
- All classes inside of `Device` conform now also to `Apodini.Content`.
- Restructures some methods.

### Testing

- Unfortunately, the correct sending of push notifications cannot be tested.
- The added database methods were tested.

### Reviewer Nudging

A good starting point is to start with the NotificationCenter` and the`NotificationCenter+Sending` extension.


awoc:
@PSchmiedmayer Unfortunately, I cannot test some of the methods of the `NotificationCenter` because it requires to send push notifications to a device. Would it be possible for you to still merge this branch to develop, if approved?

PSchmiedmayer:
@awoc Yes, would be fine with me. Did you manually test it e.g. using your example project to verify everything works?

awoc:
> @awoc Yes, would be fine with me. Did you manually test it e.g. using your example project to verify everything works?

Yes, for now I point my sample project to this branch.

PSchmiedmayer:
@awoc Great, let me know once this PR has the necessary approvals 👍

PR #206, opened 2021-02-03
Add the Ability to Pass `Status` Codes to a `Response`

### LINKED ISSUE #177: [MINOR FEATURE] Allow handle() to return Void

PSchmiedmayer:
# *Add the Ability to Pass `Status` Codes to a `Response`*

It is currently not possible to influence the status of a response for interface exporters that support this feature. In addition returning an `EventLoopFuture<Void>` or a `Void` type is currently not possible. Components as the `Delete` component therefore currently return an `Int` indicating the status code.

## :bulb: Proposed solution
This PR refactors the `Response` type to carry the `Status` code information and simplifies the handling of the `Response` for exporters by reducing the need to rely on a switch statement.

### Problem that is solved
Provides a solution and therefore closes #177 

### Implications
The public API of Apodini is completely unaffected. Exportes have been adapted to work with the new `Response` type.

### Testing
The test include new tests for the new Response type as well as adaptations to the `ResponseContainer` in the REST interface exporter.

### Reviewer Nudging
It would be interesting to expand the current range of status responses with additional status codes that would be useful. Do you have any suggestions?


awoc (commented):
The changes look great to me and are really helpful when creating a web service.

If I understand this correctly one would write `func handle() -> Response<Empty>` instead of just `Void` in `Handler`.

Maybe we could think about adding the status code 206 for partial content when using `send` for a server-side stream or a file transfer.

PSchmiedmayer:
@awoc Yes, `Response<Empty>` would be one option. While reading your comment I also had the idea to make `Status` directly conform to `ResponseTransformable` and therefor you can also use `func handle() -> Status` now 👍

I think that `206` partial content would be an interesting addition. I could even imagine that the status code as some associated values (e.g. the range and the total number of bytes). I am not 100% convinced if this is something that should be directly exposed to the developer. It could also be an internal thing that e.g. only the REST Exporter does when seeing a large chunk of data that is responded. The gRPC exporter already automatically does this by forwarding this to Vapor which automatically splits up a response. Nevertheless I will keep it in the back of my mind 💪

awoc (approved):
Sounds great 👍

hendesi (approved):
Many thanks for taking this over! Looks really cool! I just have 2 comments below, otherwise I think it's good to go 👍

Sources/Apodini/Response/Status.swift
```diff
@@ -0,0 +1,24 @@
...
+import NIO
+
+
+/// An `Status` expresses additional information that can be passed to a `Response`
+public enum Status: ResponseTransformable {
+    /// The request was handled and the response contains the expected content
+    case ok
+    /// The request was handled and a new resource has been created
+    case created
+    /// The request was handled and the response does not contain any content
```
hendesi, line 17: Maybe we could add a `delete` case here, although I think this is normally already handled by 200 or 204. Im not 100% sure, if this falls more into the error section, but what about an 304, if an update did not succeed

Sources/XCTApodini/Response+Checks.swift
```diff
@@ -0,0 +1,174 @@
...
+import XCTest
+
+extension Empty: Equatable {
+    public static func == (lhs: Self, rhs: Self) -> Bool {
+        true
+    }
+}
+
+/// Adds the possibility to easily check the of a `Handler` by investigating the `Response`
+/// - Parameters:
```
hendesi: I think there is a word missing :). Applies to all documentation in this file.

Sources/Apodini/Response/Status.swift
PSchmiedmayer: If we think about HTTP status codes this is already handled by the ones named above. Maybe we should generally think about more general status codes (e.g. `success` instead of `ok`) and some additional expressions that might be eventually matched to the same status code. For now I would suggest to remain with the three here as the `deleted` case would easily be covered with the `ok` `Status`.

An update that has been failed should be indicated by an error code. HTTP Status code 304 is mostly used to indicate that the resource requested did not change since the last request and the cache on the client is still valid (as e.g. indicated by the ETag header value). I think that a failed update is already handled by the Apodini Error. Nevertheless this might be something that the `Database` handlers might want to indicate using the `@Throws` property wrapper.

Sources/XCTApodini/Response+Checks.swift
PSchmiedmayer: Thanks a lot, I will fix this 👍

PR #207, opened 2021-02-04
GRPC bidirectional streaming

moritzschuell:
# GRPC bidirectional streaming

Currently, the GRPC exporter only supports 2 (unary and client-streaming) of Apodini's 4 communicational patterns.
This PR adds support for bidirectional streaming, which enables all 4 of the patterns.

## :bulb: Proposed solution
All 4 communicational patterns are now exposed using the same functionality. Everything (i.e. every request and response) is always treated as a stream.
This is, because unary HTTP2 requests (and responses) are basically just streams that only deliver one data frame directly followed by an end frame.

This PR also adds support for `ObservableObjects` to the GRPC exporter. 
I tried implementing the synchronization between evaluations of observations and client-frames the same way as the WebSocket exporter does (which is via the `Publisher.syncMap(..)`). However, because of how the response-streaming works, this is not suitable for my use-case (the response-stream needs to subscribe to the publisher first so that no events will get lost, but the subscription must already exist to build the response-stream callback ➡️ Henne-Ei Problem that I wasn't able to solve).
Hence, I implemented the synchronization using a `DispatchQueue`.
I would be happy about feedback on this approach.

### Implications
The existing GRPC implementations for unary and client-streaming endpoints have been removed.
Other exporters and functionality is not affected.

### Related PRs
None. You could look at the `WebSocketInterfaceExporter` for comparison, if interested.

### Testing
The existing test cases have been adjusted.
A new test case is added to test the bidirectional streaming functionality.
`Vapor.Response` does not offer a way to drain chunked responses frame by frame on the client-side. 
Thus, the new test case collects all frames received from the server and only then asserts (which is not really the "actual" way one would use the response generated from a service-streaming endpoint).
Hints on how to better test this are welcome 🤓

### Reviewer Nudging
I think it's best to directly start with `GRPCServiceHandler.swift`. This is where the main changes are. All the other changes are just adjustments to the new bidirectional handler.


Sources/Apodini/Semantic Model Builder/gRPC/GRPCService/GRPCServiceHandler.swift
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    private func handleCompletion(request: Vapor.Request,
+                                  context: inout ConnectionContext<GRPCInterfaceExporter>,
+                                  responseWriter: BodyStreamWriter,
+                                  serviceStreaming: Bool,
+                                  lastMessage: GRPCMessage) {
+        if serviceStreaming {
+            responseWriter.write(.end, promise: nil)
+        } else {
```
theMomax: Does that mean that the GRPC Exporter skips evaluating service-side streaming `Handler`s one last time with `Connection.state == .end`?

theMomax (changes requested):
There is one specific code-level question below. Other than that I think you have a conceptual problem:

You only synchronize normal code execution, not `EventLoopFutures`, i.e. **async** code. Synchronizing normal code is not even necessary if it belongs to the same client as it is executed on the same `EventLoop` (and thus thread) (see [here](https://docs.vapor.codes/4.0/async/#event-loop)).

The `GRPCStreamTestHandler` used in `testBidirectionalStreamingHandler` is synchronous, thus this problem is not captured. What you have to do is make sure that the next message isn't evaluated until the previous one's `EventLoopFuture` (the one returned from `context.handle`) has **completed**. This is was `syncMap` does for the WebSocket Exporter. Once you have tackled this problem, another one arises: You have to make sure you don't drop messages while waiting for the evaluation of a previous message to complete. I implemented `buffer` to solve this problem.

It may be of course that I read your code wrong and your implementation behaves as it should. In that case just write a test similar to `CombineSyncMapTests.testUninterruptedRun` for the GRPC Exporter featuring an **async** `Handler` (message me in case you have questions about that code).

Btw, I think you could solve your problem with `syncMap` by using `buffer` in front of it, which immediately accepts upstream values and buffers them until downstream reports demand.

PR #210, opened 2021-02-05
OAS Enhancement: add `tag` to endpoints via DescriptionModifier

### LINKED ISSUE #209: [MINOR FEATURE] Add tags to endpoints to enhance OAS
Currently, if displayed in swagger-ui interface all endpoints of the created OAS are grouped under "**default**". This is because no specifics `tags` for endpoints are provided.

image

As a consequence, after client stub generation, the respective services are grouped in one file named **`default.service`**.
image

**Propose a solution**
Provide the means to easily tag endpoints and thereby achieve better subgrouping of endpoints in swagger-ui as well as more fine-grained service stub generation.

What we would like to see instead is subgrouping in swagger-ui by tags **`game`**, **`game-action`**, and **`player`**: 
image

and separated files for generated client-side services, such as **`gameService`**, **`gameActionService`**, **`playerService`**:
image

It would be desirable if a developer can provide `tags` via modifier, but Apodini would derive reasonable `tags` if none are given.

lschlesinger:
# Add `tags` to endpoints via `TagModifier`

[OAS Operation objects](https://swagger.io/specification/#operation-object) can have `tags` which represents a "_list of tags for API documentation control_". They can be used for logical grouping of operations.
Currently, the developer is not able to make use of this OAS feature.

## :bulb: Proposed solution
New endpoint modifier  `TagModifier` with property `tags` to allow for "tagging" an endpoint and allow the developer to group endpoints.
If not explicitly given, default tags are derived from the paths by following rules:

1. If parameter in path: 
1.2. AND **preceding** (non-empty) string component in path: **last** string component **before** path parameter is used as tag
-> endpoint path:  `v1/game/{gameId}` has tag: `game`
1.2 AND **no preceding** string (non-empty) components in path: tag is set to `default`
-> endpoint path:  `/{gameId}` has tag: `default`

2. If no parameter in path
2.1 AND string component exists: **last** string component appended to path is used as tag
-> endpoint path:  `v1/player` has tag: `player`
2.2 AND no string component exists: tag is set to `default`

### Problem that is solved
#209

### Implications
With regard to OAS (see screenshots in related issue #209):
- Endpoints in swagger-ui are grouped by tags
- More fine-grained client side services are generated (one service file per tag)

I guess this can be refactored, in case some form of Meta Description DSL is implemented and integrated that provides the means to comprehensively add meta data and description to certain web service parts (i.e. endpoints, parameters etc.). 

### Testing
New test for tag modifier were added:
- Setting `tags` via modifier: `TagModifierTests`
- Deriving `tags` from paths, by rules defined above: `OpenAPIPathsObjectBuilderTests`

### Reviewer Nudging
You were randomly picked 😄 
Question is, if the `TagModifier` should be in the Apodini Modifier scope, or if it should be moved somewhere else, as currently the OAS tag is the only use case for this.


awoc (commented):
I like the feature and the code looks good 👍 

However, I think it would be better to move the `TagModifier` to the `OpenAPI` target because it is only used there. Also, this would remove autocompletion suggestions from an IDE if the user doesn't want to use OpenAPI.

lschlesinger:
> However, I think it would be better to move the `TagModifier` to the `OpenAPI` target because it is only used there. Also, this would remove autocompletion suggestions from an IDE if the user doesn't want to use OpenAPI.

 Thanks for the immediate feedback 👍 , I've changed it in 5093b26

awoc (approved):

PR #215, opened 2021-02-06
Autocorrect SwiftLint

tgymnich:
# Autocorrect SwiftLint

SwiftLint does not fix trivial violations, which is really annoying.

## :bulb: Proposed solution
Let SwiftLint autocorrect trivial violations.

### Problem that is solved
More developer productivity.

### Testing
I did a commit that should trigger a violation


bauer-andreas (commented):
I'm not entirely sure about GitHub Actions automatically (and unexpectedly) altering code. Can the autocorrect feature of SwiftLint be considered stable?
Probably up to @PSchmiedmayer to decide if we incorporate this.

The action file itself looks great though 👍

tgymnich:
@Supereg  I've used it quite a lot and have never had anything weird happen to my code. Also you can always just revert the commit and add a `swiftlint:disable ...` or even exclude the file. An alternative to this PR would be to add swiftlint to the git commit hooks (only if SwiftLint is installed).

PSchmiedmayer:
@tgymnich @Supereg I think that `swiftlint autocorrect` is stable enough to automate this step. I would be fine with automatically adding an additional commit.

One more general comment on the feature: As far as I can see from the [example commit above](https://github.com/Apodini/Apodini/pull/215/commits/893b2abdbffc2dddc777929582ebec558f2d9546) the action adds a commit to the PR but the actions are not triggered again. If I understand this correctly a developer would need to manually create an empty commit to trigger the actions so we have a full report about the status of the PR (e.g. builds and tests). 
We would trade running `swiftlint autocorrect` and committing the changes with the need to commit an empty commit to trigger the actions. It there a way to trigger the actions if the SwiftLint action added a commit? Or that the SwiftLint action runs first and all other actions are only triggered after the commit and take the new commit as the checkout point?

PSchmiedmayer (commented):
@tgymnich @Supereg I think that `swiftlint autocorrect` is stable enough to automate this step. I would be fine with automatically adding an additional commit.

One more general comment on the feature: As far as I can see from the [example commit above](https://github.com/Apodini/Apodini/pull/215/commits/893b2abdbffc2dddc777929582ebec558f2d9546) the action adds a commit to the PR but the actions are not triggered again. If I understand this correctly a developer would need to manually create an empty commit to trigger the actions so we have a full report about the status of the PR (e.g. builds and tests). 
We would trade running `swiftlint autocorrect` and committing the changes with the need to commit an empty commit to trigger the actions. It there a way to trigger the actions if the SwiftLint action added a commit? Or that the SwiftLint action runs first and all other actions are only triggered after the commit and take the new commit as the checkout point?

tgymnich:
@PSchmiedmayer  SwiftLint should only modify the syntax of our program the semantics should stay the same in 99.99% of cases  (unless we have a bug in either SwiftLint or the swift compiler).

We could use `repository_dispatch` and a personal access token to sequentialize our workflows. This would increase the time it takes for our workflows to run though. We could alternatively include this step in the git commit hooks (optionally of course).

I would hope, that compiler errors and SwiftLint bugs are so rare, that they don't justify the extra time we would have to wait when running workflows sequentially. In the worst case scenario our push workflows will catch the bug when the PR is merged.

PSchmiedmayer:
> I would hope, that compiler errors and SwiftLint bugs are so rare, that they don't justify the extra time we would have to wait when running workflows sequentially. In the worst case scenario our push workflows will catch the bug when the PR is merged.

Yes, I also suspect that the `autocorrect` changes shouldn't impact the build and test workflows. But as shown with the [
1be4d62](https://github.com/Apodini/Apodini/pull/215/commits/1be4d6237d2f8d9156f2b2ac70608daa3f526682) commit the merge checks won't work anymore as the latest commit won't have any checks must pass.
Therefore I don't really see any alternative to run the workflows sequentially if we want to work with the GitHub Actions to autocorrect SwiftLint issues.
If we can manage to do this in a nice way I would be fine with this, otherwise having it as an optional git commit hooks would also be fine to me.

tgymnich:
@PSchmiedmayer I didn't think of this. Then running the checks sequentially would not work either since the SwiftLint check for the SwiftLint commit would not be available even when we trigger all other checks afterwards. The only option is to re-trigger all checks which seems wasteful. I also looked more closely into pre-commit hooks and it turns out that they slow down the commit quite a bit and the changes will not be applied to the current commit and will be put in the staging area instead. Right now I think running the checks manually might actually be the best option.

PSchmiedmayer:
@tgymnich Thanks for investigating if we could solve this with GitHub Actions or using a commit hook. I would agree with you. Based on what you have presented running `swiftlint autocorrect` manually seems to be the best option 👍

PR #216, opened 2021-02-06
Update swift packages automatically

### LINKED ISSUE #191: [INTERNAL CHANGE] Add a scheduled GitHub Action to update the Package.resolved file
**Is your change related to a problem? Please describe.**
As a result of the review of #171 [@Supereg and @PSchmiedmayer noticed that it might be useful to automatically check that Apodini always works with the latest resolved dependencies if they break the semantic versioning contract](https://github.com/Apodini/Apodini/pull/171#pullrequestreview-579474576). To prevent such an issue we suggest an automated way to check such regressions and update the Package.resolved file if needed. 

**Propose a solution**
We [should use the schedule GitHub Action functionality](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#scheduled-events) to periodically check for regressions due to updates to dependencies. The action could run a `swift package update` command, check if there is any change in the `Package.resolved` file and if so test everything and create a PR with the updated `Package.resolved` file.

tgymnich:
# Update swift packages automatically

No automatic updates.

## :bulb: Proposed solution
Automatically update swift packages and create a pull request every week

### Problem that is solved
Outdated dependencies

### Related PRs
fixes #191


.github/workflows/spm-update.yml
```diff
@@ -0,0 +1,103 @@
...
+      run: swift --version
+    - name: Update Swift Packages
+      run: swift package update
+    - name: Build
+      if: matrix.configuration == 'release'
+      run: swift build --configuration ${{ matrix.configuration }}
+    - name: Build & Test
+      if: matrix.configuration == 'debug'
+      run: swift test --configuration ${{ matrix.configuration }} --enable-test-discovery
+
```
PSchmiedmayer, line 67: As noted here there seems to be an error with the empty line in the workflow file: https://github.com/Apodini/Apodini/actions/runs/543502065/workflow

PSchmiedmayer (commented):
@tgymnich Thanks a lot for taking a look at the issue! The overall approach of the workflow looks nice.

Is there some way to manually test the action so I can see how it works in action? 
I have seen that you [had run it before and it failed](https://github.com/Apodini/Apodini/actions/runs/543502065/workflow) but as far as I can see it can not manually trigger the workflow at the moment.

.github/workflows/spm-update.yml
tgymnich: this run is super old, basically my first commit. I developed the workflow in my own fork: https://github.com/tgymnich/apodini. Feel free to look at the pull requests and past action runs. If you want to try it yourself just fork my fork. Also I omitted `needs: [macos, linux]` in the createPR workflow to speed up testing in my fork. The workflow in this PR waits for the test on macOS and linux to finish.

PSchmiedmayer: Thanks a lot! I checked out the fork, great to see the workflow working in the fork 👍

PSchmiedmayer (approved):

Issue #223, opened 2021-02-07
[MAJOR FEATURE] Relationship Retrieval: Up Next


bauer-andreas:
Instead of doing a separate issue for every distinct relationship related feature, I though I would just create one issue, all in one place, using GitHubs checklist support.

This issue aims to highlight further improvements to the relationship retrieval, some related to pains points I encountered in may demo project.
_Edit: The list got longer than expected. If you think that it would make sense to split out certain points, feel free to suggest those_

- [ ]  1. The property used for a `reference` Relationship isn't hidden in the REST exporter response (and properly renamed on Exporters not using Relationships).
  Both requirement of  [2.3.2.1](https://github.com/Apodini/Apodini/blob/develop/Documentation/Relationship/Relationship.md#2321-relationship-definition-references) of the proposal (cited below) are currently not incorporated (e.g. REST doesn't hide the property from the response, and other exporters don't append the  `Id` suffix). This is related to lack of control in the Encoding process (#221).
**Workaround:** For WebServices specifically designed for REST, one could define custom `CodingKeys` manually hiding the 
desired property.
> Exporters which make use of relationship information (REST and GraphQL) MUST NOT incorporate the property (in the examplewrittenBy) into the response or the query schema. Instead it is replaced by the defined relationship.
Other exporters should ideally rename such fields by adding a Id suffix (e.g. writtenById).
- [ ] 2. Array return types (e.g. `[User]`) are currently treated as a distinct type, meaning any Relationship DSL definitions of `User` is ignored. This is also related to #221 as adding relationships to the response would probably require to pair them with the individual entires of the Array.
- [ ] 3. If a relationship is shadowed (relationship **name** shadowing) by a `RelationshipDestination` with `hideLink=true` the shadowed relationship is nonetheless not shown in a response. (e.g. structural relationship called `"a"` with `.hideLink()` and a inherited relationship called `"a"` would result in no `"a"` relationship to be shown in the response) 
- [ ] 4. Add ability to hide inherited relationships. This might be incorporated into the Relationship DSL, supporting explicit `Inherits` declarations and also our "auto inheritance" feature.
- [ ] 5.`Relationship` instances currently support exactly **1** destination. Instead of defining multiple `Relationship` instances with the same name, for every operation, we might allow that a `Relationship` has multiple destinations, if they are all defined under the same path having only different `Operation`s.
- [ ] 6. Add support for the `Fluent` `Model` protocol:
[...]

PR #224, opened 2021-02-07
Minor Relationship Retrieval & REST Adjustments

bauer-andreas:
# Minor Relationship Retrieval & REST Adjustments

When developing by demo project, I encountered some smaller issues with the relationship retrieval and REST _links formatting process which were quick to fix.

## :bulb: Proposed solution
This PR adds a commit for every issue fixed.

### Problem that is solved
* b855359 This changes path parameter resolvers such that explicitly defined resolvers are prioritized over automatically retrieved resolver. This addresses one particular issue of point 7. of #223 
* ab88a58 Removes that relationships overwrite relationships of other "sources" if they point to the same destination. This address one part of the issue of point 9. of #223 
* 8dc85f1 Just some typos
* aa59ff0 Resolves point 10. of #223 
* 1cf7efe As the REST exporter will consider RelationshipDestinations with any Operation, those could compete in the name shadowing process. This change ensure that the outcome is not deepened on hashing.
* 11e3d77 Some refactoring providing better semantics for formatRelationships
* 70c6cd2 Changes the stance on https://github.com/Apodini/Apodini/pull/55#issuecomment-743771845 and enforces that Handlers must always declare all  `PathParameter` contained in their path as `@Parameter`. This mitigates some issues were such PathParameters would never be retrieved and thus wouldn't be resolved in relationships.

### Testing
Tests were added were needed.

### Reviewer Nudging
It is probably best to review this PR commit by commit as those are self contained. I hope the explanation above and in the commit message helps for a fast review.


theMomax (approved):
Everything looks pretty reasonable. Only thing I am not sure about is the change of 70c6cd2 as someone might have a strong stance against this behavior (I don't). Maybe give it a second thought who might have a problem with that and request their review in case you can think of someone. Otherwise, just go ahead🚀

bauer-andreas:
> Everything looks pretty reasonable. Only thing I am not sure about is the change of [70c6cd2](https://github.com/Apodini/Apodini/commit/70c6cd20ad0f7d77d6f34ec896ea99313c33a53f) as someone might have a strong stance against this behavior (I don't). Maybe give it a second thought who might have a problem with that and request their review in case you can think of someone. Otherwise, just go ahead🚀

Thanks for the review.
I think the linked comment https://github.com/Apodini/Apodini/pull/55#issuecomment-743771845 describes the implications well. It really is an ill described Webservice. We previously just aimed for ignoring such definitions, as the user would need to fix it anyways, if they would want to make use of the parameter. But under the latest circumstances (and with the assumptions made in the Semantic Model) it makes more sense to explicitly create this restriction.

Sources/Apodini/Semantic Model Builder/InterfaceExporter/EnrichedContent.swift
```diff
@@ -33,48 +32,75 @@ public struct EnrichedContent: Encodable {
...
     ) -> Formatter.Result {
         let context = ResolveContext(content: response.wrappedValue, parameters: parameters)
 
         var result = initialValue
 
-        if let specifiedOperation = operation {
-            endpoint.selfRelationship(for: specifiedOperation)?.formatRelationship(with: formatter, result: &result, context: context)
-        } else {
-            result = endpoint.selfRelationships().formatRelationships(into: result, with: formatter, context: context)
+        for operation in Operation.allCases.sorted(by: sortKeyPath) {
```
nityanandaz, line 79: You need to move `Collection+SortedByKeyPath.swift` back into the `Apodini` target 🙂

bauer-andreas: Yeah saw it, but couldn’t yet commit it.
Is it okay to make that public? We would need to as it is used in the ProtobufferExporter.

PR #231, opened 2021-02-08
Added ParameterType for HTTP Headers

tgymnich:
# Added ParameterType for HTTP Headers

For my final project I added support for header parameters to Apodini, but I didn't end up using it. I'll just leaves this here and let you guys decide if we want to add it,

## :bulb: Proposed solution
Add header `ParameterType`

### Problem that is solved
HTTP headers can be accessed

### DIFF
--- a/Sources/Apodini/Semantic Model Builder/Model/ParameterNamespace.swift
+++ b/Sources/Apodini/Semantic Model Builder/Model/ParameterNamespace.swift
@@ -49,7 +49,7 @@ public extension Array where Element == ParameterNamespace {
     static let global: [ParameterNamespace] = [.all]
     /// With the `.individual` level, a parameter name must be
     /// unique across all parameters with the same `ParameterTyp` on the given `Endpoint`.
-    static let individual: [ParameterNamespace] = [.lightweight, .content, .path]
+    static let individual: [ParameterNamespace] = [.lightweight, .content, .path, .header]
 }
 
 extension ParameterNamespace {
--- a/Sources/Apodini/Semantic Model Builder/Model/ParameterNamespace.swift
+++ b/Sources/Apodini/Semantic Model Builder/Model/ParameterNamespace.swift
@@ -46,10 +46,10 @@ public extension Array where Element == ParameterNamespace {
     /// With the `.global` level, a parameter name must be
     /// unique across all `ParameterType`s on the given `Endpoint`.
     /// This is the default namespace when nothing is specified by the exporter.
-    static let global: [ParameterNamespace] = [.all]
+    static let global: [ParameterNamespace] = [.lightweight, .content, .path]
     /// With the `.individual` level, a parameter name must be
     /// unique across all parameters with the same `ParameterTyp` on the given `Endpoint`.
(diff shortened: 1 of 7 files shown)


bauer-andreas (commented):
The PR seems good 👍 Just not quite sure about enforcing Parameter names to be unique with including http headers. Though as Headers typically start with upper case letters, this should also be fine 👍 

Only thing, I think the `.individual` Namespace definition must be changed to include headers, so one can't define conflicting header Parameter definitions (in the case where lenient Parameter namespaces are enabled) https://github.com/Apodini/Apodini/blob/36cf037d32ec1917c3f4e614ceff6ccf174d5297/Sources/Apodini/Semantic%20Model%20Builder/Model/ParameterNamespace.swift#L52

Sources/Apodini/Semantic Model Builder/Model/ParameterNamespace.swift
```diff
@@ -46,10 +46,10 @@ public extension Array where Element == ParameterNamespace {
     /// With the `.global` level, a parameter name must be
     /// unique across all `ParameterType`s on the given `Endpoint`.
     /// This is the default namespace when nothing is specified by the exporter.
-    static let global: [ParameterNamespace] = [.all]
+    static let global: [ParameterNamespace] = [.lightweight, .content, .path]
```
bauer-andreas: This seems wrong to me. 
`.global` would require that all parameters are in the same namespace. What you are doing here is `.individual` but without a `.header`.
Replacing `.all` of the old definition would equal to `[[.lightweight, .content, .path, .header]]`.
Meaning an array of namespaces, containing one namespace (the inner array) containing all `ParameterType`s.

I would suggest to leave it as is, especially as with the current version you incorporated, names of `.header` won't ever be checked for collisions. 

EDIT: I would also propose to add some test cases for `.header` parameter types, so such scenarios would be covered and secured against regressions such as above.

bauer-andreas (changes requested):

Sources/Apodini/Semantic Model Builder/Model/ParameterNamespace.swift
```diff
@@ -46,10 +46,10 @@ public extension Array where Element == ParameterNamespace {
     /// With the `.global` level, a parameter name must be
     /// unique across all `ParameterType`s on the given `Endpoint`.
     /// This is the default namespace when nothing is specified by the exporter.
-    static let global: [ParameterNamespace] = [.lightweight, .content, .path]
+    static let global: [ParameterNamespace] = [.all]
     /// With the `.individual` level, a parameter name must be
     /// unique across all parameters with the same `ParameterTyp` on the given `Endpoint`.
-    static let individual: [ParameterNamespace] = [.lightweight, .content, .path, .header]
+    static let individual: [ParameterNamespace] = [.lightweight, .content, .path]
```
bauer-andreas: With just reverting the commit, `.header` Parameters go now unchecked for `.individual`.
Not sure about the state of the PR; you haven't requested by review again, though wanted to ensure this doesn't go unnoticed. So sorry if this is already worked on your local copy.

PR #241, opened 2021-02-11
Replace `APNS` with `APNSwift`

awoc:
# Replace `APNS` with `APNSwift`

At the moment, we're using the [APNS](https://github.com/vapor/apns) library from Vapor to send push notifications. However, this library is only an abstraction of [APNSwift](https://github.com/kylebrowning/APNSwift), therefore, we can also replace it with `APNSwift` directly. This would reduce the target size of `ApodiniNotifications` since `APNSwift` is a peer dependency of `APNS`. Furthermore, this enables us to move the `APNS` struct of Apodini from a nested struct inside of `Application` to its own entity because the name collision with the `APNS` library is resolved. This is helpful when injecting the `apns` computed property with `@Environment` since the type changes from `Application.APNS` to just `APNS`.

## :bulb: Proposed solution

Replace `APNS` with `APNSwift` and adapt the `Application+APNS` file.

### Implications

Shouldn't have any usage on the usage.

Includes the MIT License of Vapor, since the implementation to `APNS` is similar.

### Reviewer Nudging

The exisiting code base was just restructured.

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -2,21 +2,12 @@
   "object": {
     "pins": [
       {
-        "package": "apns",
-        "repositoryURL": "https://github.com/vapor/apns.git",
-        "state": {
-          "branch": null,
-          "revision": "a8bc863a718eb23e05b2ae3bfc7de66407b6b85d",
-          "version": "1.0.1"
-        }
-      },
-      {
         "package": "apnswift",
         "repositoryURL": "https://github.com/kylebrowning/APNSwift.git",
         "state": {
           "branch": null,
-          "revision": "ce8213a3e0ec0581c4c39a155e7778401ec5e24a",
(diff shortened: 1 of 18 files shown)


awoc:
@PSchmiedmayer I included the license of the [Vapor](https://github.com/vapor/vapor), however, the [APNS](https://github.com/vapor/apns) repository doesn't include its own license. Therefore, I'm not sure if this restructuring is even possible.

PSchmiedmayer:
@awoc I have added a PR to the vapor/apns repo. It looks to me like they have just forgot to add the license file: https://github.com/vapor/apns/pull/32. Once it is clear what the correct license would be we can merge this PR or continue using vapor/apns as a dependency.

PR #246, opened 2021-02-19
Add Deployment Providers

lukaskollmer:
# Add Deployment Providers

Currently, a web service can only be run as a monolithic single-process system, either via `swift run ...` or by manually launching the compiled executable. Additionally, web services currently can run only on the host platform, and there are no facilities for type-safe interactions between Handlers.

## :bulb: Proposed solution
This PR addresses all of the above: it introduces the concept of deployment providers, which allow deploying a web service to arbitrary platforms, and running them as a distributed system (i.e. multiple processes), while also taking care of interactions between Handlers.

### Problem that is solved
The web service can now run in more complex situations.

### Implications
There are new Handler types, and some additional APIs which hook into Apodini, and, in some cases, take control of the execution of the current process.

### Related PRs
- Some of these functionalities used to be initially proposed in #141, which has been closed in favor of this PR

### Testing
There are tests, but they are still somewhat lacking, primarily testing of the deployment providers, as well as testing the deployment options API on xenial ubuntu in debug builds (see #257)

### Reviewer Nudging

The two sample deployment providers can be tested in the terminal (cwd should be the root dir of the repo):
- Localhost:
  Run `swift run DeploymentTargetLocalhost . --product-name ApodiniDeployTestWebService`
- AWS Lambda:
  Run `swift run DeploymentTargetAWSLambda . --product-name ApodiniDeployTestWebService --s3-bucket-name BUCKET_NAME --aws-api-gateway-api-id API_ID`
  - There should be an aws user configured for the system, run `swift run DeploymentTargetAWSLambda --help` if not.
  - `BUCKET_NAME` should be the name of an already existing S3 bucket, to which the AWS user used for the deployment has read/write access.
  - `API_ID` should be either the id of an already existing API Gateway (HTTP), or `_createNew` to have the deployment provider create a new one.


Package.swift
```diff
@@ -251,7 +280,131 @@ let package = Package(
...
+                .target(name: "DeploymentTargetAWSLambdaRuntime"),
+                .target(name: "ApodiniREST"),
+                .target(name: "ApodiniGRPC"),
+                .target(name: "ApodiniProtobuffer"),
+                .target(name: "ApodiniOpenAPI"),
+                .target(name: "ApodiniWebSocket"),
+                .target(name: "ApodiniNotifications"),
+                .target(name: "ApodiniDeploy")
+            ]
+        ),
```
PSchmiedmayer: If possible I would suggest to combining the two TestWebServices into one that lives in its own package.

Sources/Apodini/Properties/Parameter.swift
```diff
@@ -37,6 +43,16 @@ public struct Parameter<Element: Codable>: Property {
         
         return element
     }
+    
+    
+    // This is here because, other than a `@Parameter` property itself (ie the storage), the projected value
+    // isn't private, meaning that we can use this to specify a parameter when remote-invoking a handler.
+    // We could alwo make the projectedValue return self, although it seems like a better idea to limit ourselves
```
PSchmiedmayer:
```suggestion
    // We could also make the projectedValue return self, although it seems like a better idea to limit ourselves
```

Sources/Apodini/Semantic Model Builder/SemanticModelBuilder.swift
```diff
@@ -62,9 +62,28 @@ class SemanticModelBuilder: InterfaceExporterVisitor {
...
+                    if ident1 == ident2 {
+                        return ident1
+                    } else {
+                        fatalError("""
+                            Handler '\(handler)' has multiple explicitly specified identifiers ('\(ident1)' and '\(ident2)').
+                            A handler may only have one explicitly specified identifier.
+                            This is caused by using both the 'IdentifiableHandler.handlerId' property as well as the '.identified(by:)' modifier.
+                            """
+                        )
+                    }
```
PSchmiedmayer: I would suggest to logg out a warning here (to inform the developer about the double configuration) but I would suggest that the `dslSpecifiedIdentifier` takes precedence over the `handlerSpecifiedIdentifier` in this case.

Sources/Apodini/WebService.swift
```diff
@@ -22,13 +25,16 @@ extension WebService {
...
+    
     /// This function is executed to start up an Apodini `WebService`
     @discardableResult
     static func main(waitForCompletion: Bool) throws -> Application {
         let app = Application()
         LoggingSystem.bootstrap(StreamLogHandler.standardError)
 
         main(app: app)
+        
+        //return app;
```
PSchmiedmayer:
```suggestion
```

Sources/ApodiniDeploy/ApodiniDeployInterfaceExporter+ExportWebServiceStructure.swift
```diff
@@ -0,0 +1,49 @@
...
+//
+//  Created by Lukas Kollmer on 16.02.21.
+//
+
+import Foundation
+import Apodini
+import ApodiniUtils
+// getting rid of the @testable here would require making the StorageKey (as well as its Value) public,
+// which would imply making most of the OpenAPIExporter public (which we'd ideally prefer to avoid)
+@testable import ApodiniOpenAPI
```
PSchmiedmayer: The problem is that we won't be able to compile the web service in release builds with `@testable`. I think that we don't have any other choice then making a lot of aspects of the exporter public for now. Implementing the vision of stackable exporters as discussed last week would require some sort of public interface anyways. Nevertheless, we should aim for the stackable exporter idea in the future.

Sources/ApodiniDeploy/ApodiniDeployConfiguration.swift
```diff
@@ -0,0 +1,133 @@
...
+    
+    public func configure(_ app: Application) {
+        app.storage.set(StorageKey.self, to: self)
+    }
+}
+
+
+
+
+// MARK: Pt 2
```
PSchmiedmayer: I would suggest to move the types into some separate Swift files and a subfolder to e.g. indicate that they are extension on the Apodini DSL/Handlers/Groups.

Sources/ApodiniDeploy/ApodiniDeployConfiguration.swift
```diff
@@ -0,0 +1,133 @@
...
+
+
+extension DeploymentGroup {
+    /// Creates a single deployment group containing all handlers of type `H`.
```
PSchmiedmayer, line 38: It would be nice to have some examples on how to use the different public extension as part of the code documentation once the API is stable.

Sources/ApodiniUtils/Optional.swift
```diff
@@ -0,0 +1,42 @@
...
+}
+
+
+///// Adds the shortcut `.null` if you have a double Optional type to create a Wrapped?(nil) value,
+///// meaning a `Optional` containing a `Optional.none` aka a empty `Optional`.
+//extension Optional where Wrapped: ExpressibleByNilLiteral {
+//    static var null: Self {
+//        .some(nil)
+//    }
+//}
```
PSchmiedmayer:
```suggestion
```

Sources/ApodiniUtils/Swift.swift
```diff
@@ -0,0 +1,159 @@
...
+            return try firstIndex(from: index(after: idx), where: predicate)
+        }
+    }
+}
+
+
+extension RandomAccessCollection {
+    /// Safely access the element at the specified index.
+    /// - returns: the element at `idx`, if `idx` is a valid index for subscripting into the collection, otherwise `nil`.
+    public subscript(lk_safe idx: Index) -> Element? {
```
PSchmiedmayer:
```suggestion
    public subscript(safe index: Index) -> Element? {
```

Sources/ApodiniUtils/System.swift
```diff
@@ -0,0 +1,71 @@
...
+
+extension ProcessInfo {
+    /// The group id of the current process
+    public var lk_groupId: pid_t { getpgrp() }
+    
+    /// The process id of the current process's parent
+    public var lk_parentProcessId: pid_t { getppid() }
+    
+    /// The group id of the current process's parent
+    public var lk_parentGroupId: pid_t { getpgid(lk_parentProcessId) }
```
PSchmiedmayer: please remove the `lk_` prefixes

Sources/ApodiniUtils/System.swift
```diff
@@ -0,0 +1,71 @@
...
+}
+
+
+//precondition(URL(fileURLWithPath: ProcessInfo.processInfo.arguments[0]) == Bundle(for: Self.self).executableURL!)
+
+#if canImport(MachO)
+public func LKGetCurrentExecutableUrl() -> URL {
```
PSchmiedmayer:
```suggestion
public func GetCurrentExecutableUrl() -> URL {
```

Package.swift
```diff
@@ -54,13 +64,30 @@ let package = Package(
         .package(url: "https://github.com/wickwirew/Runtime.git", from: "2.2.2"),
         // Used for testing purposes only. Enables us to test for assertions, preconditions and fatalErrors.
         .package(url: "https://github.com/mattgallagher/CwlPreconditionTesting.git", from: "2.0.0"),
-        .package(url: "https://github.com/mattpolzin/OpenAPIKit.git", from: "2.4.0"),
-        .package(url: "https://github.com/jpsim/Yams.git", from: "4.0.0")
+        .package(url: "https://github.com/jpsim/Yams.git", from: "4.0.0"),
+        
+        // Deploy
+        .package(url: "https://github.com/apple/swift-argument-parser", from: "0.3.0"),
+        .package(url: "https://github.com/swift-server/swift-aws-lambda-runtime.git", from: "0.1.0"),
```
PSchmiedmayer: We do currently get a warning as the AWS Lambda Runtime is not used at the moment:
_"dependency 'swift-aws-lambda-runtime' is not used by any target"_.

Sources/ApodiniDeploy/ApodiniDeployInterfaceExporter.swift
```diff
@@ -0,0 +1,250 @@
...
+    var lk_stableIdentity: String {
+        switch self.parameterType {
+        case .content:
+            return "c:\(self.name)"
+        case .lightweight:
+            return "l:\(self.name)"
+        case .path:
+            return "p:\(self.name)"
+        }
+    }
```
PSchmiedmayer: I would suggest to remove all `LK`/`lk` prefixes from declarations in the PR. 

I suspect you can't use the id property here as the UUIDs don't match over different instances of the web service? If so, this identifier should also be part of the AnyEndpointParameter when the `id` property is not required but we care more about the same kind of parameter (combining type and name) across different instances of a web service.

Sources/ApodiniDeploy/ApodiniDeployInterfaceExporter.swift
```diff
@@ -0,0 +1,250 @@
...
+        }
+        collectedEndpoints.append(CollectedEndpointInfo(
+            handlerType: HandlerTypeIdentifier(H.self),
+            endpoint: endpoint,
+            deploymentOptions: CollectedOptions(reducing: [
+                endpoint.handler.getDeploymentOptions(),
+                endpoint.context.get(valueFor: HandlerDeploymentOptionsSyntaxNodeContextKey.self)
+            ].flatMap { $0.compactMap { $0.resolve(against: endpoint.handler) } })
+        ))
+        app.vapor.app.add(Vapor.Route(
```
PSchmiedmayer: I am not sure if I have missed something but I think that we should restrict registering invocable routes to `Handler`s that conform to `InvocableHandler`. As far as I can see those are the only handlers that would be registered to be invocable from the outside at our special path.

Sources/ApodiniDeploy/ApodiniDeployInterfaceExporter.swift
```diff
@@ -0,0 +1,250 @@
...
+            method: .POST,
+            path: ["__apodini", "invoke", .constant(endpoint.identifier.rawValue)],
+            responder: InternalInvocationResponder(RHIIE: self, endpoint: endpoint),
```
PSchmiedmayer: What the spelled out version of `RHIIE`? I would suggest to use abbreviations sparely to make it easier to read and reason about parameters, types, variables and so on

Sources/ApodiniDeploy/ApodiniDeployInterfaceExporter.swift
```diff
@@ -0,0 +1,250 @@
...
+            responseType: InternalInvocationResponder<H>.Response.self
+        ))
+    }
+    
+    
+    public func finishedExporting(_ webService: WebServiceModel) {
+        do {
+            try performDeploymentStuff()
+        } catch {
+            fatalError("ugh \(error)")
```
PSchmiedmayer: Maybe a more detailed description should be added here. In general we need an Apodini Error that is not meant for external consumers but for internal usage. This should then be used here and then be delegated up a throwing chain to a point where we can handle the error, e.g. only work with exporters that have not thrown an error.

Sources/ApodiniDeploy/ApodiniDeployInterfaceExporter.swift
```diff
@@ -0,0 +1,250 @@
...
+        switch args[1] {
+        case WellKnownCLIArguments.exportWebServiceModelStructure:
+            let outputUrl = URL(fileURLWithPath: args[2])
+            do {
+                try self.exportWebServiceStructure(
+                    to: outputUrl,
+                    deploymentConfig: self.app.storage.get(ApodiniDeployConfiguration.StorageKey.self)?.config ?? .init()
+                )
+            } catch {
+                fatalError("Error exporting web service structure: \(error)")
```
PSchmiedmayer, line 140: I would suggest to be consistent within this method to also throw an error here and catch them in a single fatal Error (or throw them up the chain in the future).

Sources/ApodiniDeploy/ApodiniDeployInterfaceExporter.swift
```diff
@@ -0,0 +1,250 @@
...
+            case value(Any)    // the value, as-is
+            case encoded(Data) // the value, encoded
+        }
+        
+        private let parameterValues: [String: Param] // key: stable endpoint param identity
+        
+        init<H: Handler>(endpoint: Endpoint<H>, collectedParameters: [CollectedParameter<H>]) {
+            parameterValues = .init(uniqueKeysWithValues: collectedParameters.map { param -> (String, Param) in
+                let paramId = (endpoint.handler[keyPath: param.handlerKeyPath] as! AnyParameterID).value
+                let endpointParam = endpoint.parameters.first { $0.id == paramId }!
```
PSchmiedmayer: Would it make sense to fail more gracefully here and not crash the complete program on an invalid input/missing parameters? A male formatted message could (as far as I can see) easily crash the web service with a missing parameter.

Sources/ApodiniDeploy/DeploymentOptions/Condition.swift
```diff
@@ -0,0 +1,57 @@
+//
+//  File.swift
+//  
+//
+//  Created by Lukas Kollmer on 2021-02-06.
+//
+
+import Foundation
```
PSchmiedmayer, line 11: I like how this turned out after our discussions in the thesis meetings 👍

Would it make sense to move the Conditions to The ApodiniDeployBuildSupport target instead of defining them here? What would be be tradeoff we would need to make if we have all the general OptionHandling in the ApodiniDeployBuildSupport target? And I have seen that the `TestWebService` lists both targets in its dependencies: In general it would be normal to only import Apodini Deploy?

Sources/ApodiniDeploy/InternalInvocationResponder.swift
```diff
@@ -0,0 +1,114 @@
...
+            let vaporResponse = Vapor.Response(status: .ok)
+            do {
+                try vaporResponse.content.encode(
+                    Response(status: .handlerError, encodedData: try JSONEncoder().encode(handlerError.localizedDescription)),
+                    using: JSONEncoder()
+                )
+            } catch {
+                vaporResponse.status = .internalServerError
+            }
+            return vaporResponse
```
PSchmiedmayer, line 84: Errors thrown in `Handler`'s handle function should `ApodiniErrors`. So if you encounter an `ApodiniError` you should be able to use its type, reason and so on to fill in the construct. All other errors are suspected to be thrown in code outside of the developer's control. In the long term we should at least log a warning that we encountered a non `ApodiniError` here which we can not extract information from. You would even encode and decode the Apodini error to pass it further down the chain on the other side that expects the response.

Sources/ApodiniDeploy/RemoteHandlerInvocationManager.swift
```diff
@@ -0,0 +1,308 @@
...
+    }
+    
+    
+    private func _invoke<H: InvocableHandler>(
+        handlerType: H.Type,
+        handlerId: H.HandlerIdentifier,
+        collectedInputParams: [CollectedParameter<H>]
+    ) -> EventLoopFuture<H.Response.Content> {
+        guard let RHIIE = self.app.storage.get(ApodiniDeployInterfaceExporter.ApplicationStorageKey.self) else {
+            return eventLoop.makeFailedFuture(ApodiniDeployError(message: "unable to get \(ApodiniDeployInterfaceExporter.self) object"))
```
PSchmiedmayer:
```suggestion
            return eventLoop.makeFailedFuture(ApodiniDeployError(message: "Unable to get the \(ApodiniDeployInterfaceExporter.self) object"))
```

Sources/ApodiniDeploy/RemoteHandlerInvocationManager.swift
```diff
@@ -0,0 +1,308 @@
...
+///// A stand-in helper function to create `Swift.Error`-conforming objects without having to define a custom error type
+//internal func makeApodiniError(code: Int = 0, _ message: String) -> Swift.Error {
+//    NSError(domain: "Apodini", code: code, userInfo: [
+//        NSLocalizedDescriptionKey: message
+//    ])
+//}
+
+
+
+
```
PSchmiedmayer:
```suggestion
```

Sources/ApodiniDeploy/RemoteHandlerInvocationManager.swift
```diff
@@ -0,0 +1,308 @@
...
+        case .locally:
+            return targetEndpoint._invoke(withCollectedParameters: collectedInputParams, RHIIE: RHIIE, on: eventLoop)
+        case .remotely(let targetNode):
+            guard let runtime = RHIIE.deploymentProviderRuntime else {
+                return eventLoop.makeFailedFuture(ApodiniDeployError(message: "Unable to find runtime"))
+            }
+            // The targetEndpoint is on a different node, dispatch the invocation there
+            
+            // TODO if theres a guarantee that an EndpointParameter's id is the same as the id of the @Parameter it was generated from,
+            // we can combine these two sets into one
```
PSchmiedmayer: We should make sure that this is the case and if so we should also add a comment to the creation of the EndpointParameter to hint at this constraint.

Sources/ApodiniDeploy/RemoteHandlerInvocationManager.swift
```diff
@@ -0,0 +1,308 @@
...
+            let invocationParams: [HandlerInvocation<H>.Parameter] = collectedInputParams.map { collectedParam in
+                // The @Parameter property wrapper declaration in the handler
+                let handlerParamId = (targetEndpoint.handler[keyPath: collectedParam.handlerKeyPath] as! AnyParameterID).value
+                let endpointParam = targetEndpoint.findParameter(for: handlerParamId)!
+                if !alreadyProcessedParamKeyPaths.insert(collectedParam.handlerKeyPath).inserted {
+                    print("Parameter '\(endpointParam.name)' specified multiple times in remote handler invocation")
+                }
+                if !alreadyProcessedEndpointParamIds.insert(endpointParam.id).inserted {
+                    print("Endpoint parameter with id '\(endpointParam.id)' matched multiple times in remote handler invocation")
+                }
```
PSchmiedmayer: I would suggest using a logger (e.g. the one in the app storage) instead of using print statements. Same applies to other print statements in the code.

Sources/ApodiniDeploy/RemoteHandlerInvocationManager.swift
```diff
@@ -0,0 +1,308 @@
...
+                    stableIdentity: endpointParam.lk_stableIdentity,
+                    name: endpointParam.name,
+                    value: collectedParam.value as! HandlerInvocation<H>.Parameter.Value
+                )
+            }
+            let missingParamNames: [String] = targetEndpoint.parameters
+                .filter { !$0.hasDefaultValue && !alreadyProcessedEndpointParamIds.contains($0.id) }
+                .map(\.name)
+            guard missingParamNames.isEmpty else {
+                fatalError("Missing parameters in remote handler invocation: \(missingParamNames)")
```
PSchmiedmayer: I would suggest to use the `RemoteInvocationResponseError` or a similar construct here too. While most of the time this is a development error it could also be that some remote invocations depend in input provided by the consumer and there fore sometimes throwing an appropriate error would be a good idea.

Sources/ApodiniDeploy/RemoteHandlerInvocationManager.swift
```diff
@@ -0,0 +1,308 @@
...
+                let handlerInvocation = HandlerInvocation<H>(handlerId: handlerId, targetNode: targetNode, parameters: invocationParams)
+                let runtimeHandlingResult = try runtime.handleRemoteHandlerInvocation(handlerInvocation)
+                
+                switch runtimeHandlingResult {
+                case .result(let future):
+                    return future
+                case .invokeDefault(let url):
+                    let requestUrl = url
+                        .appendingPathComponent("__apodini")
+                        .appendingPathComponent("invoke")
```
PSchmiedmayer: We should have some configuration to configure these routes and use a global source of truth to store and access them.

Sources/ApodiniDeploy/RemoteHandlerInvocationManager.swift
```diff
@@ -0,0 +1,308 @@
...
+}
+
+
+
+
+extension Handler where Self: WithEventLoop {
+    // TODO what about moving the `invoke` functions from the RemoteHandlerInvocationManager?
+    // There's really no point in having this type, since, essentially, all it does it store the app/eventLoop object
+//    public func invoke
+}
```
PSchmiedmayer: I am not quite sure if this is just a leftover? If so we can remove it, otherwise feel free to clarify the problem here and I might be able to provide some input.

Sources/ApodiniDeployBuildSupport/FileManagerExtensions.swift
```diff
@@ -0,0 +1,16 @@
+//
```
PSchmiedmayer: Can be removed

Sources/ApodiniUtils/Bitset.swift
```diff
@@ -0,0 +1,59 @@
+//
+//  Bitset.swift
+//
+//
+//  Created by Lukas Kollmer on 16.02.21.
+//
+
+// TODO turn this into a proper bitset type?
+
+extension FixedWidthInteger {
```
PSchmiedmayer, line 10: Please remove the lk prefixes 👍

Sources/ApodiniUtils/FileManager.swift
```diff
@@ -0,0 +1,78 @@
...
+//  FileManager.swift
+//
+//
+//  Created by Lukas Kollmer on 16.02.21.
+//
+
+import Foundation
+
+
+extension FileManager {
```
PSchmiedmayer, line 11: Please remove the lk prefixes 👍

Sources/DeploymentTargetAWSLambdaRuntime/LambdaRuntimeHandler.swift
```diff
@@ -0,0 +1,21 @@
+//
+//  File.swift
+//  
+//
+//  Created by Lukas Kollmer on 2021-01-24.
+//
+
+//import Foundation
```
PSchmiedmayer: Can be removed

TestWebService/Sources/TestWebService/main.swift
```diff
@@ -6,26 +6,120 @@
 //
 
 import Apodini
+import NIO
```
PSchmiedmayer: The TestWebService needs some cleanups here and there to compile 👍

In addition I would suggest moving the `Handler`s in separate Swift files.

Sources/ApodiniDeployBuildSupport/DSLDeploymentConfig.swift
```diff
@@ -0,0 +1,101 @@
...
+
+
+
+extension NSError {
+    public static func apodiniDeploy(code: Int = 0, localizedDescription: String) -> NSError {
+        return NSError(domain: "ApodiniDeploy", code: code, userInfo: [
+            NSLocalizedDescriptionKey: localizedDescription
+        ])
+    }
+}
```
PSchmiedmayer: As far as Ic an see this is not used anywhere

Sources/DeploymentTargetLocalhostRuntimeSupport/LocalhostRuntimeSupport.swift
```diff
@@ -0,0 +1,42 @@
...
+        app.http.address = .hostname(nil, port: currentNodeCustomLaunchInfo.port)
+        //app.vapor.app.http.server.configuration.port = currentNodeCustomLaunchInfo.port
+    }
+    
+    
+    public func handleRemoteHandlerInvocation<H: IdentifiableHandler>(
+        _ invocation: HandlerInvocation<H>
+    ) throws -> RemoteHandlerInvocationRequestResponse<H.Response.Content> {
+        let LLI = invocation.targetNode.readUserInfo(as: LocalhostLaunchInfo.self)!
+        // TODO read hostname from app?
```
PSchmiedmayer: As discussed in the last meeting we should add a configuration that (in addition to collecting the interface and the port the web service should bind to) should also collect the hostname that we can extract from that information and use here.

Sources/DeploymentTargetLocalhostRuntimeSupport/LocalhostRuntimeSupport.swift
```diff
@@ -0,0 +1,42 @@
...
+    public required init(deployedSystem: DeployedSystemStructure, currentNodeId: DeployedSystemStructure.Node.ID) {
+        self.deployedSystem = deployedSystem
+        self.currentNodeId = currentNodeId
+        self.currentNodeCustomLaunchInfo = deployedSystem.node(withId: currentNodeId)!.readUserInfo(as: LocalhostLaunchInfo.self)!
+    }
+    
+    
+    public func configure(_ app: Apodini.Application) throws {
+        app.http.address = .hostname(nil, port: currentNodeCustomLaunchInfo.port)
+        //app.vapor.app.http.server.configuration.port = currentNodeCustomLaunchInfo.port
```
PSchmiedmayer: I think that there is already a configuration to configure the hostname and port in Apodini. We might want to reuse this functionality if I am not missing something here. This would make this implementation a bit more indented of Vapor if we decide to remove that dependency at some point.

Sources/DeploymentTargetLocalhost/main.swift
```diff
@@ -0,0 +1,131 @@
...
+                launchInCurrentProcessGroup: true,
+                environment: [WellKnownEnvironmentVariables.currentNodeId: node.id]
+            )
+            func taskTerminationHandler(_ terminationInfo: Task.TerminationInfo) {
+                if terminationInfo.exitCode == 4 && terminationInfo.reason == .uncaughtSignal {
+                    // This seems to be the combination with which a fatalError terminates a program.
+                    // If one of the children was terminated with a fatalError, we re-spawn it to keep the server running
+                    logger.warning("Restarting child for node '\(node.id)'")
+                    // TODO implement! (this is what triggers the compiler bug)
+                    // try! task.launchAsync(taskTerminationHandler)
```
PSchmiedmayer, line 104: I just uncommented the line in Xcode and ran all tests and it compiled fine. I am using Version 12.4 (12D4e). Or does the comment refer to an other line?

Sources/DeploymentTargetAWSLambdaCommon/LambdaCommon.swift
```diff
@@ -0,0 +1,85 @@
...
+//public final class LambdaHandlerOptionKey<Value: Codable>: DeploymentOptionKey<Value> {}
+//
+//
+//// All lambda-specific deployment options
+//public enum LambdaHandlerOption {
+//    /// The lambda function's memory size, in MB
+//    public static let memorySize = LambdaHandlerOptionKey<Int>(defaultValue: 128, key: "memory-size")
+//    /// The lambda function's timeout, in seconds
+//    public static let timeout = LambdaHandlerOptionKey<Int>(defaultValue: 3, key: "timeout")
+//}
```
PSchmiedmayer: Please remove these lines.

Sources/DeploymentTargetAWSLambda/main.swift
```diff
@@ -0,0 +1,337 @@
...
+set -eu
+
+executable_path=$1 # path to the built executable
+output_dir=$2      # path of the directory we should copy the object files to
+
+rm -rf "$output_dir"
+mkdir -p "$output_dir"
+# add the target deps based on ldd
+ldd "$executable_path" | grep swift | awk '{print $3}' | xargs cp -L -t "$output_dir"
+"""
```
PSchmiedmayer: I would suggest using resources for the content of these constants

Sources/DeploymentTargetAWSLambda/main.swift
```diff
@@ -0,0 +1,337 @@
...
+    let productName: String
+    let packageRootDir: URL
+    let awsProfileName: String
+    let awsRegion: String
+    let s3BucketName: String
+    let s3BucketPath: String
+    private(set) var awsApiGatewayApiId: String
+    let awsDeployOnly: Bool
+    
+    private let FM = FileManager.default
```
PSchmiedmayer: Please use lowercase names for property names.

Sources/ApodiniDeployBuildSupport/Options/Namespace.swift
```diff
@@ -0,0 +1,28 @@
...
+
+//extension OuterNamespace {
+//    static var id: String { "\(Self.self)" }
+//}
+
+
+public protocol InnerNamespace: class {
+    associatedtype OuterNS: OuterNamespace
+    static var id: String { get }
+}
```
PSchmiedmayer: If I understand this correctly the inner namespace is used e.g. to specify deployment options that are specific to one deployment provider and the generic options also have their own inner namespace that is used for them? Maybe renaming it to `Namespace` and then `Category` or `Subset` or something like this might be more fitting?

Sources/ApodiniDeployBuildSupport/Options/Namespace.swift
```diff
@@ -0,0 +1,28 @@
...
+
+
+
+public protocol OuterNamespace: class {
+    static var id: String { get }
+}
+
+//extension OuterNamespace {
+//    static var id: String { "\(Self.self)" }
+//}
```
PSchmiedmayer:
```suggestion
```

Sources/ApodiniDeployBuildSupport/Options/Namespace.swift
```diff
@@ -0,0 +1,28 @@
...
+
+
+public protocol InnerNamespace: class {
+    associatedtype OuterNS: OuterNamespace
+    static var id: String { get }
+}
+
+//extension InnerNamespace {
+//    static var id: String { "\(Self.self)" }
+//}
```
PSchmiedmayer:
```suggestion
```

Sources/ApodiniDeployBuildSupport/Options/OptionKey.swift
```diff
@@ -0,0 +1,63 @@
...
+//  OptionKey.swift
+//
+//
+//  Created by Lukas Kollmer on 2021-02-06.
+//
+
+import Foundation
+
+
+public class AnyOptionKey<OuterNS: OuterNamespace>: Codable, Hashable, Equatable, CustomStringConvertible {
```
PSchmiedmayer, line 11: A more general question: We have the Options that are defined as part of ApodiniDeploy and we have the Options that are part of the main Apodini Target. Wouldn't it make sense to move the general construct in its own target, e.g. ApodiniOptions and import this in Apodini and Apodini Deploy so we don't have two options implementations laying around? I think this would also become beneficial when we start thinking about the metadata DSL in the future. Or would there be a specific reason for keeping the two implementations separate?

In addition this might also be beneficial for the work that @eldcn is doing in the future. The current WebService Model already contains a lot of important information and maybe adding some more general context and information to that would be good as this would allow us to e.g. reuse the WebService model for the diffing of web service interfaces too.

Sources/ApodiniDeployBuildSupport/LaunchInfo.swift
```diff
@@ -0,0 +1,138 @@
...
+
+
+
+
+/// The structure of a deployed system.
+/// A deployed system is a distributed system consisting of one or more nodes.
+/// Each node implements one or more of the deployed `WebService`'s endpoints.
+/// - Note: There may be more than one instances of a node running at a given time,
+///   for example when deploying to a platform which supports scaling.
+public struct DeployedSystemStructure: Codable { // TODO or just `DeployedSystem`?
```
PSchmiedmayer: I like the name `DeployedSystem`

Sources/ApodiniDeployBuildSupport/LaunchInfo.swift
```diff
@@ -0,0 +1,138 @@
...
+//
+//
+//    public func writeTo(url: URL, options: Data.WritingOptions = []) throws {
+//        let encoder = JSONEncoder()
+//        #if DEBUG
+//        encoder.outputFormatting.insert(.prettyPrinted)
+//        #endif
+//        let data = try encoder.encode(self)
+//        try data.write(to: url, options: options)
+//    }
```
PSchmiedmayer:
```suggestion
```

Sources/ApodiniDeployBuildSupport/LaunchInfo.swift
```diff
@@ -0,0 +1,138 @@
...
+        /// The deployment options for all endpoints exported by this node
+        public func combinedEndpointDeploymentOptions() -> DeploymentOptions {
+            DeploymentOptions(exportedEndpoints.map(\.deploymentOptions).flatMap(\.options))
+        }
+        
+        
+        //public func withUserInfo(_ userInfo: _OptionalNilComparisonType) -> Self {
+            // ?? We can afford the try! here because we're passing nil, meaning that it'll never encode anything, meaning it won't crash ??
+            //return try! Node(id: self.id, exportedEndpoints: self.exportedEndpoints, userInfo: nil, userInfoType: Null.self)
+        //}
```
PSchmiedmayer:
```suggestion
```

Sources/DeploymentTargetAWSLambda/AWSStuff.swift
```diff
@@ -0,0 +1,473 @@
...
+//            if !functionsToBeDeleted.isEmpty {
+//                logger.notice("Deleting old apodini lambda functions")
+//                for function in functionsToBeDeleted {
+//                    logger.notice("[SKIPPED] - deleting \(function.functionArn!)")
+//                    //try lambda
+//                    //    .deleteFunction(Lambda.DeleteFunctionRequest(functionName: function.functionName!))
+//                    //    .wait()
+//                }
+//            }
+//        }
```
PSchmiedmayer: As noted in the last meeting this would be a good idea. In combination with getting the real hostname using using this differentiate between different lambdas of different web services this should work. 
Maybe one more general comment: I would suggest to make the more destructive actions (e.g. removing lambdas) after the deployment of new lambdas was successful. Otherwise we could have an error in the middle of the deployment and might be stuck with a system that is no longer up. I think this might also apply to some of the order the tasks happen here that I might have missed.

Sources/DeploymentTargetAWSLambda/AWSStuff.swift
```diff
@@ -0,0 +1,473 @@
...
+        tmpDirUrl: URL
+    ) {
+        self.awsProfileName = awsProfileName
+        self.awsRegion = .init(rawValue: awsRegionName)
+        self.tmpDirUrl = tmpDirUrl
+        awsClient = AWSClient(
+            credentialProvider: .configFile(profile: awsProfileName),
+            retryPolicy: .exponential(),
+            httpClientProvider: .createNew
+        )
```
PSchmiedmayer: As we no longer rely on the CLI we might also think about adopting some other credential provider techniques that soto offers. I would imagine a command linke argument that configures how the credentials are loaded.

PSchmiedmayer (changes requested):
@lukaskollmer I looked over the code and commented on the functionality, code style, and general leftover of previous iterations. The overall code looks well structured to me, and it seems to fulfill its job. 
The three bigger elements I see that we should address before merging would be:
- Think about failing more graceful in case of an error. In general, we might want to replace some of the fatalErrors with errors that can be reported to the user of the remote handler invocation API or can be transformed to ApoiniErrors along the way. Maybe this should be the first step towards internal error handling.
- The two separate Options implementations are, on the one hand, part of the Apodini core and the Apodini deploy tool. As far as I can see, you use your model to make the options encodable and decodable and provide several layers of namespacing. A question to you: Would it make sense to combine these implementations?
- Improving the test coverage and making the test web service and the release builds run as expected.
- Fix the smaller comments I mentioned as part of the review.

In summary: I like the PR, and I am looking forward to seeing the improvements and functionality in Apodini!

Package.swift
lukaskollmer: Yes good point, I'll revert that to the current structure in the develop branch

lukaskollmer: Yes good point, I'll revert that to the current structure in the develop branch

Sources/ApodiniUtils/Swift.swift
lukaskollmer: Sorry about prefixing everything, that's a remnant from my Objective-C days...

I do generally speaking prefer to prefix all custom extensions on foreign-defined types, so that it is easier to differentiate between calls to vendor-provided methods and calls to custom functions on vendor-defined types (the underlying assumption being that my implementations probably have worse performance/efficiency than what apple ships as part of the stdlib), but in this case adding my prefix all over the Apodini code base clearly is not a desirable thing.

I'll remove all of them.

Sources/ApodiniDeploy/ApodiniDeployInterfaceExporter.swift
lukaskollmer: Yes exactly. The `id`s are randomly generated when the web service is run, meaning that we cannot use a parameter's id to compare parameters across multiple instances of the same web service.
I didn't add it to `AnyEndpointParameter` because I actually wasn't sure if there is any guarantee that a `Handler` cannot have multiple `@Parameter`s with the same name.

Sources/ApodiniDeploy/ApodiniDeployInterfaceExporter.swift
lukaskollmer: Correct, these routes currently would not be accessible via the RemoteHandlerInvocation API. I figured it wouldn't hurt to add them anyway, in case there might be other reasons for which we'd want to manually invoke a specific handler in a specific instance (eg things like the observed objects, etc).

Sources/ApodiniDeploy/ApodiniDeployInterfaceExporter.swift
lukaskollmer: yeah i forgot to change some of these to the new name

Sources/ApodiniDeploy/RemoteHandlerInvocationManager.swift
lukaskollmer: Yes, i had that thought as well. I decided to underscore the first path component, because this is an internal API and i didn't want it to cause any name clashes with user-defined handlers, but some other parts of Apodini (eg the grpc exporter) already define additional endpoints which are not underscored (eg `/apodini/proto`). These probably should be unified in some way. Generally speaking I'd say that all additional endpoints (ie endpoints which are not generated from a `Handler`) should be under one common "internal" path. A notable exception to this would be the OpenAPI definition which is currently served at `/openapi`. I don't know if that is required by the standard or could also be moved into the internal group.

Sources/DeploymentTargetLocalhost/main.swift
lukaskollmer: hmmm i'll have a look at this because that's what caused the swiftc hang for me the last time i tried it (also running 12D4e)

Sources/ApodiniDeploy/DeploymentOptions/Condition.swift
lukaskollmer: Yes exactly. The idea is as follows:
- ApodiniDeploy: the "user-facing" target. a web service would import this, as well as all deployment providers it wants to use
- ApodiniDeployBuildSupport: everything shared between ApodiniDeploy and ApodiniDeployRuntimeSupport. This is the target one would import to implement a deployment provider. (eg you'll see that neither of the deployment providers imports the ApodiniDeploy target, they import only ApodiniDeployBuildSupport and ApodiniDeployRuntimeSupport

I think moving the conditions into ApodiniDeployBuildSupport should be possible, but that would require us to make either `AnyOption.resolve` or `ConditionalOption` public (because the interface exporter needs to access these to resolve the conditions), which in turn would also make them public to the web service importing ApodiniDeploy. I wasn't sure if that would be a good idea, since the resolve function and the ConditionalOption types are implementation details which are not relevant to (and should not be used by) the user.

Sources/DeploymentTargetAWSLambda/main.swift
lukaskollmer: oh wow i had completely missed that apple finally added resource support to SPM. will move to that.

Sources/DeploymentTargetAWSLambda/main.swift
lukaskollmer: ok. Are you, in general, fine with abbreviating `FileManager` to "fm" or would you prefer the name be spelled out? Asking because i also added an exception to the swiftlint rules to allow this one specific identifier, which you may not like.

Sources/DeploymentTargetAWSLambda/AWSStuff.swift
lukaskollmer: Yes that's one of the items on my to-do list, as part of the AWS ressource mgmt thing.
My current idea is to make the --aws-profile-name optional (which i think technically speaking it already is), and then try to detect AWS credentials dynamically:
- If the CLI parameter is defined, that value always takes precedence and we try to use that profile in the config file
- If the CLI parameter is not defined, and we find `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` environment variables, we use these
- If the CLI parameter is not defined, and there are no environment variables, we try to use the "default" profile from the config file

Sources/ApodiniUtils/Swift.swift
PSchmiedmayer: I already guessed that this is the case. The Objective-C days ;D
Sounds good 👍

Sources/ApodiniDeployBuildSupport/Options/OptionKey.swift
lukaskollmer: Yea there's definitely a lot of duplication between the two Options APIs.

I did look into simply using @Supereg's implementation, but there are a couple of limitations to my use case which the other implementation doesn't cover.

I did look into using @Supereg's implementation, but there's a couple of constraints which apply to my use case, which do not apply to the existing implementation in the Apodini target.
The main issue is that my options API needs to be 100% codable: since the options are instantiated in one process (the web service) and then used in another (the deployment provider), we have to somehow transfer them across that barrier w/out any loss in type safety.
It would probably be possible to rewrite the existing options API in a way to make it Codable, but that would come at the cost of a worse API usage experience.
Example: The existing API currently uses `ObjectIdentifier` to compare option keys, meaning you can define option keys ([eg the `.http` option](https://github.com/Apodini/Apodini/blob/0653bb4f07dd67bbe45ebffb3e6b7d07a16f6903/Sources/Apodini/Properties/Options/HTTPParameterMode.swift)) simply by defining a static property. (This works because the `.http` object being used when specifying an `@Parameter` option is then the same one being used later to read that option.)
In my case, because these two objects (the one used when specifying the option in the web service, and the one used to read the option in the deployment provider) are completely distinct objects, using a static property and comparing the object's identity simply would not work, and i instead had to add string keys to my `AnyOptionKey` type: its still strongly typed when you use it locally and when you de/encode it, but the actual comparison of option keys has to fall back to the untyped strings used in the background.

Obviously we could just add the concept of explicitly defined keys to the existing options implementation, but that would result in an objectively worse API: you can already define options without having to key them, so why require adding keys if they're not actually required for the in-process usage of the options api?

There's a couple of other issues which are specific to my implementation, mainly the concept of unresolved options (the conditional ones), which may or may not contribute to the full set of options, depending on a handler-based condition. I'm not really sure if/how this could be best integrated into the existing API.

Sources/ApodiniDeployBuildSupport/Options/Namespace.swift
lukaskollmer: Yes pretty much. I tried to design the Options API as abstract as possible, hence the two-level namespacing.
The outer namespace creates scope for options, and the inner namespace is used to differentiate between otherwise-duplicate options within that scope.
Currently we have only one outer namespace (the one for deployment options), but in theory if eg we wanted to move the `@Parameter` options to this API, we could simply define a second outer namespace for ParameterOptions, and there would be a type-system-level check to make sure you cannot pass a deployment option to a parameter and vice versa.

I'll see what i can do about the naming. I agree that the Outer/InnerNamespace names aren't exactly perfect, but tbh I'm not sure `Category` or `Subset` are much better, since the inner namespace neither defines a category, nor a subset.

Sources/ApodiniDeploy/ApodiniDeployInterfaceExporter.swift
PSchmiedmayer: It can be that an Endpoint has multiple `@Parameter`s with the same name. The semantic model checks for this based on the collected `parameterNamespace` of each exporter. A normal exporter has a `.global` namespace indicating that parameter names must be unique. But some exporters might have a weaker namespace rule, e.g. the RESTful exporter. You can see the possible name spaces in `ParameterNamespace.swift`. As these name space definitions currently correspond to the main supported HTTP Options using this information and moving this into the AnyEndpointParameter should be fine 👍

Sources/ApodiniDeploy/ApodiniDeployInterfaceExporter.swift
PSchmiedmayer: I see. I think we should somehow make this configurable. It might be useful for debug purposes or to try out something but I think it would be an unexpected behavior if a production system would expose more routes then expected (even though it would just be duplicates as the `Handler` might already be exposed by an other interface exporter)

Sources/DeploymentTargetAWSLambda/AWSStuff.swift
PSchmiedmayer: That sounds like a good idea 🚀

Sources/ApodiniDeploy/DeploymentOptions/Condition.swift
PSchmiedmayer: If I understood this correctly the functionality of `ConditionalOption` is not really needed in `ApodiniDeployBuildSupport` because you can define an option without the need to implement anything on top to make it conditional. So every option that is defined can be made conditional and that functionality is only needed in ApodiniDeploy. If I understood this correctly I am fine with keeping it in the ApodiniDeploy target 👍

Sources/ApodiniDeploy/RemoteHandlerInvocationManager.swift
PSchmiedmayer: As far as I know there is not standard path that an OpenAPI specification has to be served on. I would propose to move every internal functionality behind a reserved `/apodini/` route. So `/apodini/openapi`, `/apodini/swaggerui`, `/apodini/proto`, and `/apodini/invoke` would all be located at these routes.
I have documented this in an issue #248

Sources/DeploymentTargetAWSLambda/main.swift
PSchmiedmayer: I would suggest to use the spelled out version (`fileManager`) to make it easier to read for anyone taking a first look at the codebase.

Sources/ApodiniDeployBuildSupport/Options/Namespace.swift
PSchmiedmayer: I think moving the `Parameters` over to the new Options API would be a good idea, I like that you already had this extensibility in mind 👍
I agree that `Outer`/`InnerNamespace` or `Category` and `Subset` are not the best names, maybe we can find a better name along the way.

Sources/ApodiniDeployBuildSupport/Options/OptionKey.swift
PSchmiedmayer: Thanks for the detailed overview! As mentioned in a comment above maybe the other way around would work? So instead of using the Options API that we have at the moment for the Parameters we could use the more generalized API that you have at the moment using the inner and outer namespaces. I could even imagine that the conditional API would be a very nice addition in the future to enable the usage of conditionals e.g. in the Metadata DSL. And if the Parameters Options and all other Options follow the same API it would me even simpler to move some of the aspects into the Metadata DSL in the future.

Sources/ApodiniDeploy/DeploymentOptions/Condition.swift
lukaskollmer: Yeah pretty much. There's an extension `when(_ condition: AnyHandlerCondition)` on `AnyOption`, which allows you to turn any option into a conditional option (eg: `.memory(.mb(12)).when(...)`).
I initially defined the API in a way that you'd put the condition directly into the option (eg: `.memory(.mb(12), when: ...)`), but that'd require the person defining a new option to also add the necessary overloads to support conditions, which just results in a lot of unnecessary boilerplate code

Sources/ApodiniDeploy/ApodiniDeployInterfaceExporter.swift
lukaskollmer: In this specific case, we end up here after launching the web service with the `WellKnownCLIArguments.exportWebServiceModelStructure` argument, which tells the web service to write its model to the file and then terminate (the web service is running with the sole purpose to perform the initial run of the semantic model builder), so I'm not sure delegating the error to the caller would be a good idea, since the web service should be terminated anyway as soon as possible after writing the model file.

I do agree with you that in general, errors should be able to be caught somewhere up the call stack, and the other branch (`WellKnownCLIArguments.launchWebServiceInstanceWithCustomConfig`) does in fact do that exact thing.

Sources/ApodiniDeploy/ApodiniDeployInterfaceExporter.swift
lukaskollmer: Nope, this works the other way around: we're mapping the set of collected parameters (i.e. the parameters passed to the remote invocation API) to their corresponding `EndpointParameter`s.
Because the parameter objects (think of these as key-value pairs mapping a parameter key path to the value passed for that parameter) you pass to the remote handler invocation API can only be you can only constructed by referencing an `@Parameter`'s key path, we know that every collected parameter has some corresponding `@Parameter`.
And because every `@Parameter` has a corresponding `EndpointParameter`, we can guarantee that this mapping from collectedParameters to EndpointParameters will always succeed.

summary: we're mapping from a subset to a superset, therefore we will always find the matching EndpointParameter and the force unwrap should never fail

Sources/ApodiniDeploy/RemoteHandlerInvocationManager.swift
lukaskollmer: In this case we technically wouldn't throw, but return a failed future.
I'm currently a bit careful about returning failed futures, because that's also how errors thrown in handlers are returned to the invoking handler, and otherwise there'd be no distinction between "the remote handler invocation API returned a failed future because the called handler threw an error" and "the remote handler invocation API returned a failed future because something else internally went wrong".
For example, none of the remote-handler-invocation functions (most of them are called something alone the lines of `invoke`) currently throw, because otherwise the user would have to handle both the remoteHandlerAPI throwing, as well as it returning a failed future

Sources/ApodiniDeployBuildSupport/Options/Namespace.swift
lukaskollmer: Yes I think that might work, although you'd lose some convenience because you's have to manually specify keys instead of using pointer equality to compare options (which is how the parameter options API currently works).

I'll have a look.

Sources/ApodiniDeploy/ApodiniDeployInterfaceExporter.swift
lukaskollmer: Yes that's a good point, i'll limit it to `InvocableHandler`s for the time being, and, if at some point in the future there is an API for defining exporter configurations (the way i'm imagining this it would be part of the stackable exporters, since the internal invocation API would probably be used by some other exporter or part of Apodini) we can bring it back via that.

Sources/ApodiniDeploy/ApodiniDeployInterfaceExporter.swift
PSchmiedmayer: Thanks for the explanation, sounds good to me 👍

Sources/ApodiniDeploy/ApodiniDeployInterfaceExporter.swift
PSchmiedmayer: I see, this makes sense. I would suggest to add a code comment with the reasoning mentioned above to the source code so this is pronely documented and a future reader can follow your logic more easily 👍

Sources/ApodiniDeploy/RemoteHandlerInvocationManager.swift
PSchmiedmayer: > otherwise there'd be no distinction between "the remote handler invocation API returned a failed future because the called handler threw an error" and "the remote handler invocation API returned a failed future because something else internally went wrong".

I understand, this makes sense. From a perspective of a developer using the invocation API the only really way to check this would be to inspect the error and see if the type is an `ApodiniError `(error thrown by a `Handler` if it follows the Apodini conventions) or any other error that would point to an error that most likely is based on an implementation mistake. I think this is a general thing that he exporters should consider, e.g. only forward ApodiniErrorrs to an API consumer and log all other errors internally and only return e.g. an 500 internal server error code for REST to the consumer. It would be good to note this in an GitHub issue so we take a closer look at this in the future.

PSchmiedmayer (approved):
As discussed in the thesis meeting we will merge this PR in its current state which already looks great! Thanks for all the work @lukaskollmer 💪
More tests, a consequently larger test coverage and other improvements will follow in a next PR.

PR #247, opened 2021-02-24
Binding

theMomax:
# Binding

Until now, re-using `Handler`s was only possible when the context was exactly the same at two points in your service.

### Problem that is solved
In general a `Handler` doesn't care where the information relevant for generation the output comes from. That is only relevant for your service-definition. A `Handler` doesn't care if it gets a `String` from a `PathParameter`, a normal `Parameter` or the `Environment`.

## :bulb: Proposed solution

```swift
struct HelloWorld: WebService, EnvironmentAccessible {
    struct Greeter: Handler {
        @Binding var country: String?
        
        func handle() -> String {
            country ?? "World"
        }
    }
        
    var featured: String? = "Italy"
    
    @PathParameter var selectedCountry: String
    @Environment(\BindingTests.featured) var featuredCountry
    
    var content: some Component {
        Greeter(country: .constant(nil))
        Group("default") {
            Greeter(country: Binding<String?>(.constant("USA")))
        }
        Group("optional") {
            Greeter(country: Parameter<String?>().projectedValue)
        }
        Group("featured") {
            Greeter(country: $featuredCountry)
        }
        Group("country", $selectedCountry) {
            Greeter(country: Binding<String?>($selectedCountry.projectedValue))
        }
    }
}
```

### Implications
Everything is backwards compatible. The `@Binding` property wrapper was added. `@Parameter` and `@Environment` now have a `projectedValue` returning the respective `Binding`.

In order to implement this I had to adapt `Properties`/`Traversable` and `DynamicProperty`. They now enable the user to define a `namingStrategy` so we can choose which name to the framework should use.

E.g. for the following example the `wrapped` `Parameter` would be exported as `country`, now `wrapped`.
```swift
struct Wrapper: DynamicProperty {
    @Parameter var wrapped: String

    func namingStragegy(_ names: [String]) -> String? {
        names[names.count - 2]
    }
}

let country = Wrapper()
```

### Testing
I added tests for both, `Binding` and `namingStrategy`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,130 @@
 
 A declarative, composable framework to build web services using Swift.
 
-The framework is in a very early alpha phase. You can inspect the current development manifestos describing the framework in the [documentation folder](Documentation/)
+## Getting Started
+
+### Installation
+
+Apodini uses the Swift Package Manager:
+
+Add it as a project-dependency:
+```swift
+dependencies: [
+    .package(url: "https://github.com/Apodini/Apodini.git", .branch("develop"))
+]
+```
+
+Add the base package and all exporters you want to use to your target:
+```swift
+targets: [
+    .target(
+        name: "Your Target",
+        dependencies: [
+            .product(name: "Apodini", package: "Apodini"),
+            .product(name: "ApodiniREST", package: "Apodini"),
+            .product(name: "ApodiniOpenAPI", package: "Apodini")
+        ])
+]‚
+
+```
+
+### Hello World
+
+Getting started is really easy:
+
+```swift
+import Apodini
+import ApodiniREST
+
+struct Greeter: Handler {
+    @Parameter var country: String?
+
+    func handle() -> String {
+        "Hello, \(country ?? "World")!"
+    }
+}
+
+struct HelloWorld: WebService {
+    var configuration: Configuration {
+        ExporterConfiguration()
+            .exporter(RESTInterfaceExporter.self)
+    }
+
+    var content: some Component {
+        Greeter()
+    }
+}
+
+try HelloWorld.main()
+
+// http://localhost:8080/v1 -> Hello, World!
+// http://localhost:8080/v1?country=Italy -> Hello, Italy!
+```
+
+Apodini knows enough about your service to automatically generate OpenAPI docs. Just add the respective exporter:
+
+```swift
+import ApodiniOpenAPI
+...
+struct HelloWorld: WebService {
+    var configuration: Configuration {
+        ExporterConfiguration()
+            .exporter(RESTInterfaceExporter.self)
+            .exporter(OpenAPIInterfaceExporter.self)
+    }
+    ...
+}
+
+// JSON definition: http://localhost:8080/openapi
+// Swagger UI: http://localhost:8080/openapi-ui
+```
+
+With `Binding`s we can re-use `Handler`s in different contexts:
+```swift
+struct Greeter: Handler {
+    @Binding var country: String?
+
+    func handle() -> String {
+        "Hello, \(country ?? "World")!"
+    }
+}
+
+struct HelloWorld: WebService {
+    var configuration: Configuration {
+        ExporterConfiguration()
+            .exporter(RESTInterfaceExporter.self)
+            .exporter(OpenAPIInterfaceExporter.self)
+    }
+    
+    @PathParameter var country: String
+
+    var content: some Component {
+        Greeter(country: .constant(nil))
+            .description("Say 'Hello' to the World.")
+        Group("country") {
+            CountrySubsystem()
+        }
+    }
+}
+
+struct CountrySubsystem: Component {
+    @PathParameter var country: String
+    
+    var content: some Component {
+        Group($country) {
+            Greeter(country: Binding<String?>($country.projectedValue))
+                .description("Say 'Hello' to a country.")
+        }
+    }
+}
+
+// http://localhost:8080/v1 -> Hello, World!
+// http://localhost:8080/v1/country/Italy -> Hello, Italy!
+```
+
+## Documentation
+
+The framework is in early alpha phase. You can inspect the current development manifestos describing the framework in the [documentation folder](Documentation/)
 
 You can find a generated technical documentation for the different Swift types at [https://apodini.github.io/Apodini](https://apodini.github.io/Apodini)
 
--- a/README.md
+++ b/README.md
@@ -99,8 +99,6 @@ struct HelloWorld: WebService {
             .exporter(RESTInterfaceExporter.self)
             .exporter(OpenAPIInterfaceExporter.self)
     }
-    
-    @PathParameter var country: String
 
     var content: some Component {
         Greeter(country: .constant(nil))
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ struct HelloWorld: WebService {
     }
 
     var content: some Component {
-        Greeter(country: .constant(nil))
+        Greeter(country: nil)
             .description("Say 'Hello' to the World.")
         Group("country") {
             CountrySubsystem()
@@ -114,7 +114,7 @@ struct CountrySubsystem: Component {
     
     var content: some Component {
         Group($country) {
-            Greeter(country: Binding<String?>($country.projectedValue))
+            Greeter(country: Binding<String?>(_country.binding))
                 .description("Say 'Hello' to a country.")
         }
     }
--- a/Sources/Apodini/Properties/Binding.swift
+++ b/Sources/Apodini/Properties/Binding.swift
@@ -87,7 +87,7 @@ extension Binding where Value: Codable {
 
 extension Binding {
     /// Creates a binding by projecting the base value to an optional value.
-    public init<V: Codable>(_ base: Binding<V>) where Value == V? {
+    public init<V>(_ base: Binding<V>) where Value == V? {
         self.store = base.store
         switch base.retrieval {
         case .constant(let value):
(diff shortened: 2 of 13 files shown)


Sources/Apodini/Properties/Properties.swift
```diff
@@ -16,19 +16,37 @@ import Foundation
 public struct Properties: Property {
     internal var elements: [String: Property]
     
+    internal var namingStrategy: ([String]) -> String? = Self.defaultNamingStrategy
```
PSchmiedmayer, line 19: I think that the index based naming strategy is a bit confusing and not easy to understand when encountering it in the source code. As far as I can see there are only two base strategies that I could find based on your comments:
1. Use the name of the `@Parameter` itself (`names[names.count-1]` or `names.last`)
2. The name of the wrapping `Properties` (`names[names.count-2]`).

With multiple `@Properties` applying a naming strategy for them to use the parent's name would result in naming collisions. As far as I can see, using this strategy only really applies to cases where you only have one `@Parameter` in the `Properties` construct, is this correct? 

In general it might be desirable to have a naming strategy that is dependent on the key in the `elements` dictionary. So defining a function `@escaping (String) -> NamingStrategy` with `NamingStrategy` being an enum `propertyName` or `parentName` would enable a developer to express the same functionality. And this would then allow you to recursively move up a potential `Properties` chain if there is more than one layer.

Sources/Apodini/Properties/Binding.swift
```diff
@@ -0,0 +1,99 @@
...
+        self.store = Properties()
+        self.retrieval = .constant(constant)
+    }
+    
+    /// Create a `Binding` that always returns the given `value`.
+    public static func constant(_ value: Value) -> Binding<Value> {
+        Binding(constant: value)
+    }
+}
+
```
PSchmiedmayer, line 47: I would suggest to add more convenience initializers for the `Binding` that all relate to the `ExpressibleBy...` protocols: https://developer.apple.com/documentation/swift/swift_standard_library/initialization_with_literals

This can enable to write
```swift
Greeter(country: .constant(nil))
```
as
```swift
Greeter(country: nil)
```
Similar functionality can be achieved for Strings (`ExpressibleByStringLiteral`) which makes the API more compact and easier to read.

Tests/ApodiniTests/PropertiesTests/BindingTests.swift
```diff
@@ -0,0 +1,106 @@
...
+    }
+        
+    var featured: String? = "Italy"
+    
+    @PathParameter var selectedCountry: String
+    @Environment(\BindingTests.featured) var featuredCountry
+    
+    @ComponentBuilder
+    var testService: some Component {
+        Greeter(country: .constant(nil))
```
PSchmiedmayer: As commented above with the following extension
```swift
extension Binding: ExpressibleByNilLiteral where Value: ExpressibleByNilLiteral {
    public init(nilLiteral: ()) {
        self.init(constant: nil)
    }
}
```
allows you to write:
```suggestion
        Greeter(country: nil)
```

Tests/ApodiniTests/PropertiesTests/BindingTests.swift
```diff
@@ -0,0 +1,106 @@
...
+        Group("default") {
+            Greeter(country: Binding<String?>(.constant("USA")))
```
PSchmiedmayer: It would be great to use:
```suggestion
            Greeter(country: "USA")
```
here. We should investigate if we can even support string interpolation here but I think that just enabling the usage of String Literals if the type is a string would be great!
Similar things apply to bools, floats and so on

Tests/ApodiniTests/PropertiesTests/BindingTests.swift
```diff
@@ -0,0 +1,106 @@
...
+        }
+        Group("optional") {
+            Greeter(country: Parameter<String?>().projectedValue)
```
PSchmiedmayer: I remember that we had several discussions about allowing `@Parameter`s in `Component`s. As far as I remember the conclusion was that they should not be used in `Component`s as a `Component` doesn't have a handle function. But from looking at the idea of `Bindings` I think it might make sense to have a `Component` like this:
```swift
struct AComponent: Component {
    @Parameter var test: String?
    
    var content: some Component {
        Greeter(country: $test)
    }
}
```
and similarly it would make sense to have a `@Parameter` here and pass the projected value to the `Greeter`. On the other hand this might be something that could be a bit confusing as the `@Paramater` is not used in the context of the `Component` itself but is only passed down to the handling `Handler`. But as we can not really restrict the usage of `@Parameter`s on compile time and the usage of property wrappers for local variables is on the way with Swift 5.4 we can take advantage of this and might also want to test that behavior in a unit test.

Tests/ApodiniTests/PropertiesTests/BindingTests.swift
```diff
@@ -0,0 +1,106 @@
...
+        }
+        Group("featured") {
+            Greeter(country: $featuredCountry)
+        }
+        Group("country", $selectedCountry) {
+            Greeter(country: Binding<String?>($selectedCountry.projectedValue))
```
PSchmiedmayer: It would be cool to think about a way to make this a bit cleaner. Maybe a `binding` or an `optionalBinding` property that does the same but hides it behind a property of a `@Parameter` or `@PathParameter` so we could write:
```suggestion
            Greeter(country: $selectedCountry.projectedValue.optionalBinding)
```
or even better:
```suggestion
            Greeter(country: $selectedCountry.optionalBinding)
```
I could imagine that could make the `public init<V: Codable>(_ base: Binding<V>) where Value == V? { /* ... */}` initializer internal in this process.

Sources/Apodini/Properties/Binding.swift
```diff
@@ -0,0 +1,99 @@
...
+//
+//  Created by Max Obermeier on 23.02.21.
+//
+
+import Foundation
+
+private enum Retrieval<Value> {
+    case constant(Value)
+    case storage((Properties) -> Value)
+}
```
PSchmiedmayer, line 13: I looked over the name `namingStrategy` code and came by the idea of just storing a `Property` in the `storage` instead of a `Porperties` instance which would in turn reduce the need for the `namingStrategy` as we would not have the nesting anymore.
I am sure you already thought about the idea and using `Porperties` and consequentially implementing the `namingStrategy` was the solution. What was the reason for not using a `Property` here? The type casting and everything around it should also work without using the key based index in the `Properties` struct.

README.md
```diff
@@ -5,7 +5,128 @@
 
 A declarative, composable framework to build web services using Swift.
 
-The framework is in a very early alpha phase. You can inspect the current development manifestos describing the framework in the [documentation folder](Documentation/)
+## Getting Started
```
PSchmiedmayer, line 8: Thanks a lot for a first README for the Apodini project 🚀

PSchmiedmayer (changes requested):
I really like the feature, thanks a lot @theMomax!

I had a few comments here and there in the source code that would make the feature even easier to use. But apart from these comments and one more general question I had this looks really nice 💪

Sources/Apodini/Properties/Binding.swift
theMomax: I agree...adding some of those for the most important cases makes it a lot easier to read👍🏻

Tests/ApodiniTests/PropertiesTests/BindingTests.swift
theMomax: I don't think this is possible. We cannot have two implementations for one protocol even with different associated type requirements and the compiler isn't smart enough to coerce from string literal to Binding<String> and then to Binding<String?>. Best we can get is `Binding<String?>("USA")`.

Sources/Apodini/Properties/Properties.swift
theMomax: > With multiple @Properties applying a naming strategy for them to use the parent's name would result in naming collisions.

Yes, but naming collisions have been possible before, e.g.
```swift
struct A: DynamicProperty {
    // x is used as the name
    @Parameter var x: String
}

struct B: Handler {
    var a: A
    var b: A
}
```
The user is responsible for avoiding those.

> As far as I can see, using this strategy only really applies to cases where you only have one @Parameter in the Properties construct, is this correct?

No. Currently this is the case in the only use-case, but I designed it so that it provides as much flexibility as possible. I could imagine a _shortest unique postfix_ strategy could be useful too in some cases. This would not be possible with your simplified approach.

I consider this code not to be part of the _standard_ API that one would have to use on a daily basis, thus I think it is ok to make it a bit more complex and give more power to the user.

Tests/ApodiniTests/PropertiesTests/BindingTests.swift
theMomax: Done with 5195cbc354c8f508b37339215035bdad1e4997ab

Tests/ApodiniTests/PropertiesTests/BindingTests.swift
theMomax: For some reason `Binding<String?>("USA")` won't work, however `.constant("USA")` on its own does work. I also created a `.some("USA")` that does basically the same as `constant`, except the naming and typing is adapted to the special case of optional-wrapping.

Tests/ApodiniTests/PropertiesTests/BindingTests.swift
theMomax: > But from looking at the idea of Bindings I think it might make sense to have a Component like this: ...

I agree. I changed the test to reflect this.

Tests/ApodiniTests/PropertiesTests/BindingTests.swift
theMomax: I added an `binding` computed property on `@PathParameter` and an `asOptional` one on `Binding`. This makes it a lot cleaner indeed: `_selectedCountry.binding.asOptional`.

> I could imagine that could make the public init<V: Codable>(_ base: Binding<V>) where Value == V? { /* ... */} initializer internal in this process.

I See no point in doing so. The signature is an exact copy of SwiftUI's `Binding`. I think having a familiar element there might help some developers.

Sources/Apodini/Properties/Binding.swift
theMomax: The point is that I would have needed an `Optional<Property>` which cannot be traversed and thus its content is never found. I could also have implemented `Traversable` for `Optional` or have created a `None` property, but I didn't want to create any new primitives and instead use what we already have. If you prefer either of the mentioned versions, I could also do it that way, however, note that neither of them works without the `namingStrategy`.

Tests/ApodiniTests/PropertiesTests/BindingTests.swift
PSchmiedmayer: Looks good 🚀

Sources/Apodini/Properties/Binding.swift
PSchmiedmayer: I understand. I will merge the current version in develop, I think that the naming strategy is good first step here. We might want to generalize this in the future but the current implementation is well rounded and therefore good to be merged in the main develop branch 🚀

Sources/Apodini/Properties/Properties.swift
PSchmiedmayer: I see. As commented [here](https://github.com/Apodini/Apodini/pull/247#discussion_r599526726) this sounds good to me 👍

PSchmiedmayer (approved):
@theMomax Thanks for the feature and the work you put into the PR, I will merge it in develop now 🚀

PR #252, opened 2021-03-14
Serve Static Files

nerdsupremacist:
# Serve Static Files

There's currently no way to serve files via HTTP while using Apodini

## :bulb: Proposed solution
We should allow the developer to make any directory of static files public. Available options:
- Which directory is public
- Is there an http path prefix (for serving multiple different folders)

Due to the fact that serving files is http specific and not protocol specific, my proposed solution is to add this functionality as a configuration:

```swift
struct TestWebService: Apodini.WebService {
    var content: some Component {
        ...
    }
    
        ...
        StaticFiles(publicDirectory: "Public", at: "files", "public")
    }
}
```

### Problem that is solved
The configuration struct allows the developer to serve the directory via http, regardless of whether or not they're using REST, gRPC or GraphQL 

### Implications
No particular implication

### Related PRs
None

### Testing
I tested with the Test Web Service both the case with and without a prefix

### Reviewer Nudging
- In all honesty I'm not super happy with the naming so far. I'm happy to discuss better naming.
- Also the current implementation checks if the folder you're making public resides in the folder of the Swift Package, and then checks in the CWD. I did this since Xcode will set the CWD to the derived data folder with the executable. I thought this would be a great default especially for the dev environment, but feel free to tell me why it's a bad idea

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -24,8 +24,8 @@
         "repositoryURL": "https://github.com/swift-server/async-http-client.git",
         "state": {
           "branch": null,
-          "revision": "ba845ee93d6ea10671e829ebf273b6f7a0c92ef0",
-          "version": "1.2.4"
+          "revision": "0a8dddbe15cd72f7bb5e47951fb7c1eb0836dfc2",
+          "version": "1.2.2"
         }
       },
       {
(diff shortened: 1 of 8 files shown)


Package.resolved
```diff
@@ -24,17 +24,17 @@
         "repositoryURL": "https://github.com/swift-server/async-http-client.git",
         "state": {
           "branch": null,
-          "revision": "ba845ee93d6ea10671e829ebf273b6f7a0c92ef0",
-          "version": "1.2.4"
+          "revision": "0a8dddbe15cd72f7bb5e47951fb7c1eb0836dfc2",
+          "version": "1.2.2"
```
tgymnich, line 28: why are all packages downgraded?

tgymnich (commented):
Have you considered using the bundle/ spm resources to store static html files? As far as I remember spring boot als uses some resource folder inside the  package.

PR #260, opened 2021-03-29
Eldicano/webservice model

eldcn:
# Web service structure and comparison logic

## :bulb: Proposed solution
Adds a generic structure of the web service, capturing most relevant interface related details. The model of the web service structure is not final, should serve only as a POC for the comparison logic. All elements of the current structure can be compared with an other version and the change is rendered in a change object. Two exemplary `json` files are added in `TestWebService` package for the web service structure and a change object.

### Problem that is solved
*Provide a description and link issues that are solved*

### Implications
- Pull request to migration-guide base branch

-

### Related PRs
-

### Testing
Two basic test classes that test building of schemas and comparison logic

### Reviewer Nudging
I would suggest that you first start with `ComparableTests.swift` to have a first look how the comparison works, then `ApodiniDelta/Compare` and afterwards with how the elements of web service structure and schemas conform to the `Comparable` protocols


Sources/ApodiniDelta/Change/ChangeType.swift
```diff
@@ -0,0 +1,18 @@
...
+//  File.swift
+//  
+//
+//  Created by Eldi Cano on 27.03.21.
+//
+
+import Foundation
+
+/// Distinct change types that might occur
+enum ChangeType: String, Value {
```
PSchmiedmayer, line 11: It seems to me that the type of change is manifested in two different ways: The inheritance of the `Change` class and the `changeType` property of a `Change`. What is the reason for this redundancy? As far as I could see this was not yet used in the code but I suspect that there is a valid reason for it.

Sources/ApodiniDelta/Change/CompositeChange.swift
```diff
@@ -0,0 +1,62 @@
...
+}
+
+// MARK: - Array extension
+extension Array where Element: Change {
+    func equalsIgnoringOrder(to other: Self) -> Bool {
+        guard count == other.count else {
+            return false
+        }
+
+        for change in self where other.filter({ $0.isEqual(to: change) }).isEmpty {
```
PSchmiedmayer: You can use `contains` here:
```suggestion
        for change in self where other.contains(where: { $0.isEqual(to: change) }) {
```

Sources/ApodiniDelta/Compare/ChangeContextNode.swift
```diff
@@ -0,0 +1,52 @@
...
+
+    /**
+     These functions can be removed if the type checking for ComparableCollection would work as expected
+     */
+
+    /// Used to register the result of comparison between two collections of comparable objects
+    /// - Parameters:
+    ///   - result: result of comparison
+    ///   - type: The type of the `Element` of the collection
+    func register<O: ComparableObject>(result: CollectionChangeContextNode<O>, for type: O.Type = O.self) {
```
PSchmiedmayer: To make the function call a bit less transparent for the caller, would it be possible to remove the argument name and have the same function signature as above or does this then confuse the compiler on what function to call?
```suggestion
    func register<O: ComparableObject>(_ result: CollectionChangeContextNode<O>, for type: O.Type = O.self) {
```

Sources/ApodiniDelta/Compare/ChangeContextNode.swift
```diff
@@ -0,0 +1,52 @@
...
+        if result.containsChange {
+            precondition(changes[O.identifier] == nil, "Attempting to override changes of \(type).")
+            changes[O.identifier] = result
+        }
+    }
+
+    /// Collection change retrieval for a specific type
+    func change<O: ComparableObject>(comparable: O.Type) -> CollectionChangeContextNode<O>? {
```
PSchmiedmayer, line 44: Same as the comment above

Sources/ApodiniDelta/Compare/ComparableCollection.swift
```diff
@@ -0,0 +1,71 @@
+//
```
PSchmiedmayer: Please remove commented files

Sources/ApodiniDelta/Compare/ComparableObject.swift
```diff
@@ -0,0 +1,39 @@
...
+/// Base protocol that all comparable objects conform to
+protocol ComparableObject: _Comparable {
+    typealias Result = ChangeContextNode
+
+    var deltaIdentifier: DeltaIdentifier { get }
+
+    func evaluate(result: Result, embeddedInCollection: Bool) -> Change?
+}
+
+extension ComparableObject {
```
PSchmiedmayer, line 17: Nice shortcut 👍

Sources/ApodiniDelta/Configuration/DeltaConfiguration.swift
```diff
@@ -0,0 +1,49 @@
...
+public class DeltaConfiguration: Configuration {
+    var webServiceStructurePath: String?
+    var strategy: DeltaStrategy = .create
+
+    public init() {
+        self.webServiceStructurePath = nil
+
+        #if Xcode
+        runShellCommand(.killPort(8080))
+        #endif
```
PSchmiedmayer, line 37: I think it would even make sense to add this to the default main function and get the port from the HTTP configuration. I would suggest to check if we are running in debug mode and if we do so kill the port independent of if we run in Xcode or not

Sources/ApodiniDelta/Exports.swift
```diff
@@ -0,0 +1,9 @@
+//
+//  File.swift
+//  
+//
+//  Created by Eldi Cano on 19.03.21.
+//
+
+@_exported import ApodiniTypeReflection
```
PSchmiedmayer: What APIs of the Type reflection API are commonly used that support the decision to export the complete interface of `ApodiniTypeReflection`?

Sources/ApodiniDelta/Extensions/Array+Extensions.swift
```diff
@@ -0,0 +1,14 @@
+//
```
PSchmiedmayer, line 1: Some of the content found in Extensions could move to ApodiniUtils if this an extension that is generally useful for other parts of the project. Is up to you what you would consider generally useful or what is only used in ApodiniDelta.

Sources/ApodiniDelta/Extensions/Node+Extensions.swift
```diff
@@ -0,0 +1,54 @@
...
+        }
+    }
+}
+
+// MARK: - ReflectionInfo Node extensions
+extension Node where T == ReflectionInfo {
+    // MARK: - Properties
+    var isPrimitive: Bool {
+        let type = value.typeInfo.type
+        return isSupportedScalarType(type) || type == UUID.self
```
PSchmiedmayer: Maybe we should think adding UUID as a scalar type

Sources/ApodiniDelta/Schema/PrimitiveType.swift
```diff
@@ -0,0 +1,37 @@
...
+//  
+//
+//  Created by Eldi Cano on 20.03.21.
+//
+
+import Foundation
+
+extension PrimitiveType: ComparableProperty {}
+
+enum PrimitiveType: String, RawRepresentable, Codable {
```
PSchmiedmayer: We also have some logic about scalar types in Apodini. Could this be merged with this `PrimitiveType` implementation or maybe there is some overlap there?

Sources/ApodiniDelta/Schema/Schema.swift
```diff
@@ -0,0 +1,140 @@
...
+//  File.swift
+//  
+//
+//  Created by Eldi Cano on 20.03.21.
+//
+
+import Foundation
+
+class SchemaName: PrimitiveValueWrapper<String> {}
+class IsEnum: PrimitiveValueWrapper<Bool> {}
```
PSchmiedmayer: Why is it necessary to store these values in the wrapper types? Could it also be possible that the `Primitive` protocol also inherits the requirements of `ComparableProperty` if this is needed here?

Sources/ApodiniDelta/Schema/Schema.swift
```diff
@@ -0,0 +1,140 @@
...
+
+    /// Properties of the schema
+    let properties: Set<SchemaProperty>
+
+    /// Indicates whether the schema is an enumeration
+    /// If that is the case, the properties represent the cases as strings
+    let isEnumeration: IsEnum
+
+    /// The reference to the own schema
+    var reference: SchemaReference {
```
PSchmiedmayer: If we go for the name as a unique key we might want to consider the edge chase that there are two types named exactly the same but they are defined in different modules. I am not quire sure how to handle the edge case, can we get the name of the module where the type was defined too? If this is too completed or not even relevant for now we don't have to invest too much time in this.

Sources/ApodiniDelta/Schema/SchemaProperty.swift
```diff
@@ -0,0 +1,132 @@
...
+//  File.swift
+//  
+//
+//  Created by Eldi Cano on 20.03.21.
+//
+
+import Foundation
+
+class PropertyName: PrimitiveValueWrapper<String> {}
+class PropertyOffset: PrimitiveValueWrapper<Int> {}
```
PSchmiedmayer: Same Comment as in `Schema.swift`

Sources/ApodiniDelta/Schema/Schema.swift
```diff
@@ -0,0 +1,140 @@
+//
```
PSchmiedmayer, line 1: The general question would be if it might even be useful to move the complete `Schema` code in the ApodiniTypeReflection target. I am not sure if `Schema` is the best word here as you more or less describe the structure of a type if I got this correctly. I hope I got this correct: As far as I can see you create a high level wrapper around the `TypeReflection` information that is containing more sanitized and compact information about the types that are used. 
@theMomax will tackle the type reflection problem as part of his thesis so it might make sense to merge and combine the logic that you use here and the work in the type reflection part (that will move away from type reflection) into a more general "Type description" part that can then be used to compare types (your work) but also can be used to provide type information to the exporters. What do you both think (@theMomax @eldcn)? Or am I going in the wrong direction with this?

Sources/ApodiniDelta/Schema/SchemaReference.swift
```diff
@@ -0,0 +1,31 @@
...
+        SchemaReference(schemaName: named)
+    }
+
+    static func reference(_ primitiveType: PrimitiveType) -> SchemaReference {
+        .reference(primitiveType.description)
+    }
+
+    static var empty: SchemaReference {
+        .reference("")
+    }
```
PSchmiedmayer: What is an Empty SchemaReferenced used for? We actually have an `Empty` type in Apodini but I don't think that it makes sense to return `"Empty"` here as this is a differentiation between having no type information and encountering the `Empty` typer here isn't it?

Sources/ApodiniDelta/Shared/Primitive.swift
```diff
@@ -0,0 +1,22 @@
...
+//  File.swift
+//  
+//
+//  Created by Eldi Cano on 29.03.21.
+//
+
+import Foundation
+
+/// A helper protocol that all primitive types (as considered in Apodini) conform to
+protocol Primitive: Value {}
```
PSchmiedmayer: As noted above we might want to merge this code with the scalar type code that we have in Apodini already we don't have primitives and scalar types in Apodini which seem very similar to me.

Sources/ApodiniDelta/Shared/PrimitiveValueWrapper.swift
```diff
@@ -0,0 +1,25 @@
...
+//  File.swift
+//  
+//
+//  Created by Eldi Cano on 29.03.21.
+//
+
+import Foundation
+
+/// An abstract class that already conforms to ComparableProperty that wrapps values of a certain primitive type
+class PrimitiveValueWrapper<P: Primitive>: ComparableProperty {
```
PSchmiedmayer: As noted in the comments above I am not 100% sure what the exact need for this type is. I remember talking about this in the last meeting but I must be honest that I don't remember the exact use case and making a type that conforms to `Primitive` also conform to `ComparableProperty` by defining `Primitive` like this:
```swift
protocol Primitive: ComparableProperty { /*...*/ }
```
would achieve the same result at first glance.

Sources/ApodiniDelta/WebServiceStructure/Service.swift
```diff
@@ -0,0 +1,92 @@
...
+//
+
+import Foundation
+
+extension Apodini.Operation: ComparableProperty {}
+class ServicePath: PrimitiveValueWrapper<String> {}
+class HandlerName: PrimitiveValueWrapper<String> {}
+
+/// Represents an endpoint
+struct Service: Codable {
```
PSchmiedmayer: As far as I can see a `Service` is a more compact representation of an `Endpoint` that is used to compare different `Endpoints`. I find the name `Service` not really suiting for this abstraction but most be honest that I don't have a suggestion for a better name at the moment. While I see the appeal for creating a new data structure to compare two different endpoints and their parameters I am a bit concerned that this is going to lead to some code duplication and difficulties when we start to evolve the semantic model including simplifying `Endpoints` in the future.

I would like to get your honest opinion on this: Wouldn't it also be possible to make endpoint conform to `Codable` and `ComparableObject` in an extension as part of this target without creating a new type? I think that cleaning up an `Endpoint` and reducing it to the necessary properties and lazily computing a lot of different properties (e.g. out of the context) might most possible benefit the system as a whole a lot more. Maybe also @lukaskollmer @theMomax and @Supereg 's option on this would be very interesting! I could see that a way to make the existing `WebServiceModel` and `Endpoints` `Codable` would also reduce the need for a dedicated exporter. Not only for your work but also for the work that @lukaskollmer did with this exporter.

PSchmiedmayer: The same also applies to `ServiceParameter` and `WebServiceStructure`. I think that the current semantic model grew very organically and that focusing on this would be a good think that @theMomax is also addressing. I already noted this in a meeting with him: I would be fine to reduce the number of currently supported exporters to REST/JSON API and an OpenAPI exporter for now if this makes it easier to better structure the WebServiceModel and the complete semantic model and try to integrate gRPC, WebSockets, and GraphQL in a next step but keep them in the back of our heads when improving the inner workings of Apodini.

Sources/ApodiniUtils/Shell.swift
```diff
@@ -0,0 +1,41 @@
...
+//  File.swift
+//
+//
+//  Created by Eldi Cano on 30.03.21.
+//
+
+import Foundation
+
+/// Contains predefined shell command cases
+public enum ShellCommand {
```
PSchmiedmayer, line 11: As far as I remember @lukaskollmer also has some Schell execution abstraction in his Deployment Provider Kit, maybe moving this here into ApodiniUtlils and merging it with your code might be a nice idea to share it with more modules.

PSchmiedmayer (changes requested):
@eldcn Thanks for the great PR. I really like the way you write code and how well documented everything is and how it follows a clear and easy-to-understand style! I would suggest direly merge the code you introduce here into the development branch if possible. I think the only thing stopping us might be the test coverage, but I am sure that some more unit tests would bring it up to speed with the current Apodini level.

I have a few comments throughout the code that were mostly some smaller remarks and suggestions. Nevertheless, there are several parts where I have the feeling that code could be moved to another target or there are existing structures in Apodini that could be merged with your code. In most instances, I think that the existing structures could greatly benefit from your good structured code and could be replaced/augmented with your current code. 
I would suggest discussing the individual elements where I noted this in the review comments. I am looking forward to your options about the suggestions. 🚀

Sources/ApodiniDelta/WebServiceStructure/Service.swift
lukaskollmer: (disclaimer: i only skimmed the code of this PR, so I could be missing something here)

---

@PSchmiedmayer 

I did look into making Apodini's "core" data structures `Codable` as part of my code, it's not really possible.
Some thoughts on why (this is essentially a summary of my reasoning as to why I went with the implementation approach I ended up using)

- Making `Endpoint` conform to `Codable`: this is effectively impossible, because a) it is a generic type, and b) most of its properties cannot be encoded.
- Making the `WebServiceModel` conform to `Codable`:
  - The type itself is not generic, but it is a reference type (which in itself does make not conforming to `Codable` impossible, but greatly complicates things, especially if there's recursion involved, which I don't think is the case here).
  - But: the individual `EndpointTreeNode`s in the web service model store the `Endpoint`s (in the form of `_AnyEndpoints`). Now, the `_AnyEndpoint` protocol is not somehow generic, but it does define several non-encodable properties (response transformers, guards, etc. all of these are function types and therefore cannot be encoded).
- I don't know much about the context in which @eldcn's code would be running (is it running directly inside Apodini (i.e. the same process & address space), or is it running in a separate process?)
  - If it is running in the same process, tbh I don't see why anything would need to be `Codable` in the first place.
  - If it is running in a different process, and you need the `Codable` conformance in order to "access" the web service's model from a process different to the one in which Apodini is running, what I ended up doing was that I defined [a couple of `Codable` structs](https://github.com/Apodini/Apodini/blob/0f9d7cb794042c8460e88f9e225b3ee0a4a0bd2f/Sources/ApodiniDeployBuildSupport/WebServiceStructure.swift) for collecting the subset of information from the web service model I was interested in: the set of endpoints (for which I cared about each `Endpoint`'s static type, handler id, and deployment options) of the web service as well as deployment options.
- With regards to the "dedicated exporter" point (this is still assuming that @eldcn's code is running out-of-process), I think it would be a really interesting idea to have one exporter which doesn't export any client-accessible endpoints or services or anything, but instead is used by Apodini-extending components (such as eg ApodiniDeploy) to get a "copy" of the structure of the web service. This would also make it a lot easier to define such components. What I'm envisioning here is some kind of "I have the file path and SPM target name of some Apodini web service, and I want to get a data structure containing as much information as possible about the web service as possible" API, kinda like what the [`DeploymentProvider` protocol currently implements](https://github.com/Apodini/Apodini/blob/0f9d7cb794042c8460e88f9e225b3ee0a4a0bd2f/Sources/ApodiniDeployBuildSupport/DeploymentProvider.swift#L94)


@PSchmiedmayer also regarding the number of exporters: i'd guess that in most cases, if you're using the REST exporter, chances are you also want the OpenAPI interface exporter (so there's an implication `REST → OpenAPI`), but I can't really see why someone would want to use the OpenAPI exporter w/out also using the REST exporter (so we also have `OpenAPI !→ REST`.
What I'm getting at here is that I think, if reducing the number of interface exporters is the goal, we could merge the REST and OpenAPI interface exporters into one (eg the existing REST IE), and add an `includeOpenAPIDef` option on that.

Sources/ApodiniUtils/Shell.swift
lukaskollmer: @PSchmiedmayer My child process API (see [Task.swift](https://github.com/Apodini/Apodini/blob/0f9d7cb794042c8460e88f9e225b3ee0a4a0bd2f/Sources/ApodiniUtils/Task.swift)) is already in the ApodiniUtils target.

It conceptually works a bit different from the `ShellCommand` API implemented here (this one seems to assume that the command is run in zsh, whereas mine allows launching any executable as a child command), but I think it should be possible to merge the two. Especially since, at the end of the day, both are using the same underlying system API (NSTask)

Sources/ApodiniDelta/Change/ChangeType.swift
eldcn: It only serves as a helper type to distinguish different changes currently. I will need it in the future to filter certain change types, and since some Change objects are generics, having this enum avoids using a type erasure.

Sources/ApodiniDelta/Compare/ChangeContextNode.swift
eldcn: Yep, compiler complaining that's why I had to add an argument label there

Sources/ApodiniDelta/Exports.swift
eldcn: `ApodiniTypeReflection` does not offer much actually, besides of `ReflectionInfo` struct and some `handle` methods, but I updated and imported those only in the files that use that logic

Sources/ApodiniDelta/Schema/PrimitiveType.swift
eldcn: I already moved the `Primitive` protocol to the scalar type logic. This enum is again a convenience helper type that makes sense to stay in `ApodiniDelta`

Sources/ApodiniDelta/Schema/Schema.swift
eldcn: So, the logic of registering a change in the current context node, requires that all comparable properties of an object are of different types, since I am not using some additional context keys there, but use the object identifier of the type directly. If an object would have two `String` properties, when registering the changes of the properties of that object, the change would be overwritten. By using the primitive properties with the wrapper e.g. `class SchemaName: PrimitiveValueWrapper<String> {}` guarantees unique types also for a primitive properties

Sources/ApodiniDelta/Schema/Schema.swift
eldcn: Good point, I now use `String(reflecting:)` api which also contains the module name, e.g. `Apodini.WebServiceModel`, and use that as the identifier of the schema. The problem in that case would be if the type would move in a new module. However, I plan to introduce an additional logic to handle additions and removals in general in another step, for instance comparing other properties even though the identifier did not match

Sources/ApodiniDelta/Schema/Schema.swift
eldcn: I do not think that it would make sense currently to move it there. There is some redundancy happening there currently, between my `Schema`, `ProtobufferMessage` and `JSONSchema` of `OpenAPIKit` used in `OpenApiIE`. All these types rely on `ReflectionInfo` and extract the infos used in their exporter. That's what I am also doing. But for sure, I will discuss this issue with Max

Sources/ApodiniDelta/Schema/SchemaReference.swift
eldcn: Good point, this will fall away once I have a centralised error handling logic. Currently instead of throwing I return `nil` and `empty` is a fallback if `ReflectionInfo` fails to obtain type info

Sources/ApodiniDelta/Shared/Primitive.swift
eldcn: 👍

Sources/ApodiniDelta/WebServiceStructure/Service.swift
eldcn: We will have to discuss the options to make the core data structures `Codable`, but as Lukas also said, it is impossible currently. @lukaskollmer I need to persist all interface related infos of the endpoints and compare different versions, thats why I need `Codable` and the core context runs in the same process that Apodini runs

PR #267, opened 2021-04-08
Binding Improvements

PSchmiedmayer:
# *Binding Improvements*

In the moment the usage of the `projectedValue` of a `PathParameter` and a `Parameter` is inconsistent. The `projectedValue` of a `PathParameter` returns an `Parameter` that can be passed as a `PathComponent` to a `Group`.

## :bulb: Proposed solution
This PR simplifies this setup by always emitting a `Binding` for the projected value for `PathParameter`s and `Parameter`s.
This enables code like the following setup while only relying on the projected value.
```swift
struct TransactionComponent: Component {
    @PathParameter var transactionId: Transaction.ID
    
    
    var content: some Component {
        Group("transactions") {
            GetAllTransactions()
            Group($transactionId) {
                GetTransaction(transactionId: $transactionId)
            }
        }
    }
}

struct GetTransaction: Handler {
    @Binding var transactionId: Transaction.ID
    
    
    func handle() throws -> Transaction {
        // ...
    }
}
```

This enables a consistent usage of the projected value. Using this enables you to write the projected value for the following two cases. This can make it a bit easier to reason about for newcomers who already use Swift and SwiftUI:
```swift
struct MyComponent: Component {
    @PathParameter var id: Int


    var content: some Component {
         Group($id) {
             MySecondComponent(id: $id)
         }
    }
}


struct MySecondComponent: Component {
    @Binding var id: Int


    var content: some Component {
         Group($id) {
             MyHandler(id: $id)
         }
    }
}

struct MyHandler: Handler {
    @Binding var id: Int


    func handle() -> Empty { /*...*/ }
}
```

### Implications
This as source breaking implications if the developer has relied on the projected value of a `@PathParameter` to be a `@Parameter`. Web services need to migrate to the usage of `Binding`s instead.

### Related PRs
#247 

### Testing
Adapted the tests to reflect the new usage of the projected value for `PathParameter`s and `Parameter`s

### DIFF
--- a/Sources/Apodini/Properties/Binding.swift
+++ b/Sources/Apodini/Properties/Binding.swift
@@ -45,6 +45,10 @@ public struct Binding<Value>: DynamicProperty, PotentiallyParameterIdentifyingBi
             return retriever(store)
         }
     }
+    
+    public var projectedValue: Self {
+        self
+    }
 }
 
 
--- a/Sources/Apodini/Properties/Binding.swift
+++ b/Sources/Apodini/Properties/Binding.swift
@@ -32,12 +32,11 @@ private enum Retrieval<Value> {
 /// or an `@Environment`. The latter `Binding`s for the latter two are contained in their
 /// `projectedValue`s.
 @propertyWrapper
-public struct Binding<Value: Codable>: DynamicProperty, PotentiallyParameterIdentifyingBinding {
+public struct Binding<Value>: DynamicProperty, PotentiallyParameterIdentifyingBinding {
     private let store: Properties
     private var retrieval: Retrieval<Value>
     let parameterId: UUID?
-    let parameterOptions: PropertyOptionSet<ParameterOptionNameSpace>
-
+    
     
     public var wrappedValue: Value {
         switch self.retrieval {
@@ -51,18 +50,19 @@ public struct Binding<Value: Codable>: DynamicProperty, PotentiallyParameterIden
     public var projectedValue: Self {
         self
     }
-    
-    
-    func parameterOption<Option>(for key: Parameter<Value>.OptionKey<Option>) -> Option? {
-        parameterOptions.option(for: key)
-    }
 }
+
 
 // MARK: PathComponent
 
-extension Binding: _PathComponent {
+extension Binding: PathComponent & _PathComponent where Value: Codable {
     func append<Parser: PathComponentParser>(to parser: inout Parser) {
-        parser.visit(self)
+        guard let parameter = store.wrappedValue["parameter"] as? Parameter<Value> else {
+            assertionFailure("Only bindings created from a `Parameter` or `PathParameter` can be used as a path component")
+            return
+        }
+        
+        parser.visit(parameter)
     }
 }
 
@@ -74,7 +74,6 @@ extension Binding {
         store = Properties()
         retrieval = .constant(constant)
         parameterId = nil
-        parameterOptions = PropertyOptionSet()
     }
     
     /// Create a `Binding` that always returns the given `value`.
@@ -82,6 +81,27 @@ extension Binding {
         Binding(constant: value)
     }
 }
+
+
+// MARK: Environment
+
+extension Binding {
+    private init<K: EnvironmentAccessible>(environment: Environment<K, Value>) {
+        store = Properties(wrappedValue: ["environment": environment])
+        retrieval = .storage { store in
+            guard let parameter = store.wrappedValue["environment"] as? Environment<K, Value> else {
+                fatalError("Could not find Environment object in store. The internal logic of Binding is broken!")
+            }
+            return parameter.wrappedValue
+        }
+        parameterId = nil
+    }
+
+    internal static func environment<K: EnvironmentAccessible>(_ environment: Environment<K, Value>) -> Binding<Value> {
+        Binding(environment: environment)
+    }
+}
+
 
 // MARK: Parameter
 
@@ -97,13 +117,13 @@ extension Binding where Value: Codable {
             return parameter.wrappedValue
         }
         parameterId = parameter.id
-        parameterOptions = parameter.options
     }
     
     internal static func parameter(_ parameter: Parameter<Value>) -> Binding<Value> {
         Binding(parameter: parameter)
     }
 }
+
 
 // MARK: Optional Wrapping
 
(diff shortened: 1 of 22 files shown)


Sources/Apodini/Properties/Binding.swift
```diff
@@ -45,6 +47,23 @@ public struct Binding<Value>: DynamicProperty, PotentiallyParameterIdentifyingBi
             return retriever(store)
         }
     }
+    
+    public var projectedValue: Self {
+        self
```
theMomax, line 51: Why do you want to expose the `Binding` itself via the `projectedValue`? We can already use the underscore `_someBinding` for quick access...why do we need `$someBinding` on top of that? This only results in a breaking change later on if we wanted to use the `projectedValue` for something else.

Sources/Apodini/Properties/Binding.swift
```diff
@@ -32,10 +32,12 @@ private enum Retrieval<Value> {
 /// or an `@Environment`. The latter `Binding`s for the latter two are contained in their
 /// `projectedValue`s.
 @propertyWrapper
-public struct Binding<Value>: DynamicProperty, PotentiallyParameterIdentifyingBinding {
+public struct Binding<Value: Codable>: DynamicProperty, PotentiallyParameterIdentifyingBinding {
     private let store: Properties
     private var retrieval: Retrieval<Value>
-    fileprivate let parameterId: UUID?
+    let parameterId: UUID?
+    let parameterOptions: PropertyOptionSet<ParameterOptionNameSpace>
```
theMomax: I dislike how those properties are stored redundantly. There should be a single source of truth. Please modify the `Retrieval` enum to differentiate between `Environment` and `Parameter`, so you can obtain the `Parameter` and its required properties from there.

Sources/Apodini/Properties/Binding.swift
```diff
@@ -32,10 +32,12 @@ private enum Retrieval<Value> {
 /// or an `@Environment`. The latter `Binding`s for the latter two are contained in their
 /// `projectedValue`s.
 @propertyWrapper
-public struct Binding<Value>: DynamicProperty, PotentiallyParameterIdentifyingBinding {
+public struct Binding<Value: Codable>: DynamicProperty, PotentiallyParameterIdentifyingBinding {
```
theMomax: Why constrain the `Value` to `Codable`? If we should decide to use `Codable` for collecting `Property`s on `Handler`s, the only option to have a non-`Codable` "regular" property (for parametrization or whatever) on a `Handler` would be through a `Binding`. Thus, I would prefer not to constrain `Binding` here as I think it isn't really necessary.

theMomax (commented):
> In addition this PR removes the ability to get a binding from an @Environment. @Environments should be injected directly in the Handler.

Could you elaborate on why you want to remove this feature? I think allowing to exchange `@Environment` and `@Parameter` on `Handler`s makes a lot of sense. I see that an `@Environment` could be variable, which doesn't work particularly well when used on `Group`. However, I consider `Handler`-reusability more important than `Component`-reusability.

I also made some notes below. As you can tell I'm not really happy with where this seems to be going. But maybe that's just because the PR isn't ready, yet. I just stumbled across it and had some spare time, so I gave it a look.

Sources/Apodini/Properties/Binding.swift
PSchmiedmayer: The goal of the projected value was to provide the same "just add a $ and you have a binding" (https://developer.apple.com/documentation/swiftui/binding/projectedvalue) approach that SwiftUI is approaching for `@State` and `@Binding`. I was preparing a session to show Apodini to the iPraktikum session tomorrow and noted this divergence. At the moment I don't see any other usage for the projected value.
Using this enables you to write the projected value for the following two cases. I am convinced this makes it a bit easier to reason about for newcomers:
```swift
struct MyComponent: Component {
    @PathParameter var id: Int


    var content: some Component {
         Group($id) {
             MySecondComponent(id: $id)
         }
    }
}


struct MySecondComponent: Component {
    @Binding var id: Int


    var content: some Component {
         Group($id) {
             MyHandler(id: $id)
         }
    }
}

struct MyHandler: Handler {
    @Binding var id: Int


    func handle() -> Empty { /*...*/ }
}
```

Sources/Apodini/Properties/Binding.swift
PSchmiedmayer: This is a good point, I can just use the `Parameter` stored in the store. I will change the implementation to reflect the suggestion

Sources/Apodini/Properties/Binding.swift
PSchmiedmayer: I changed this as the `PathComponentParser`'s `mutating func visit<T>(_ parameter: Parameter<T>)` requires that the type conforms to Codable. I will check how this could be circumvented so we don't need the constraint to `Codable` here.

PSchmiedmayer: I have addressed the feedback by making the conformance to `PathComponent` and `_PathCompoent` if value conforms to `Codable`:
```swift
extension Binding: PathComponent & _PathComponent where Value: Codable {
    func append<Parser: PathComponentParser>(to parser: inout Parser) {
        guard let parameter = store.wrappedValue["parameter"] as? Parameter<Value> else {
            assertionFailure("Only bindings created from a `Parameter` or `PathParameter` can be used as a path component")
            return 
        }
        
        parser.visit(parameter)
    }
}
```

PSchmiedmayer:
@theMomax I have updated the PR to reflect your feedback. The diff reduced by a lot and is mostly centered on the fact that the `@PathParameter`'s `projectedValue` is a `Binding`. The ability to get a `Binding` from an `Environment` is still there. 

I will merge the PR when all tests pass as I will introduce Apodini to the iPraktikum developers tomorrow morning and would like to present the usage of the `@Bindings` there.

theMomax:
@PSchmiedmayer thanks for incorporating my suggestions. I just noticed this PR didn't update the readme, so it's outdated now. It is just a tiny inconsistency, but it could confuse some of the iPraktikum participants. #268 contains a fix.

PSchmiedmayer:
@theMomax Thanks a lot for the quick fix of the README, I have approved and merged the PR 🚀

PR #270, opened 2021-04-12
Add a first draft of the encoder based type descriptor

eldcn:


Sources/ApodiniTypeDescriptor/InstanceCreator.swift
```diff
@@ -0,0 +1,122 @@
...
+        }
+        return instance
+    }
+}
+
+extension InstanceCreator {
+    static func dictionaryInstance(key: PrimitiveType, valueType: Any.Type) throws -> Any {
+        let valueInstance = try InstanceCreator(type: valueType).instance
+        switch key {
+        case .int: return [Int(): valueInstance]
```
theMomax, line 105: Just testing for primitive types here seems really restraining. I haven't tested it, but wouldn't it be possible to do something like this?

```swift
static func dictionaryInstance(keyType: Any.Type, valueType: Any.Type) throws -> Any {
    let valueInstance = try InstanceCreator(type: valueType).instance
    let keyInstanceCandidate = try InstanceCreator(type: keyType).instance
    guard let keyInstance = keyInstanceCandidate as? Hashable else {
        throw InstanceCreatorError.nonSupportedDictionaryKey(keyType)
    }
    return [keyInstance: valueInstance]
}
```

theMomax, line 105: Just testing for primitive types here seems really restraining. I haven't tested it, but wouldn't it be possible to do something like this?

```swift
static func dictionaryInstance(keyType: Any.Type, valueType: Any.Type) throws -> Any {
    let valueInstance = try InstanceCreator(type: valueType).instance
    let keyInstanceCandidate = try InstanceCreator(type: keyType).instance
    guard let keyInstance = keyInstanceCandidate as? Hashable else {
        throw InstanceCreatorError.nonSupportedDictionaryKey(keyType)
    }
    return [keyInstance: valueInstance]
}
```

Tests/ApodiniTests/ApodiniTypeDescriptorTests/TypeDescriptorTests.swift
```diff
@@ -0,0 +1,78 @@
...
+        let users: [String:User]
+        let grades: [Float]
+        let name: String?
+        let surname: String
+    }
+    
+    func testStudentTypeDescriptor() throws {
+        let descriptor = try typeDescriptor(Student.self)
+        
+        XCTAssertEqual(descriptor.debugDescription, expectedResult().debugDescription)
```
theMomax, line 40: Looks about right. I would suggest to add a test with a custom-conformance to `Codable`, though.

theMomax (commented):
I think the general direction is right. Good solution with the `InstanceCreator`. I think there is no _perfect_ solution for this problem, but this approach should catch all regular cases. I see a problem with custom conformances to `Encodable`/`Decodable` that skip properties conditionally (dependent on another property). Consider the following:

```swift
struct A {
    let child_present: Bool
    let child: B?

    func encode(to encoder: Encoder) throws {
        // ... encode child_present .... 
        if child_present {
             // ... encode child ...
        }
    }

    init(from decoder: Decoder) throws {
        // ...
    }
}
```
Your code would initialize `A(child_present: false, child: B())`, which encodes to
```json
{
    "child_present": false
}
```

But as I said, I don't think there is a way to catch such scenarios, while using `Codable` or `Runtime`. Maybe one could define custom protocols that refine `Codable` by providing a new way to conform to `Codable` in a  in a **declarative** way. That is totally out of scope, though.


On an other note, I would try to trim the implementation of the actual "Coder` down. Maybe we can cut some corners there. Also, don't forget to credit the creators of the JSONEncoder as required by their license.

Sources/ApodiniTypeDescriptor/InstanceCreator.swift
eldcn: I restricted to those primitive types, since the current logic in the project only accepts those types for dictionary keys. We could extend it to accept hashables using AssociatedTypeRequirementsVisitor since Hashable has Self from Equatable protocol.

theMomax: Ah, right I forgot about Equatable...but I‘d still go for it even if it requires AssociatedTypeRequirementsVisitor...but for a prototype it is fine as it is right now, of course.

PR #276, opened 2021-05-09
Blackboard-Pattern for cleaner Semantic Model

theMomax:
# Blackboard-Pattern for cleaner Semantic Model

The `SemanticModelBuilder` invokes three other builders (TypeIndex, WebServiceModel and Relationship) as well as some other methods to fill the `Endpoint` struct (as well as the WebServiceModel) with content. This content is an arbitrary and mostly unstructured set of information that somehow proved to be necessary to build one or the other `Exporter`. However, not all `Exporter`s need all of the provided information.

## :bulb: Proposed solution
`Endpoint` and `WebServiceModel` now provide access to a `Blackboard`.

The blackboard-pattern aggregates some form of knowledge on a blackboard using knowledge-sources. The aggregation is controlled by a controller. In my implementation, the knowledge-items are **instances** of `KnowledgeSource`s. The static type defines how the instance can be created from other knowledge that is available on the `Blackboard`. 

The blackboard-pattern is implemented is not an exact replica of the [standard pattern](https://en.wikipedia.org/wiki/Blackboard_(design_pattern). There is no defined controller, instead some initial basic knowledge is placed on the `Blackboard`s by the `SemanticModelBuilder`. Afterwards, all other `KnowledgeSource`s are initialized lazily based on the `Exporter`s' demand.

A `KnowledgeSource` can furthermore specify if it is to be initialized on a local (i.e. `Endpoint`-scoped) or global (i.e. `Application`-scoped) `Blackboard`. `LocalBlackboard` transparently also provides access to the `GlobalBlackboard`, whereas `GlobalBlackboard` provides a basic `KnowledgeSource`, which contains a list of all `LocalBlackboard`s present on the `Application`. 

### Problem that is solved
The Semantic Model becomes more extensible and its components become more reusable. Furthermore, for lightweight `Exporter`-configurations, startup-time should improve.

### Implications
A lot of files had to be modified to access information through a `Blackboard` instead of a property on `Endpoint`. `WebServiceModel` is now deprecated. New code should build on top of `WebServiceComponent`. 

I added a SwiftLint config for `Tests/` which allows force unwrapping and force casts as I don't see why it shouldn't be allowed there.

I didn't manage to fully re-build `TypeIndexBuilder` and `RelationshipBuilder` on top of the Blackboard-Pattern yet. I tried to refactor them a couple of times, but that turned out to be very difficult. Re-implementing from scratch with re-use of some logical (not structural) components seems easier (it worked well for `WebServiceComponent`). I also think the PR is big enough already without those additional changes. This also means, that `WebServiceModel`, `TypeIndexBuilder` and `RelationshipBuilder` are still present and active on the `SemanticModelBuilder`. When the Blackboard-based replacements are there, they can be removed. Knowledge-items are referenced using the `KnowledgeSource`'s type. 

### Related PRs
None

### Testing
The pattern's core is tested by previous tests. I created a new test for `WebServiceComponent`.

### Reviewer Nudging
Take a look at the `Blackboard` and `KnowledgeSource` protocols first. Understand, that each `Blackboard` comes with a defined set of `KnowledgeSource`s already initialized. Those _special_ `KnowledgeSource`s can be found at the bottom of `KnowledgeSource.swift`. All `KnowledgeSource`s are derived from those basics. Next, get a closer look at the implementations of `Blackboard`, how they play together and how they are used in the `SemanticModelBuilder`. Finally, take a look at how `WebServiceComponent`/`WebServiceRoot` are implemented and how they play together.


Sources/Apodini/Semantic Model Builder/PathModelBuilder.swift
```diff
@@ -62,6 +65,36 @@ struct PathModelBuilder: PathComponentParser {
...
+    
+    mutating func visit<Type: Codable>(_ parameter: Parameter<Type>) {
+        let identifyingType = parameter.option(for: PropertyOptionKey.identifying)
+
+        let pathParameter: AnyEndpointPathParameter
+        if let optionalParameter = parameter as? EncodeOptionalPathParameter {
+            pathParameter = optionalParameter.createPathParameterWithWrappedType(id: parameter.id, identifyingType: identifyingType)
+        } else {
+            pathParameter = EndpointPathParameter<Type>(id: parameter.id, identifyingType: identifyingType)
+        }
```
bauer-andreas: Is there a way to not 1:1 duplicate the code of `PathModelBuilder.visit(_ parameter: Parameter<Type>)`?

Sources/Apodini/Semantic Model Builder/Model/Endpoint.swift
```diff
@@ -43,7 +20,7 @@ public protocol AnyEndpoint: CustomStringConvertible {
 
     /// Creates a set of `RelationshipDestination` which ensures that relationship names
     /// are unique (for all collected destination for a given `Operation`)
-    /// - Parameter operation :The `Operation` of the Relationship destination to create a unique set for.
+    /// - Parameter operation:The `Operation` of the Relationship destination to create a unique set for.
```
bauer-andreas:
```suggestion
    /// - Parameter operation: The `Operation` of the Relationship destination to create a unique set for.
```

lol

bauer-andreas (approved):
Looks good to me, great work 👍 
Only a minor thing from my side regarding some code duplication in the PathComponent builders.

eldcn:
@theMomax Great job 👍 I would have some minor suggestions:
-  `MockBlackboard`, better define it in `ApodiniTests` target, and add a `convenience` initializer to `WebServiceModel` in the target:
```swift
extension WebServiceModel {
    convenience init(mockBlackboard: Blackboard = MockBlackboard()) {
        self.init(mockBlackboard)
    }
}
```
- The only issue would be the `deprecated` `WebServiceModule`, can it be removed?
- It should not happen in this PR, but just to keep in mind to add a `.md` in Documentation folder with an explanation of the new architecture.
- Other than that, chapeau

eldcn (approved):

Sources/Apodini/Semantic Model Builder/PathModelBuilder.swift
theMomax: Good point. Fixed with 1e37a18585ae9cd101de65f02ffb0f8b2892b4ac

Sources/Apodini/Semantic Model Builder/Model/Endpoint.swift
theMomax: 1e37a18585ae9cd101de65f02ffb0f8b2892b4ac xD

theMomax:
@Supereg I addressed your remarks.

@eldcn Thanks for your feedback!

> The only issue would be the deprecated WebServiceModule, can it be removed?

Yes, `WebServiceModule` can be removed, I think. I actually meant to mark `WebServiceModel` as deprecated (as I wrote in the PR description), which cannot be removed yet, as `RelationshipBuilder`/`TypeIndexBuilder` depend on it.

> MockBlackboard, better define it in ApodiniTests target, and add a convenience initializer to WebServiceModel in the target

Good point. I agree that it is ill-placed right now. However, I don't think moving it to `ApodiniTests` works, as I also need it in some other test targets and I don't think we can use test-targets as dependencies, can we?

> It should not happen in this PR, but just to keep in mind to add a .md in Documentation folder with an explanation of the new architecture.

Yes, I should definitely do that👍🏻

bauer-andreas:
> I addressed your remarks.

Seen it 👍 

> However, I don't think moving it to ApodiniTests works, as I also need it in some other test targets and I don't think we can use test-targets as dependencies, can we

Would putting it into the `XCTApodini` target be a desirable solution?

theMomax:
> Would putting it into the `XCTApodini` target be a desirable solution?

@Supereg This was my first intuition too. What should we move there?

* `Tests/ApodiniTests/Helper` (I don't think it makes much sense to define `Models` in XCTApodini)
* `Tests/ApodiniTests/Helper/Mocks`
* `Tests/ApodiniTests/Helper/Mocks` + * `Tests/ApodiniTests/Helper/Components`
* Only `Tests/ApodiniTests/Helper/Mocks/MockBlackboard.swift`

eldcn:
> Would putting it into the `XCTApodini` target be a desirable solution?

@theMomax  @Supereg With ApodiniTests I meant `XCTApodini` actually 😅, which is used by almost all test targets, I think it should work

theMomax:
@eldcn Still leaves the question if we should move the whole `Mocks` folder...

theMomax:
@eldcn I decided to move the entire `Mocks` folder to `XCTApodini`: 59d44586bef9aa7165c65d5d8b0a2b33a93434a0

PR #278, opened 2021-05-14
assert unique handler identifiers

eldcn:
# *Asserting unique handler identifier*

In case that the handler identifier is specified via `.identified(by:)` modifier, current implementation does not check for uniqueness of the specified identifier.

## :bulb: Proposed solution
Insert a unique id check in `SemanticModelBuilder`

### Problem that is solved
Deployment Provider implementation and my work relies on unique identifiers

### DIFF
--- a/Sources/Apodini/Components/Handler/HandlerIdentifier.swift
+++ b/Sources/Apodini/Components/Handler/HandlerIdentifier.swift
@@ -36,7 +36,7 @@ private struct AllAnyHandlerIdentifiers: KnowledgeSource {
                 if ident1 == ident2 {
                     rawValue = ident1.rawValue
                 } else {
-                    fatalError("""
+                    preconditionFailure("""
                         Handler '\(blackboard[HandlerName.self].name)' has multiple explicitly specified identifiers ('\(ident1)' and '\(ident2)').
                         A handler may only have one explicitly specified identifier.
                         This is caused by using both the 'IdentifiableHandler.handlerId' property as well as the '.identified(by:)' modifier.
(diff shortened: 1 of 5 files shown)


theMomax (changes requested):
Thanks for fixing my closures👍

I think it would be better to implement this as a global KnowledgeSource. Many exporters don’t need AnyHandlerIdentifier and if we can avoid the opposite it’s best to keep semanticmodelbuilder clean. Refactor the AnyHandlerIdentifier KS to be initialized by a new private global KS that first collects all handler identifiers from the Blackboards KS, then makes the uniqueness check and then initializes the local AnyHandlerIdentifier KSs on each blackboard. Refer to the last section of the Semantic Model Documentation for more detail on the implementation pattern. You can also look at the code of webserviceroot and webservicecomponent which employ the same technique. This way all AnyEndpointIdentifiers are initialized and the check is applied when AnyHandlerIdentifier is accessed the first time on an endpoint.

Sources/Apodini/Components/Handler/HandlerIdentifier.swift
```diff
@@ -8,35 +8,66 @@
 
 import Foundation
 
+private struct AllAnyHandlerIdentifiersAnchor: TruthAnchor {}
 
-/// An `AnyHandlerIdentifier` object identifies a `Handler` regardless of its concrete type.
-open class AnyHandlerIdentifier: Codable, RawRepresentable, Hashable, Equatable, CustomStringConvertible, KnowledgeSource {
-    public let rawValue: String
+/// A `KnowledgeSource` that initializes and holds all `AnyHandlerIdenfier`, by also asserting the uniqueness of them.
+struct AllAnyHandlerIdentifiers: KnowledgeSource {
```
theMomax: Just a suggestion, but why didn’t you make it private? I don’t see a reals use case where you would want to access all identifiers (especially on a global (not exporter specific) scope.

theMomax (approved):
Thanks for implementig it via the blackboard pattern👍 Well done!

PR #280, opened 2021-05-17
Metadata DSL

bauer-andreas:
# Metadata DSL

The `Metadata DSL` idea was a concept to deal with incorporating the formulation WebService Requirements into the WebService definition itself. It should provide a unified way of defining Metadata on all individual WebService components.

## :bulb: Proposed solution
This PR adds the initial ground work for the `Metadata DSL`. Specifically it adds the DSL part of the `Metadata DSL`.
Storage is currently based on the `Context` and `ContextKey`s (thus currently not fully supporting what our vision is for specifying Metadata on `Content` types as well as not supporting scenarios where it is required that the Metadata definitions are Codable, like for DeploymentOptions).

Apodini provides very little built in Metadata. Currently this the `Description` Metadata and the `Empty` Metadata.
The preexisting `.description` modifier was migrated to be based on the `DescriptionMetadata`.
In the future it might make sense to move more of Apodinis HandlerModifier to be `MetadataDefinitions`.

Furthermore, this PR completely removes the Relationship DSL and integrates it directly into the Metadata DSL.

### Testing
As the Relationship DSL was integrated into the Metadata DSL system, we already have some preexisting test vectors for `ContentMetadataDefinition`s ensuring the Metadata DSL fulfills the expectations set in the past for the Relationship DSL.

Further, the PR adds extensive testing for all kinds of different Metadata types.

### Reviewer Nudging

The reviewer should probably start with making theirselves familiar with the fundamental building blocks and concepts of the Metadata DSL system.
It might make sense to go through the files `AnyMetadata`, `MetadataDefinition`, `MetadataGroup`, `MetadataNamespace` and `RestrictedMetadataGroup`, all providing extensive documentation and examples.
You may also look into the predefined Metadata in `DescriptionMetadata` and `EmptyMetadata` for examples of Metadata Definitions.
Additionally the Metadata Tests provided in the `ApodiniTests` target might be helpful to see how Metadata is used.


Sources/Apodini/Metadata/MetadataDefinition.swift
```diff
@@ -0,0 +1,118 @@
+//
+// Created by Andreas Bauer on 21.05.21.
+//
+
+/// A `MetadataDefinition` represents a specific type of Metadata which can be declared
+/// on appropriate locations, like in the Metadata Declaration blocks of a `Component`.
+/// To create a `MetadataDefinition` use on of the following protocols, depending on where
```
theMomax: *one

Sources/Apodini/Metadata/MetadataNamespace.swift
```diff
@@ -0,0 +1,208 @@
+//
+// Created by Andreas Bauer on 16.05.21.
+//
+
+/// The `HandlerMetadataNamespace` can be used to define a appropriate
```
theMomax: *an

Sources/Apodini/Metadata/MetadataNamespace.swift
```diff
@@ -0,0 +1,208 @@
...
+/// extension HandlerMetadataNamespace {
+///     public typealias Description = DescriptionHandlerMetadata
+/// }
+/// ```
+///
+/// Refer to `TypedHandlerMetadataNamespace` if you need access to the generic `Handler`
+/// Type where the Metadata is used on.
+public protocol HandlerMetadataNamespace {}
+
+/// The `ComponentOnlyMetadataNamespace` can be used to define a appropriate
```
theMomax: *an

Sources/Apodini/Metadata/MetadataNamespace.swift
```diff
@@ -0,0 +1,208 @@
...
+/// of `ComponentMetadataGroup`s.
+///
+/// This is only necessary if you additionally have equivalent or similar Component-Only and Handler and/or WebService Metadata
+/// which additionally share the same name in the `ComponentOnlyMetadataNamespace`, `HandlerMetadataNamespace` and
+/// `WebServiceMetadataNamespace` (Such example is the `EmptyComponentMetadata`).
+public typealias ComponentMetadataGroupNamespace = ComponentMetadataGroup
+
+
+/// As the `HandlerMetadataNamespace` is accessible from `HandlerMetadataGroup`s,
+/// there is no way to know the type of the `Handler` the Metadata will be used one
```
theMomax: *on

Sources/Apodini/Metadata/RestrictedMetadataGroup.swift
```diff
@@ -0,0 +1,128 @@
+//
+// Created by Andreas Bauer on 21.05.21.
+//
+
+/// A `RestrictedMetadataGroup` is a  `AnyMetadataGroup` which is restricted to only contain
+/// a specific Type of `AnyMetadata` (`RestrictedMetadataGroup`s support arbitrary nesting, meaning
+/// they always can contain themselves).
+///
+/// In order to support multiple `AnyMetadata` which is allowed inside a `RestrictedMetadataGroup`,
+/// use a class as the base Metadata type, creating subclasses for all allowed Metadata.
```
theMomax: Wouldn't that also be possible using structs/enums/classes conforming to a base-protocol?

Sources/Apodini/Metadata/Modifier/HandlerMetadataModifier.swift
```diff
@@ -0,0 +1,70 @@
...
+/// Modifiers to add arbitrary Metadata to a `Handler`.
+///
+/// Furthermore `HandlerMetadataModifier` serves as a build block to easily create a custom
+/// `HandlerModifier` for your `HandlerMetadataDefinition` without much overhead.
+/// In order to create a Modifier declare a `Handler` extension as usual, returning a
+/// `HandlerMetadataModifier` instantiated via `HandlerMetadataModifier.init(modifies:with:)`:
+/// ```swift
+/// extension Handler {
+///     public func myModifier(_ value: ExampleValue) -> HandlerMetadataModifier<Self> {
+///         HandlerMetadataModifier(modifies: self, with: ExampleHandlerMetadata(value))
```
theMomax, line 17: It would be cool to have a chaining syntax here. E.g.

```swift
MyHandler()
    .description("...")
    .responseTime(20)
```
instead of
```swift
MyHandler().metadata {
    Description("...")
    ResponseTime(20)
}
```

I guess that's personal preference, though...

Tests/ApodiniTests/MetadataTests/BuildinMetadata/DescriptionMetadataTests.swift
```diff
@@ -0,0 +1,134 @@
...
+        }
+    }
+
+    struct TestHandler: Handler {
+        func handle() -> TestContent {
+            TestContent()
+        }
+
+        var metadata: Metadata {
+            Description("The description inside the TestHandler")
```
theMomax, line 21: Is there a situation where one **must** declare their metadata like this directly on the handler? I think from a handler-reusability perspective the other approach, where you add the metadata to the handler-instance in the component-tree, is preferable. It still makes sense to support both, but if one **had** to declare certain aspects here, that would be a bummer.

Tests/ApodiniTests/MetadataTests/ComponenetMetadataTests.swift
```diff
@@ -0,0 +1,287 @@
...
+    var metadata: Metadata {
+        TestInt(0)
+
+        if state {
+            TestInt(1)
+        }
+
+        Empty()
+
+        Collect {
```
theMomax: I'm not sure if `Collect` is intuitive. I get that you don't want to use `Group` to avoid confusion with the `Component` `Group`. However, `Collect` as a verb seems kind of misplaced to me in a declarative context. I'd suggest `All`, `Many`, `Collection`, `Set` or even stick with `Group` if swift allows that.

theMomax (approved):
Amazing work! I really like the flexibility of the namespace-approach. I added some comments here and there, mostly just spelling.

Two general points:

Since the setup is so generic, I guess it takes some effort to create a personal metadata-subsystem such as the Relationship DSL. Maybe it would be good to add a guide for that to the Documentation folder. There is a lot of documentation in the code, but it is still something different than a start-to-finish guide.

This is not really a job for this PR, but I think we should start embracing a core<->foundation strategy which the former only containing core interfaces and processing logic, whereas the latter provides standard implementations such as the Documentation metadata-tag or the Text handler and much more.

Sources/Apodini/Metadata/RestrictedMetadataGroup.swift
bauer-andreas: To my testing, you can't pass a protocol as a generic type there.

Sources/Apodini/Metadata/Modifier/HandlerMetadataModifier.swift
bauer-andreas: Thats exactly what `HandlerMetadataModifier` and `ComponentMetadataModifier` are there for.
Though, you will always have to define the extension for Modifier yourself. `HandlerMetadataModifier` and `ComponentMetadataModifier` just help to reduce the work needed to bring Modifier support for your Metadata.

If you look at the `Description` metadata, you can observe how it creates the extension for the `.description` modifier and returns an instance of `ComponentMetadataModifier`.

Tests/ApodiniTests/MetadataTests/BuildinMetadata/DescriptionMetadataTests.swift
bauer-andreas: You are pretty flexible in that regard. You may have seen those `.metadata` modifiers (either via ResultBuilder or directly passing a single Metadata instance) which can be used to overwrite Metadata on Handlers.
Additionally as mentioned before, you might choose to add a regular HandlerModifier for your Metadata.

One drawback currently is, that the `HandlerMetadataNamespace` is not available inside `Component`s, thus you can't really pass Metadata which is only defined for Handlers via the `.metadata` Modifier directly. However, as a workaround you can define a custom `HandlerMetadataGroup` and pass that via the `.metadata` Modifier.

Not sure if we should add the `HandlerMetadataNamespace` to the `Component` protocol as well. I think this could introduce some complications in regards to type checking and Metadata with similar names.

Tests/ApodiniTests/MetadataTests/ComponenetMetadataTests.swift
bauer-andreas: Yeah, that's one of the things I still wanted to discuss in the Meeting tomorrow.
The name was really just created because `Group` was already taken.

Sources/Apodini/Metadata/MetadataNamespace.swift
```diff
@@ -0,0 +1,208 @@
...
+/// extension WebServiceMetadataNamespace {
+///     public typealias Description = DescriptionWebServiceMetadata
+/// }
+/// ```
+///
+/// Refer to `TypedWebServiceMetadataNamespace` if you need access to the generic `WebService`
+/// Type where the Metadata is used on.
+public protocol WebServiceMetadataNamespace {}
+
+/// The `ComponentMetadataNamespace` can be used to define a appropriate
```
eldcn: Cmd+F `a approapriate` -> replace `an approapriate` 😅

Sources/Apodini/Components/Component/Component.swift
```diff
@@ -33,10 +45,31 @@ extension Component {
...
         if let visitable = self as? SyntaxTreeVisitable {
+            // This cases covers any Components conforming to SyntaxTreeVisitable.
+            // Most commonly this are Modifiers, but also Components like `Group`
+            // or special purpose Components like `TupleComponent`
+
+            // As stated above, this might be a Modifier and the Metadata of Modifiers can't be accessed.
+            // So we only start parsing the metadata if in fact we know that it isn't a Modifier.
+            if StandardModifierVisitor()(self) != true {
+                let metadata: Self.Metadata = self.metadata
+                (metadata as! AnyMetadata).accept(visitor)
```
eldcn, line 57: Why isn't the linter complaining here? I think you can directly go with `(metadata as? AnyMetadata)?.accept(visitor)` without the need of declaring a new variable on line 57

Sources/Apodini/Components/Component/Component.swift
```diff
@@ -33,10 +45,31 @@ extension Component {
...
         } else {
-            HandlerVisitorHelperImpl(visitor: visitor)(self)
+            let visited = HandlerVisitorHelperImpl(visitor: visitor)(self) != nil
+
+            if !visited {
+                // Covering components which are not Handlers and don't conform to `SyntaxTreeVisitable`.
+                // Such Components are typically constructed by users.
+                // Executed before we enter the content below.
+                let metadata: Self.Metadata = self.metadata
+                (metadata as! AnyMetadata).accept(visitor)
```
eldcn, line 68: Same here

eldcn (approved):
Great job 👍

Sources/Apodini/Components/Component/Component.swift
bauer-andreas: Though it might make sense to make it explicit why `metadata.accept(...)` does not work, because Swift seems to think, that `Self.Metadata` is not a `AnyComponentOnlyMetadata` (see line 20). Not doing the cast would always result in calling `.accept` on the default Empty Metadata defined in line 33.
But I'm also fine merging it into one line, regression testing is in place.

Though not sure if I want to use optional chaining, after all it shouldn't be ignored silently if this fails.

bauer-andreas: And regarding SwiftLint, has been really unreliable for me as well. Missed e.g. some spots where it should have complained with `missing_docs`. Not sure what's up with that.

Tests/ApodiniTests/MetadataTests/ComponenetMetadataTests.swift
bauer-andreas: For future reference, in the Meeting we decided to go with `Block`

Sources/Apodini/Components/Component/Component.swift
bauer-andreas: Resolved d59f0bc

Issue #228, opened 2021-02-07
[MAJOR FEATURE] Composition of Handlers


theMomax:
This issue was previously discussed as part of some files in the `Documentation` folder. Thus the documentation here mostly consists of links.

**Describe your use case**
The feature is important whenever a `ObservedObject`'s `ObservableObject` depends on another `Property` e.g. an `@Parameter`.

Currently the only option to do this is to create a `ObservableObject` for each `Handler`-instance and then pass it to the service-layer on the initial evaluation, which then has to store it using a `weak` reference:

```swift
struct SomeHandler: Handler {
    @ObservedObject var observed = SomeObservableObject()

    @State var initial: Bool = true

    @Parameter var id: UUID

    func handle() -> ... {
         defer { initial = false }
         if initial {
              SomeService.setObservable(observed, for: id)
         }
         ...
    }
}
```

A full implementation of this pattern can be seen [here in a sample project](https://github.com/theMomax/apodini-websocket-sample-project-automations/blob/baa601a2c12275f221fbf464c727c228b927916c/Sources/apodini-websocket-sample-project-automations/ChannelHandler.swift#L70).

Additionally, the logic of some `Hanlder`s can get quite verbose as many different cases have to be considered (e.g. [here](https://github.com/theMomax/apodini-websocket-sample-project-automations/blob/baa601a2c12275f221fbf464c727c228b927916c/Sources/apodini-websocket-sample-project-automations/ChannelHandler.swift#L36)). In those cases many `Properties` have to be used and it would be nice to be able to group them into smaller chunks that provide more structure to the logic.

**Describe the problem**
The problem is described [here](https://github.com/Apodini/Apodini/blob/develop/Documentation/Communicational%20Patterns/2.%20Tooling/2.1.%20Tooling.md#layers-of-logic).

**Propose a solution**
The solution is described [here](https://github.com/Apodini/Apodini/blob/develop/Documentation/Communicational%20Patterns/2.%20Tooling/2.1.%20Tooling.md#composition-of-handlers).

**Describe alternatives you've considered**
The layers-of-logic-problem is most critical for `ObservedObject` and could partially be solved by making the `ObservedObject` variable, so one could change the `ObservableObject` observed by the `Handler` based on service logic.

```swift
struct SomeHandler: Handler {
    @ObservedObject var observed: SomeObservableObject?

    @State var initial: Bool = true

    @Parameter var id: UUID

    func handle() -> ... {
         defer { initial = false }
         if initial {
[...]

PR #281, opened 2021-05-20
Generalised Handler Delegation

theMomax:
# Generalised Handler Delegation

Currently, there is no generic form of delegation in writing `Handler`s. There are three specialised options:

* `Guard` allows to perform an action before a `Handler` is prepared and executed and to abort execution.
* `ResponseTransformer` allows to modify a `Handler`'s output.
* `DynamicProperty` allows for bundling multiple `Property`s into a functional group.

## :bulb: Proposed solution
`Delegate<D>` is a lazy version of `DynamicProperty`. It works exactly the same way, except it allows to delay injection (and thus verification) of input into `@Parameter`s.

We can use it to implement `ResponseTransformer`s and `Guard`s. This could be a generic wrapper for authentication using an API Token:

```swift
struct Authenticator<H: Handler>: Handler {
    @Parameter var token: String

    @Throws(.forbidden, "Invalid Token") var badToken

    @Environment(\.tokenValidator) var tokenValidator
    
    let handler: Delegate<H>

    init(_ handler: H) {
        // We manually choose `Optionality.required` here. The default value would be `.optional`.
        // This doesn't affect the runtime behaviour, but it indicates to static exporters that the only
        // reason why this delegate might not be called, is that the delegating `Handler`'s `handle()` throws.
        // Thus, e.g. OpenAPI Exporter still marks all non-optional, no-default `@Parameter`s as required.
        // A delegating `Handler` which delegates to either of two `Delegate`s depending on some input, would
        // keep the default `.optional` and thus even non-optional, no-default `@Parameter`s would be
        // described as optional.
        self.handler = Delegate(handler, .required)
    }

    func handle() throws -> H.Response {
        guard tokenValidator.isValid(token) else {
             throw badToken
        }

        // `Delegate` can be called as a function - this injects
        // `@Parameter`s and returns the injected struct afterwards
        return try handler().handle()
    }
}
```

However, we can also just use a struct containing some `@Parameter`s that might be expensive to decode as a `Delegate`. `Delegate` doesn't require its content to be a `Handler`. We can also use two different `Delegate`s that take different input and delegate to either of them depending on some initial input.
 

### Problem that is solved
This implements the core concept presented under "alternatives considered" in #228.
Furthermore, this PR introduces an option for decoding input lazily.
This PR also provides the basis for a rewrite of `Guard` and `ResponseTransformer`, where they are just special forms of delegating Handlers. This should massively reduce code-duplication in some core components of Apodini.

### Implications
Input validation is now performed lazily. `ValidatedRequest` was changed to `ValidatingRequest`. To enable that, I also had to finally make the `inject` method of `RequestInjectable` throwing. I also refactored `Parameter` and `Environment` to use the `Activatable` pattern for convenience reasons. 

There is no impact on existing web service code. `Guard` and `ResponseTransformer` have remained unchanged. Refactoring those is to be done in a future PR.

This PR does not solve the _layers of logic_ existing between `@Parameter` and `@ObservedObject`. This is still to be done.

### Testing
I added new tests for checking the runtime-behaviour of `Delegate`. The `Optionality` is checked in an OpenAPIExporter-Test I adapted.

### Reviewer Nudging
The core changes are in `Validator.swift` (lazy validation) and `Traversable.swift` (delay parameter injection for `Delegate`).

Also, what do you think about `Delegate` behaving exactly the same as `DynamicProperty` for everything except for parameter injection? That also implies `ObservedObject`s on a `Delegate` could fire even before the `Delegate` was called. This would cause nothing more or less than the delegating `Handler` to be evaluated. This is no problem for the Apodini framework, but would the developer expect this to happen. Same applies to the time where the initial value of a `@State` would be initialized. We could change this behaviour so that preparation of all other `Properties` is also delayed to the point where the `Delegate` is called.


Sources/Apodini/Semantic Model Builder/Model/EndpointRelationship.swift
```diff
@@ -200,6 +200,7 @@ public struct RelationshipDestination: CustomStringConvertible, Hashable {
     /// Resolves any path parameters contained in the `destinationPath` if
     /// there is a resolver and a value for it.
     internal mutating func resolveParameters(context: ResolveContext) {
+        print(self.destinationPath)
```
bauer-andreas: I assume some leftover debug print?

Sources/Apodini/Semantic Model Builder/Traversable.swift
```diff
@@ -194,15 +189,15 @@ private func execute<Element, Target>(
 
         switch child {
         case let target as Target:
-            assert(((try? typeInfo(of: property.type).kind) ?? .none) == .struct, "\(Target.self) \(property.name) on element \(info.name) must be a struct")
+            assert(((try? typeInfo(of: property.type).kind) ?? .none) != .class, "\(Target.self) \(property.name) on element \(info.name) must be a struct")
```
bauer-andreas: Seems like the assert messages doesn't really reflect the assert condition anymore(?) [applies to all asserts in this file?].
Also, to my knowledge, asserts are not present in the final release build, are they(?). Might be sensible to switch to [precondition](https://developer.apple.com/documentation/swift/1540960-precondition)?

bauer-andreas (commented):
Couldn't quite finish my review in time today. But for my first impression this looks pretty solid. 

For further clarifitcations, the `Delegate` is really only to somewhat "group" and lazily decode Parameters right? Its not a mechanism to delegate handling logic (something I expected when I read the name "Delegate")?

Sources/Apodini/Semantic Model Builder/Model/EndpointRelationship.swift
theMomax: yes -> 7e0d4fad4bd71cb3aac747d6211991f331057f97

Sources/Apodini/Semantic Model Builder/Traversable.swift
theMomax: I forgot to update the message. Should be fixed now.

Yes, asserts are not executed in the release build and that is on purpose. The conditions contain extra `typeInfo` calls, which are quite expensive. Thus I thought it would not be worth it. I mean if someone really manages to introduce a class type Handler to a release build without ever starting the application in dev mode, it is their fault I guess🤷‍♂️

theMomax:
I updated the PR: Lifetime of `@State` and `@ObservedObject` now starts right before the first preparation of the delegate, not the delegating handler!

bauer-andreas (approved):
Looks good to me 👍 Seems like a good way to do lazy parameter decoding in way that it is reflected in the code.
Should we eventually move to writing a Apodini Wiki, it might be be great to add some notes here and there to highlight the existence and functionality of `Delegate`.

theMomax:
Yes, I will add docs with one of the follow-up PRs, once the feature is more mature.

PR #285, opened 2021-06-02
Negative Compile tests

bauer-andreas:
# Negative Compile tests

Apodini defines an internal domain-specific language to create web service. We have extensive testing to ensure that all DSL features do in fact work. However we currently don't have any way to test that a certain expression does not compile as it may be an illegal expression in our DSL (such restrictions are typically enforced via the Swift type system, allowing for compile time checking).

## :bulb: Proposed solution

This PR adds the possibility for those additional test vectors, enabling us to verify that certain expressions result in a compile time error.
The PR adds support for so called `NegativeCompileTests` targets.
In those test targets you can place code that deliberately does not compile (while annotating what lines do not compile and what compiler error is expected). The `NegativeCompileTestsRunner` (a newly created test target which uses the XCTest framework to bootstrap the Runner)
will then try to build those negative compile tests targets and check the build log against the expected compiler errors.

### Defining a new Negative Compiler Tests Target

To add a new Negative Compile Test target you have to do two things:

1. Define a new SPM `.testTarget` inside your `Package.swift` (while adding the special `Cases` folder the list of excluded source files)
2. Configure your Target inside `NegativeCompileTestsRunner/configurations.swift` so the Runner knows about it.

So, step 1. would look like the following:
```swift
.testTarget(
    name: "ApodiniNegativeCompileTests",
    exclude: ["Cases"]
),
```

Step 2.:
```swift
let configurations: TestRunnerConfiguration = [
    .target(name: "ApodiniNegativeCompileTests")
]
```

Lastly you want to create two files inside your test target: the `Cases` folder and an empty source file in the root of the target (preferably called `Empty.swift`; this is used to make the compiler stop complain about an empty target).

### Create Test Cases inside your Test Target

Once the target is set up, you can define your test cases inside the `Cases` folder, either by playing a source file directly into the `Cases` folder, or by creating subfolders.

You can then start writing source files which deliberately do not compile.
In order for the Runner to be able to validate that all of the expected compiler errors did in fact fire and that nothing more than the expected compiler errors occurred, you need to declare each an every expected compiler error.
You do this by adding special comments above the line which contains the error.

Adhere to the following format: `// error: <message>`.
An example compiler error declaration might look like the following:
```swift
var i = 0
// ...
// error: cannot find operator '++' in scope; did you mean '+= 1'?
i++
// ...
```

### Problem that is solved

This PR adds the possibilities for additional test vectors, enabling us to verify that certain expressions result in a compile time error.
Specifically the tests for the Metadata DSL enable us to ensure that expressions prohibited by the DSL are tested and possibilities for regressions are reduced.

--

### Related PRs
--

### Testing

This PR adds negative compile tests for the Metadata DSL, to ensure Metadata Definitions cannot be used
in Metadata Declaration Blocks for which they are not defined.

### Reviewer Nudging

The above describes how negative compile tests are created. Additionally, you may look at the example of the `ApodiniNegativeCompileTests` (A similar guide was included in the `Documentation` folder of the repository).
All the Runner logic is placed inside `NegativeTestRunner` which is initialized in the `XCTBootstrap.testRunner()` XCTestCase.


theMomax (approved):
Looks quite solid...pretty sure compiler-logs are not guaranteed to be stable, though. So we might adapt parsing from time to time...and also the error-messages can change with compiler-versions. Have you tested if something like the following would work?

```swift
#if swift(<=6.0)
// error: cannot find operator '++' in scope; did you mean '+= 1'?
#endif
i++
```

Otherwise we might have to implement our own parsing-logic, e.g.

```swift
// error [swift(<=6.0)]: <message>
i++
```
But I guess we can also take care of this once the problem actually comes up.

bauer-andreas:
> Looks quite solid...pretty sure compiler-logs are not guaranteed to be stable, though. So we might adapt parsing from time to time...and also the error-messages can change with compiler-versions.

Agree. This approach may need further maintenance in the future.
While there might be (somehow) the possibility to directly interface with the compiler, like Xcode does it, it currently isn't worth the time cost associated with it.

> Have you tested if something like the following would work?
> 
> ```swift
> #if swift(<=6.0)
> // error: cannot find operator '++' in scope; did you mean '+= 1'?
> #endif
> i++
> ```
> 
> Otherwise we might have to implement our own parsing-logic, e.g.
> 
> ```swift
> // error [swift(<=6.0)]: <message>
> i++
> ```
> 
> But I guess we can also take care of this once the problem actually comes up.

My approach currently just parses the file line for line, so a `#if swift(...)` directive would not work out of the box. Should we need something like that, I would propose using a custom syntax as you proposed in your second approach.
Not sure if we should realize such an approach already in this PR.

Tests/NegativeCompileTestsRunner/NegativeTestRunner.swift
```diff
@@ -0,0 +1,389 @@
...
+
+        let task = Task(
+            executableUrl: swiftBinary,
+            arguments: arguments.split(separator: " ").map { String($0) },
+            workingDirectory: workingDirectory,
+            captureOutput: true,
+            redirectStderrToStdout: true,
+            launchInCurrentProcessGroup: false,
+            environment: environment,
+            inheritsParentEnvironment: false // even though we pass false, our construction made above inherits environment from parent!
```
lukaskollmer: The reason for this is that the `environment` you're passing to the initialiser was initially copied from `ProcessInfo.processInfo.environment`.

The `Task` API, if `inheritsParentEnvironment` is false, will set the child's environment to what was passed to the initialiser (or set after initialising but before launching the process).

Or am I missing something here?

bauer-andreas: Thats exactly the intention here.
We want to inherit the environment form the parent process, but need to unset a certain environment variable in certain conditions, see here https://github.com/Apodini/Apodini/blob/96d12e977afd0f1105d135a10976ee2b8cff1734/Tests/NegativeCompileTestsRunner/NegativeTestRunner.swift#L320-L325 (this is to workaround an issue when running the TestRunner inside Xcode).

To me it seemed that with `inheritsParentEnvironment` the Task API only supports adding additional environment variables on top, but doesn't have an option to **unset** certain variables, or am I wrong?

lukaskollmer: The way it's implemented (or, was intended to be implemented) is that, based on the value of `inheritsParentEnvironment`, the child will get either only the environment passed to the initialiser (if it shouldn't inherit), or the current environment merged with what was passed to the task (if it should inherit).

https://github.com/Apodini/Apodini/blob/7b26ea0397b5e66eee2fc7cf035636a7e778e5fe/Sources/ApodiniUtils/Task.swift#L179-L186

It's correct that if `inheritsParentEnvironment` is true, you wouldn't be able to unset something from the parent (due to how the merging works, given that the `environment` dictionary's `Value` is `String` instead of `String?`.
But if `inheritsParentEnvironment` is false, it simply sets the task's environment dictionary as the child's environment, which (at least when I tested it locally when I wrote it) should completely ignore the parent's Process.environment.

Feel free to change this though (I think simply changing `Task.environment`'s type to `[String: String?]` and thus using nil values to make sure the for loop deletes existing values should suffice??)

If you want to see the actual environment being passed to the child you can have a look at `CApodiniUtils.c`, that's where the pseudo-child is replaced with the proper child (note that this only applies to Tasks where `launchInCurrentProcessGroup == true`)

https://github.com/Apodini/Apodini/blob/7b26ea0397b5e66eee2fc7cf035636a7e778e5fe/Sources/CApodiniUtils/CApodiniUtils.c#L45

lukaskollmer: Are you running this in Xcode or via the command line?
I think that might also make a difference since Xcode/lldb in some cases seem to interfere with the process structure

bauer-andreas: > The way it's implemented (or, was intended to be implemented) is that, based on the value of inheritsParentEnvironment, the child will get either only the environment passed to the initialiser (if it shouldn't inherit), or the current environment merged with what was passed to the task (if it should inherit).

Yes I noticed that. 

> It's correct that if inheritsParentEnvironment is true, you wouldn't be able to unset something from the parent (due to how the merging works, given that the environment dictionary's Value is String instead of String?.
But if inheritsParentEnvironment is false, it simply sets the task's environment dictionary as the child's environment, which (at least when I tested it locally when I wrote it) should completely ignore the parent's Process.environment.

> Feel free to change this though (I think simply changing Task.environment's type to [String: String?] and thus using nil values to make sure the for loop deletes existing values should suffice??)

You are right, changing that to a Dictionary with optional Strings could probably be used to encode "unset variable" using `nil`. Haven't really though about changing the `Task` internals. As this is probably a pretty unique use case and also just used for another workaround I'm not sure if we really need to adapt the `Task` API and instead leave it as is(?).

> Are you running this in Xcode or via the command line?

Well, yes 🙃 Both should work. In the CI (or if you wish to run it locally) it should be executable via command line. But tests are also often executed within Xcode (or any other IDE), therefore running inside Xcode is also supported.
Both work right now.

lukaskollmer: No, `Task` should absolutely be changed to include this functionality (after all, what's the point of having a child process API with support for setting the child's environment, if it doesn't let you set the child's environment?).

It is, however, still weird that this is happening, since you're setting `inheritsParentEnvironment` to false, meaning it never enters the first branch of the if stmt (the one w/ the loop) and instead just sets the environment to the dictionary passed in (which does not contain the key bc it was deleted by the caller)
Maybe swiftc modifies its own environment to re-set the key?

I can have a look at this if you want.
You could also try changing it to launch in the current process group, update the C file linked above to dump its envp, and run the test in the terminal. That'd give you the definitive environment passed to the child, since you'd be reading the child's envp in the child itself, as actually passed by Foundation's Process (i.e. after anyone could change it)

bauer-andreas: > It is, however, still weird that this is happening [...]

Not sure what you are talking about? Everything works as expected.

> No, Task should absolutely be changed to include this functionality (after all, what's the point of having a child process API with support for setting the child's environment, if it doesn't let you set the child's environment?).

I'll check if I can easily add this functionality.

lukaskollmer: The way I understood this the issue was that, despite passing in a custom environment with a specific key-value pair deleted, the child still ended up being launched with that key-value pair present?
At least that's how I interpreted this comment:
https://github.com/Apodini/Apodini/blob/96d12e977afd0f1105d135a10976ee2b8cff1734/Tests/NegativeCompileTestsRunner/NegativeTestRunner.swift#L335

bauer-andreas: Nope.
This is just a disclaimer/heads up to the reader who shouldn't be fooled by the `inheritsParentEnvironment: false`, as we **do** inherit from the parent environment, because we construct that ourselves above (we just do not really on the environment construction done by `Task`).
And the reason we construct the parent environment ourselves was because of the introduce workaround to delete a specific key form the environment in certain situations.

lukaskollmer: Ooh ok well in that case you probably can ignore everything I said above (excl the making it `[String: String?]` thing)

bauer-andreas: Solved via c9be844

PR #286, opened 2021-06-03
Generic Handler Delegation: Environment Injection + Variable ObservedObject + Guard/ResponseTransformer Replacement

theMomax:
# Generic Handler Delegation: Environment Injection + Variable ObservedObject + Guard/ResponseTransformer Replacement

Note: The description here is only a short overview. ~~I will create a Wiki pages to fully document the issue.~~ There are now wiki pages on [how `ObservedObject` works](https://github.com/Apodini/Apodini/wiki/Handling-Unsolicited-Events) and [Delegating Handlers](https://github.com/Apodini/Apodini/wiki/Delegating-Handlers).

For the basics of handler-delegation, please refer to #281. Handler-Composition and the layer-of-logic problem of `ObservedObject` was previously discussed in #228.

Another problem is that `Guard` and `ResponseTransformer` - while being very similar to `Handler` - currently require significant amounts of specialised code in the area of `Modifier`, `SemanticModelBuilder`, `Endpoint`, `EndpointInstance` and `RequestHandler`.

## :bulb: Proposed solution

1. `ObservedObject` allows for switching out the `ObservableObject`
2. `Delegate` allows for injecting values into the local environment and setting `ObservedObject` and `Binding` to a different value.
3. `DelegationModifier` allows for applying a delegating `Handler` to all endpoints of a `Component` or directly on a `Handler`.

### Problem that is solved
#228 is fixed with this PR. Furthermore, this PR re-builds `Guard` and `ResponseTransformer` as delegating `Handler`s. Thus, I was able to remove a lot of guard/transformer-specific logic.

This PR also modifies the `@ObservedObject` and `@Environment` properties and introduces `@EnvironmentObject`. They all have the same functionality in that they provide variable access to an element and observe it, in case it is an  `ObservableObject` (`@ObservedObject` only works with `ObservableObject`s, the others can contain any type). They differ in the way the element is obtained. `@ObservedObject` requires an initial value, `@Environment` identifies the element based on a keypath and `@EnvironmentObject` identifies the element solely by its type.

### Implications
While the public `Modifier` types implementing `Guard` and `ResponseTransformer` have changed, users should not notice anything. The `.guard()` and `.response()` functions should still work in any context they worked before.

Projects that use `@ObservedObject(\Some.keyPath) var x` are going to break. However, a simple regex-replace to `@Environment(\Some.keyPath) var x` should resolve all errors.

Projects that use `@Environment(\Some.keyPath) var x` , where the referenced object is an `ObservableObject` perceive a change in behaviour. Previously, `@Environment` would not subscribe to the `ObservableObject` and not cause evaluations of the `Handler`. Now, they do. `@Environment(\Some.keyPath, observed: false) var x` can be used to go back to the old functionality.

### Related PRs
#281 Implemented the basics of the `Delegate` type.

### Testing
I adapted tests for `ObservedObject` to use `Environment` where necessary and added a test-case where the `ObservableObject` is switched out and a `TriggerEvent` is cancelled.
I also created additional tests for `EnvironmentObject` for testing if it observes correctly and if it works with `Binding`.
Finally, I added tests that cover the `Delegate`'s injection/setter logic.

Existing `Guard` and `ResponseTransformer` tests cover the `DelegationModifier` part.

### Reviewer Nudging
Read the Wiki. Make sure you get the concept behind the newly introduced `TriggerEvent` and how it enables `ObservedObject` to be variable. Afterwards, check out `DelegationModifier` and `DelegatingHandlerInitializer` in particular. Refer to `ResponseTransformingHandler` and `ResponseTransformingHandlerInitializer` to see it in action.

### Help Wanted

In
```swift
public struct ResponseTransformingHandlerInitializer<T: ResponseTransformer>.instance<D>(for delegate: D) throws -> SomeHandler<Response> where D: Handler
```
I have to assert that `D.Response.Content == T.InputContent` so I can create a
```swift
internal struct ResponseTransformingHandler<D, T>: Handler where D: Handler, T: ResponseTransformer, D.Response.Content == T.InputContent
```

Can this be done more elegantly? My current implementation:

```swift
public struct ResponseTransformingHandlerInitializer<T: ResponseTransformer>: DelegatingHandlerInitializer {
    public typealias Response = Apodini.Response<T.Content>
    
    let transformer: T
    
    public func instance<D>(for delegate: D) throws -> SomeHandler<Response> where D: Handler {
        if let transformingHandler =
            (TransformerCandidate(
                transformer: transformer,
                delegate: delegate) as? Transformable)?() as? SomeHandler<Response> {
            return transformingHandler
        }
        
        fatalError("Cannot use response transformer \(transformer) with handler of type \(D.self) because content types do not match.")
    }
}

private struct TransformerCandidate<Transformer: ResponseTransformer, Delegate: Handler> {
    let transformer: Transformer
    let delegate: Delegate
}

extension TransformerCandidate: Transformable where Transformer.InputContent == Delegate.Response.Content {
    func callAsFunction() -> Any {
        SomeHandler<Response<Transformer.Content>>(ResponseTransformingHandler<Delegate, Transformer>(
                                                    transformed: Apodini.Delegate(delegate),
                                                    transformer: Apodini.Delegate(transformer)))
    }
}

private protocol Transformable {
    func callAsFunction() -> Any
}
```

## 🛠 Migration Guide

The environment based initializer for `ObservedObject` (e.g. `@ObservedObject(\Some.keyPath) var x`) was removed.
The same functionality can now be achieved using `@Environment(\Some.keyPath) var x`.

 `Environment` now by default observes its wrapped value if it is an `ObservableObject` just as `ObservedObject` did before. Previously, `Environment` would not subscribe to the `ObservableObject` and not cause evaluations of the `Handler`. Now, they do. The new initializer `@Environment(\Some.keyPath, observed: false) var x` can be used to go back to the old functionality.

Refer to the Wiki for a detailed guide on how the new system around handling unsolicited events works.


Sources/Apodini/Modifier/DelegationModifier.swift
```diff
@@ -0,0 +1,146 @@
...
+// MARK: DelegationModifier
+
+public struct DelegationModifier<C: Component, I: DelegatingHandlerInitializer>: Modifier {
+    public typealias ModifiedComponent = C
+    
+    public let component: C
+    private let initializer: I
+    private let prepend: Bool
+    
+    public var content: some Component { EmptyComponent() }
```
bauer-andreas: Is there a reason why this Modifier adds the content property?
Modifier sets the `Content` type alias to Never if I'm not wrong? 🤔

Sources/Apodini/Modifier/ResponseTransformer.swift
```diff
@@ -0,0 +1,88 @@
...
+}
+
+extension Handler {
+    /// A `response` modifier can be used to transform the output of a `Handler`'s response to a different type using a `ResponseTransformer`
+    /// - Parameter responseTransformer: The `ResponseTransformer` used to transform the response of a `Handler`
+    /// - Returns: The modified `Handler` with a new `ResponseTransformable` type
+    public func response<T: ResponseTransformer>(
+        _ responseTransformer: T
+    ) -> DelegationModifier<Self, ResponseTransformingHandlerInitializer<T>> where Self.Response.Content == T.InputContent {
+        self.delegated(by: ResponseTransformingHandlerInitializer(transformer: responseTransformer))
```
bauer-andreas, line 86: Regarding your type question. The problem really is that the input parameter of type `D` of the `instance(for:)` is completely decoupled from the initialization going on here.
I don't think you can "solve" this completely, only really option would be trying to make it "less" "ugly"?
Can you somehow pass the generic type `Self` over to the `ResponseTransformingHandlerInitializer` (with a `where` enforcing that output type equals inout type)? Then you could just check if the both generic types are still the same in `instance(for:)`, but not sure if this makes anything better 🤷

bauer-andreas (approved):
Looks good to me 👍 Great work 🚀 

As discussed, it would be great if you could add a tests case with a Handler and a Delegate to `HandlerMetadataTests.swift`, such that Metadata parsing is validated in such cases.

Sources/Apodini/Modifier/ResponseTransformer.swift
theMomax: Yes, exactly - it is just about about making it less ugly. But I don't thing what you suggest would lead to a solution. It isn't nice, but it works, so I'll leave it as is for now.

Sources/Apodini/Modifier/DelegationModifier.swift
theMomax: You're right, that doesn't make sense. It doesn't really have an effect either, though. I guess I just copied that part form `ResponseTransformer` or `Guard` stuff...removed it

theMomax:
@Supereg Thanks for the quick review and for the reminder regarding the metadata test! The order was indeed wrong🙃 Fixed with fb7c46c.

PR #288, opened 2021-06-04
Run Tests in release configuration

bauer-andreas:
# Run Tests in release configuration

As of now, our test targets all run in `debug` configuration, running without any compiler optimizations. This is of course wanted in development to easily debug those tests.
However, Apodini has some logic (mainly anything that is based on the AssociatedTypesRequirementsKit) which can break release builds if not used properly. As shown below there are also other use cases besides logic surrounding AssociatedTypesRequirementsKit.

## :bulb: Proposed solution

This PR proposes to execute our unit test with release configuration.

### Problem that is solved

We added another test vector, ensuring that everything what is tested in debug builds, works as expected in released builds as well.

### Implications

In order to run tests in release configuration, you have to pass the `-c release` argument to the `swift test` command.
Further, in order to be able to use `@testable` we have to pass the `-Xswiftc -enable-testing` compiler flag.
Lastly, we define the custom Active Compilation Condition `RELEASE_TESTING` using `-Xswiftc -DRELEASE_TESTING` in order to compile `#if` conditional compilation blocks.

#### Discussion: Active Compilation Condition: `DEBUG` vs custom `RELEASE_TESTING`

In order for `XCTApodini` to compile properly in release configuration, we need to wrap code which depends on `@testable` into `#if` conditional compilation blocks. Previously this was done using `#if DEBUG` statements, relying on the `DEBUG` condition which is automatically handled by spm in debug configuration.

This condition is not set in release configuration, therefore we have to specify it manually.

Now we have to the choice between manually specifying the existing `DEBUG` condition or creating a separate one (like `RELEASE_TESTING`) representing that exact state: testing with release configuration.

You can argue for either way. Using `DEBUG` would reduce the need to always write both: `#if DEBUG || RELEASE_TESTING`.
In some cases we have added further debug statements inside the `#if DEBUG` blocks, this would perfectly fit, to print those extended debug logging in release test builds as well.

However, one might also use `DEBUG` to add differing logic, meaning using a `#else` branch to do execute slightly different logic than in the `DEBUG` case. Those cases are still not executed when running tests in release configuration, therefore not setting `DEBUG` and creating a custom `RELEASE_TESTING` condition would make sense in those cases.

As we don't have a lot of those cases, I currently went with just setting `DEBUG` condition in release tests as well.

But I'm happy to hear about your opinion on this.

This PR already demonstrates why this additional test vector is helpful:
* The ApodiniDeployInterfaceExporter would crash in release builds, as it used an `AssociatedTypeRequirementsVisitor` visitor in its export method which wasn't setup properly (see https://github.com/Apodini/Apodini/pull/171#discussion_r563343324)
* When using `Binding` as PathParameters without encapsulating a `@Parameter` Apodini would just silently ignore them in release configuration only printing a proper error message in debug configuration.

### Related PRs
--

### Testing
This PR doesn't add any additional test cases

The Bamboo jobs currently do not execute tests in release configuration. Though I have verified locally that running them on macOS works as well. My plan would be to adjust those jobs after the PR has merged. If you can make branch dependent changes to the Bamboo Jobs I would also be happy to make those changes before merging.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -50,7 +50,7 @@ jobs:
       run: swift build -c release
     - name: Release Build & Test
       if: matrix.configuration == 'release_testing'
-      run: swift test -c release --enable-test-discovery -Xswiftc -enable-testing -Xswiftc -DDEBUG
+      run: swift test -c release --enable-test-discovery -Xswiftc -enable-testing -Xswiftc -DRELEASE_TESTING
     - name: Debug Build & Test
       if: matrix.configuration == 'debug'
       run: swift test -c debug --enable-test-discovery
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -45,13 +45,13 @@ jobs:
     - name: Check Swift version
       run: swift --version
 
-    - name: Build Release # Ensuring release build runs successfully without -enable-testing flag
+    - name: Release Build # Ensuring release build runs successfully without -enable-testing flag
       if: matrix.configuration == 'release'
       run: swift build --configuration ${{ matrix.configuration }}
-    - name: Build & Test
+    - name: Release Build & Test
       if: matrix.configuration == 'release'
       run: swift test --configuration ${{ matrix.configuration }} --enable-test-discovery -Xswiftc -enable-testing -Xswiftc -DDEBUG
-    - name: Build & Test
+    - name: Debug Build & Test
       if: matrix.configuration == 'debug'
       run: swift test --configuration ${{ matrix.configuration }} --enable-test-discovery
     - name: Run Deployment Provider Tests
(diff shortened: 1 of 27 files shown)


bauer-andreas:
@PSchmiedmayer fyi, once this PR is fully reviewed and discussed it may require your admin powers to merge. On the one hand CodeCov complains about diff coverage (due to some modifications in the DeploymentProvider), on the other hand, as I made some changes to the GitHub CI jobs, GitHub currently expects required checks which don't exist anymore.

Further, not sure when to best adjust Bamboo builds for macOS builds (see above).

theMomax (approved):
LGTM👍🏻

I don't really have a strong opinion on the DEBUG vs custom flag topic. Would adding a custom flag require any local IDE configuration changes for the usual workflow to work? If so, I'd prefer sticking with DEBUG.

PSchmiedmayer:
@Supereg Thanks for the PR, a nice addition!
@theMomax Thanks for the review 👋

Regarding the debug flag: I would argue for adding a custom `RELEASE_TESTING` flag. As far as I can see this shouldn't affect the easy open and build configuration in e.g. Xcode if we use `#if DEBUG || RELEASE_TESTING` in XCTApodini. This would allow us to e.g. log additional information and do more verification in the debug mode. If we have the `RELEASE_TESTING` flag we can verify that e.g. additional logging is actually skipped when running release tests using unit tests.

PR #289, opened 2021-06-04
Renew ExporterConfiguration and CLI arguments of the WebService

philippzagar:
# Update ExporterConfiguration and CLI-arguments for WebService

At the moment, `InterfaceExporter`s are configured in the `Configuration` of the `WebService` pretty statically. There is no possibility to configure specific exporters without a global configuration (see `OpenAPIConfiguration`) or to create associations between dependent exporters (eg. `RESTInterfaceExporter` and `OpenAPIInterfaceExporter`). 

```swift
        OpenAPIConfiguration(outputFormat: .json, 
                             outputEndpoint: "oas", 
                             swaggerUiEndpoint: "oas-ui",
                             title: "The great TestWebService - presented by Apodini")        

         ExporterConfiguration()
             .exporter(RESTInterfaceExporter.self)
             .exporter(GRPCInterfaceExporter.self)
             .exporter(ProtobufferInterfaceExporter.self)
             .exporter(OpenAPIInterfaceExporter.self)
             .exporter(WebSocketInterfaceExporter.self)
}
```

Furthermore, currently there isn't an option to modify the `Configuration` of the `WebService` via CLI arguments (eg. add/remove exporters). CLI arguments are currently handled in the specific configuration classes (eg. `HTTPConfiguration`) via CommandLine.arguments. Therefore, the user has no way to know what CLI arguments are actually supported by the `WebService`, except going through all the different `Configuration` files

## :bulb: Proposed solution
See the gist of @theMomax and @PSchmiedmayer, especially the final comment of @theMomax: [GitHub Gist](https://gist.github.com/theMomax/70adab5cf9297e952638f49ca0e2edc2#gistcomment-3741100)
My implementation is pretty close to the proposed solution.

```swift
import Apodini
import ArgumentParser

struct SomeWebService: WebService {
    @Argument var useRESTfulXML: Bool
    @Argument var openAPIOutputEndpoint: String

    // Not actually needed, just for demonstration purposes
    @Flag(help: "Include a counter with each repetition.")
    var includeCounter = false

    @Option(name: .shortAndLong, help: "The number of times to repeat 'phrase'.")
    var count: Int?

    @Argument(help: "The phrase to repeat.")
    var phrase: String

        if useRESTfulXML {
            RESTInterfaceExporter(
                encoder:  XMLEncoder(),
                decoder: XMLDecoder())
        } else {
            RESTInterfaceExporter(
                encoder: JSONEncoder(),
                decoder: JSONDecoder()) {
                    OpenAPIExporter(
                        outputEndpoint: openAPIOutputEndpoint,
                        swaggerUiEndpoint: "/ui/swagger",
                        title: "The Game - Endangered Nature Edition, built with Apodini")
                }
         }
    }
}
```

### Problem that is solved

- Exporters are now configurable (exporter-specific)
- Dependent exporters can share configuration (eg. `RESTInterfaceExporter` and `OpenAPIInterfaceExporter`)
- Enforce at compile time which Exporter-combinations are possible (eg. `RESTInterfaceExporter` and `OpenAPIInterfaceExporter`, but NOT `GRPCInterfaceExporter` and `OpenAPIInterfaceExporter`)
- CLI arguments can be used to configure `WebService` at startup and passed to other `Configuration` classes
- CLI arguments aren't scattered over multiple files anymore, now amassed in one central file (`WebService`)
- Type-safe CLI arguments provided by Swift Argument parser

I split all exporters into a `Configuration`-conformant class (which is public and used in the configuration of the `WebService`) and an `InterfaceExporter`-conformant class (which is internal and should not be accessible by the developer, since there is no need to). To me, this was a very elegant solution, but of course one could also combine these two classes into one class (but then eg. the `Application` instance in the exporter would have to be an Optional since it only gets passed during the configuration of the `WebService` at startup etc.). 
Please feel free to let me know what additionally should be configureable for the respective `InterfaceExporter`s. The possible configurations are restricted to the most basic things for now.

### Related PRs
@Supereg implemented a slimmed down version of the exporter configuration in order to be able to specify encoding/decoding strategies for the JSONEncoder/JSONDecoder of the `RESTInterfaceExporter`: #112
A couple of weeks ago (my first attempts of walking in Apodini) I extended the POC so that it would at least be functional: [Test-Branch](https://github.com/philippzagar/Apodini/tree/philippzagar/test)

### Testing
Basic test cases of the new Exporter Configuration are located in the `ExporterConfigurationTests.swift` file. Lots of other test cases had to be adapted to the new Exporter Configuration strategy.
Regarding CLI-Arguments: I didn't write test cases for this feature since I wasn't able to find a way to specify "fake" CLI arguments for test cases (please let me know if there is a way). I tested the feature with a small custom project, so I could pass actual CLI arguments to the `WebService`. On the other hand, I doubt that there is a need to write extensive test cases since the argument parsing is solely done by Apple's own Swift ArgumentParser.

### Reviewer Nudging
The actual "relevant" code changes aren't too big. 
I would suggest to start with the `RESTInterfaceExporter` and `_RESTInterfaceExporter` to get an idea how I implemented the proposed concept of @theMomax and @PSchmiedmayer. Then take a look at the `OpenAPIInterfaceExporter` to see how a dependent exporter would look like and how the configuration of the "parent" exporter is handed down. Similar procedure for other exporters. Also review my `AnyEncoder` and `AnyDecoder` implementation that conform to the `ContentEncoder`/`ContentDecoder` protocol. This is needed for the Coder configuration of the `RESTInterfaceExporter` (which relies on Vapor).
In order to be able to inject CLI arguments into the `WebService` it must conform to the `ParsableCommand` protocol of the Swift ArgumentParser which itself conforms to `Decodable`. Therefore, every property that could be stated in the `WebService` must be decodable from now on. I added conformance to `Decodable` to some classes, please let me know if this must be extended to some other classes as well.  
Please take a closer look at the changes in the DeploymentProviders regarding the CLI-argument parsing. I'm not too familiar with this area of Apodini (or actually any area, yet) and I fear that I did some pretty silly changes in order to integrate the Swift ArgumentParser.


philippzagar:
SwiftLint is crashing because of the added result builders (see comment on rocketchat from @Supereg).
I already included an exemption for these files in the .swiftlint.yml, but these changes only take effect once the PR is merged.

bauer-andreas:
> SwiftLint is crashing because of the added result builders (see comment on rocketchat from @Supereg).
> I already included an exemption for these files in the .swiftlint.yml, but these changes only take effect once the PR is merged.

To my experience SwiftLint always uses the configuration of the current branch.

Looking at the CI log you probably missed `Sources/Apodini/Configurations/StaticConfigurations/RESTDependentStaticConfigurationBuilder.swift` on the exclude list.
Note the "Static", the non static builder is on the list of excluded files, this one isn't. Been there too 😄

philippzagar:
> > SwiftLint is crashing because of the added result builders (see comment on rocketchat from @Supereg).
> > I already included an exemption for these files in the .swiftlint.yml, but these changes only take effect once the PR is merged.
> 
> To my experience SwiftLint always uses the configuration of the current branch.
> 
> Looking at the CI log you probably missed `Sources/Apodini/Configurations/StaticConfigurations/RESTDependentStaticConfigurationBuilder.swift` on the exclude list.
> Note the "Static", the non static builder is on the list of excluded files, this one isn't. Been there too 😄

Oh wow, that happens when I'm so silly and don't rename the file name to the actual class name 😄
Thanks for the info!

philippzagar:
Close PR since I moved the code to the main repo because forks can't be properly built by Bamboo (macOS builds etc.)

PR #292, opened 2021-06-07
Renew ExporterConfiguration and CLI arguments of the WebService

philippzagar:
# Renew ExporterConfiguration and CLI-arguments for WebService

At the moment, `InterfaceExporter`s are configured in the `Configuration` of the `WebService` pretty statically. There is no possibility to configure specific exporters without a global configuration (see `OpenAPIConfiguration`) or to create associations between dependent exporters (eg. `RESTInterfaceExporter` and `OpenAPIInterfaceExporter`). 

```swift
        OpenAPIConfiguration(outputFormat: .json, 
                             outputEndpoint: "oas", 
                             swaggerUiEndpoint: "oas-ui",
                             title: "The great TestWebService - presented by Apodini")        

         ExporterConfiguration()
             .exporter(RESTInterfaceExporter.self)
             .exporter(GRPCInterfaceExporter.self)
             .exporter(ProtobufferInterfaceExporter.self)
             .exporter(OpenAPIInterfaceExporter.self)
             .exporter(WebSocketInterfaceExporter.self)
}
```

Furthermore, currently there isn't an option to modify the `Configuration` of the `WebService` via CLI arguments (eg. add/remove exporters). CLI arguments are currently handled in the specific configuration classes (eg. `HTTPConfiguration`) via CommandLine.arguments. Therefore, the user has no way to know what CLI arguments are actually supported by the `WebService`, except going through all the different `Configuration` files

## :bulb: Proposed solution
See the gist of @theMomax and @PSchmiedmayer, especially the final comment of @theMomax: [GitHub Gist](https://gist.github.com/theMomax/70adab5cf9297e952638f49ca0e2edc2#gistcomment-3741100)
My implementation is pretty close to the proposed solution.

```swift
import Apodini
import ArgumentParser

struct SomeWebService: WebService {
    @Argument var useRESTfulXML: Bool
    @Argument var openAPIOutputEndpoint: String

    // Not actually needed, just for demonstration purposes
    @Flag(help: "Include a counter with each repetition.")
    var includeCounter = false

    @Option(name: .shortAndLong, help: "The number of times to repeat 'phrase'.")
    var count: Int?

    @Argument(help: "The phrase to repeat.")
    var phrase: String

        if useRESTfulXML {
            RESTInterfaceExporter(
                encoder:  XMLEncoder(),
                decoder: XMLDecoder())
        } else {
            RESTInterfaceExporter(
                encoder: JSONEncoder(),
                decoder: JSONDecoder()) {
                    OpenAPIExporter(
                        outputEndpoint: openAPIOutputEndpoint,
                        swaggerUiEndpoint: "/ui/swagger",
                        title: "The Game - Endangered Nature Edition, built with Apodini")
                }
         }
    }
}
```

### Problem that is solved

- Exporters are now configurable (exporter-specific)
- Dependent exporters can share configuration (eg. `RESTInterfaceExporter` and `OpenAPIInterfaceExporter`)
- Enforce at compile time which Exporter-combinations are possible (eg. `RESTInterfaceExporter` and `OpenAPIInterfaceExporter`, but NOT `GRPCInterfaceExporter` and `OpenAPIInterfaceExporter`)
- CLI arguments can be used to configure `WebService` at startup and passed to other `Configuration` classes
- CLI arguments aren't scattered over multiple files anymore, now amassed in one central file (`WebService`)
- Type-safe CLI arguments provided by Swift Argument parser

I split all exporters into a `Configuration`-conformant class (which is public and used in the configuration of the `WebService`) and an `InterfaceExporter`-conformant class (which is internal and should not be accessible by the developer, since there is no need to). To me, this was a very elegant solution, but of course one could also combine these two classes into one class (but then eg. the `Application` instance in the exporter would have to be an Optional since it only gets passed during the configuration of the `WebService` at startup etc.). 
Please feel free to let me know what additionally should be configureable for the respective `InterfaceExporter`s. The possible configurations are restricted to the most basic things for now.

### Related PRs
@Supereg implemented a slimmed down version of the exporter configuration in order to be able to specify encoding/decoding strategies for the JSONEncoder/JSONDecoder of the `RESTInterfaceExporter`: #112
A couple of weeks ago (my first attempts of walking in Apodini) I extended the POC so that it would at least be functional: [Test-Branch](https://github.com/philippzagar/Apodini/tree/philippzagar/test)

### Testing
Basic test cases of the new Exporter Configuration are located in the `ExporterConfigurationTests.swift` file. Lots of other test cases had to be adapted to the new Exporter Configuration strategy.
Regarding CLI-Arguments: I didn't write test cases for this feature since I wasn't able to find a way to specify "fake" CLI arguments for test cases (please let me know if there is a way). I tested the feature with a small custom project, so I could pass actual CLI arguments to the `WebService`. On the other hand, I doubt that there is a need to write extensive test cases since the argument parsing is solely done by Apple's own Swift ArgumentParser.

### Reviewer Nudging
The actual "relevant" code changes aren't too big. 
I would suggest to start with the `RESTInterfaceExporter` and `_RESTInterfaceExporter` to get an idea how I implemented the proposed concept of @theMomax and @PSchmiedmayer. Then take a look at the `OpenAPIInterfaceExporter` to see how a dependent exporter would look like and how the configuration of the "parent" exporter is handed down. Similar procedure for other exporters. Also review my `AnyEncoder` and `AnyDecoder` implementation that conform to the `ContentEncoder`/`ContentDecoder` protocol. This is needed for the Coder configuration of the `RESTInterfaceExporter` (which relies on Vapor).
In order to be able to inject CLI arguments into the `WebService` it must conform to the `ParsableCommand` protocol of the Swift ArgumentParser which itself conforms to `Decodable`. Therefore, every property that could be stated in the `WebService` must be decodable from now on. I added conformance to `Decodable` to some classes, please let me know if this must be extended to some other classes as well.  
Please take a closer look at the changes in the DeploymentProviders regarding the CLI-argument parsing. I'm not too familiar with this area of Apodini (or actually any area, yet) and I fear that I did some pretty silly changes in order to integrate the Swift ArgumentParser.


Documentation/Exporters/OpenAPI.md
```diff
@@ -48,17 +49,17 @@ This means that per default a `JSON` representation of the OAS is served at `/op
...
-            swaggerUiEndpoint: "/ui/swagger",
-            title: "The Game - Endangered Nature Edition, built with Apodini"
-        )
+        RESTInterfaceExporter { 
+            /// Adds configuration to the OpenAPIExporter
+            OpenAPIInterfaceExporter(outputFormat: .yaml,
+                                     outputEndpoint: "/docs/openapi",
+                                     swaggerUiEndpoint: "/ui/swagger",
+                                     title: "The Game - Endangered Nature Edition, built with Apodini")
+        }
```
PSchmiedmayer: In general I would suggest the following formatting:
```suggestion
            OpenAPIInterfaceExporter(
                outputFormat: .yaml,
                 outputEndpoint: "/docs/openapi",
                 swaggerUiEndpoint: "/ui/swagger",
                 title: "The Game - Endangered Nature Edition, built with Apodini"
            )
```

Package.swift
```diff
@@ -125,7 +128,8 @@ let package = Package(
                 .product(name: "NIOSSL", package: "swift-nio-ssl"),
                 .product(name: "Logging", package: "swift-log"),
                 .product(name: "Runtime", package: "Runtime"),
-                .product(name: "ConsoleKit", package: "console-kit")
+                .product(name: "ConsoleKit", package: "console-kit"),
```
PSchmiedmayer: Do we still need to import ConsoleKit?

Sources/ApodiniREST/RESTInterfaceExporter.swift
```diff
@@ -3,37 +3,101 @@
...
+    
+    /**
+     Initializes the configuration of the `RESTInterfaceExporter` with (default) `AnyEncoder` and `AnyDecoder`
+     - Parameters:
+     - encoder: The to be used `AnyEncoder`
+     - decoder: The to be used `AnyDecoder`
+     */
+    public init(encoder: AnyEncoder = defaultEncoder,
+                decoder: AnyDecoder = defaultDecoder) {
+        self.configuration = RESTExporterConfiguration(encoder: encoder, decoder: decoder)
```
theMomax: Isn't this the actual `convenience init`?

Sources/ApodiniUtils/AnyEncodable.swift
```diff
@@ -6,7 +6,7 @@
 //
 
 import Foundation
-
+import Vapor
```
theMomax, line 9: Please move any Vapor-dependent code to ApodiniREST. Add the `AnyEncoder: ContentEncode` conformance as an extension there.

Sources/ApodiniREST/RESTInterfaceExporter.swift
```diff
@@ -3,37 +3,101 @@
...
+    static let parameterNamespace: [ParameterNamespace] = .individual
+    
+    let app: Vapor.Application
+    let exporterConfiguration: RESTConfiguration
+    
     /// Initialize `RESTInterfaceExporter` from `Application`
-    public required init(_ app: Apodini.Application) {
+    required init(_ app: Apodini.Application, _ exporterConfiguration: ExporterConfiguration = RESTExporterConfiguration()) {
+        guard let castedConfiguration = dynamicCast(exporterConfiguration, to: RESTExporterConfiguration.self) else {
+            fatalError("Wrong configuration type passed to exporter, \(type(of: exporterConfiguration)) instead of \(Self.self)")
```
theMomax: This is really ugly and quite unnecessary. Remove the `init` from the `BaseInterfaceExporter` protocol and delete the `ExporterConfiguration` protocol. `RESTInterfaceExporter` calls `_RESTInterfaceExporter.init`, not the framework, therefore we don't have to make any artificial restrictions here.

Sources/Apodini/Configurations/StaticConfigurations/RESTDependentStaticConfiguration.swift
```diff
@@ -0,0 +1,30 @@
+//
+//  RESTDependentStaticConfiguration.swift
```
theMomax: Please move these files (and the accompanying builders) to their respective targets.

Sources/Apodini/Configurations/StaticConfigurations/RESTDependentStaticConfiguration.swift
```diff
@@ -0,0 +1,30 @@
...
+//
+//  Created by Philipp Zagar on 27.05.21.
+//
+
+/// `RESTDependentStaticConfiguration`s are used to register static services dependend on the `RESTInterfaceExporter`
+public protocol RESTDependentStaticConfiguration: StaticConfiguration {}
+
+/// The default configuration is an `EmptyRESTDependentStaticConfiguration`
+public struct EmptyRESTDependentStaticConfiguration: RESTDependentStaticConfiguration {
+    public func configure(_ app: Application, parentConfiguration: ExporterConfiguration) { }
```
theMomax: This is the same as for the initialiser on `BaseInterfaceExporter`. `RESTInterfaceExporter` knows `RESTDEpendentStaticConfiguration`. Delete `StaticConfiguration` and replace `ExporterConfiguration` by `RESTExporterConfiguration`. Same applies for gRPC, of course.

theMomax (changes requested):
General concept looks good! I added some comments regarding the new configuration protocols - we don't need them. The exporters know their own configuration-types. Using the protocols forces the exporter-developer to do some ugly type assertions.

One other point: there were a couple of files that would be better placed in the respective exporter-package.

Please fix these minor problems, other than that, looks good👍🏻

Sources/ApodiniREST/RESTInterfaceExporter.swift
philippzagar: My current `convenience init` additionally initializes the staticConfigurations (if the Encoder/Decoder is JSON) and then uses the initializer above. Therefore my current implementation should be correct, right?
Please let me know if I have an error in my reasoning

theMomax: `convenience` initialisers usually provide less customisability than non-convenience ones. The convenience is that you have to type less.

theMomax: Ah sorry, I didn't notice your `convenience init` is restricted to `JSONEncoder` instead of `AnyEncoder`...makes total sense now👍🏻

philippzagar:
Thanks for the feedback @theMomax ! I think it improved the quality of the PR a lot, especially the clear separation of the different modules (exporters) 👍

Sources/ApodiniREST/RESTInterfaceExporter.swift
```diff
@@ -8,32 +8,94 @@ import NIO
...
+// swiftlint:disable type_name
+final class _RESTInterfaceExporter: InterfaceExporter {
+    static let parameterNamespace: [ParameterNamespace] = .individual
+    
+    let app: Vapor.Application
+    let exporterConfiguration: RESTConfiguration
+    
     /// Initialize `RESTInterfaceExporter` from `Application`
-    public required init(_ app: Apodini.Application) {
+    required init(_ app: Apodini.Application,
```
theMomax: This is not `required` anymore, is it?

Sources/Apodini/WebService.swift
```diff
@@ -57,10 +62,17 @@ extension WebService {
...
+    /**
+     This function is provided to start up an Apodini `WebService`. The `app` parameter can be injected for testing purposes only. Use `WebService.start()` to startup an Apodini `WebService`.
+     - Parameters:
+         - app: The app instance that should be injected in the Apodini `WebService`
+         - webService: The instanciated `WebService` by the Swift ArgumentParser
+     */
+    static func start(app: Application, webService webServiceTemp: Self? = nil) {
+        /// If `WebService` isn't already instanciated by the Swift ArgumentParser, manually create an instance here
+        let webService = webServiceTemp ?? Self()
+
```
PSchmiedmayer: I think that we can simplify the signature a bit:
```suggestion
    static func start(app: Application, webService: Self = Self()) {
```

Sources/Apodini/WebService.swift
```diff
@@ -29,19 +29,24 @@ public extension WebService {
...
     
     /// This function is executed to start up an Apodini `WebService`
     @discardableResult
-    static func main(waitForCompletion: Bool) throws -> Application {
+    static func start(waitForCompletion: Bool, webService: Self? = nil) throws -> Application {
         let app = Application()
         LoggingSystem.bootstrap(StreamLogHandler.standardError)
 
-        main(app: app)
+        start(app: app, webService: webService ?? Self())
```
PSchmiedmayer: Can probably be simplified in one function like this:
```swift
    /// This function is executed to start up an Apodini `WebService`
    @discardableResult
    static func start(waitForCompletion: Bool = true, webService: Self = Self()) throws -> Application {
        let app = Application()
        LoggingSystem.bootstrap(StreamLogHandler.standardError)

        start(app: app, webService: webService)
        // ...
```

Sources/Apodini/Application/Application+Exporters.swift
```diff
@@ -59,7 +59,8 @@ extension Application {
             return self.application.storage[Key.self]!
         }
 
-        var semanticModelBuilderBuilder: (SemanticModelBuilder) -> (SemanticModelBuilder) {
+        /// The `SemanticModelBuilder` to register services upon
+        public var semanticModelBuilderBuilder: (SemanticModelBuilder) -> (SemanticModelBuilder) {
```
PSchmiedmayer: To be honest I think that the current handling of the SemanticModelBuilder as well as how it is extended in this PR is very complicated and making it public is also an addition to the API surface. As far as I can see the reason for making the SemanticModelBuilder public is to easily register the interface exporters in the configure method.

I would suggest to following which doesn't require the SemanticModelBuilder to be public as this is really a class that a normal user shouldn't really need to care about. I would suggest the following:
Add a computed property to `Application` that stores a collection of `AnyInterfaceExporter` and maybe add convenience methods to application such as `addInterfaceExporter(...)` for `InterfaceExporter` and `StaticInterfaceExporter`. This way the interface exporters are collected in the application, the SemanticModelBuilder doesn't need to be passed around and the `SemanticModelBuilder` can be created after all configurations have been run and gets the interface exporter collection passed in in the initializer or accesses it using the `Application`:
```swift
extension Application {
    private final class InterfaceExporterStorage {
        var exporters: [AnyInterfaceExporter] = []
        
        func append(_ exporter: AnyInterfaceExporter) {
            exporters.append(exporter)
        }
    }
    
    private enum InterfaceExporerKey: StorageKey {
        typealias Value = InterfaceExporterStorage
    }
    
    var interfaceExporters: [AnyInterfaceExporter] {
        get {
            storage[InterfaceExporerKey.self]?.exporters ?? []
        }
    }
    
    private var interfaceExporterStorage: InterfaceExporterStorage {
        get {
            guard let exporterStorage = storage[InterfaceExporerKey.self] else {
                let exporterStorage = InterfaceExporterStorage()
                storage[InterfaceExporerKey.self] = exporterStorage
                return exporterStorage
            }
            return exporterStorage
        }
    }
    
    /// Registers an `InterfaceExporter` instance on the model builder.
    /// - Parameter instance: The instance to register.
    /// - Returns: `Self`
    public func register<T: InterfaceExporter>(exporter instance: T) {
        interfaceExporterStorage.append(AnyInterfaceExporter(instance))
    }
    
    /// Registers an `StaticInterfaceExporter` instance on the model builder.
    /// - Parameter instance: The instance to register.
    /// - Returns: `Self`
    public func register<T: StaticInterfaceExporter>(exporter instance: T) {
        interfaceExporterStorage.append(AnyInterfaceExporter(instance))
    }
}
```

The `SemanticModelBuilder` drops the with methods and just gets the interface exporters in the initializer.

The `start` method changes to something like this:
```swift
    static func start(app: Application, webService: Self = Self()) {
        /// Configure application and instanciate exporters
        webService.configuration.configure(app)
        
        /// If no specific address hostname is provided we bind to the default address to automatically and correcly bind in Docker containers.
        if app.http.address == nil {
            app.http.address = .hostname(HTTPConfiguration.Defaults.hostname, port: HTTPConfiguration.Defaults.port)
        }
        
        webService.register(
            SemanticModelBuilder(app, interfaceExporters: app.interfaceExporters)
        )
    }
```

This can then also remove the logic in this file here.

Sources/ApodiniDeploy/ApodiniDeployExporterConfiguration.swift
```diff
@@ -11,29 +11,28 @@ import Apodini
 import ApodiniDeployBuildSupport
 import ApodiniDeployRuntimeSupport
 
-
-public struct ApodiniDeployConfiguration: Apodini.Configuration {
-    struct StorageKey: Apodini.StorageKey {
-        typealias Value = ApodiniDeployConfiguration
-    }
-    
+struct ApodiniDeployExporterConfiguration {
```
PSchmiedmayer: I think this is a breaking change for the deployment provider logic and probably results in failed test cases if we would enable the deployment tests (CCL @lukaskollmer)

I think it might be reasonable to add the functionality of subcommands of the Swift Argument parser to Apodini: https://github.com/apple/swift-argument-parser/blob/main/Documentation/03%20Commands%20and%20Subcommands.md

I would suggest to extend `Apodini.Configuration` with an additional property:
```swift
import ArgumentParser


/// `Configuration`s are used to register services to Apodini.
/// Each `Configuration` handles different kinds of services.
public protocol Configuration {
    var subcommands: [ParsableCommand.Type] { get }
    
    /// A method that handles the configuration of a service which is called by the `main` function.
    ///
    /// - Parameter
    ///    - app: The `Vapor.Application` which is used to register the configuration in Apodini
    func configure(_ app: Application)
}

extension Configuration {
    public var subcommands: [ParsableCommand.Type] {
        []
    }
}

// ...

    public var subcommands: [ParsableCommand.Type] {
        var subcommands: [ParsableCommand.Type] = []
        forEach {
            subcommands.append(contentsOf: $0.subcommands)
        }
        return subcommands
    }
    
        forEach {
            $0.configure(app)
        }
    }
}
```

The logic currently found in the WebService could move to a `Start` `ParsableCommand` that is the default parsable command in the WebService:
```swift

public extension WebService {
    var configuration: CommandConfiguration {
        CommandConfiguration(
            abstract: "A utility for performing maths.",
            subcommands: [Start.self] + self.configuration.subcommands,
            defaultSubcommand: Start.self
        )
    }
}

struct Start: ParsableCommand {
    // ...
}
```


This might be relevant for 
- `Apodini Deploy`: @hendesi & @lukaskollmer The deployment provider would add a deploy subcommand and then e.g. an export sub subcommand that triggers the export functionality so you can call it like this: `MyWebService deploy export --mode ... --file ... --node` and so on. Adding this to the main web service using Arguments doesn't really make sense but enabling configurations to add this as subcommands is definitely a welcome thing. Please get in contact with @lukaskollmer as this current change definitely breaks something in the deployment providers and we need to make sure they still work after the PR. @hendesi will continue this work with his deployment providers and I can imagine that these subcommand structure might also become helpful for him as we have discussed incorporating deployment provider right in the main web service binary. You could then deploy your web service to AWS Lambda e.g. executing the web service with `MyWebService deploy awslambda --username ...`
- `Apodini Database`: Relevant for @hendesi : It would be good to have a command like `MyWebService database migrate` or `MyWebService database revert` to migrate and revert database schemas without starting the actual web service. The command could then exit after the migration/revert with a success exit code and then migrating database changes is easier then starting up the web service to run the migrations and currently reverting a migration is not even possible.

Sources/ApodiniREST/RESTInterfaceExporter.swift
```diff
@@ -8,32 +8,94 @@ import NIO
...
+                            decoder: JSONDecoder = defaultDecoder as! JSONDecoder,
+                            @RESTDependentStaticConfigurationBuilder staticConfigurations: () -> [RESTDependentStaticConfiguration] = { [] }) {
+        self.init(encoder: encoder, decoder: decoder)
+        self.staticConfigurations = staticConfigurations()
+    }
+}
 
+/// Internal Apodini Interface Exporter for REST
+// swiftlint:disable type_name
+final class _RESTInterfaceExporter: InterfaceExporter {
```
PSchmiedmayer: To be honest I am not too much of a fan of the underscores being added to all exporters. Maybe a more pragmatic way: What about naming the configuration `REST`? I think that the average Apodini user has no clue what an InterfaceExporter is anyways and just naming the configuration `REST` probably makes it quite easy to find. Same applies to `OpenAPI`, `GRPC` and so on ...

Sources/ApodiniREST/RESTExporterConfiguration.swift
```diff
@@ -0,0 +1,29 @@
...
+//  
+//
+//  Created by Philipp Zagar on 21.05.21.
+//
+
+import Foundation
+import Apodini
+
+/// Configuration of the `RESTInterfaceExporter`
+public struct RESTExporterConfiguration {
```
PSchmiedmayer: I would suggest to name this `ExporterConfiguration` and maybe make it a nested type so it is in the scope of the `REST` configuration, so you from the outside you refer to it as `REST.ExporterConfiguration`.

PSchmiedmayer (changes requested):
First of all thanks for the time and effort your put in this PR @philippzagar even though it is not at the core if your goal of adding logging, metrics and traces to Apodini. I greatly appreciate this.
The overall direction looks good and I added a few comments there and there about smaller things. 
Nevertheless I had three bigger comments that would improve the PR:
- I have added a large comment sharing an approach that would not require making the `SemanticModelBuilder` public.
- The naming of the configurations and interface exporters could be simplified (e.g. the configuration is named `REST`, leaving the name `RESTInterfaceExporter` available so we don't have to use underscores.
- I think that it would make sense for some exporters to offer subcommands. I have added an extensive comment in the PR review about this.

A small comment: I noted some non `@_implementationOnly` Vapor imports. After the feedback they are only located in the ApodiniREST target, which should be fine @Supereg ?

Package.swift
philippzagar: I actually think we can remove the overall console-kit dependency from Apodini, Swift Argument Parser should completely replace it.

Sources/Apodini/WebService.swift
philippzagar: During the implementation, I was actually wondering why the "first" main() function (which doesn't take any parameters) is required :)

Sources/Apodini/Application/Application+Exporters.swift
philippzagar: Maybe it would even make sense to merge the passed `InterfaceExporters` in the initializer of `SemanticModelBuilder` with the `InterfaceExporters` in Application.interfaceExporters? Of course, we would need to check for duplicates (eg. not two RESTExporters, even if they have different configurations) and prioritize the passed `InterfaceExporters` over Application.interfaceExporters

Sources/ApodiniREST/RESTExporterConfiguration.swift
philippzagar: I expanded this idea to all exporters, now all ExporterConfigurations are accessible via <Exporter>.ExporterConfiguration

Sources/ApodiniDeploy/ApodiniDeployExporterConfiguration.swift
philippzagar: First of all, I really like the idea to have different subcommands that trigger various functionality in Apodini (eg. migrating DB without startup of whole Apodini WebService)!

However, there are a few problems I would like to mention:

- You are declaring the `var configuration: CommandConfiguration` in the `WebService` as non-static and as a computed property. However, a look into the documentation shows that the configuration first needs to be static (the easy thing to solve) and second it MUST be a stored property (i tried it out myself, a computed property apparently isn't recognized by the ArgumentParser and therefore the `run` method in `Start` is never called, only the original `run` in the `WebService`). A stored property in a protocol extension isn't possible, therefore I guess we have a problem here.
image
I guess the developers have to specify a `CommandConfiguration` themselves in the `WebService` struct, a very unpleasant solution. But I guess there could be some workarounds, like https://stackoverflow.com/questions/38885622/swift-protocol-extensions-property-default-values

- Name conflict between the new `configuration: CommandConfiguration` and the old `configuration: Configuration`, we would have to rename the `configuration: Configuration` to something like `ExporterConfiguration`.
- What logic from the `WebService` should move to the `Start` exactly? I would argue for as little as possible, so just the `run` method required by the `ArgumentParser` as well as the `start` functions from Apodini. So, ground rules established. Now, to make the CLI-Arguments in the Start CMD configurable by the user (clarification: if we execute `webserice start --test=abc`, the parameter is only injected into the `Start` struct), the `Start` struct must actually be a protocol. I see two problems here: First, if the `configuration: Configuration` stays in the `WebService`, how would the injected arguments in `Start` be used in the `configuration: Configuration` in the `WebService`? Second, if `Start` is a protocol, the actual struct instantiation can be named arbitrarily. How do we reference this struct then in the `CommandConfiguration`? (If we require the developer to implement a `CommandConfiguration` manually in the WebService as well as a struct `StartCMD` that conforms to `Start` which conforms to `ParsableCommand` and provides a default implementation for the `run` function, then it would be possible)
- Another option would be to move the `configuration: Configuration` to `Start` (so conformance to the `ConfigurationCollection)`. In my opinion a very unpleasant solution. Now it would be possible to use the passed CLI arguments in the configuration etc. But again, how do we specify the actual `Start` struct in the `CommandConfiguration` in the `WebService` then? And, I think it breaks LOTS of things if we don't have a configuration in the `WebService`. I also think that this would be the completely wrong direction, it really doesn't make sense to specify the configuration in the `Start` struct and not in the `WebService`
- Probably some more problems I'm not seeing right away

It's very well possible that there is a very easy solution here that I'm not seeing. If that's the case, please let me know ASAP :)

Sources/Apodini/Application/Application+Exporters.swift
PSchmiedmayer: I see, I missed that you already pass the app instance in the `SemanticModelBuilder`. Just calling `SemanticModelBuilder(app)` is sufficient and you can also remove the interface builder related property and functions in the `SemanticModelBuilder`, this is then just duplicated and everything should be stored in the `app` instance and should be retrieved from there.

Sources/ApodiniDeploy/ApodiniDeployExporterConfiguration.swift
PSchmiedmayer: You are right, this is not that easy. I have provided a possible solution showcases in the following PR: #298 ... ideally we could get the subcommands from the configurations using Runtime but as noted in #298 this is currently blocked by features enabled by Runtime.

Sources/ApodiniDeploy/ApodiniDeployInterfaceExporter.swift
```diff
@@ -43,15 +43,10 @@ public final class ApodiniDeploy: Configuration {
     
     public init(runtimes: [DeploymentProviderRuntime.Type] = [],
                 config: DeploymentConfig = .init(),
-                mode: String? = nil,
-                fileURL: String? = nil,
-                node: String? = nil) {
+                mode: String? = nil) {
```
lukaskollmer: I don't think the `mode` parameter is still necessary.

Sources/ApodiniDeploy/ApodiniDeployInterfaceExporter.swift
```diff
@@ -151,12 +146,17 @@ class ApodiniDeployInterfaceExporter: InterfaceExporter {
     
     
     private func performDeploymentRelatedActions() throws {
-        guard let mode = self.exporterConfiguration.mode, let fileURL = self.exporterConfiguration.fileURL else {
+        // If those enironment variable do not exist, return immediatly
+        guard let mode = ProcessInfo.processInfo.environment[WellKnownEnvironmentVariables.executionMode] else {
+            return
+        }
+        
+        guard let fileURL = ProcessInfo.processInfo.environment[WellKnownEnvironmentVariables.fileUrl] else {
```
lukaskollmer: You can also combine the two `guard`s into one:

```swift
let env = ProcessInfo.processInfo.environment
guard
    let mode = env[WellKnownEnvironmentVariables.executionMode],
    let fileUrl = env[WellKnownEnvironmentVariables.fileUrl]
else {
    return
}

```

Sources/DeploymentTargetAWSLambda/AWSIntegration.swift
```diff
@@ -182,11 +182,15 @@ class AWSIntegration { // swiftlint:disable:this type_body_length
...
             try fileManager.setPosixPermissions("rw-r--r--", forItemAt: launchInfoFileUrl)
             
+            
             do {
                 // create & add bootstrap file
+                // TODO: Not entirly sure if this passing of ENVs works
                 let bootstrapFileContents = """
                 #!/bin/bash
-                ./\(lambdaExecutableUrl.lastPathComponent) --mode=\(WellKnownCLIArguments.launchWebServiceInstanceWithCustomConfig) --fileurl=./\(launchInfoFileUrl.lastPathComponent)
+                \(WellKnownEnvironmentVariables.executionMode)=\(WellKnownEnvironmentVariableExecutionMode.launchWebServiceInstanceWithCustomConfig)
```
lukaskollmer: I don't see why this shouldn't work (but then i haven't tested it), but i think it might be a better idea to use lambda's environment handling to set these. (That's what we're already doing with the node id)

You can leave the code as is, and I'll make that change in my branch once this is merged

Sources/DeploymentTargetAWSLambda/main.swift
```diff
@@ -285,8 +285,9 @@ struct LambdaDeploymentProvider: DeploymentProvider {
         try runInDocker(
             imageName: dockerImageName,
             bashCommand: [
-                "swift", "run", productName,
-                "--mode=" + WellKnownCLIArguments.exportWebServiceModelStructure, "--fileurl=.build/\(tmpDirName)/\(filename)" // can't use self.tmpDirUrl here since that's an absolute path but we need a relative one bc this is running in the docker container which has a different mount path
+                "\(WellKnownEnvironmentVariables.executionMode)=\(WellKnownEnvironmentVariableExecutionMode.exportWebServiceModelStructure)",
```
lukaskollmer: same as above, this should be working fine. However, I'd prefer we instead add an `environment` parameter to the `runInDocker` function, and use docker's `--env` option to specify the container's environment variables. I'll implement that in my branch (unless of course you want to make that change)

Tests/ApodiniDeployTests/ApodiniDeployTestCase.swift
```diff
@@ -40,8 +40,9 @@ class ApodiniDeployTestCase: XCTestCase {
     
     
     static var shouldRunDeploymentProviderTests: Bool {
-        ProcessInfo.processInfo.environment["ENABLE_DEPLOYMENT_PROVIDER_TESTS"] != nil
-            || ProcessInfo.processInfo.environment["AD_APODINI_SOURCE_ROOT"] != nil
+//        ProcessInfo.processInfo.environment["ENABLE_DEPLOYMENT_PROVIDER_TESTS"] != nil
```
lukaskollmer: this needs to be changed back to the actual implementation

Sources/ApodiniDeployBuildSupport/WebServiceStructure.swift
```diff
@@ -11,25 +11,28 @@ import ApodiniUtils
...
-
 /// Well-known environment variables, i.e. environment variables which are read by Apodini and used when performing certain tasks.
 /// Note: environment variables which are used in a lambda context must satisfy the regex `[a-zA-Z]([a-zA-Z0-9_])+`
 public enum WellKnownEnvironmentVariables {
     /// Key for an environment variable specifying the current instance's node id (relative to the whole deployed system).
     /// This environment variable is only set of the web service is running as part of a managed deployment.
     public static let currentNodeId = "ApodiniDeployCurrentNodeId"
+    
+    /// Key for an environment variable specifying the execution mode of ApodiniDeploy, ether dump the WebService's model structur or launch the WebService with custom config
+    public static let executionMode = "ApodiniDeployExecutionMode"
```
lukaskollmer: I'd change these to match what seems to be the general style used for environment variables: all uppercase and snake case:

- `AD_EXECUTION_MODE`
- `AD_INPUT_FILE_PATH`

lukaskollmer (changes requested):
(Commenting only on the CLI param handling part, as it relates to ApodiniDeploy)

I think this looks good, and, since we're just moving the params from one place to another, shouldn't have too much (i.e. none at all) impact on the other parts of the system.

There's some places where I commented improvement ideas as to how exactly the arguments are passed when invoking child processes, but as far as i can see these are optional, and it should work just fine with the current implementation

Sources/DeploymentTargetAWSLambda/AWSIntegration.swift
philippzagar: 👍

Tests/ApodiniDeployTests/ApodiniDeployTestCase.swift
philippzagar: Was already done, i updated the branch after I sent you the commit :)

Sources/DeploymentTargetAWSLambda/main.swift
philippzagar: I'll take a closer look at that in the evening :)

Sources/ApodiniDeploy/ApodiniDeployExporterConfiguration.swift
philippzagar: I will implement this feature in a separate PR once this one is merged

Sources/Apodini/WebService.swift
```diff
@@ -29,19 +29,24 @@ public extension WebService {
...
+    /**
+     This function is executed to start up an Apodini `WebService`
+     - Parameters:
+         - waitForCompletion: Indicates whether the `Application` is launched or just booted. Defaults to true, meaning the `Application` is run
+         - webService: The instanciated `WebService` by the Swift ArgumentParser containing CLI arguments.  If `WebService` isn't already instanciated by the Swift ArgumentParser, automatically create a default instance
+     - Returns: The application on which the `WebService` is operating on
+     */
     @discardableResult
-    static func main(waitForCompletion: Bool) throws -> Application {
+    static func start(waitForCompletion: Bool = true, webService: Self = Self()) throws -> Application {
```
theMomax, line 43: Does it make sense to keep the `waitForCompletion` option around? As far as I can see, `WebService` always is a `ParsableCommand`, thus `run` always gets executed, which in turn always uses the default `true`. Is there a way to manually execute `start` in a sensible manner?

Sources/Apodini/WebService.swift
```diff
@@ -57,10 +62,14 @@ extension WebService {
     }
     
 
-    /// This function is provided to start up an Apodini `WebService`. The `app` parameter can be injected for testing purposes only. Use `WebService.main()` to startup an Apodini `WebService`.
-    /// - Parameter app: The app instance that should be injected in the Apodini `WebService`
-    static func main(app: Application) {
-        let webService = Self()
+    /**
+     This function is provided to start up an Apodini `WebService`. The `app` parameter can be injected for testing purposes only. Use `WebService.start()` to startup an Apodini `WebService`.
+     - Parameters:
```
theMomax: We use single line comments throughout the project...I'd prefer to keep it that way.

Sources/ApodiniDeploy/ApodiniDeployInterfaceExporter+ExportWebServiceStructure.swift
```diff
@@ -1,5 +1,5 @@
 //
-//  ApodiniDeployInterfaceExporter+ExportWebServiceStructure.swift
+//  _ApodiniDeployInterfaceExporter+ExportWebServiceStructure.swift
```
theMomax: I think you can remove the underscore again.

Sources/ApodiniDeploy/ApodiniDeployInterfaceExporter+ExportWebServiceStructure.swift
```diff
@@ -15,9 +15,9 @@ import OpenAPIKit
 
 
 extension ApodiniDeployInterfaceExporter {
-    func exportWebServiceStructure(to outputUrl: URL, apodiniDeployConfiguration: ApodiniDeployConfiguration) throws {
+    func exportWebServiceStructure(to outputUrl: URL, apodiniDeployConfiguration: ApodiniDeploy.ExporterConfiguration) throws {
         let deploymentConfig = apodiniDeployConfiguration.config
-        guard let openApiDocument = app.storage.get(OpenAPIStorageKey.self)?.document else {
+        guard let openApiDocument = app.storage.get(OpenAPI.StorageKey.self)?.document else {
```
theMomax, line 20: I like the usage of child-types!

Sources/ApodiniDeploy/ApodiniDeployInterfaceExporter.swift
```diff
@@ -213,10 +238,10 @@ public class ApodiniDeployInterfaceExporter: InterfaceExporter {
 }
 
 
-// MARK: ApodiniDeployInterfaceExporter.ExporterRequest
+// MARK: _ApodiniDeployInterfaceExporter.ExporterRequest
```
theMomax: Again, underscore can be removed, I guess.

Sources/ApodiniREST/AnyCodable.swift
```diff
@@ -0,0 +1,48 @@
...
+import ApodiniUtils
+import Vapor
+
+public protocol AnyEncoder: ApodiniUtils.AnyEncoder, ContentEncoder {}
+
+/// Default implementation of the `ContentEncoder` protocol that sets the content type to JSON
+extension AnyEncoder {
+    /**
+     Encodes the given object which conform to `Encodable` into the `Bytebuffer`
+     - Parameters:
```
theMomax: Again, please use single line comments for consistency.

Sources/ApodiniUtils/AnyCodable.swift
```diff
@@ -7,7 +7,6 @@
 
 import Foundation
```
theMomax, line 8: Name at the top must be changed to `AnyCodable.swift`.

Tests/ApodiniTests/OpenAPITests/OpenAPIDocumentBuilderTests.swift
```diff
@@ -125,9 +125,9 @@ final class OpenAPIDocumentBuilderTests: ApodiniTests {
 
         let builtDocument = documentBuilder.build()
 
-        XCTAssertNoThrow(try builtDocument.output(.json))
-        print(try document.output(.json)!)
-        print(try builtDocument.output(.json)!)
+        XCTAssertNoThrow(try builtDocument.output(configuration: exporterConfiguration))
+        print(try document.output(configuration: exporterConfiguration)!)
+        print(try builtDocument.output(configuration: exporterConfiguration)!)
```
theMomax: Left over debug prints! Please also search the rest of the project...maybe I oversaw some...

theMomax (approved):
Looks good! It really makes everything look a lot cleaner...the web service declarations as well as the exporters' public interfaces.

I noted down some minor details (mostly formatting really) that should be addressed before merging. Also, there seem to be two empty files you should delete: `Sources/ApodiniJobs/Job.swift` and `Tests/ApodiniJobsTests/JobsTests.swift`.

While the general changes seem fine, I haven't checked the details related to ApodiniDeploy. Maybe @lukaskollmer can give final approval on that part. Please wait for his feedback.

PSchmiedmayer (commented):
Once @theMomax and @lukaskollmer have approved the PR and their comments are addressed we can merge the PR from my side. From a high level perspective it looks like all my comments were addressed, looks good to me 🚀

PSchmiedmayer (approved):

Tests/ApodiniTests/OpenAPITests/OpenAPIDocumentBuilderTests.swift
philippzagar: I actually didn't insert those prints (I think they got in via Metadata DSL or smt. like that), I just edited how the configuration is passed

theMomax: Ah, I see...that happens...please remove them anyways.

Sources/Apodini/WebService.swift
philippzagar: I also thought about that, but this option is used (not the default `true`) in the `WebserviceTests.swift` test. Should I leave it in there or remove it?

theMomax: Oh, in that case you can leave it!

philippzagar:
> Looks good! It really makes everything look a lot cleaner...the web service declarations as well as the exporters' public interfaces.
> 
> I noted down some minor details (mostly formatting really) that should be addressed before merging. Also, there seem to be two empty files you should delete: `Sources/ApodiniJobs/Job.swift` and `Tests/ApodiniJobsTests/JobsTests.swift`.
> 
> While the general changes seem fine, I haven't checked the details related to ApodiniDeploy. Maybe @lukaskollmer can give final approval on that part. Please wait for his feedback.

The "empty" files in the PR (`Job.swift`, `JobsTest.swift`) seem like a GitHub bug.. I also talked to Max, he thinks that the error lies in the change-view, therefore the merge should be fine.

Changed files view (seems to be displaying wrong information):
image

Actual file in branch in the repo:
image

lukaskollmer (approved):

PR #293, opened 2021-06-08
AWS resource management, and other improvements

lukaskollmer:
# AWS resource management, other improvements

The current implementation of the AWS deployment provider lacks the functionality to manage the resources created when running the deployment provider. While it currently is possible to re-use existing lambda functions which were created as part of previous deployments, other resources (e.g. IAM roles) were re-created for every deployment, and then just left around in the AWS account.

## :bulb: Proposed solution
The Lambda deployment provider now supports two additional subcommands for deleting resources in the AWS account. This is meant to be used in combination with the deploy command (the primary subcommand), to make sure an AWS account contains only the resources which are actually required for running a deployed web service.
This is achieved by encoding the id of the API Gateway into the names of all AWS resources created as part of a deployment. In the absence of an "official" web service id, this is the next best thing we can use to identify a deployed web service.

### Implications
It's better than before.

- Both new subcommands have a `--dry-run` option, which is pretty neat for checking what would be deleted w/out actually changing anything in the account.
- @PSchmiedmayer I removed the "delete existing lambda functions" code block from the AWSIntegration class. I think you uncommented that when you merged the PR, but it didn't actually work, and the same functionality is now also provided by the new dedicated commands, so IMO there's no point in keeping it.
- There also is an option not to delete the API Gateway (but delete everything else, including all routes in the gateway). The idea behind this is that it allows us to keep the gateway id (i.e. the id of the deployed web service) stable across test runs, which means we can run the cleanup command as part of the test script (outside XCTest) to make sure there's no remnants from failed tests remaining in the AWS account.
- @PSchmiedmayer I also created a second, more simple, command (`swift run DeploymentTargetAWSLambda delete-iam-roles`). This one goes through the AWS account and deletes _all_ apodini-related IAM roles, regardless of whether the deployment using that role is still active. It should allow you to delete the roles created in the testing AWS account. There's a `--dry-run` option to get a list of all the roles it'd delete, so that you can check it won't touch any unrelated resources.

### Testing
The changes in this PR do not affect the tests

### Reviewer Nudging
Run a deployment w/ the lambda DP so that the AWS account is populated, and then run `swift run DeploymentTargetAWSLambda remove-deployment --api-gateway-api-id ID --keep-api-gateway=true --dry-run` to play around w/ the new sub-command.

---

@Supereg I made some changes to the Task.environment handling to make the code a bit more elegant.


bauer-andreas (approved):
Looks good to me (though with a bit limited understanding of ApodiniDeploy internals)

PSchmiedmayer (approved):
@lukaskollmer The changes look good to me, thanks for the improvements.

Unfortunately we currently have the case that the deployment provider tests for AWS Lambda seem to fail for us in Bamboo and therefore I have disabled them for now except for this branch so we can merge PRs that are unrelated to the AWS deployment.
I have added a condition that checks the branch name and only runs the deployment provider job for this branch. Once the tests work again we can merge this branch and remove the condition from the Bamboo task. I will drop you a message on chat.tum.de about some more details.

PSchmiedmayer:
@lukaskollmer One thing that I noticed while reviewing an other PR (#292): The deployment provides have a few imports of Vapor that are not `@_implementationOnly`. The only problem making it `@_implementationOnly` would be `HandlerInvocation.encodeValue`. It would be good if we can change the function signature in a way that all imports of Vapor at `@_implementationOnly`.

lukaskollmer:
@PSchmiedmayer Regarding the import in `HandlerInvocation.swift`: that one can be made `@_implementationOnly` fairly easy. The 2nd overload of the `encodeValue` function currently is not actually used by anything, meaning that we can simply remove it and thus stop exposing Vapor APIs as part of the public interface.
In fact, we can (and should, I'll update this PR w/ that change) remove the Vapor import entirely from that file.

(Iirc I initially wrote that function back when Vapor was still part of Apodini-proper's exported interface, and then kept it around when we removed Vapor, to allow deployment providers to access the parameter encoding logic, in case a deployment provider needs to encode a parameter it gets into a Vapor ContentContainer. It's not used anywhere atm, and i'd argue that if a deployment provider needs this kind of functionality, it'd be responsible for implementing it itself)

There are some other non-`@_implementationOnly` imports of Vapor in the deployment providers, but I'd say that these are fine, since they're not part of Apodini's public interface (and neither of ApodiniDeploy's). Same way like how the RESTInterfaceExporter also gets away with its Vapor imports.

PSchmiedmayer:
@lukaskollmer I agree, sounds good 👍

PR #294, opened 2021-06-08
Update Package to Swift 5.4

PSchmiedmayer:
# Update Package to Swift 5.4
## :recycle: Current situation & Problem
We do currently only build on Swift 5.3 on Linux and our package versions are based on the Swift 5.3 toolchain.
In addition we are using Bamboo to host the macOS builds as GitHub Action was lacking macOS 11 support until now.

## :bulb: Proposed solution
Update the CI, use macOS 11 build on GitHub Actions and move the minimal Swift version requirement to Swift 5.4. We can rely on all Swift 5.4 features and are better prepared for Swift 5.5.

The Swift 5.4.1 compiler seems to crash when compiling OpenAPIKit. I have created an issue (mattpolzin/OpenAPIKit#224) and demonstrated it using GitHub Actions here: https://github.com/PSchmiedmayer/OpenAPIKit/actions/runs/919770569 .
There are two bugs.swift.org issues tracking the swift compiler crash:
- https://bugs.swift.org/browse/SR-14680
- https://bugs.swift.org/browse/SR-14743

https://bugs.swift.org/browse/SR-14743 documents a possible work around: `swift build -Xswiftc -Xfrontend -Xswiftc -sil-verify-none YOUR-REGULAR-ARGUMENTS`. 

In addition the PR relied on some fixes in https://github.com/vapor-community/vapor-aws-lambda-runtime/pull/18 that prepare the Swift 5.5 support and fixes in the Runtime framework added by @Supereg: https://github.com/wickwirew/Runtime/pull/93

## :gear: Release Notes 
This pull requests introduces breaking changes for users using Swift 5.3.
Package.swift files having Apodini as a dependency and use Swift tools version 5.3 (`// swift-tools-version:5.3`)
will have to update the Swift tools version to Swift 5.4. The Package.swift file must begin like this (or a later version of Swift e.g. 5.5, ...):
```
// swift-tools-version:5.4
```
Apodini Web Services that use the `OpenAPI` interface exporter will have to add the `-Xswiftc -Xfrontend -Xswiftc -sil-verify-none` flags to any `swift run`, `swift build`, or `swift test` commands followed by the regular arguments. E.g. 
```
swift build -c release
```
is changed to
```
swift build -Xswiftc -Xfrontend -Xswiftc -sil-verify-none -c release
```

PR #295, opened 2021-06-11
Update Experimental Async/Await Features to not Break Build on Xcode13

theMomax:
# Update Experimental Async/Await Features to not Break Build on Xcode13

As the Xcode 13 beta ships with Swift 5.5 and $AsyncAwait enabled by default, the previous mechanism for disabling experimental async await functionality in Apodini did no longer function when using Xcode13 beta, even with `experimentalAsyncAwait = false`.

## :bulb: Proposed solution

I added a custom compiler-flag `APODINI_EXPERIMENTAL_ASYNC_AWAIT`.

### Problem that is solved

When working with DocC we need Xcode 13, so that should now be possible.

### Implications
Nothing changes for Xcode 12 /  <Swift 5.5 users. For Xcode 13 users the project compiles with `experimentalAsyncAwait = false`. With `experimentalAsyncAwait = true`, the project does not compile, as I could not find a version of swift-nio's _NIOConcurrency that actually compiles with Xcode13.

I also adapted `Handler.swift` so that it should compile with `experimentalAsyncAwait = true` once _NIOConcurrency does.

### Related PRs
#265 introduced the async/await syntax to `Handler`.

### Testing
I did not add any tests. Functionality for Xcode 12 is covered by existing tests and the rest is experimental.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -7,7 +7,7 @@ import PackageDescription
 /// Configures the Package for usage of the experimental `async`/`await` syntax as introduced by
 /// https://github.com/apple/swift-evolution/blob/main/proposals/0296-async-await.md
 /// When set to `true`, a recent commit from the **main** branch of **swift-nio** is used. Furthermore, the
-/// swift compiler is configured to enable this feature. Swift 5.4 is required for this to work. You may need to reset
+/// swift compiler is configured to enable this feature. Swift 5.5 is required for this to work. You may need to reset
 /// your package caches for this to take effect.
 let experimentalAsyncAwait = false
 
@@ -17,7 +17,8 @@ var apodiniSwiftSettings: [SwiftSetting] {
             .unsafeFlags(
                 [
                     "-Xfrontend",
-                    "-enable-experimental-concurrency"
+                    "-enable-experimental-concurrency",
+                    "-DAPODINI_EXPERIMENTAL_ASYNC_AWAIT"
                 ]
             )
         ]
@@ -85,7 +86,7 @@ let package = Package(
         .package(url: "https://github.com/OpenCombine/OpenCombine.git", .upToNextMinor(from: "0.11.0")),
         // Event-driven network application framework for high performance protocol servers & clients, non-blocking.
         experimentalAsyncAwait
-                    ? .package(url: "https://github.com/apple/swift-nio.git", .revision("4220c7a16a5ee0abb7da150bd3d4444940a20cc2"))
+                    ? .package(url: "https://github.com/apple/swift-nio.git", .revision("67f084365315b8470cd22eb161d855755b3e2748"))
                     : .package(url: "https://github.com/apple/swift-nio.git", from: "2.18.0"),
         // Bindings to OpenSSL-compatible libraries for TLS support in SwiftNIO
         .package(url: "https://github.com/apple/swift-nio-ssl.git", from: "2.8.0"),
--- a/Sources/Apodini/Components/Handler/Handler.swift
+++ b/Sources/Apodini/Components/Handler/Handler.swift
@@ -6,7 +6,7 @@
 //
 
 import NIO
-#if compiler(>=5.4) && $AsyncAwait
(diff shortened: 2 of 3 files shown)


Sources/Apodini/Components/Handler/Handler.swift
```diff
@@ -6,7 +6,7 @@
 //
 
 import NIO
-#if compiler(>=5.4) && $AsyncAwait
+#if compiler(>=5.5) && $APODINI_EXPERIMENTAL_ASYNC_AWAIT
```
bauer-andreas, line 9: Does it need the dollar sign? afaik you just pass a active compilation condition via `-DAPODINI_EXPERIMENTAL_ASYNC_AWAIT` and I never used a dollar sign for those 🙃 so just checking, seems unfamiliar to me.

Package.swift
```diff
@@ -17,7 +17,8 @@ var apodiniSwiftSettings: [SwiftSetting] {
             .unsafeFlags(
                 [
                     "-Xfrontend",
-                    "-enable-experimental-concurrency"
+                    "-enable-experimental-concurrency",
+                    "-DAPODINI_EXPERIMENTAL_ASYNC_AWAIT"
```
bauer-andreas, line 21: Something not entirely related, but maybe something we should discuss as we are seemingly introducing more and more custom Active Compilation Conditions (like in this PR with `APODINI_EXPERIMENTAL_ASYNC_AWAIT`).
In order to optimize build time of our CI suite, we somehow need to match compiler arguments for those test suites which do custom compilation steps (this are ApodiniDeploy and NegativeCompileTests at the moment; See [Note to CI maintainers](https://github.com/Apodini/Apodini/blob/develop/Documentation/Testing/NegativeCompileTests.md#note-to-ci-maintainers)).
My current approach was to manually hardcode those (see [here](https://github.com/Apodini/Apodini/blob/1183e58ebdeba14959ae3840e88cf7853c93f8e1/Tests/NegativeCompileTestsRunner/NegativeTestRunner.swift#L268-L285)) as those shouldn't change that frequently, but I can see where such a change may go unnoticed in the future. Maybe we can find a way to automatically/dynamically grab the compiler arguments passed to the compiler?

Anyways, as we currently don't compile tests with async await enabled this isn't a problem. Though, maybe it makes sense to add support for the `APODINI_EXPERIMENTAL_ASYNC_AWAIT` in the NegativeTestRunner already(?).

bauer-andreas (approved):
Seems good to me 👍  seems like you finally got Xcode 13 running 🚀

bauer-andreas:
Could we stop using auto merge, maybe even disallow it globally?

theMomax:
> Could we stop using auto merge, maybe even disallow it globally?

Sorry, I'm definitely guilty on that one...didn't expect in-depth feedback on this one.😅 Definitely add a discussion point for next meeting!

bauer-andreas:
Expect the unexpected 😎

PSchmiedmayer:
Regarding auto merge: Yes, we can disable the feature. I would suggest to add it to the meeting agenda for the next meeting.

In addition I think that we should replace the `#if compiler` statements with `#if swift` statements. Compiling code with the Swift 5.5 compiler in Swift 4.X mode would try to compile the async await code which fails as the feature is not backwards compatible.

A more general comment: I think we should either do the following:
1. Remove the feature flag, compilation condition, and make async await the default behavior if you compile on Swift 5.5. This includes:
   - Make sure that the package actually compiles in Swift 5.5 (I get compile errors if I set `experimentalAsyncAwait` to true
   - All unit test complete successfully on Swift 5.5 with the default behavior of async await being available
   - All this would require additional work on that front and probably some more `#if swift(>=5.5)` statements.
2. Remove async await for now until we do 1. at a later point.

theMomax:
@PSchmiedmayer I don't see how we could get it working without using a fork of swift-nio. Thus I'd suggest we follow option two. The feature will be removed with #294. I'll probably give it another try in a month or so, maybe the situation will be more stable then.

PR #298, opened 2021-06-15
Add the ability to parse subcommands

PSchmiedmayer:
# Add the ability to parse subcommands

We currently can't add subcommands to a web service

## :bulb: Proposed solution
Add the ability to add subcommands to a web service.
Currently the following approach is possible: Add a static `subcommand` property to a `WebService`:
```swift
static var subcommands: [ParsableCommand.Type] {
    [Test.self]
}
```

It would be desirable if Configurations could also pass their `subcommands` directly to the `WebService` but this would require us to create an instance of a `WebService` using Runtime which currently is blocked by the issue documented in: https://github.com/wickwirew/Runtime/pull/91 . If this is solved we can create an instance of a web service using runtime and query static properties of the `Configuration` instances to get all subcommands for all configurations. 

### Problem that is solved
Problem described here: https://github.com/Apodini/Apodini/pull/292#discussion_r648257873

### Testing

No unit tests as this is just a demo. You can try it out using the ApodiniDeployTestWebService.
You can e.g. pass arguments to it using Xcode: 
Screen Shot 2021-06-15 at 8 01 22 PM

The help command prints out the two subcommands as expected:
```
USAGE: start <subcommand>

OPTIONS:
  -h, --help              Show help information.

SUBCOMMANDS:
  test
  web-service (default)

  See 'start help <subcommand>' for detailed help.
```
The test subcommand works as expected.

### Related PRs
#292


theMomax:
@PSchmiedmayer @philippzagar Just a heads up...there is a PR open on swift argument parser that could make this obsolete: https://github.com/apple/swift-argument-parser/pull/319 (may also be relevant for #292 )

PSchmiedmayer:
@theMomax Thanks for sharing that link, nice addition to the Swift Argument Parser and a good find!
I looked over the code and it only seems to find subcommands in types that are nested which won't be the case for e.g. `Configuration`s or deployment providers which are defined in different modules or files so it doesn't address the problem we face here at the moment in automatically detecting subcommands that are defined outside of the scope of the type implementing `WebService`.

PR #299, opened 2021-06-16
Remove Relationship Code from `SemanticModelBuilder`

theMomax:
# Remove Relationship Code from `SemanticModelBuilder`

#276 introduced the `Blackboard` pattern to the `SemanticModelBuilder` and exporter API. In the same step, `WebServiceModel` was deprecated. `WebServiceModel`, nevertheless, was still an integral part of the exporter API and fully built in the `SemanticModelBuilder`. `WebServiceModel` originally only provided insight into the structure of a web service, enabling exporters to explore the global context of single endpoints. #276 introduced `WebServiceRoot`/`WebServiceComponent`, which is a lazy, `Blackboard`-based structure that provides the same functionality. However, the `WebServiceModel` is tightly integrated with `RelationshipBuilder` and `TypeIndexBuilder`. Together, they used to build up relationship-related information on the `Endpoint` struct. In order to maintain this functionality, `WebServiceModel` and the builders were kept as a core part of the `SemanticModelBuilder` in #276.

Furthermore, all exporters used to work on a `EnrichedContent` type as a response from a `Handler`'s evaluation. This type packed the actual type-erased `Response` as well as additional relationship-information. This information depended on the `Endpoint`'s relationship-information as well as the request's parameter-values. Most exporters don't make use of this extra information.

## :bulb: Proposed solution
All relationship-information is removed from `Endpoint`. This data is now stored on a separate structure `RelationshipEndpoint`. `WebServiceModel` was renamed `RelationshipWebServiceModel`. `RelationshipWebServiceModel` works on `RelationshipEndpoint`s now. A new `WebServiceModel` was introduced, which is only a lightweight wrapper around a global `Blackboard`.

`RelationshipWebServiceModel` is now solely available as a global `KnowledgeSource`. One can also directly access the according `AnyRelationshipEndpoint` on a local `Blackboard`.

`ConnectionContext` no longer returns an `EnrichedContent`, but the typed `Response` of the `Handler`. This also means `ConnectionContext` now has an additional generic parameter - the `Handler` the connection evaluates. Exporters that don't want this extra information can use `AnyConnectionContext` instead. The request's parameters required for the `RESTInterfaceExporter` to build the `EnrichedContent` are directly returned by the `ConnectionContext` along with the `Response` in a new alternative to `.handle()`. This is not elegant, but it works for now. A future PR of mine will remodel the entire exporter API and present a more elegant solution for this problem.

### Problem that is solved
Building the `WebServiceModel` was a rather expensive startup-operation, which, now,  is only performed if an exporter is configured that makes use of the `RelationshipWebServiceModel`-related `KnowledgeSource`s. That is only the case for `RESTInterfaceExporter` and `OpenAPIInterfaceExporter`.

All exporters other than `RESTInterfaceExporter` have less runtime-overhead as they don't have to build `EnrichedContent` anymore.

#### API

I had to adapt the exporters a bit to work with the typed `Response` and `ConnectionContext`. Furthermore, I had to adapt `RESTInterfaceExporter` and `OpenAPIInterfaceExporter` to use the new `KnowledgeSource`s instead of the `WebServiceModel` provided by the `SemanticModelBuilder`.

#### Runtime Behavior

This PR changes the runtime-behavior of Apodini in two ways.

Firstly, Apodini no longer has a structured view on a web service's endpoints. Instead, they are just treated as a list. This also means the order in which the endpoints are exported might change. Previously, endpoints were exported recursively starting from the root. Now, they are exported in the same order as they are parsed by the `SyntaxTreeVisitor`. We didn't explicitly guarantee the order in which endpoints were exported, but that doesn't mean no-one relied on it. If necessary, an exporter can use `WebServiceRoot` in `finishedExporting(_:)` to export all endpoints in a structured manner.

Secondly, Apodini used to have an "autocomplete" feature for path parameters. That is, if you were to define a `Handler` with one `@Parameter(.http(.path)) var id: UUID` on the `WebService`'s root, Apodini would automatically add this path-parameter to the endpoint's path. In this example, the path of the endpoint would be `v1/{id}`. This mechanism was part of the old `WebServiceModel` and thus applied to all exporters equally, even though it was an `.http()` modifier. This PR restricts this mechanism to those exporters that explicitly use `EndpointPathComponentsWithHTTPParameterOptions` `KnowledgeSource` to build their path. I configured `RESTInterfaceExporter` and `OpenAPIInterfaceExporter` to do so.

### Related PRs
#276

### Testing
There are no new tests. I had to adapt some tests which manually used the old `WebServiceModel` to obtain an `Endpoint` from a `Handler`.

### Reviewer Nudging
Take a look at what information remains in `Endpoint` and what I moved to `RelationshipEndpoint`. Compare the old version of `SemanticModelBuilder` and the new `RelationshipKnowledge.swift` to check that those `KnowledgeSource`s should work as intended. Then, take a quick look at the changes in `ConnectionContext.swift` and `RequestHandler.swift`. Finally, check out `PathKnowledge.swift` to see the difference between path-parameter "autocomplete" and the new default "manual" behavior.


Sources/Apodini/Semantic Model Builder/InterfaceExporter/ConnectionContext.swift
```diff
@@ -160,15 +223,73 @@ class EndpointSpecificConnectionContext<I: InterfaceExporter, H: Handler>: Conne
         }
     }
 
-    override func register<Listener: ObservedListener>(listener: Listener) where Listener.Exporter == I {
+    /// Register a listener that will be notified once an observed object did change in the handler
+    /// that is being used in this connection.
+    public func register<Listener: ObservedListener>(listener: Listener) {
         // register the given listener for notifications on the handler
         for object in collectObservedObjects(from: endpoint.handler) {
             self.observations.append(object.register { [weak self] triggerEvent in
```
eldcn: There is no need to weakify `self` since it is not getting used in the closure anymore

Sources/Apodini/Semantic Model Builder/KnowledgeSources/LoggerKnowledge.swift
```diff
@@ -0,0 +1,17 @@
...
+//  LoggerKnowledge.swift
+//  
+//
+//  Created by Max Obermeier on 14.06.21.
+//
+
+import Foundation
+import Logging
+
+extension Logger: KnowledgeSource {
```
eldcn, line 11: 👍

Sources/Apodini/Semantic Model Builder/Model/Endpoint.swift
```diff
@@ -3,34 +3,18 @@
...
-    func relationships(for operation: Operation) -> Set<RelationshipDestination>
-
-    /// Returns the special "self" Relationship for all `Operation`s.
-    func selfRelationships() -> Set<RelationshipDestination>
-
-    /// Returns the special "self" Relationship for a given `Operation`
-    /// - Parameter for: The `Operation` for the desired destination.
-    func selfRelationship(for: Operation) -> RelationshipDestination?
+/// A ``ParameterCollection`` provides random access to an endpoint's ``EndpointParameter``s.
+public protocol ParameterCollection {
```
eldcn, line 7: Looking at the name of the protocol `ParameterCollection`, one would expect to have a parameters collection requirement. I would suggest to adjust it with a `var parameters: [AnyEndpointParameter] { get }` and provide a default implementation for `findParameter(for:)`

Sources/Apodini/Semantic Model Builder/Model/EndpointReference.swift
```diff
@@ -22,16 +22,16 @@ struct EndpointReference: CustomStringConvertible, CustomDebugStringConvertible
     /// Holds the response type of the `Endpoint`
     let responseType: Any.Type
 
-    init<H: Handler>(on node: EndpointsTreeNode, off endpoint: Endpoint<H>) {
+    init<H: Handler>(on node: EndpointsTreeNode, off endpoint: RelationshipEndpoint<H>) {
```
eldcn: typo -> `of endpoint`, not yours but since you are there 😅

Sources/Apodini/Semantic Model Builder/PathModelBuilder.swift
```diff
@@ -71,7 +71,7 @@ private struct PathComponentElementParser: PathComponentParser {
 
 // MARK: Helpers
 
-private extension Parameter {
```
eldcn: I think this extension can remain private since it is only getting used in the scope of this file

Sources/Apodini/Semantic Model Builder/SemanticModelBuilder.swift
```diff
@@ -70,78 +59,16 @@ class SemanticModelBuilder: InterfaceExporterVisitor {
...
-            self.typeIndexBuilder.indexContentType(content: content,
-                                                   reference: reference,
-                                                   markedDefault: markedDefault,
-                                                   pathParameters: pathParameters,
-                                                   operation: operation)
-        }
+        // We first only build the blackboards and the `Endpoint`. The validation and exporting is done at the
+        // beginning of `finishedRegistration`. This way `.global` `KnowledgeSource`s get a complete view of
+        // the web service even when accessed from an `Endpoint`.
+        endpointsToExport.append(Endpoint(
```
eldcn: I really like how the `SemanticModelBuilder` has trimmed down now. Great job with the `Blackboard` pattern

Sources/Apodini/Semantic Model Builder/SemanticModelBuilder.swift
```diff
@@ -14,22 +14,11 @@ class SemanticModelBuilder: InterfaceExporterVisitor {
...
     private let allowLenientParameterNamespaces = true
-
-    let webService: WebServiceModel
-    let rootNode: EndpointsTreeNode
-
-    var relationshipBuilder: RelationshipBuilder
-    var typeIndexBuilder: TypeIndexBuilder
     
-    private var onRegistrationDone: [() -> Void] = []
+    var endpointsToExport: [_AnyEndpoint] = []
```
eldcn: perhaps `parsed-` or `collectedEndpoints`?

eldcn (commented):
Looks good to me, well done 👍  You will have to resolve merge conflicts since develop changed yesterday

eldcn (approved):

Sources/Apodini/Semantic Model Builder/SemanticModelBuilder.swift
theMomax: Yeah, sounds better, thanks!

Sources/Apodini/Semantic Model Builder/SemanticModelBuilder.swift
theMomax: Thanks!

Sources/Apodini/Semantic Model Builder/InterfaceExporter/ConnectionContext.swift
theMomax: Thanks, fixed it!

Sources/Apodini/Semantic Model Builder/Model/Endpoint.swift
theMomax: Seems reasonable👍🏻

PR #300, opened 2021-06-16
Binary Large Objects

PSchmiedmayer:
# Binary Large Objects

An Apodini web service can currently not return a **b**inary **l**arge **ob**ject (**blob**) directly. Use cases for this would be to return an HTML page that is displayed in a browser, offer `Handler`s that return files or return images from a `Handler`.
Apodini needs a convenient way to return these blobs from a `Handler` and attach e.g. a MIME type to the blob so it can be correctly interpreted by a client.

## :bulb: Proposed solution

This pull request introduces a `Blob` type that allows a developer to return a `blob` including a MIME type.
The PR includes a common set of MIME types as listed on https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types .

### Implications

The `Blob` is an additive change. It has implications on the existing and future exporters. In addition to handling the special `Empty` type exporters should now also consider handling the `Blob` type. The `Blob` type includes a default `Encodable` implementation that enables Exporters that don't include code to specially handle the `Blob` type to continue working.

### Related Packages
Can be used by the new [ApodiniLeaf](https://github.com/Apodini/ApodiniLeaf) package. 

### Testing
Includes test to cover responses including a `Blob` and testing the RESTExporter as well as the OpenAPI Exporter

## 🛠 Migration Guide

Binary Large Objects (`Blob`) allow you to return binary files from a `Handler` that are not encoded by exporters. This can be used to return files or return already encoded files or binary data such as HTML pages or any other content that should be displayed by a browser. A `Blob` can be initialized using a `Data` or `ByteBuffer` instance as well as a `MimeType` that indicates the file type expressed with the `Blob`.

The following example shows a `Handler` that returns a `Blob`:
```swift
struct BlobResponseHandler: Handler {
    func handle() -> Blob {
        let data: Data = // ...
        let mimeType: MimeType = .image(.gif)
        return Blob(data, type: mimeType)
    }
}
```
You can get an overview of all `MimeTypes` that are implemented in Apodini in [MimeType.swift](https://github.com/Apodini/Apodini/blob/develop/Sources/Apodini/Response/MimeType.swift)

### DIFF
--- a/Sources/Apodini/Application/Application.swift
+++ b/Sources/Apodini/Application/Application.swift
@@ -62,6 +62,22 @@ extension Application {
 
 /// Configuration and state of the application
 public final class Application {
+    private static var latestApplicationLogger: Logger?
+    /// A global loggerr that can be used when no  `Application` instance is available.
+    ///
+    /// - Note: In `Handler`s you shpuld rely on the `Logger` injected in the `@Environment`.
+    public static var logger: Logger {
+        latestApplicationLogger ?? {
+            var newLogger = Logger(label: "Pre-startup")
+            #if DEBUG
+            newLogger.logLevel = .debug
+            #endif
+            latestApplicationLogger = newLogger
+            return newLogger
+        }()
+    }
+    
+    
     /// Decides how EventLoopGroups are created
     public let eventLoopGroupProvider: EventLoopGroupProvider
     /// The EventLoopGroup for the application
@@ -142,10 +158,13 @@ public final class Application {
         }
         self.locks = .init()
         self.didShutdown = false
+        
         self.logger = .init(label: "org.apodini.application")
         #if DEBUG
         self.logger.logLevel = .debug
         #endif
+        Application.latestApplicationLogger = self.logger
+        
         self.storage = .init(logger: self.logger)
         self.lifecycle = .init()
         self.isBooted = false
--- a/Sources/Apodini/Application/Application.swift
+++ b/Sources/Apodini/Application/Application.swift
@@ -63,7 +63,7 @@ extension Application {
 /// Configuration and state of the application
 public final class Application {
     private static var latestApplicationLogger: Logger?
-    /// A global loggerr that can be used when no  `Application` instance is available.
+    /// A global logger that can be used when no  `Application` instance is available.
     ///
     /// - Note: In `Handler`s you shpuld rely on the `Logger` injected in the `@Environment`.
     public static var logger: Logger {
--- a/Sources/Apodini/Response/Blob.swift
+++ b/Sources/Apodini/Response/Blob.swift
@@ -28,8 +28,7 @@ public struct Blob: Encodable, ResponseTransformable {
     ///   - data: The `Data` representation of the `Blob`
     ///   - type: The MIME type associated with the `Blob`
     public init(_ data: Data, type: MimeType? = nil) {
-        self.byteBuffer = ByteBuffer(data: data)
-        self.type = type
+        self.init(ByteBuffer(data: data), type: type)
     }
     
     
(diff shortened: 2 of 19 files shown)


Sources/ApodiniVaporSupport/HTTPStatus+Status.swift
```diff
@@ -0,0 +1,34 @@
+//
+//  File.swift
```
theMomax: Wrong filename…

Tests/ApodiniTests/ResponseTests/BlobTests.swift
```diff
@@ -0,0 +1,102 @@
+//
+//  File.swift
```
theMomax: Wrong filename…

Sources/Apodini/Response/Blob.swift
```diff
@@ -0,0 +1,57 @@
...
+    public init(_ byteBuffer: ByteBuffer, type: MimeType? = nil) {
+        self.byteBuffer = byteBuffer
+        self.type = type
+    }
+    
+    
+    public func encode(to encoder: Encoder) throws {
+        print(
+            """
+            Information: The used Exporter currently doesn't support Blob's as a content of an Apodini Reponse.
```
theMomax, line 47: Maybe we should wrap this in a if DEBUG? Otherwise this could cause quite some spam on production systems, especially since it is not explicitly _wrong_ for an exporter to use this strategy. It would also be great if we could use the Application’s Logger here…should we provide a static property on Application, which always references rhe last Application that was initialized?

Sources/Apodini/Response/Empty.swift
```diff
@@ -0,0 +1,11 @@
+//
+//  Empfy.swift
+//  
+//
+//  Created by Paul Schmiedmayer on 5/26/21.
+//
+
+
+public struct Empty: Encodable {
```
theMomax, line 9: Is there a reason why we don’t use Void for this?

theMomax (commented):
I just added some comments…looks quite good so far. One suggestion: Currently, middleware developers are not confronted with the Blob type; they must know it exists and then explicitly detect the case. We could add specialized methods to the InterfaceExporter protocol along with a phantom type for choosing the implementation strategy. For ApodiniError we employ a similar strategy with StandardErrorCompliantExporter. (Same applies for Empty if course)

Sources/Apodini/Response/Empty.swift
PSchmiedmayer: Yes, we have considered `Void` back in the days when we introduced `Empty` but you can not extend `Void` to e.g. conform to `Encodable`:
```swift
extension Void: Encodable {}
```
Results in the following compiler error: Non-nominal type 'Void' cannot be extended

Sources/Apodini/Response/Blob.swift
PSchmiedmayer: Good idea, I have added a global logger as a static property of the `Application`

PSchmiedmayer:
@theMomax Thanks for your feedback! I agree that the Exporter interface should reflect the mapping of `Empty` as well as `Blob` and the `ApoiniError`. I would suggest that a new "Version 2" Exporter interface requires the exporters to handle all these cases, this would somehow be a breaking change but would set this on a solid foundation.

Sources/Apodini/Application/Application.swift
```diff
@@ -62,6 +62,22 @@ extension Application {
 
 /// Configuration and state of the application
 public final class Application {
+    private static var latestApplicationLogger: Logger?
+    /// A global loggerr that can be used when no  `Application` instance is available.
```
theMomax: typo

theMomax (approved):
Looks good! If you don't want to add an extra add-on `InterfaceExporter` protocol with this PR, I can also set this on my TODO list for the exporter API refactoring👍🏻

PSchmiedmayer:
@theMomax Yes, I would suggest to tackle this problem once we rework the exporter API 👍

PR #301, opened 2021-06-16
Information for Requests and Responses

PSchmiedmayer:
# Information for Requests and Responses

Interacting with a browser to e.g. set cookies requires a developer to return more information in a response. This information can e.g. be mapped to HTTP Headers for HTTP based exporters or can be encoded as part of the response in non HTTP based exporters.

HTTP Headers or other information is currently passed in to a Handler using `@Parameters`. This injection mechanism can sometimes feel burdensome and does not really fit well in the other `@Parameter` types that we have at the moment. It would be ideal if additional information e.g. stored in HTTP headers is also attached to the connection context similarly to how information is attached to a `Response`.

## :bulb: Proposed solution

In addition the `Response` type now includes the possibility to pass in additional `information` when creating a `Request`. The `information` is consisting of a set of predefined `InformationKey`s and can be extended with custom keys. We orient ourselves on a common set of HTTP Header types, the most common concept the Information is mapped to: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers

This PR makes `Information` that is passed in to an available using the `ConnectionContext` found in the environment. Using the `ConnectionContext` information can be accessed and is automatically decoded and provided to the developer implementing a `Handler`.

### Implications

The restructuring related to the `Information` is a breaking change. Existing usages of the `@Parameter(.http(.header))` property wrappers has to be replaced with accessing the `ConnectionContext` using the `@Environment` property wrapper and accessing the `ConnectionContext.information` property.

### Related Packages
Can be used by the new [ApodiniLeaf](https://github.com/Apodini/ApodiniLeaf) package. 

### Testing
Includes test to cover the `Information` type, `InformationKey`s, and the interaction with Requests and Responses.

## 🛠 Migration Information

Apodini now includes the functionality to easily access context information from incoming requests in `Handler`s and pass information to a `Response`. `Information` is mapped to the corresponding wire format for the different exporters, an REST Exporter e.g. maps `Information` instances to HTTP Header fields.

  The `@Parameter` option `.http(.header)` is removed and replaced with the `.information` property of the `Connection` of type `Set<Information>`.

  The following code:
  ```swift
  struct InformationHandler: Handler {
      @Parameter("Authorization",.http(.header))
      var authorization: String

      func handle() -> String {
          // Parse the basic authorization information from the Header String ...
          // Check if Paul is signed in ...
          return "Hello Paul 👋"
      }
  }
  ```
  
  Can be migrated to:
  
   ```swift
   struct _InformationHandler: Handler {
       @Apodini.Environment(\.connection) var connection: Connection
    
    
       func handle() -> String {
           guard let basicAuth = connection.information[Authorization.self]?.basic else {
               // Throw error ...
           }
        
           let username = basicAuth.username
           let password = basicAuth.password
        
           // Check if Paul is signed in ...
           return "Hello Paul 👋"
       }
   }
   ```

   The example shows how `Information` can be accessed in a type safe way using the Information types implemented in Apodini. Custom information can be accessed using a string based index, e.g. `let rawValue =  connection.information["MyInformation"]`

  The following example shows how `information` can be passed to a `Response` instance in a `Handler:`
  
  ```swift
  struct InformationHandler: Handler {
      func handle() throws -> Response<Int> {
          .send(42, status: .ok, information: [
             .authorization(.basic(username: "PaulSchmiedmayer", password: "SuperSecretPassword")),
             .cookies(["test": "value"]),
             .etag("ABCD", isWeak: true),
             .expires(Date(timeIntervalSince1970: 1623843720)),
             .redirectTo(try XCTUnwrap(URL(string: "https://ase.in.tum.de/schmiedmayer"))),
             .custom(key: "Test", rawValue: "ATest")
         ])
      }
  }
  ```

  The example above details all implemented `Information` types. The `.custom(key: String, rawValue: String)` `Information` instance can be used to pass any information to a `Response`.


Sources/Apodini/Information/Set+Information.swift
```diff
@@ -0,0 +1,40 @@
+//
+//  File.swift
```
philippzagar: Rename to "Set+Information.swift"

Sources/Apodini/Response/Response.swift
```diff
@@ -18,32 +18,33 @@ public struct Response<Content: Encodable>: ResponseTransformable {
         Response<Content>(connectionEffect: .close)
     }
     
-    
-    public static func send(_ content: Self.Content, status: Status? = nil) -> Response<Self.Content> {
-        Response<Self.Content>(status: status, content: content, connectionEffect: .open)
+    public static func send(_ content: Self.Content, status: Status? = nil, information: Set<AnyInformation> = []) -> Response<Content> {
```
philippzagar: You could remove the Self from Self.Content here as well to stay consistent with the other Content usages. Same for line 29

Sources/Apodini/Semantic Model Builder/InterfaceExporter/ValidatingRequest.swift
```diff
@@ -29,22 +29,26 @@ struct ValidatingRequest<I: InterfaceExporter, H: Handler>: Request {
     let endpointValidator: EndpointValidator<I, H>
     let storedEndpoint: Endpoint<H>
     let eventLoop: EventLoop
-    let remoteAddress: SocketAddress?
+    
+    var remoteAddress: SocketAddress? {
+        exporterRequest.remoteAddress
+    }
+    var information: Set<AnyInformation> {
```
philippzagar, line 36: Nice! I will do something similar for the Logger implementation 👍

Sources/Apodini/Information/Authorization.swift
```diff
@@ -0,0 +1,123 @@
...
+    
+    public init(_ value: Value) {
+        self.value = value
+    }
+}
+
+
+extension AnyInformation {
+    /// An `Information` carrying authorization information
+    /// - Parameter authorization: /// The content of an `Authorization` `Information`
```
philippzagar: Remove the second '///'

Sources/Apodini/Information/Authorization.swift
```diff
@@ -0,0 +1,123 @@
...
+        "Authorization"
+    }
+    
+    
+    public private(set) var value: Value
+    
+    
+    public var rawValue: String {
+        value.rawValue
+    }
```
philippzagar: I guess I would prefer to always have these properties at the top of the struct, even if a custom Value is defined (it's clearer and more consistent with the other Information structs). But I guess that's a matter of taste. Would also apply to eg. `ETag.swift`

Package.swift
```diff
@@ -241,6 +241,7 @@ let package = Package(
             dependencies: [
                 .target(name: "Apodini"),
                 .target(name: "ApodiniVaporSupport"),
+                .target(name: "ApodiniREST"),
```
philippzagar, line 244: This dependency is also introduced with my renewed ExporterConfiguration, therefore fits perfectly fine 👍

philippzagar (approved):
Really nice feature addition to the RESTful exporter, just some minor changes left

Sources/Apodini/Information/Authorization.swift
PSchmiedmayer: I am trying to have a consistent style in the project that is as followed:
- Nested Types
- Static Properties (First Stored then Computed)
- Instance Properties (First Stored then Computed)
- Subsctipts
- Initializer
- Static Functions
- Instance Functions

But I agree that the long nested type makes it a bit hard to read. I have added it below the main type in an extension, this might make it a bit clearer how the code is structured.
In the long term we could enforce this file order rule using SwiftLint, We should probably take a look at that.

PSchmiedmayer:
@philippzagar Thanks for the feedback, I have incorporated it all in the latest commit 💪

PR #302, opened 2021-06-17
Fix Propagation of Delegate Optionality

theMomax:
# Fix Propagation of Delegate Optionality
Currently, a non-optional Delegate, which delegates to an optional Delegate would override the `Optionality` to `.required` even though logically it should be `.optional`.

E.g. in the test example, `Optionality` would have been `.required`:

```swift
struct OptionalDelegate {
      @Parameter var name: String
}

struct RequiredDelegatingDelegate {
      let delegate = Delegate(OptionalDelegate())
}

struct SomeHandler: Handler {
      let delegate = Delegate(RequiredDelegatingDelegate(), .required)

      func handle() throws -> some ResponseTransformable {
          try delegate().delegate().name
      }
}
```

## :bulb: Proposed solution
`Delegate` now pays attention to the parameter's own `Optionality` (which may have been set by a child-`Delegate`).

### Problem that is solved
Behavior should now be as expected.

### Testing
I added two test cases, that cover both sides of the problem.

### Reviewer Nudging
Should be quick...


theMomax:
@PSchmiedmayer could you merge this PR with your admin powers once it is approved? There is only one out of three lines not covered in the patch. We currently don't need mutating access on `AnyParameter`, thus this line is not covered...

bauer-andreas (approved):
Nothing to add. Good catch 👍🏼

philippzagar (approved):
LGTM!

PR #303, opened 2021-06-21
Rework ContextNode infrastructure to accommodate for extended requirements

bauer-andreas:
# Rework ContextNode infrastructure to accommodate for extended requirements

The current implementation of the `ContextNode` provided only limited flexibility for storing context values.
Its rather simple design was specifically engineered for the early requirements of the Apodini DSL.

The Metadata DSL extended the requirements expected from the `ContextNode` functionality.
It has shown that the current systems simply can't deal with the following two scenarios:
- Context Values added within the Handler (such additions to the context would be treated as Modifiers and be therefore be stored in reverse order)
- Modifiers which add multiple values of the same `ContextKey` (the system was built with the assumption that every context key is only added once per Modifier. There assuming one context value per modifier)

Below is a example declaration of a `Handler` which would break the current system: 

```swift
struct TestHandler: Handler {
    // ...
    var metadata: Metadata {
        Example(1)
        Example(2)
    }
}

TestHandler()
    .metadata(Example(3))
    .metadata {
        Example(4)
        Example(5)
    }

// expected order is: 1, 2, 3, 4, 5
// currently it would result in: 2, 1, 3, 5, 4
```

## :bulb: Proposed solution

The current `ContextNode` shall be reengineered such that it is capable of preserving the correct order.  

## :gear: Release Notes 

This PR doesn't introduce any user facing changes.

However it introduces several breaking changes for Apodini Framework Developers in regards to how Modifiers are created:

- **ContextKey**: Previously, the `defaultValue` of a ContextKey was always passed as the initial value to the `ContextKey.reduce(...)` function. With this PR, the `defaultValue` will not anymore be passed to the `reduce` function,
 and is only used inside the `Context.get(valueFor:)` method when no value is found. It is extremely unlikely that your implementations depends on that, but should it you need to adjust your implementation.
- **ConteytKey**: The signature of the `reduce` function was changed. The second argument `nextValue` is not passed as a closure anymore! Change your signature from `static func reduce(value: inout Type, nextValue: () -> Type)` to `static func reduce(value: inout Type, nextValue: Type)`.
- **Modifier**: In order to distinguish parsing steps of multiple `Modifier`s, additionally logic has to be added to the `SyntaxTreeVisitable.accept(_:)` method to signal the start of a new modifier to the `ContextNode`. This is done fully automatically by providing a default implementation for the `SyntaxTreeVisitable.accept(_:)` for every `Modifier`,
but requires that preexisting Modifiers remove their `SyntaxTreeVisitable` conformance and implementation, and instead replace it with the implementation of the new `Modifier.parseModifier(_:)` method, called from the default implementation, used to make any calls to `SyntaxTreeVisitor.addContext(...)`. Below is a detailed code snippet illustrating the required changes 

<details>
  <summary>Migrating existing Modifier Definitions. Click to expand!</summary>

```swift
public struct OperationModifier<H: Handler>: HandlerModifier {
    public let component: H
    let operation: Operation

    init(_ component: H, operation: Operation) {
        self.component = component
        self.operation = operation
    }
}

// MARK: SyntaxTreeVisitable
extension OperationModifier: SyntaxTreeVisitable {
    public func accept(_ visitor: SyntaxTreeVisitor) {
        visitor.addContext(OperationContextKey.self, value: operation, scope: .current)
        component.accept(visitor)
    }
}
```

In order to migrate you need to remove the conformance to `SyntaxTreeVisitable` and implement the new 
`parseModifier(_:)` function. With that you can (and MUST) omit the `Component.accept()` call on the second line of the accept method.

After the migration the modifier looks as follows:
```swift
public struct OperationModifier<H: Handler>: HandlerModifier {
    public let component: H
    let operation: Operation

    init(_ component: H, operation: Operation) {
        self.component = component
        self.operation = operation
    }

    public func parseModifier(_ visitor: SyntaxTreeVisitor) {
        visitor.addContext(OperationContextKey.self, value: operation, scope: .current)
    }
}
```
</details>

### Problem that is solved

The results of parsing Metadata and Modifiers is now arranged as expected.
Delegating `Handler`s have higher precedence than delegated `Handler`s, Metadata Blocks are parsed in order,
(Metadata) `Modifier` have higher precedence than context values declared inside a `Handler`,
and the order is properly preserved when parsing `Modifier`s which add multiple values for the same `ContextKey`.

### Implications

This PR introduces the breaking changers as illustrated above.

Besides the reworked `ContextNode` which properly reflects ordering in all scenarios, this PR has some additional neat features:
* For `OptionalContextKey`s and `ContextKey`s with an `Array` based value type, a new default implementation for the `reduce(...)` function was added concatenating all the context values. This was the typical logic for array based `ContextKey`s. Therefore this changes reduces the effort creating such ContextKeys and reduces some code duplication for such preexisting `ContextKey`s.
* The reduction step of Context values is now done at the end, once parsing is finished. This change was required in order to properly preserve ordering in all scenarios. It may allow for the user to specify custom reduction ordering if wanted.

### Related PRs
- #280 Introduction of the Metadata DSL

### Testing

Most of the PR changes were already covered with preexisting tests. The Metadata tests were adapted to now expect the correct ordering.
Further I added two new test cases testing the new default implementation of the `ContextKey.reduce(...)` function (specifically for array based ContextKeys).

### Reviewer Nudging

You may want to make yourself aware of the requirements set against the `ContextNode` system by reading through the inline docs of `ContextNode`. I highlighted some of the problematic parsing scenarios.
This may provide you the necessary background to understand the reasoning behind the infrastructure changes.


Sources/Apodini/SyntaxTreeVisitor/Context/ContextEntryCollection.swift
```diff
@@ -0,0 +1,55 @@
...
+
+/// A type erased version of `ContextEntryCollection`.
+protocol AnyContextEntryCollection {
+    /// Append a `AnyContextEntry` to the collection.
+    /// - Parameters:
+    ///   - entry: The `AnyContextEntry` to be added.
+    ///   - derivedFromModifier: Defines if the supplied `AnyContextEntry`
+    func add(entry: AnyContextEntry, derivedFromModifier: Bool)
+
+    /// as
```
bauer-andreas: Missing docs

bauer-andreas: fixed in 11fb8fd

Sources/Apodini/Metadata/Extensions/Context+Metadata.swift
```diff
@@ -0,0 +1,21 @@
...
+    /// Retrieves the value for a given `MetadataDefinition`.
+    /// - Parameter contextKey: The `MetadataDefinition` to retrieve the value for.
+    /// - Returns: Returns the stored value or `nil` if it does not exist on the given `Context`.
+    func get<Metadata: MetadataDefinition>(valueFor metadata: Metadata.Type = Metadata.self) -> Metadata.Key.Value? {
+        get(valueFor: metadata.Key.self)
+    }
+
+    /// Retrieves the value for a given `MetadataDefinition`.
+    /// - Parameter contextKey: The `MetadataDefinition` to retrieve the value for.
+    /// - Returns: Returns the stored value or `nil` if it does not exist on the given `Context`.
```
philippzagar: Adapt comment

philippzagar (approved):
LGTM!
A more general question: Why differentiate between overriding a normal value completely and overriding an array via appending to the origin array? Maybe it's just me but the automatic appending seems kind of unintuitive to me and may lead to unexpected behavior from the user's perspective. Are there actual use cases for that?

bauer-andreas:
> Why differentiate between overriding a normal value completely and overriding an array via appending to the origin array?

Well, looking at our existing `ContextKey`s **all** of them which are array based collect all the encountered values instead of overriding them. In my opinion it was actually the more intuitive way. Cause typically when doing array based ContextKeys you want to exploit the "inheriting" part of it (like Handler inherits ContextKey values from the upper Component tree, or multiple Metadata declarations get reduced into a single array).
Consequentially, doing so resulted in some reduced code duplication.

But yeah, if you guys feel it is more unintuitive that way, I have no strong opinion (other than that I spent way to long figuring out on how to best incorporate a second default implementation 🙃) about that.

Sources/Apodini/Metadata/Extensions/Context+Metadata.swift
bauer-andreas: Great catch 👍🏼

philippzagar:
> But yeah, if you guys feel it is more unintuitive that way, I have no strong opinion (other than that I spent way to long figuring out on how to best incorporate a second default implementation 🙃) about that.

In this aspect I trust your judgment definitely more than my own, since today was actually the first time I took a closer look at how modifiers work in Apodini 😂
If @hendesi has a similar opinion on this subject (like me), we can maybe revisit the topic in the Apodni group meeting. Otherwise, the PR would have my full support 👍

bauer-andreas:
> If @hendesi has a similar opinion on this subject (like me), we can maybe revisit the topic in the Apodni group meeting. Otherwise, the PR would have my full support 👍

Would love to get the pr merged soon, as my next works are already in the pipeline and kinda depend on it. But yeah, I see that this topic might be discuss-worthy.

Sources/Apodini/Metadata/Extensions/Context+Metadata.swift
bauer-andreas:
```suggestion
    /// - Returns: Returns the stored value or the `ContextKey.defaultValue` if it does not exist on the given `Context`.
```

workaround to not stash all my unprepared changes 🙃

philippzagar:
> > If @hendesi has a similar opinion on this subject (like me), we can maybe revisit the topic in the Apodni group meeting. Otherwise, the PR would have my full support 👍
> 
> Would love to get the pr merged soon, as my next works are already in the pipeline and kinda depend on it. But yeah, I see that this topic might be discuss-worthy.

No objections from my side if you want to merge the PR immediately 👍

bauer-andreas:
@philippzagar @hendesi I would propose to merge this PR once the checks have completed successfully and incorporate any changes resulting from discussions in the next meeting to be implemented via a follow up PR, if that is acceptable.

bauer-andreas:
fyi, the PR description now also contains a Migration guide on how to migrate existing Modifier definitions. This isn't relevant for Apodini users, but for those potentially providing their own Apodini framework extensions and all Apodini Maintainers.

Issue #133, opened 2021-01-14
[MINOR FEATURE] Support multiple .content Parameter in the REST Exporter


bauer-andreas:
### Is your feature request related to a problem?

The Apodini DSL allows the definition of multiple `.content` `Parameter` definitions inside a `Handler`, like

```swift
@Parameter(.http(.body)) var userId: String
@Parameter var post: Post
@Parameter var other: Other
```

One would then expect that the following json body of a request will be parsed accordingly

```json5
{
  "userId": "...."
  "post": {
    // ...
  },
  "other": {
    // ...
  }
}
```

However the REST Interface Exporter currently only supports retrieval of one `.content` `Parameter` (or multiple with the same type holding the same value).

### Propose a solution

If the REST Exporter detects that there are multiple `.content` `Parameter`s defined for a given `Endpoint`, it will need to parse the first level of the json body, in such a way, that it can supply the properly typed values for the parameters.

If it detects that only one `.content` parameter exists on the `Endpoint`, the Exporter can just delegate decoding to Vapor as usual.

Issue #217, opened 2021-02-06
[MINOR FEATURE] REST: Handle thrown ApodiniErrors


bauer-andreas:
**Is your feature request related to a problem? Please describe.**
The `RESTInterfaceExporter` currently doesn't handle `ApodiniError`s as it doesn't conform to `StandardErrorCompliantExporter` (Not sure if that is needed or what exactly this is for, but seemingly this has some relevancy?).

**Propose a solution**
Add support for handling `ApodiniError` providing some standard JSON representation.

**Additional context**
The thrown is just passed onto Vapor, which returns **some** error to the user.
The `ApodiniError` is just turned into a string like the following.

```json
{
  "error": true,
  "reason": "ApodiniError(type: Apodini.ErrorType.notFound, reason: Optional(\"User not found\"), description: nil, options: Apodini.PropertyOptionSet<Apodini.ErrorOptionNameSpace>(options: [:]))"
}
```

The http error code is `500` by default, ignoring whatever is set in `ApodiniError`.

PR #307, opened 2021-07-01
Rework Exporter API

theMomax:
# Rework Exporter API

There exist a couple of problems with the current exporter API:
1. It provides little to no standardized solutions for anything beyond what is done by `ConnectionContext`, i.e. parameter validation and evaluating the `Handler`. This results in little similarities between our existing exporters and significant effort for implementing anything that goes beyond a request-response pattern (see WebSocket exporter). All of this introduces a huge potential for buggy code.
2. `ConnectionContext` packs "one solution that fits it all". This also means it performs operations that might not be necessary (e.g. validation of parameter-mutability, which is wasted on endpoints that only accept one request anyways). Furthermore, many exporters use higher level libraries, which might already do some of those validation steps on their own.
3. The process of observing `ObservableObject`s with `ConnectionContext` is pretty manual requiring quite some boilerplate code that should be basically the same for all exporters.
4. It exposes multiple public types on the `Apodini` target that are not relevant to _users_ of Apodini. 
5. It doesn't reflect special types such as `Blob`.

## :bulb: Proposed solution
There is a shift in the way we provide functionality to exporters: Previously, we exposed types that provide **functionality**, but also depend on functionality that had to be made available via the `InterfaceExporter` protocol. Now, the `Apodini` target only exposes types that provide **information**. We reuse the public `Delegate` type from the `Apodini` target as a new core for evaluating `Handler`s. The steps to and from this evaluation are facilitated using a multitude of protocols and standard implementations contained in the new target `ApodiniExtension`.

Some concrete consequences are:
* `InterfaceExporter` does no longer require `ExporterRequest` and `retrieveParameter()`
* There is no differentiation between `StaticInterfaceExporter` and `InterfaceExporter` anymore
* `ConnectionContext`, `Validator`, `ValidatingRequest` and `RequestHandler` are gone
* For testing, there is a `MockConnectionContext` in `XCTApodini` that packs the same functionality as the old `ConnectionContext`
* In `ApodiniExtension` there is a `LegacyInterfaceExporter` protocol that allows for interfacing with the new decoding API while using the old `retrieveParameter()` function 

In order to see the new constructs in action, take a look at the new HTTP interface exporter in `ApodiniHTTP`, which provides implementations for all for communicational patterns and `Blob`.

The new pipeline for handling a request looks about as follows:

Upon exporting, the exporter decides on a `DecodingStrategy` for transforming its own request type (e.g. `Vapor.Request`, or `Data`) to an Apodini `Request` (more specifically, how the decode parameters from its internal request type). 

```swift
ParameterTypeSpecific(
            .path,
            // is used if the `ParameterType` of the parameter is `.path` (works on `Vapor.Request`)
            using: PathStrategy(),
            otherwise: ParameterTypeSpecific(
                .lightweight,
                // is used if the `ParameterType` of the parameter is `.lightweight` (works on `Vapor.Request`)
                using: LightweightStrategy(), 
                // is used if the `ParameterType` of the parameter is `.content` (works on `Data`)
                //
                // uses `configuration.decoder` to decode either directly from the base
                // container (if there is only one `.content` parameter defined on `endpoint`), or
                // on a sub-container keyed by the parameter's name (if there are multiple `.content`
                // parameters defined on `endpoint`
                otherwise: NumberOfContentParameterAwareStrategy
                                .oneIdentityOrAllNamedContentStrategy(configuration.decoder, for: endpoint)
                                // interfaces between the inner strategy (which is based on `Data`) and the outer
                                // (which is based on `Vapor.Request`) by extracting the `Vapor.Request`'s `body` data
                                .transformedToVaporRequestBasedStrategy())
            )
```

Upon arrival of a request:
* The exporter gets its internal request (previously named `ExporterRequest`)
* The exporter creates a new `Delegate`
* The exporter uses the predefined `DecodingStrategy` to obtain an Apodini `Request´ from its internal request.
* The exporter wraps the obtained `DecodingRequest` in validating requests (e.g. one that inserts default values or one that validates the `Mutability` property)
* If necessary we subscribe to `ObservableObject`s
* The exporter evaluates the request on the `delegate`
* The exporter `transform`s the resulting `Response` to its internal response type using a `ResultTransformer`

For the request-response pattern we use a pipe of `EventLoopFuture`s. For anything else we use an OpenCombine `Publisher` pipe (although we will probably convert that to `AsyncSequence`s once Swift 5.5 is released).

HTTP exporter examples:

```swift
// build request-response pattern request-handling closure:
        let strategy = singleInputDecodingStrategy(for: endpoint)
        
        let transformer = VaporBlobResponseTransformer()
        
        return { (request: Vapor.Request) in
            var delegate = Delegate(endpoint.handler, .required)
            
            return strategy
                .decodeRequest(from: request, with: request, with: request.eventLoop)
                .insertDefaults(with: defaultValues)
                .cache()
                .evaluate(on: &delegate)
                .transform(using: transformer)
        }

// build service-side streaming pattern request-handling closure:
        let strategy = singleInputDecodingStrategy(for: endpoint)
        
        let abortAnyError = AbortTransformer<H>()
        
        return { (request: Vapor.Request) in
            var delegate = Delegate(endpoint.handler, .required)
            
            return Just(request) // we start a pipeline that just publishes our single request
                .map { request in (request, request) }
                .decode(using: strategy, with: request.eventLoop)
                .insertDefaults(with: defaultValues)
                .validateParameterMutability()
                .cache()
                .subscribe(to: &delegate) // we subscribe to the delegate, making it a pipeline of `Event`s instead of `Request`
                .evaluate(on: &delegate)
                .transform(using: abortAnyError)
                .cancel(if: { response in
                    response.connectionEffect == .close // once we receive a `.final` or `.end` `Response` we cancel subscription and evaluation
                })
                .collect() // we collect all `Response`s into an array that is published once the evaluation was cancelled
                .tryMap { (responses: [Apodini.Response<H.Response.Content>]) in
                    // we build one response form all the responses by taking the status and information from the last and
                    // building an array from the (non empty) contents
                    let status: Status? = responses.last?.status
                    let information: Set<AnyInformation> = responses.last?.information ?? []
                    let content: [H.Response.Content] = responses.compactMap { response in
                        response.content
                    }
                    let body = try configuration.encoder.encode(content)
                    
                    return Vapor.Response(status: HTTPStatus(status ?? .ok),
                                          headers: HTTPHeaders(information),
                                          body: Vapor.Response.Body(data: body))
                }
                // we transform the publisher-pipeline to an `EventLoopFuture` by taking the first (and only) element
                .firstFuture(on: request.eventLoop)
                .map { optionalResponse in
                    optionalResponse ?? Vapor.Response()
                }
        }
```

## :gear: Release Notes 
* Web Service Developers: no changes
* `StaticInterfaceExporter` Developers:
   * conform to `InterfaceExporter` instead
   * add an export function for `Blob`:
   ```swift
    func export<H>(blob endpoint: Endpoint<H>) where H: Handler, H.Response.Content == Blob {
      export(endpoint) // if nothing changed for your associated exporter you can just forward to the normal export
    }
   ```
* `InterfaceExporter` Developers: easiest way to adopt the new system is through `LegacyInterfaceExporter`
   * conform to `LegacyInterfaceExporter` instead of `InterfaceExporter`
   * you can obtain a `DecodingStrategy` like this:
   ```swift
   let decodingStrategy = InterfaceExporterLegacyStrategy(self).applied(to: endpoint)
   ```
   * to get the same results as you had with `ConnectionContext`, you should use the following `Request`-wrappers:
   ```swift
   .insertDefaults(with: defaultValueStore)
   .validateParameterMutability()
   .cache()
   ```
   * you will need to adapt to the usage of `Delegate`, `evaluate(on:)` and - depending on your needs also - `subscribe(to:)` instead of `ConnectionContext`

This PR starts a process of trimming the public interface of the `Apodini` target down to what really is required for building a web service with Apodini. Future PRs should always hesitate to add additional public types to Apodini which are not part of the "frontend". Instead, exporter-related types should be defined in `ApodiniExtension`. The long term goal is to move everything starting from the `SemanticModelBuilder` (or even `SyntaxTreeVisitor`) out of the `Apodini` target. Methods to facilitate this are the following:
* add a static function to the public `_Internal` enum
* make an internal protocol (e.g. `RequestInjectable`) public, but prefix it and all its requirements with an underscore (`_RequestInjectable`, `_inject(using:)`)

### Testing
There are already some tests for `ApodiniHTTP` and I will add further coverage as necessary. Some components which re-implement older concepts (e.g. `ValidatingRequest`) already have pretty good coverage.

### Reviewer Nudging
I think the best way is to look at the new HTTP exporter to get a feel for how the new API is used. Take a look at `EndToEndTests.swift` in parallel to see how the endpoints actually behave. Explore the single components of the pipeline from there. You can take a look at WebSocketExporter to see a `LegacyInterfaceExporter`. An interesting detail might be to inspect the new extensions on `_Internal` to see what was necessary to make this work in a different target.


Package.resolved
```diff
@@ -24,8 +24,8 @@
         "repositoryURL": "https://github.com/swift-server/async-http-client.git",
         "state": {
           "branch": null,
-          "revision": "8e4d51908dd49272667126403bf977c5c503f78f",
-          "version": "1.5.0"
+          "revision": "8ccba7328d178ac05a1a9803cf3f2c6660d2f826",
+          "version": "1.3.0"
```
bauer-andreas: Seems like you reverted the project configuration to include older dependencies 🤔

Sources/XCTApodini/Mocks/MockConnectionContext.swift
```diff
@@ -0,0 +1,153 @@
...
+
+
+#if DEBUG || RELEASE_TESTING
+@testable import Apodini
+@testable import ApodiniExtension
+import struct Foundation.UUID
+@testable import ApodiniREST
+import ApodiniVaporSupport
+import Vapor
+import ApodiniREST
```
bauer-andreas: Minor: duplicate import of REST

Sources/ApodiniHTTP/InterfaceExporter.swift
```diff
@@ -0,0 +1,408 @@
...
+        
+        /// Insert exporter into `InterfaceExporterStorage`
+        app.registerExporter(exporter: exporter)
+    }
+}
+
+
+// MARK: Exporter
+
+struct Exporter: InterfaceExporter {
```
bauer-andreas, line 55: This file seems bit lengthy to me. Would it make sense to break out some stuff into multiple extensions making it easier to oversight?

Sources/ApodiniHTTP/InterfaceExporter.swift
```diff
@@ -0,0 +1,408 @@
...
+    /// - Parameters:
+    ///    - encoder: The to be used `AnyEncoder`, defaults to a `JSONEncoder`
+    ///    - decoder: The to be used `AnyDecoder`, defaults to a `JSONDecoder`
+    ///    - caseInsensitiveRouting: Indicates whether the HTTP route is interpreted case-sensitivly
+    public init(encoder: AnyEncoder = defaultEncoder, decoder: AnyDecoder = defaultDecoder, caseInsensitiveRouting: Bool = false) {
+        self.configuration = ExporterConfiguration(encoder: encoder, decoder: decoder, caseInsensitiveRouting: caseInsensitiveRouting)
+    }
+    
+    public func configure(_ app: Apodini.Application) {
+        /// Instanciate exporter
```
bauer-andreas:
```suggestion
        /// Instantiate exporter
```

Sources/ApodiniHTTP/InterfaceExporter.swift
```diff
@@ -0,0 +1,408 @@
...
+    /// The default `AnyDecoder`, a `JSONDecoder`
+    public static var defaultDecoder: AnyDecoder {
+        JSONDecoder()
+    }
+    
+    /// Initializes the configuration of the ``HTTP`` exporter with (default) `AnyEncoder` and `AnyDecoder`
+    /// - Parameters:
+    ///    - encoder: The to be used `AnyEncoder`, defaults to a `JSONEncoder`
+    ///    - decoder: The to be used `AnyDecoder`, defaults to a `JSONDecoder`
+    ///    - caseInsensitiveRouting: Indicates whether the HTTP route is interpreted case-sensitivly
```
bauer-andreas:
```suggestion
    ///    - caseInsensitiveRouting: Indicates whether the HTTP route is interpreted case-sensitively
```

Sources/Apodini/Metadata/BuiltinMetadata/CommunicationalPatternMetadata.swift
```diff
@@ -0,0 +1,74 @@
...
+extension HandlerMetadataNamespace {
+    /// Name Definition for the `ComponentDescriptionMetadata`
+    public typealias Pattern = CommunicationalPatternMetadata
+}
+
+/// The ``CommunicationalPatternMetadata`` can be used to explicitly declare a ``Handler``'s
+/// ``CommunicationalPattern``.
+///
+/// The framework can automatically detect ``CommunicationalPattern/requestResponse`` if the
+/// ``Handler/handle()-94skg``'s return type is no ``Response``. If that is not the case and you don't
```
bauer-andreas, line 39: Is it me or is the -94skg weird lol

Sources/Apodini/Metadata/BuiltinMetadata/CommunicationalPatternMetadata.swift
```diff
@@ -0,0 +1,74 @@
...
+/// explicitly specify the ``HandlerMetadataNamespace/Pattern``, the ``Handler`` is assumed to use
+/// ``CommunicationalPattern/bidirectionalStream``.
```
bauer-andreas, line 41: Looking at the `CommunicationalPattern` Knowledge Source (specifically `AutomaticCommunicationalPattern`) this statement isn't entirely true is it?
Rather the default is `.requestResponse` and `.bidirectionalStream` where some Apodini `Response` is returned.

Does that make sense to assume bidirectional streams when returning `Response`?

Sources/ApodiniExtension/KnowledgeSources/CommunicationalPatternKnowledge.swift
```diff
@@ -0,0 +1,44 @@
+//
+//  KommunicationalPatternKnowledge.swift
```
bauer-andreas:
```suggestion
//  CommunicationalPatternKnowledge.swift
```

Sources/ApodiniExtension/KnowledgeSources/CommunicationalPatternKnowledge.swift
```diff
@@ -0,0 +1,44 @@
...
+    init<H, B>(from handler: H, _ blackboard: B) throws where H: Handler, B: Blackboard {
+        if H.Response.self is CustomizableResponse.Type {
+            self.value = .bidirectionalStream
+        } else {
+            self.value = .requestResponse
+        }
+    }
+}
+
+private protocol CustomizableResponse { }
```
bauer-andreas, line 46: Even though its internal, maybe it makes sense to add some docs explaining **that** it is used for and **how** it is used for inferring the default Communication Pattern?

Sources/Apodini/Metadata/BuiltinMetadata/CommunicationalPatternMetadata.swift
```diff
@@ -0,0 +1,74 @@
...
+///
+/// The Metadata is available under the ``HandlerMetadataNamespace/Pattern`` name and can be used like the following:
+/// ```swift
+/// struct ExampleHandler: Handler {
+///     // ...
+///     var metadata: Metadata {
+///         Pattern(.clientSideStream)
+///     }
+/// }
+/// ```
```
bauer-andreas, line 55: Very detailed documentation 💪🏽

Sources/Apodini/Semantic Model Builder/InterfaceExporter/InterfaceExporter.swift
```diff
@@ -90,16 +72,3 @@ extension InterfaceExporter {
         visitor.visit(exporter: self)
     }
 }
-
-
-/// Any InterfaceExporter creating a representation of the WebService must conform to this protocol.
-///
-/// Such exporters do not actively create a accessible WebService themselves but rather a static representation
-/// of the WebService (e.g. a Endpoint serving documentation of the WebService).
-public protocol StaticInterfaceExporter: BaseInterfaceExporter {}
```
bauer-andreas, line 99: Really like the simplification of the InterfaceExporter protocol, unifying it again 👍

Sources/ApodiniExtension/AsyncEvaluation.swift
```diff
@@ -0,0 +1,282 @@
...
+            return result
+        }
+    }
+}
+
+
+// MARK: Handling Subscription
+
+public extension Publisher where Output: Request {
+    /// A `Publisher` that takes care of subscribing to `TriggerEvent`s emmited
```
bauer-andreas:
```suggestion
    /// A `Publisher` that takes care of subscribing to `TriggerEvent`s emitted
```

Sources/ApodiniExtension/AsyncEvaluation.swift
```diff
@@ -0,0 +1,282 @@
...
+    /// by the given `Delegate` as well as converting the upstream's `Request`s to
+    /// `Event`s.
+    ///
+    /// In detail, this `Publisher` does four things:
+    ///     - it maps all upstream `Request`s to an ``Event/request(_:)``
+    ///     - if successfull (`finished`), it maps the upstream's completion to an
```
bauer-andreas:
```suggestion
    ///     - if successful (`finished`), it maps the upstream's completion to an
```

Sources/ApodiniExtension/AsyncEvaluation.swift
```diff
@@ -0,0 +1,282 @@
...
+        })
+        .store(in: &cancellables)
+        
+        return downstream
+    }
+}
+
+/// A `Publisher` that also is a `Cancellable`.
+///
+/// The ``CancellablePublisher`` wrapps a given `Publisher` `P` and
```
bauer-andreas:
```suggestion
/// The ``CancellablePublisher`` wraps a given `Publisher` `P` and
```

Sources/ApodiniExtension/AsyncEvaluation.swift
```diff
@@ -0,0 +1,282 @@
...
+        _publisher.receive(subscriber: subscriber)
+    }
+    
+    public func cancel() {
+        cancelCallback()
+    }
+}
+
+extension Publisher {
+    /// Wrapps this `Publisher` into a ``CancellablePublisher`` that executes the given
```
bauer-andreas:
```suggestion
    /// Wraps this `Publisher` into a ``CancellablePublisher`` that executes the given
```

Sources/ApodiniHTTP/InterfaceExporter.swift
```diff
@@ -0,0 +1,408 @@
...
+    
+    let configuration: HTTP.ExporterConfiguration
+    
+    let logger: Logger
+    
+    init(_ app: Apodini.Application, _ configuration: HTTP.ExporterConfiguration) {
+        self.app = app
+        self.configuration = configuration
+        var logger = app.logger
+        logger.logLevel = .trace
```
bauer-andreas: Does trace as hardcoded default make sense?

Sources/ApodiniExtension/RequestDecoding/EndpointDecodingStrategy.swift
```diff
@@ -0,0 +1,294 @@
...
+///
+/// E.g. for `Element` being `String` and the parameter's name being `"name"`, the following JSON
+/// would be a valid input:
+///
+/// ```json
+/// {
+///     "name": "Max"
+/// }
+/// ```
+public struct AllNamedStrategy: EndpointDecodingStrategy {
```
bauer-andreas: This in combination with `NumberOfContentParameterAwareStrategy` should solve https://github.com/Apodini/Apodini/issues/133 right?

Sources/ApodiniExtension/RequestDecoding/EndpointDecodingStrategy.swift
```diff
@@ -0,0 +1,294 @@
+//
+//  EndpointDecodingStrategy.swift
```
bauer-andreas: I think this is another one of those files which would benefit from splitting it up into multiple files (maybe locating them into a sub dir).
I think this would the overview a lot.
This probably applies to other files as well eg. `AsyncEvaluation` ...

Sources/ApodiniExtension/RequestDecoding/EndpointDecodingStrategy.swift
```diff
@@ -0,0 +1,294 @@
...
+    /// given `parameter`.
+    ///
+    /// - Note: Due to the nature of `EndpointParameter` the `Element` is of a different type
+    /// than the one of ``DecodingStrategy/strategy(for:)`` if the Apodini `Parameter` has
+    /// an `Optional` type. Thus, the ``AnyParameterDecodingStrategy`` returned by this function
+    /// might encounter the case that it finds a `nil` value where `Element` is not an `Optional`. In that
+    /// case the ``ParameterDecodingStrategy/decode(from:)`` should throw an
+    /// `DecodingError.valueNotFound`.
+    func strategy<Element: Decodable>(for parameter: EndpointParameter<Element>) -> AnyParameterDecodingStrategy<Element, Input>
+}
```
bauer-andreas: I think this is a well constructed infrastructure to support parameter decoding, specifically for reusing decoding strategies. Great work 🚀

Sources/ApodiniExtension/RequestDecoding/EndpointDecodingStrategy.swift
```diff
@@ -0,0 +1,294 @@
...
+    }
+    
+    public func strategy<Element>(for parameter: EndpointParameter<Element>)
+        -> AnyParameterDecodingStrategy<Element, P.Input> where Element: Decodable, Element: Encodable {
+        if parameter.parameterType == self.parameterType {
+            return primary.strategy(for: parameter)
+        } else {
+            return backup.strategy(for: parameter)
+        }
+    }
```
bauer-andreas: Would it maybe make more sense to rely on e.g. a Dictionary based lookup for the selection of parameter specific decoding?
Not sure about runtime expensiveness but a direct dictionary lookup seems more practical to me(?).
While the syntax has quite a "sweetness", supplying a decoding strategy for all possible types seems to always involve quite heavy nesting (see HTTP Exporter) which in my opinion doesn't really benefit readability.

I think performance and readability wise the old style switch case statement was the best approach, bust isn't fitting obviously. Random though was that we maybe could use the resultBuilder with their switch-cases, but can't really imagine a good solution. So to me the Dictionary seems like the best idea syntax-wise, though I start to see why you didn't want to do it, as the current approach presumably forces you to core all parameter types?

Sources/ApodiniExtension/RequestDecoding/ParameterDecodingStrategy.swift
```diff
@@ -0,0 +1,224 @@
...
+}
+
+
+// MARK: TransformingStrategy
+
+/// A ``ParameterDecodingStrategy`` that forms a transition between a wrapped ``ParameterDecodingStrategy`` `S`
+/// and a parent ``ParameterDecodingStrategy`` with ``ParameterDecodingStrategy/Input`` type `I`.
+///
+/// This strategy allows for reuse of ``ParameterDecodingStrategy`` with a more basic ``ParameterDecodingStrategy/Input`` than
+/// the one you are acutally using. E.g. many of the predefined strategies have input type `Data`.
```
bauer-andreas:
```suggestion
/// the one you are actually using. E.g. many of the predefined strategies have input type `Data`.
```

Sources/ApodiniExtension/RequestDecoding/ParameterDecodingStrategy.swift
```diff
@@ -0,0 +1,224 @@
...
+        try strategy.decode(from: transformer(input))
+    }
+}
+
+
+// MARK: Implementations
+
+/// A ``ParameterDecodingStrategy`` that ignores its actual input and instead always
+/// return the same `element`.
+public struct GivenStrategy<E: Decodable, I>: ParameterDecodingStrategy {
```
bauer-andreas: Seems to be unused currently? Is there a use for it?

Same for the `ThrowingStrategy` below?

Sources/ApodiniExtension/RequestDecoding/RequestDecoding.swift
```diff
@@ -0,0 +1,129 @@
...
+
+
+// MARK: Request Decoding
+
+extension DecodingStrategy {
+    /// Generates an Apodini `Request` based on the given `basis` and `eventLoop` which lazily decodes
+    /// the given `input` by applying this strategy.
+    ///
+    /// When working with a `Publisher` pipeline, use `Publisher`'s `decode(using:with)`.
+    public func decodeRequest(from input: Input, with basis: RequestBasis, with eventLoop: EventLoop) -> DecodingRequest<Input> {
```
bauer-andreas, line 86: I think it would make sense to document the parameters of this function. On first sight I didn't really know what each of them are (and not sure if I really know it yet).

Also, it seems to me that on most occasions (e.g. in the HTTP exporter) the `input` parameter is equal to the `basis` parameter (except form the gRPC exporter and the RHI manager when the second parameter wraps the input in a  `DefaultRequestBasis`). Would it make sense to create a shortcut for that, as I think it reads odd when passing the same parameter twice (and the external parameter not really explaining it, at least for me).

Sources/ApodiniExtension/RequestProcessing.swift
```diff
@@ -0,0 +1,261 @@
...
+        self.map { request in
+            defaults.insertDefaults(request)
+        }
+    }
+}
+
+public extension Request {
+    /// Wrapps this `Request` into a ``DefaultValueStore/DefaultInsertingRequest`` using
+    /// the given `defaults`.
+    func insertDefaults(with defaults: DefaultValueStore) -> DefaultValueStore.DefaultInsertingRequest {
```
bauer-andreas: Do I understand it correctly that unless the exporter calls `insertDefaults`, default values of `@Parameter`s aren't inserted?
Seems like it produces some code duplication, as every exporter ha to query the DefaultValueStore at some point, and to me I can't set a use case where you don't want to insertDefault values?

Sources/ApodiniExtension/RequestProcessing.swift
```diff
@@ -0,0 +1,261 @@
...
+        if nilIsValidValue {
+            return .some(Type?.none as Any)
+        } else {
+            return .none
+        }
+    }
+}
+
+public extension Publisher where Output: Request {
+    /// Wrapps each incoming `Request` into a ``DefaultValueStore/DefaultInsertingRequest`` using
```
bauer-andreas:
```suggestion
    /// Wraps each incoming `Request` into a ``DefaultValueStore/DefaultInsertingRequest`` using
```

Sources/ApodiniExtension/RequestProcessing.swift
```diff
@@ -0,0 +1,261 @@
...
+    /// the given `defaults`.
+    func insertDefaults(with defaults: DefaultValueStore) -> OpenCombine.Publishers.Map<Self, DefaultValueStore.DefaultInsertingRequest> {
+        self.map { request in
+            defaults.insertDefaults(request)
+        }
+    }
+}
+
+public extension Request {
+    /// Wrapps this `Request` into a ``DefaultValueStore/DefaultInsertingRequest`` using
```
bauer-andreas:
```suggestion
    /// Wraps this `Request` into a ``DefaultValueStore/DefaultInsertingRequest`` using
```

Sources/ApodiniExtension/RequestProcessing.swift
```diff
@@ -0,0 +1,261 @@
...
+    func insertDefaults(with defaults: DefaultValueStore) -> DefaultValueStore.DefaultInsertingRequest {
+        defaults.insertDefaults(self)
+    }
+}
+
+
+// MARK: Mutability Validation
+
+public extension Publisher where Output: Request {
+    /// Wrapps all incoming `Request`s in ``MutabilityValidatingRequest``s, which share
```
bauer-andreas:
```suggestion
    /// Wraps all incoming `Request`s in ``MutabilityValidatingRequest``s, which share
```

Sources/ApodiniExtension/RequestProcessing.swift
```diff
@@ -0,0 +1,261 @@
...
+    /// a common storage for providing stateful validation.
+    func validateParameterMutability() -> OpenCombine.Publishers.Map<Self, MutabilityValidatingRequest<Output>> {
+        self.reduce()
+    }
+}
+
+
+/// A `Request` which wrapps another `Request` and forwards its properties but performs
+/// a statful validation of the Apodini `Mutability` of the requested parameters before forwarding
```
bauer-andreas:
```suggestion
/// A `Request` which wraps another `Request` and forwards its properties but performs
/// a stateful validation of the Apodini `Mutability` of the requested parameters before forwarding
```

Sources/ApodiniHTTP/Configuration.swift
```diff
@@ -0,0 +1,39 @@
...
+        /// Indicates whether the HTTP route is interpreted case-sensitivly
+        public let caseInsensitiveRouting: Bool
+        
+        
+        /// Initializes the configuration of the ``HTTP`` exporter
+        /// - Parameters:
+        ///    - encoder: The to be used `AnyEncoder`, defaults to a `JSONEncoder`
+        ///    - decoder: The to be used `AnyDecoder`, defaults to a `JSONDecoder`
+        ///    - caseInsensitiveRouting: Indicates whether the HTTP route is interpreted case-sensitivly
+        public init(encoder: AnyEncoder = HTTP.defaultEncoder,
```
bauer-andreas, line 31: Just out of curiosity, can the OpenAPI exporter be easily adapted to also depend on the plain http exporter, providing an OpenAPI specification for it as well?

Sources/ApodiniHTTP/InterfaceExporter.swift
```diff
@@ -0,0 +1,408 @@
...
+            app.vapor.app.on(knowledge.method,
+                             knowledge.path,
+                             use: buildClientSideStreamingClosure(for: endpoint, using: knowledge.defaultValues))
+        case .bidirectionalStream:
+            logger.info("Exporting Bidirectional-Streaming Pattern on \(knowledge.method): \(knowledge.path)")
+            app.vapor.app.on(knowledge.method,
+                             knowledge.path,
+                             use: buildBidirectionalStreamingClosure(for: endpoint, using: knowledge.defaultValues))
+        }
+    }
```
bauer-andreas, line 98: I think this exporter is really showcases the benefits of this new way of building request handling strategies. With very little effort, and without the need for spaghetti code or a "hugely differing" code base you can support all the different communication patterns in a very composable style.
Chapeau, great work 🚀

Sources/ApodiniVaporSupport/ApodiniError+AbortError.swift
```diff
@@ -0,0 +1,81 @@
...
+//
+//  Created by Max Obermeier on 30.06.21.
+//
+
+import Apodini
+import Vapor
+
+// MARK: AbortError
+
+extension ApodiniError: AbortError {
```
bauer-andreas, line 13: I assume this solves https://github.com/Apodini/Apodini/issues/217 right? just to keep track of it.

Sources/ApodiniHTTP/InterfaceExporter.swift
```diff
@@ -0,0 +1,408 @@
...
+                .firstFuture(on: request.eventLoop)
+                .map { optionalResponse in
+                    optionalResponse ?? Vapor.Response()
+                }
+        }
+    }
+    
+    // MARK: Bidirectional Streaming Closure
+    
+    private func buildBidirectionalStreamingClosure<H: Handler>(
```
bauer-andreas: So to my understanding this isn't a real bidirectional streaming that e.g. builds on HTTP/2 but really only multiplexes multiple requests and response into one HTTP/1.1 request response right?

Would using HTTP/2 for that work though? Maybe a great way to showcase that with a example implementation 🤟🏼

Tests/ApodiniTests/RelationshipTests/Infrastructure/RelationshipExporter.swift
```diff
@@ -5,14 +5,28 @@
 import XCTest
 import XCTApodini
 @testable import Apodini
+@testable import ApodiniREST
 
 class RelationshipExporter: MockExporter<String> {
-    var endpoints: [(AnyEndpoint, AnyRelationshipEndpoint)] = []
+    // swiftlint:disable:next large_tuple
+    var endpoints: [(AnyEndpoint, AnyRelationshipEndpoint, (String, [Any??], Application) throws -> EnrichedContent)] = []
```
bauer-andreas: Well I think the SwiftLint warning has its place here, I don't understand shit 😅 maybe a sweet struct would do the job?

bauer-andreas (commented):
Great work there 🚀
All in all those are some welcome improvements to the way how exporters are defined and written. 

While there are quite some comments from my side, most are grammar correction or hints where I think understandability could be improved either by adding more documentation or by splitting some stuff up into more files as otherwise some stuff might feel a bit overwhelming to the first reader.

I did my best to give a in-depth review as possible. As I'm not really that familiar with the internals of OpenCombine, those parts weren't really a strong focus of my review. Hope that's okay.

Sources/ApodiniVaporSupport/ApodiniError+AbortError.swift
theMomax: ~~Not entirely. REST exporter still has to make sure it converts all errors thrown to an `ApodiniError` using `Error.apodiniError`, but that should be not a big problem. I can include that with this PR.~~

I guess so, yes. The rest exporter does currently not map each and every `Error` to an `ApodiniError` before passing it to `Vapor`. But every `@Throws` in a `Handler` would be interpreted as an `Vapor.AbortError`. If a `Handler` just throws random errors, those would still be treated the same.

Sources/Apodini/Metadata/BuiltinMetadata/CommunicationalPatternMetadata.swift
theMomax: That is a DocC thing. There are four different declarations/implementations for `Handler/handle()`, so it needs the tag to differentiate for the link. But yes, it looks pretty weird in in-code docs😐

Sources/Apodini/Metadata/BuiltinMetadata/CommunicationalPatternMetadata.swift
theMomax: I don't quite get what you mean. Both reads the same for me. If the response does not make use of `Response`, we assume it to be `.requestResponse`, otherwise we assume the "worst" which is `.bidirectionalStream`. Note that there is a **no** in "return type is **no** ``Response``".

Sources/ApodiniExtension/KnowledgeSources/CommunicationalPatternKnowledge.swift
theMomax: lol

Sources/Apodini/Metadata/BuiltinMetadata/CommunicationalPatternMetadata.swift
bauer-andreas: just noticed I overhead the whole first sentence 😅 Seemingly got a bit too confused by the 94skg 😂

Sources/ApodiniHTTP/InterfaceExporter.swift
theMomax: Nope, not at all

Sources/ApodiniExtension/RequestDecoding/EndpointDecodingStrategy.swift
theMomax: It are exactly the right tools to fix #133, correct. Currently, the REST exporter doesn't use them, though.

Package.resolved
theMomax: Yeah...development of this was everything but linear...

Sources/ApodiniHTTP/InterfaceExporter.swift
theMomax: Good point 👍🏻

Sources/ApodiniExtension/RequestDecoding/EndpointDecodingStrategy.swift
theMomax: Yeah, this probably applies to almost the whole ApodiniExtension target🙃

Sources/ApodiniExtension/RequestDecoding/ParameterDecodingStrategy.swift
theMomax: This is partially a remaining of the time where I still had the input type fixed on `Data`, but they are actually pretty useful for testing, thus I kept them around.

Sources/ApodiniExtension/RequestProcessing.swift
theMomax: Yes, you are right. The use-case for that is exporters that use rather high-level frameworks for exporting the WebService that know the concept of default-values. I think GraphQL does, so you would already provide the default value to that framework, which then inserts it if noting else is provided.

Sources/ApodiniHTTP/Configuration.swift
theMomax: That would definitely still require some work, I guess. I didn't want to open that door tbh, but that would definitely be a topic for future work. The implementation of the OpenAPI exporter still pretty "manually" tries to re-build what the REST exporter does at runtime. One would have to establish a generic model at the OpenAPI exporter first.

Sources/ApodiniHTTP/InterfaceExporter.swift
theMomax: Yes, correct. Of course, it would make absolute sense to implement versions that make use of HTTP/2 features, but my intention here was more to showcase how the Publisher API can be used to nicely provide a "downgraded" endpoint, where the protocol just doesn't allow for a direct representation. Even if we were to add HTTP/2 implementations for the patterns that can be implemented properly using HTTP/2, we would still require the current implementations for clients that only support HTTP/1.1.

Sources/ApodiniExtension/RequestDecoding/EndpointDecodingStrategy.swift
theMomax: I think it's actually easiest and most elegant to just add a third generic so there is one strategy for each type:

```swift
ParameterTypeSpecific(
      lightweight: LightweightStrategy(),
      path: PathStrategy(useNameAsIdentifier: false),
      content: AllIdentityStrategy(exporterConfiguration.decoder).transformedToVaporRequestBasedStrategy()
)
```

Tests/ApodiniTests/RelationshipTests/Infrastructure/RelationshipExporter.swift
theMomax: You got me there🙈

Sources/ApodiniHTTP/InterfaceExporter.swift
bauer-andreas: I like how each closure for each communication pattern got its own file 👍

Sources/Apodini/Metadata/BuiltinMetadata/CommunicationalPatternMetadata.swift
bauer-andreas: Okay weird. But thanks for pointing out. Now a kinda understand your concern in todays meeting about Clion not supporting that docs style yet.
Seems like I have to switch between editors now 😅

Sources/ApodiniExtension/KnowledgeSources/CommunicationalPatternKnowledge.swift
bauer-andreas: Thanks for adding the extended documentation 👍

Sources/ApodiniExtension/RequestDecoding/EndpointDecodingStrategy.swift
bauer-andreas: Would it be huge work to incorporate support for that in the REST exporter? I think this is generally a long awaited feature? But a follow up PR would probably also be okay, as long as it doesn't get lost.
I remember also @eldcn asking if REST would support decoding multiple content parameters.

Sources/ApodiniExtension/RequestDecoding/EndpointDecodingStrategy/ParameterTypeSpecific.swift
```diff
@@ -0,0 +1,44 @@
...
+//  
+//
+//  Created by Max Obermeier on 06.07.21.
+//
+
+import Apodini
+
+/// An ``EndpointDecodingStrategy`` that chooses between two given strategies based
+/// on the `parameter`'s `parameterType`.
+public struct ParameterTypeSpecific<
```
bauer-andreas: I think docs might be outdated? it doesn't choose between **two** right?

Sources/ApodiniExtension/RequestDecoding/EndpointDecodingStrategy.swift
bauer-andreas: Thanks for the change, it think is really improves the overview of what belongs together 👍

Sources/ApodiniExtension/RequestDecoding/EndpointDecodingStrategy.swift
bauer-andreas: Probably due to doing the review so late in the night, that I couldn't see the simple solutions 😅 

Like that new version a lot more 👍

Sources/ApodiniExtension/RequestProcessing.swift
bauer-andreas: Would it make sense to have an overload where the Exporter doesn't need to explicitly need to pass the `DefaultValueStore` if it just takes the default one as provided by Apodini?
Not sure how easy it is to implement that in regards to required internal changes for that.

Sources/ApodiniHTTP/Configuration.swift
bauer-andreas: Alright, thanks for explaining. Just was curios.

Sources/ApodiniVaporSupport/ApodiniError+AbortError.swift
bauer-andreas: Not really sure on what exactly we decided on in the meeting today? I think Apodini should not make the decision for the exporter to automatically map all errors to ApodiniError.
But would it make sense for REST to map all errors to ApodiniErrors? Not sure if this would result in strangely differing output. Would there be even a difference?

I have no strong opinion on this anyways.

Sources/ApodiniHTTP/InterfaceExporter.swift
bauer-andreas: Understood. And completely reasonable.

Tests/ApodiniTests/RelationshipTests/Infrastructure/RelationshipExporter.swift
```diff
@@ -8,13 +8,26 @@ import XCTApodini
...
+        let evaluateCallback: (_ request: String, _ parameters: [Any??], _ app: Application) throws -> EnrichedContent
+        
+        internal init(_ endpoint: AnyEndpoint,
+                      _ relationshipEndpoint: AnyRelationshipEndpoint,
+                      _ evaluateCallback: @escaping (String, [Any??], Application) throws -> EnrichedContent) {
+            self.endpoint = endpoint
+            self.relationshipEndpoint = relationshipEndpoint
+            self.evaluateCallback = evaluateCallback
+        }
+    }
```
bauer-andreas, line 23: Way better :+1: 😘

bauer-andreas (approved):
Thanks for incorporating the feedback.
I think breaking some of the code up into their individual files really benefits code oversight and navigability. 👍 

Apart of the failing CI and one little comment I have nothing to add 🤪

I think this is a well developed improvement to the Apodini Exporter API 🚀

Sources/ApodiniExtension/RequestDecoding/EndpointDecodingStrategy.swift
theMomax: Not for the REST exporter, but for OpenAPI...

bauer-andreas: So I assume, in a potential follow up PR one need to find a solution to allow e.g. OpenAPI to retrieve the parameter decoding strategy (or potentially the whole decoding strategy) of e.g. the REST exporter for this to work seamlessly (e.g. to also allow OpenAPI to be applied to a multitude of exporters), right?

theMomax: Yes, exactly. Basically every exporter should publicly provide a KnowledgeSource that contains all information it uses for an endpoint (including the DecodingStrategy). The OpenAPI exporter must then derive everything from that KnowledgeSource.

Sources/ApodiniExtension/RequestProcessing.swift
theMomax: I don't think that would help really, as that overload would still require an `Endpoint` instance to work. I also see room for improvement there, but I am currently out of ideas how to approach this. Maybe I'll come up with a solution in the AsyncSequence follow up PR, until then it is up to three lines of code duplication for the exporters...

theMomax:
Merging this now as the PR already is pretty big. I will create a follow up PR that takes care of the async/await and `AsyncSequence` part.

PR #309, opened 2021-07-02
[Deployment Provider] Store WS structure in memory and add deployment as sub command

hendesi:
# [Deployment Provider] Store WS structure in memory and add deployment as sub command

To be able to retrieve a web services structure, the deployment provider builds the web service again in a separate process and runs it with custom flags to retrieve its structure by saving it to a JSON file. This can be very time consuming.

## :bulb: Proposed solution
Although for some deployments, e.g. AWS lambda, we still need to persist the web service structure for other processes, we can simplify this for the `LocalhostDeploymentProvider` by saving the web service structure in memory. This removes the need for a separate process to build and start the web service and thereby saves a lot of time. 

## :gear: Release Notes 
-  Reworks the current deployment CLIs and integrates them to the new Apodini web service CLI
Add it to a web service by setting the `CommandConfiguration` property:
```
    static var configuration: CommandConfiguration {
        CommandConfiguration(subcommands: [DeploymentCLI<Self>.self])
    }
```

Besides the in-memory storage of the web service structure, this PR refactores also the `Localhost` and `AWSLambda` CLIs and bundles it in a new target `DeploymentCLI`. This offers the deployment for the different targets as sub commands:
```
OVERVIEW: Apodini deployment provider

Deploys an Apodini web service to the specified target.

USAGE: test-web-service deploy <subcommand>

OPTIONS:
  --version               Show the version.
  -h, --help              Show help information.

SUBCOMMANDS:
  local                   Localhost Apodini deployment provider
  aws                     AWS Lambda Apodini deployment provider

  See 'test-web-service help deploy <subcommand>' for detailed help.
```

### Testing
No new real test cases. I created a new target `ApodiniDeployTestWebServiceExecutable` that contains the `main()` of `ApodiniDeployTestWebService` to be able to initialise the `TestWebService` as an object. Also added a helper method to conform the the new in-memory web service structure storage.

### Reviewer Nudging
The changes by this PR are not too big. Have a look at `DeploymentCLI` and the changes that have been made to `ApodiniDeployInterfaceExporter`.


Sources/ApodiniDeployBuildSupport/DeploymentProvider.swift
```diff
@@ -60,12 +75,24 @@ public extension DeploymentProvider {
 }
 
 
-struct ApodiniDeployBuildSupportError: Swift.Error {
+public struct ApodiniDeployBuildSupportError: Swift.Error {
     let message: String
+    
+    public init(message: String) {
```
philippzagar, line 80: Is this initializer actually necessary?

Sources/ApodiniDeploymentCLI/AWSLambda/AWSLambdaCLI.swift
```diff
@@ -0,0 +1,87 @@
+//
+//  File.swift
```
philippzagar: Change to AWSLambdaCLI

Sources/ApodiniDeploymentCLI/Localhost/LocalhostCLI.swift
```diff
@@ -0,0 +1,48 @@
+//
+//  File.swift
```
philippzagar: LocalHostCLI.swift

Sources/DeploymentTargetAWSLambda/AWSLambdaDeploymentProvider.swift
```diff
@@ -1,5 +1,5 @@
 //
-//  LambdaDeploymentProviderCLI.swift
+//  LambdaDeploymentProvider.swift
```
philippzagar: AWSLambdaDeploymentProvider.swift (sorry to be that annoying :D )

lukaskollmer (commented):
Some questions before I review this more thoroughly:
- Does this maintain the hard separation between "dependencies needed in order to deploy" and "dependencies needed in order to run when deployed" that currently is in place?
- From looking at the code, it seems like the `ApodiniDeploymentCLI` target (which seems to belong to Apodini) [statically lists the two deployment providers](https://github.com/Apodini/Apodini/blob/f0481eefa19d4072aa71e5f26b0c43edce7a2c6c/Sources/ApodiniDeploymentCLI/DeploymentCLI.swift#L24). But then the DeploymentTestWebService [lists them again](https://github.com/Apodini/Apodini/blob/f0481eefa19d4072aa71e5f26b0c43edce7a2c6c/Sources/ApodiniDeployTestWebService/ApodiniDeployTestWebService.swift#L45). Is that intentional?
- If I want to run a web service normally (i.e. w/out any of the deployment stuff), would the deployment-related code still be present in the executable of the web service? (This is, at least IMO, actually somewhat big of a deal: currently, if you want to run a web service w/out the deployment stuff, none of the deployment-related code/packages/dependencies/etc get compiled and none of it ends up in the web service. That's also the reason why the deployment tasks take longer, because the dependencies only get compiled when they're actually needed.)
- Would it still be possible, at least theoretically, to add/remove deployment providers w/out having to change the source code of Apodini itself? (It seems to me that this is lo longer the case, since the ApodiniDeploymentCLI target explicitly lists all deployment providers, and that target seems to be part of Apodini-proper, rather than some individual web service.) Currently, this is possible. The fact that the deployment providers are located in the same SPM package as Apodini itself is really just a coincidence, these targets are 100% separate from each other. The deployment providers know that Apodini exists, but Apodini doesn't know that/which deployment providers exist.

Sources/ApodiniDeploy/ApodiniDeployInterfaceExporter+ExportWebServiceStructure.swift
```diff
@@ -46,4 +46,37 @@ extension ApodiniDeployInterfaceExporter {
             encoderOutputFormatting: [.prettyPrinted, .withoutEscapingSlashes]
         )
     }
+    
+    func exportWebServiceStructure(apodiniDeployConfiguration: ApodiniDeploy.ExporterConfiguration) throws {
```
philippzagar: Lots of duplicate code here

Sources/ApodiniDeploymentCLI/Localhost/LocalhostCLI.swift
```diff
@@ -0,0 +1,48 @@
...
+            commandName: "local",
+            abstract: "Localhost Apodini deployment provider",
+            discussion: """
+            Deploys an Apodini web service to localhost, mapping the deployed system's nodes to independent processes.
+            """,
+            version: "0.0.2"
+        )
+    }
+    
+    @OptionGroup
```
philippzagar: Nice to use option group 👍

Sources/Apodini/WebService.swift
```diff
@@ -78,6 +78,18 @@ extension WebService {
...
+    /// Runs the `SyntaxTreeVisitor` without starting the web service.
+    public func runSyntaxTreeVisitor() {
+        let app = Application()
+        app.logger.info("Running SynatxTreeVisitor to retrieve web service structure.")
+        app.logger.logLevel = .error
+        self.configuration.configure(app)
+        self.register(
+            SemanticModelBuilder(app)
+        )
+        app.logger.logLevel = .debug
```
philippzagar: Is there a reason for the switch of the loglevel (and back of course)?

Sources/ApodiniDeployBuildSupport/DeploymentProvider.swift
```diff
@@ -24,6 +24,21 @@ public struct DeploymentProviderID: RawRepresentable, Hashable, Equatable, Codab
     }
 }
 
+public struct DeploymentStorage: MemoryStorage {
```
philippzagar: Nice to add this static storage! Maybe add the 'Memory' to the `DeploymentStorage`, so `DeploymentMemoryStorage`. Therefore the purpose is immediately clear. But of course that's just my opinion, if you think otherwise we can definitely leave this as it is.

Sources/ApodiniDeployBuildSupport/DeploymentProvider.swift
```diff
@@ -24,6 +24,21 @@ public struct DeploymentProviderID: RawRepresentable, Hashable, Equatable, Codab
...
+    public static var current = DeploymentStorage()
+    
+    private var object: Object?
+    public typealias Object = WebServiceStructure
```
philippzagar: As a Swift newbie: Couldn't we omit the `Object` typealias completely and let Swift infer the type? Or did you do that on purpose?

philippzagar (commented):
Just noticed: I started reviewing the PR slightly before @lukaskollmer left his remarks. Since he actually has a clue about ApodiniDeploy's inner workings, please refer to his more "conceptual" questions. I just left some comments regarding code improvements

Sources/Apodini/WebService.swift
bauer-andreas: I assume the idea was to restore the previous logLevel state, right?
In that case we would need to properly capture the previous logLevel and not just blindly assume `.debug`.

Sources/ApodiniDeploy/ApodiniDeployInterfaceExporter+ExportWebServiceStructure.swift
```diff
@@ -46,4 +46,37 @@ extension ApodiniDeployInterfaceExporter {
...
+        let deploymentConfig = apodiniDeployConfiguration.config
+        guard let openApiDocument = app.storage.get(OpenAPI.StorageKey.self)?.document else {
+            #if !DEBUG && !RELEASE_TESTING
+                throw ApodiniDeployError(message: "Unable to get OpenAPI document")
```
bauer-andreas: why are we unable to retrieve the OpenAPI document when compiled in release?

Tests/ApodiniDeployTests/LocalhostDeploymentProviderTests.swift
```diff
@@ -11,6 +11,7 @@ import FoundationNetworking
 #endif
 import XCTApodini
 import ApodiniUtils
+import ApodiniDeployTestWebService
```
bauer-andreas: Not sure if this import gets used?

Tests/ApodiniDeployTests/WebServiceStructureExportTests.swift
```diff
@@ -14,6 +14,7 @@ import ApodiniUtils
 @testable import ApodiniDeployBuildSupport
 import DeploymentTargetLocalhostCommon
 import DeploymentTargetAWSLambdaCommon
+import ApodiniDeployTestWebService
```
bauer-andreas: Am I right that the only reason why the `ApodiniDeployTestWebService` got split into `ApodiniDeployTestWebService` library and `ApodiniDeployTestWebServiceExcutable` is for those testing purposes here?

I think the only reason why we have a dedicated deployment test web service is for e.g. the `LocalhostDeploymentProviderTests` aiming at a complete test coverage from compilation to execution.

Now, that all this happens in a single process, can this be simplified? 
For one, I don't think we need to further split the target into two. Just define a simple TestWebService testing against the web service structure retrieval.
Second, can we rewrite the tests such that the compilation step isn't necessary anymore (and therefore the sprat test executable can be removed completely) or did I misunderstand this refactoring?

bauer-andreas (commented):
I think this refactoring has the potential to greatly simplifying the ApodiniDeploy API.
Though I think @lukaskollmer already listed some points for which we need to balance the pros and cons.

Additionally, @lukaskollmer, can we potentially with the structural refactoring this PR makes get to a point where the DeploymentProvider (terminology?) does not need to execute a `swift build` command at all?
I'm not sure of the implications, but can we in that case already rly on the fact that the executable target got already compiled (or maybe otherwise force the user to supply the url to an already compiled target)?
I think doing the `swift build` command transparently removes a lot of flexibility. It already shows as currently no single compiler flag is forwarded to the command (e.g. release configuration custom `-Xswiftc` arguments etc). With #294 I even think, that ApodiniDeploy doesn't really work anymore, cause with the required workaround (for linux platform) introduced there it isn't possible to specify those compiler arguments(?).
Also the fact to not require the swift package manager to be in place might also introduce some room for flexibility?

Sources/Apodini/WebService.swift
philippzagar: But why change it in the first place? Was the goal to "don't care" about the logs if we just require the structure and don't startup the webservice?
As of now, we specify .debug as the log level for the application logger (non-configurable). Therefore I don't see a huge problem in assuming .debug, but of course the cleaner and more future-proof solution would be to save the previous level and then restore it :)

bauer-andreas: > But why change it in the first place?

I think, it's a matter of silencing otherwise confusing logs? But not sure. Might be confusing if you try to deploy the web service and see log information which implies a started web service, but actually doesn't start the web service(?).

> As of now, we specify .debug as the log level for the application logger (non-configurable).

I think its only `.debug` if compiled in debug configuration? In release configuration it is a less chatty logger, if I remember correctly and it wasn't changed 🙃

philippzagar: Yep i thought so too. And in the "old" implementation we didn't care about the, as you put it, chatty logger, since this was executed in its own process?

You're right, i forgot about that. Then the storing of the old level is definitely required.

bauer-andreas: >  And in the "old" implementation we didn't care about the, as you put it, chatty logger, since this was executed in its own process?

Actually I have no clue. But thinking about it it might be that it was indeed not silenced. @lukaskollmer ?
Maybe we just wait for @hendesi to get the rationale behind that 😅

hendesi:
@lukaskollmer 
> Does this maintain the hard separation between "dependencies needed in order to deploy" and "dependencies needed in order to run when deployed" that currently is in place?

I'm not 100% sure if i understand you correctly, but as far as I can see, yes it does. `ApodiniDeploymentCLI` should only need the deployment target libraries to work.
 
> From looking at the code, it seems like the ApodiniDeploymentCLI target (which seems to belong to Apodini) statically lists the two deployment providers. But then the DeploymentTestWebService lists them again. Is that intentional?

The idea was to give the user the possibility to manually add the deployment cli to their custom defined `CommandConfiguration`. But I get your point, it should be possible to automatically add the cli commands for the deployment runtimes defined in the deployment configuration.

> If I want to run a web service normally (i.e. w/out any of the deployment stuff), would the deployment-related code still be present in the executable of the web service? (This is, at least IMO, actually somewhat big of a deal: currently, if you want to run a web service w/out the deployment stuff, none of the deployment-related code/packages/dependencies/etc get compiled and none of it ends up in the web service. That's also the reason why the deployment tasks take longer, because the dependencies only get compiled when they're actually needed.)

If you don't add the deployment cli or a deployment configuration, none of the deployment code should be needed or compiled.

> Would it still be possible, at least theoretically, to add/remove deployment providers w/out having to change the source code of Apodini itself? (It seems to me that this is lo longer the case, since the ApodiniDeploymentCLI target explicitly lists all deployment providers, and that target seems to be part of Apodini-proper, rather than some individual web service.) Currently, this is possible. The fact that the deployment providers are located in the same SPM package as Apodini itself is really just a coincidence, these targets are 100% separate from each other. The deployment providers know that Apodini exists, but Apodini doesn't know that/which deployment providers exist.

That is a good point. Currently I just added the providers to the deployment cli by hard, but this is something that already crossed my mind. There should definitely be the possibility to add custom providers to use them in the cli. I will provide an interface for this. Regarding the dependencies, Apodini still doesn't know about any deployment providers, but we try to bundle the already available providers in `ApodiniDeploymentCLI ` to use it with the new Apodini cli. As already said, there should definitely be a way to extend it with other providers though.

Sources/Apodini/WebService.swift
hendesi: The idea was to minimise the output during the synatax tree visit. Since this is just to retrieve the web service structure and not actually starting the web service, i wanted it to be as quite as possible 😅 
But I don't have a super strong opinion about this, so i can also just revert this. If we keep it, then you're right @Supereg  , it should be reverted to the correct level

Sources/ApodiniDeployBuildSupport/DeploymentProvider.swift
hendesi: You're right. Will remove that!

Sources/ApodiniDeployBuildSupport/DeploymentProvider.swift
lukaskollmer: Yes, because the `message` property is internal (which will synthesize an `internal init(message:)`), but we want the initializer to be public

Sources/ApodiniDeploy/ApodiniDeployInterfaceExporter+ExportWebServiceStructure.swift
hendesi: Some tests were failing because this error was thrown. I just added this as a quick fix to investigate this later, which i seemingly forgot. Thanks for reminding me :D. Will have a closer a look at the test cases, also regarding your comment below

lukaskollmer:
@Supereg As for not using `swift build` and instead using an already-compiled executable. That's already possible although I think that it currently is used only internally for testing the localhost deployments. 
The main issue here is that this is specific to each deployment provider: w/ the localhost one it'd be relatively easy to get rid of the `swift build` step (and instead using a user-supplied executable); but the lambda one would not be that simple, since then the user would be responsible for compiling the web service executable for the lambda architecture (this is the step currently taking place inside a docker container)

lukaskollmer:
> > Does this maintain the hard separation between "dependencies needed in order to deploy" and "dependencies needed in order to run when deployed" that currently is in place?
> 
> I'm not 100% sure if i understand you correctly, but as far as I can see, yes it does. `ApodiniDeploymentCLI` should only need the deployment target libraries to work.

@hendesi
There is an additional aspect to it: Currently, since the deployment provider and the web service are distinct programs, they can have separate dependencies (example: the lambda DP depends on a bunch of AWS packages. These packages are linked into the lambda deployment provider's executable, but they are not linked into the web service executable uploaded to the Lambda functions (since these two are entirely separate programs and therefore do not share their dependencies)).
The idea behind this was to compile everything in a way that every executable/library contains only the dependencies required for its specific use case. That's why, for example, the deployment providers are split into two components (this is realised by implementing them as two separate SPM targets), with one being responsible for performing the deployment tasks (i.e. determining the structure of the deployed web service, deploying it, etc) and the other then being responsible for handling how it behaves at run-time.

Note that I'm not saying that this current structure should remain in place (tbh I'm quite fond of the idea of turning the deployment tasks into SPM-package-supplied commands, which can then be dynamically loaded into a web service), but I do think it's important to consider the trade-offs here

Sources/Apodini/WebService.swift
hendesi: I just reverted the changes made to the log level. There is still an `info` msg that indicates the reason for the output, so this should be fine, I guess

hendesi:
@philippzagar @Supereg @lukaskollmer 
I think I addressed all requests and comments. Can you give this please another review?

@lukaskollmer 
Regarding the bundling of the dependencies that is introduced by this PR:
We talked about this and the tradeoffs in the Apodini group meeting on tuesday and came to the conclusion, that have a unified and clear CLI that also includes deployment options outweighs the downsides of having to bundle all dependencies. Especially in the deployment context, a slightly larger binary does not make too much of a difference. 
What do you think about this? Are you ok with this or is there something you highly disagree with?

lukaskollmer:
@hendesi I'm fine with this. One remaining question though: we recently added some subcommands to the AWS deployment provider, would these still work with this refactor?
In particular, two of the subcommands are intended to be run when there's no web service present (`AWSDeploymentProvider remove-deployment` and `AWSDeploymentProvider delete-iam-roles`).
How would these commands need to be integrated with the new system in this PR? Would it still be possible to run them even if there's no web service present?

hendesi:
@lukaskollmer 
As of now the new commands are integrate with deployment cli of this PR and can be used by `deployment lambda remove-deployment .. `. But since the deployment cli uses the `WebService.configuration` it would only work with a web service. 
I see that this is not optimal, so my first suggestion would be to create a separate target that contains all web service agnostic  commands. Since we wanted to create some database commands as well that don't need a web service, we could bundle them in some `ApodiniCLI` target.

Sources/DeploymentTargetLocalhost/LocalhostDeploymentProvider.swift
```diff
@@ -115,6 +88,10 @@ public struct LocalhostDeploymentProvider: DeploymentProvider {
                 }
             }
             try task.launchAsync(taskTerminationHandler)
+//            let stdioObserverHandle = task.observeOutput({ (type, data, task) in
+//                let text = String(data: data, encoding: .utf8)
+//                print(text)
+//            })
```
bauer-andreas, line 94: Please remove commented out code.
Didn't we have a swift lint rule for that?

bauer-andreas (approved):
Looks good to me 👍

But @lukaskollmer is probably better in judging the ApodiniDeploy specific stuff ☺️

philippzagar (approved):
From my side, the PR is ready, but to repeat the argument of @Supereg, @lukaskollmer`s review is probably way more significant :)

PR #310, opened 2021-07-04
Infrastructure preparations for Apodini Authorization: Metadata with Delegating Handlers + Information API

bauer-andreas:
# Infrastructure preparations for Apodini Authorization: Metadata with Delegating Handlers + Information API

As of now Apodini fails to meet the following two requirements needed for the Apodini Authorization package:
- The `DelegatingHandlerInitializerVisitor` doesn't account for new values added to the `DelegatingHandlerContextKey` when visiting the individual `Handler`s. This is required for ApodiniAuthorization, as the Metadata must be capable of bootstrapping a `DeleatingHandler` responsible for enforcing the authorization requirements.
- The current state of the Information API #301 does not deliver the expected flexibility and extensibility. Details are explained below.
At this point the ApodiniAuthorization framework doesn't really directly build upon the Information API. However this changes benefit the project in any way.

### Shortcomings of the Information API

* In general a global string-based identifier namespace is not easy to maintain and avoid collisions across first-party and third-party packages (specifically as potential third-party packages are not under our control).
* In addition, it is even harder as the current approach is very HTTP dependent and shares its namespace with the HTTP Header namespace.
* Therefore, any addition to the Information API must follow the same rules as imposed for HTTP Header naming (X prefix etc)
* Further, the approach doesn't allow to easily extend the set of preexisting HTTP Header implementation (currently including `Authorization`, `RedirectTo`, etc ...) with currently non existent HTTP Header implementations. Any addition requires a change in the initializer of the `AnyInformation` struct, always requires a change in the Apodini Core package.
* Consequentially, this approach exposes HTTP specific implementations and concepts (like Basic Authentication, Bearer Authentication, Cookies, ETags) in the core Apodini Package.
* Adding support for non HTTP Header based Information is unintuitive for the above illustrated reasons (adhering to the HTTP Header namespace and conventions and declaring those Information instances in the core Apodini Package and further requiring that the value of a Information can always be constructed to and from a string).
* The REST exporter maps Information instances 1:1 to HTTP Headers (in both directions). Meaning every Information must be a valid HTTP Header. On the other hand, there are several exporters which do not have a according mapping for Information to the appropriate wire protocol (WebSockets, gRPC, ...).

## :bulb: Proposed solution

This PR addresses those issues with the following changes:
- The `DelegatingHandlerInitializerVisitor` was adjusted to recalculated the value for the `DelegatingHandlerContextKey` and checking for changes. Further the two protocols `DefinitionWithDelegatingHandler` and `DefinitionWithDelegatingHandlerKey` are provided to easily add `DelegatingHandler`s to a `MetadataDefinition`. The _Provider Guide_ was updated to include coverage of those additions. Those changes are reflected in commit 
d07478e. Further see https://github.com/Apodini/Apodini/pull/310#issuecomment-877351945.
- The Information API was rethought to address the above-illustrated points. The changes are highlighted below in greater details. Those changes are reflects in commit 13eeb9a.

### Addressing the Information API

* The core Apodini Package was refactored to only include the necessary infrastructure code for the Information API. An Information instance builds upon the modified `Information` protocol. Instead of relying on a String for unique identification, the `Information` protocol requires a type based identification via `InformationKey` (A `InformationKey` is required to be `Hashable` for internal storage).
* The PR introduces the concept of `Information Classes`. Every `Information` instance must conform to at least one `InformationClass`.
`InformationClasses` are Exporter (or rather protocol) defined, as they model a dedicated Protocol Metadata (like HTTP Headers). Having this protocol based `InformationClass` approach allows, that any Information implementation may conform to any of the available Information classes.
* For `HTTPHeaders` specifically, the VaporSupport package introduces the `HTTPHeaderInformationClass`, along with the `AnyHTTPInformation` `Information` implementation (which is is part of the `HTTPHeaderInformationClass`).
In regards to migration, `AnyHTTPInformation` pretty much replaces the previous `AnyInformation`.
Further, there are several preexisting http header implementations (like `Authorization`, `Cookies`, `ETag`, `Expires`, `RedirectTo`) which represent `InformationInstantiatable` (transitively using the `HTTPInformation`).
`InformationInstantiatable` represent a way of providing a typed implementation for a predefined `Information` instance (in this case the `AnyHTTPInformation`).
The great thing about how this approach was implemented is, that anyone can write their own `HTTPInformation` conforming implementation (which can be placed in any third-party package) and use it seamlessly to retrieve a type representation of an HTTP header of the request. This addresses the non-extensibility of the current approach.
* Lastly, this PR adds support for passing Information instances to `@Throws` and `ApodiniError` instances, which was previously impossible.

## :gear: Release Notes 

This PR introduces some breaking changes in the Information API.

**Accessing Information**

<details>
<summary>Click to expand!</summary>

Previously you would write 
```swift
import Apodini

struct SomeHandler: Handler {
  @Environment(\.connection) var connection

  func handle() -> Response {
    let informationTyped = connection.information[Authorization.self]
    let untypedInformation = connection.information["SomeHTTPHeader"]
  }
}
```

This changed to:
```swift
import Apodini
// access to the typed version of e.g. Authorization requires an import now
import ApodiniREST // or import the respective HTTP based exporter you are using

struct SomeHandler: Handler {
  @Environment(\.connection) var connection

  func handle() -> Response {
    let informationTyped = connection.information[Authorization.self]

    // for accessing http header Information via a string identifier, the external parameter name `httpHeader` is now required
    let untypedInformation = connection.information[httpHeader: "SomeHTTPHeader"]
  }
}
```

</details>

**Returning Information**

<details>
<summary>Click to expand!</summary>

Previously you would write 
```swift
import Apodini

struct SomeHandler: Handler {
  func handle() throws -> Response<Int> {
    .send(42, status: .ok, information: [
        .authorization(.basic(username: "PaulSchmiedmayer", password: "SuperSecretPassword"))
    ])
  }
}
```

This changed to:
```swift
import Apodini
// access to the typed version of e.g. Authorization requires an import now
import ApodiniREST // or import the respective HTTP based exporter you are using

struct SomeHandler: Handler {
  func handle() throws -> Response<Int> {
    .send(42, status: .ok, information: [
        // instead of the method accessor, you just call the initializer directly
        Authorization(.basic(username: "YourUsername", password: "SuperSecretPassword"))
    ])
  }
}
```

</details>

**Writing your own typed version of an HTTP Header**

This PR added the ability to add your own Typed version of a protocol metadata (e.g. an HTTP header implementation).
For the example of HTTP Headers you just need to conform to the `HTTPInformation` protocol and then you can use your type as you would do any other Information instance.

### Related PRs
* This PR builds upon the #281 and the #301 

### Testing
- New test cases were added covering Metadata which bootstraps `DelegatingHandler`s.
- Additional test vectors for changes in the Information API are to be added.

### Reviewer Nudging

The commits d07478e and 13eeb9a can be reviewed individually.

- **Delegating Handlers**: I have added an additional section to the Metadata `ProviderGuide` detailing how to use the new additions. Specifically the addition in the `DelegatingHandlerInitializerVisitor` contains the important changes: https://github.com/Apodini/Apodini/blob/125df1550a119ce8e32d1c45853a00366ab34d62/Sources/Apodini/SyntaxTreeVisitor/DelegatingHandlerInitializerVisitor.swift#L45-L64
- **Information API**: A review of the changes to the Information API might require to make oneself familiar with the original version #301. Further make sure you understand the problems illustrated above. The propose solution section might already give you some guidance what the solution domain tried to achieve and how it works. For the review it might make sense to go along the package hierarchy, e.g. starting to review all changes in the `Apodini` package, then `VaporSupport` and lastly `REST` and so on.


bauer-andreas:
@PSchmiedmayer as you introduced the original Information API, it might make sense if you have a look at the PR (if you want) as it contains some infrastructural changes to the API. Not sure if I could preserve your vision behind that. Just as an fyi.
Depending on your point of view we might want to further discuss this change in the next Meeting.

Sources/Apodini/Information/AnyInformationEntry.swift
```diff
@@ -0,0 +1,164 @@
...
+            fatalError("Tried typing an AnyInformation which was instantiated from key only")
+        }
+
+        return information as? T
+    }
+
+    /// Returns the type version of the wrapped `Information` instance.
+    /// - Parameter type: The `Information` type
+    /// - Returns: Returns the casted `Information` instance.
+    public func typed<T: Information>(to type: T.Type = T.self) -> T {
```
theMomax: Unnecessary code duplication with `tryTyped`

Sources/Apodini/Information/AnyInformationEntry.swift
```diff
@@ -0,0 +1,164 @@
...
+        }
+        guard let typed = information as? T else {
+            fatalError("Tried typing AnyInformation with type \(information.self) to \(T.self)")
+        }
+
+        return typed
+    }
+
+    /// - Returns: Returns the wrapped information instance as a type erased `AnyInformation`
+    public func any() -> AnyInformation {
```
theMomax:
```suggestion
    public var information: AnyInformation {
```

Sources/Apodini/Information/DynamicInformationInstantiatable.swift
```diff
@@ -0,0 +1,24 @@
+//
+// Created by Andreas Bauer on 03.07.21.
+//
+
+/// A `DynamicInformationInstantiatable` represents a special kind of `Information`
+/// which can be instantiated from a predefined `InformationWithDynamicKey`.
+///
+/// This can be used to e.g. create a `InformationWithDynamicKey` which holds the
```
theMomax: Definitely not mandatory yet, but you could consider starting to use the new DocC syntax already. Basically just make it double backticks (only works if the type is defined in the same target, though).

Sources/Apodini/Information/Information.swift
```diff
@@ -7,36 +7,54 @@
 
 import Foundation
 
+/// Describes a type erased version of an `Information` instance.
+public protocol AnyInformation {
+    /// Internal method turning an `Information` instance into a `AnyInformationEntry`.
+    func entry() -> AnyInformationEntry
```
theMomax: It is public, though. Couldn't you conform both `Information` and `InformationKey` to `AnyInformationEntry`? That could also potentially eliminate the need for using AssociatedTypeRequirementsVisitor once more.

Sources/Apodini/Information/Information.swift
```diff
@@ -7,36 +7,54 @@
...
+    subscript<T>(dynamicMember keyPath: KeyPath<Value, T>) -> T {
         value[keyPath: keyPath]
     }
 }
+
+// MARK: Default Key
+public extension Information {
+    /// By default, the `ObjectIdentifier` of the implementing `Information` type will be used.
+    static var key: ObjectIdentifier {
+        ObjectIdentifier(Self.self)
```
theMomax: Is that a sensible choice? Is there any `Information` that doesn't have to be encoded to the network stack? The `ObjectIdentifier` changes at latest when you recompile, making it pretty much useless for network communication. As an alternative on could use the string of the type name or just not provide a default implementation.

Sources/Apodini/Information/Set+Information.swift
```diff
@@ -5,36 +5,99 @@
...
+        fatalError("""
+                   A Information with a dynamic key cannot be queried using the type. \
+                   Instead use one of the `InformationKey` based subscripts or a application defined one.
+                   """)
+    }
+
+    /// Returns the value of an `InformationWithDynamicKey` instance based
+    /// on the passed `DynamicInformationInstantiatable` type.
+    /// - Parameter key: The `DynamicInformationInstantiatable` type that is requested.
+    /// - Returns: An instance of `DynamicInformationInstantiatable.Value` if a according `InformationWithDynamicKey`
```
theMomax: an

Sources/Apodini/Metadata/MetadataWithDelegatingHandler.swift
```diff
@@ -0,0 +1,65 @@
...
+//
+
+/// A type erased `DefinitionWithDelegatingHandler`.
+public protocol AnyDefinitionWithDynamicDelegatingHandler {
+    /// Internal method used to add the `DelegatingHandlerInitializer` provided
+    /// by the `DefinitionWithDelegatingHandler` to the `ContextNode`.
+    ///
+    /// Note, this method does nothing, if self also conforms to `DefinitionWithDelegatingHandlerKey`.
+    /// - Parameter visitor: The `SyntaxTreeVisitor`.
+    func addInitializerContextValue(_ visitor: SyntaxTreeVisitor)
```
theMomax, line 12: If I get this correctly, this is just for metadata definitions that want to manipulate both, some other ContextKey and the DelegatingHandlerContextKey? If so, wouldn't it be better to provide a more generic way for a metadata definition to manipulate any number of different ContextKeys? Also, maybe it would be better to prefix this protocol with an underscore if it can't be omitted or made internal.

Sources/Apodini/Properties/Throws.swift
```diff
@@ -11,34 +11,65 @@ import Foundation
...
+    /// - Parameter `description`: The **internal** description of this error. This will only be exposed in `DEBUG` mode.
+    /// - Parameter `information`: Possible array of `Information` entries attached to the `Response`.
+    /// - Parameter `options`: Possible exporter-specific options that provide guidance for how to handle this error.
+    public init(
+        _ type: ErrorType,
+        reason: String? = nil,
+        description: String? = nil,
+        information: AnyInformation...,
+        options: ApodiniError.Option...
+    ) {
```
theMomax, line 65: I am fine with adding an extra convenience initializer where you can pass in information directly, but I would suggest to make InformationSet a PropertyOption, so we keep the type as slim as possible. We have so many strong generic patterns in Apodini but tend to not use them for implementing our internal features.

Sources/Apodini/Response/ApodiniError.swift
```diff
@@ -31,20 +31,29 @@ public struct ApodiniError: Error {
...
     
     private let `type`: ErrorType
-    
+
     private let reason: String?
-    
+
     private let description: String?
+
+    public let information: InformationSet
```
theMomax, line 39: This is the same as for `Throws`. I would suggest to make the `InformationSet` part of the `options`.

Sources/Apodini/SyntaxTreeVisitor/DelegatingHandlerInitializerVisitor.swift
```diff
@@ -0,0 +1,65 @@
+//
+// Created by Andreas Bauer on 02.07.21.
+//
+
+import ApodiniUtils
+
+class DelegatingHandlerInitializerVisitor: HandlerVisitor {
```
theMomax, line 9: Seems about right👍🏻

theMomax (changes requested):
While I really like the new capabilities introduced with this PR (both in regards to Metadata bootstrapping DelegatingHandler AND Information), I think that some parts could be implemented in an even more generic way or could be implemented so that they expose less internals.

Specifically, you could treat Information as a Map instead of a Set, which should help to get rid of AnyInformationEntry. 

I find that explicit differentiation between `InformationWithDynamicKey` and `Information` pretty unintuitive, thus I would suggest to make everything dynamic. However, if you need this differentiation to provide the implicit unwrapping of AnyHTTPInformation, then I'd suggest to keep that feature over my initial concern.

The last point is about the integration of how Metadata injects `DelegatingHandlerInitializer`s into the Context. I'd like to see a more generic approach, that generally allows Metadata to manipulate multiple different context keys.

Sources/Apodini/Information/Information.swift
bauer-andreas: AnyInformationEntry is a struct not a protocol.
Though it should ne possible to just remove this requirement in the public protocol definition and just declare the extension with he implementation internal, as this func is used internally only.

Further, when I rethink the AnyInformationEntry, I may be able to remove the need for this function anyways 🤔

Sources/Apodini/Information/DynamicInformationInstantiatable.swift
bauer-andreas: I'll try to address that 👍

Sources/Apodini/Information/Information.swift
bauer-andreas: That is exactly the point I flagged as "Up to discuss" in my `Addressing the Information API` section of the PR description.

With this PR, we allow for `Information` instances which might not be mapped directly to the corresponding wire format (simply stemming from the fact that we don't know the format of the wire format). Therefore an Exporter will need to explicitly support every `Information` type.

Giving some examples:
* HTTP based exporters provide `AnyHTTPInformation` instances, which HTTP based exporters will obviously map to Headers and response (and vice versa in requests). However other exporters potentially can't make that 1:1 mapping (simply as HTTP Headers are a protocol feature). One could image a WebSocket exporter to map `AnyHTTPInformation` to some json field. For Exporters like an GraphQL or gRPC exporter these informations might not make sense at all(?).  
Therefore this PR makes it clear, that **not** mapping a certain Information is perfectly okay. Handler code should/has to/can be written in a way that accommodates for cases where a certain Information is missing (simply as the subscript returns an Optional). So this direction probably makes sense. The reverse direction might be the one which can be a little bit unintuitive, as there is no guarantee that an Information is mapped into the Response **for all** exporters.  
But that was the case already, in the current version the REST exporter is the only one which maps `Information` instances into the response, all others ignore them. This PR just tries to make that clear a priori.
* Another example, which directly addresses your comment of why `ObjectIdentifier` makes sense even though there is no direct way of mapping it to the respective wire protocol, are `Information` instances which are the result of some translation/mapping from or to the respective wire protocol.
My use case with `ApodiniAuthorization` would be a dedicated `ApodiniAuthorizationInformation` (has nothing to do with the http based `Authorization` information), which can be used to map any of the support authentication types (e.g. a token based or credential based auth) to a generic Information instance which is processable by `ApodiniAuthorization`. Analogous the `ApodiniAuthorization` framework would inject a corresponding Information instance into the response, which the exporter needs to map to the according wire format (the response might be used to transport some multi step authorization information or to annotate e.g. an error with information about the failed authorization).
Again there is the argument, that the Exporter is required to manually add support for this `Information` type (if the respective wire format has in-built authorization mechanism, like http does).

Sources/Apodini/Metadata/MetadataWithDelegatingHandler.swift
bauer-andreas: I have already thought about this as well. I don't think there is a good way of doing that elegantly.
Of course there is always the way of overriding the `accept` yourself. So its not impossible to do it, its just that it remove a lot of convenience if you have to always implement the `accept` method yourself. And as Metadata will typically contribute exactly one value (or e.g. merge multiple values in a struct) I chose to make that typical use case convenient, but the untypical use case is still not impossible.

But feel free to leave any suggestions.

EDIT: An alternative would be to create individual Metadata Definitions for each ContextKey and compose them via a MetadataBlock.

Sources/Apodini/Properties/Throws.swift
bauer-andreas: Not sure about that one. I think it is important to distinguish the uses of Options and Information.
I looked at `Options` (in the case of ApodiniError) always as a way of supplying metadata for the exporters which gives guidance **how** to handle a given request, as it states in the documentation: https://github.com/Apodini/Apodini/blob/864a13505d2a5cdedca487a1e49dde2cf4cbd094/Sources/Apodini/Response/ApodiniError.swift#L53

In contrast `Information` actually delivers **content** for the generated response. They provide something instead of just giving guidance.

So I'm not sure about declaring that Informations are an Option just for the sake of removing a stored property(?).

Sources/Apodini/Information/Information.swift
bauer-andreas: @theMomax Maybe to reflect on the second point presented here. Could the same approach be realized using the `Environment`, e.g. using the `@EnvironmentObject` or some keyed environment entry? I assume it would be possible to pass the generic `Authorization` object down the handler evaluation using the environment, but the environment cannot be used to "return" such an environment object back to the Exporter right?

Though rethinking my thought (lol), would @EnvironmentObject be able to gracefully handle cases where the according object isn't present (e.g how the subscript always returns the `Optional<Value>` and the respective handler needs to properly handle cases where the Information is not present).

Sources/Apodini/Information/Information.swift
theMomax: Yes, that was more of the direction I was thinking off. Either make `AnyInformationEntry` a protocol or remove it entirely by using a map for storage instead of a set.

Sources/Apodini/Information/Information.swift
theMomax: Just to mention the result from today's discussion here. The most desirable approach would be to have only basic information on the InformationSet (e.g. a token, properties required for http basic auth). A stack of delegating handlers would then one-by-one access this basic information and transform it (in steps to be defined by you) to something like an authenticated user object. Each delegating handler can either use `@Environment` or `@EncvironmentObject`s to deliver the generated information to the upper `Handler`s.

The nature of the Information-keys is not really resolved. @PSchmiedmayer would vote for just using `String`. I would propose a system where each exporter defines a public protocol of Keys it can handle. E.g. WebSocket would accept any key that can provide a `String` representation. All the HTTP based exporters would probably want a `ValidHTTPHeaderStringProvidingKey` protocol or similar. When defining an `Information`, the creator of this `Information` should conform it to the "well known" key-protocols that are used by exporters where it actually makes sense to expose this type of information.

Sources/Apodini/Metadata/MetadataWithDelegatingHandler.swift
theMomax: Maybe with our new idea of having a stack of delegating Handlers (mentioned [here](https://github.com/Apodini/Apodini/pull/310#discussion_r664470138)), the MetadataBlock would indeed be the best solution. Otherwise, I'd be fine with leaving it as is, since it is generic to `DelegatingHandlerInitializer`s and we don't have discovered this requirement for other Metadata yet.

Sources/Apodini/Properties/Throws.swift
theMomax: Good point. I guess the lines are a bit blurry around things like status codes, but taking that argument into account I am fine with leaving it as is.

Sources/Apodini/Information/Information.swift
bauer-andreas: @PSchmiedmayer I would agree with @theMomax that an approach that is based on InformationKey "classes" would solve the above described issues but would be generic enough at the same time.
A key might be mapped to string or something else. In any case through the introduction of those classes we can easily achieve namespace separation even when the resulting value type (e.g. String) is the same.

The key of a concrete Information instance would then declare conformance to one or multiple key classes, making it available in the according exporters. After all the actual Information implementation is Exporter specific (or at least middleware specific) anyways.

Information basically maps certain metadata of a dedicated protocol into Apodini. Therefore any type of Wire format would declare a InformationKey class if applicable (e.g. the VaporSupport package would declare one for e.g. HTTP Headers).

Every Exporter would then provide the according Information Instances for the applicable protocol Metadata (e.g. HTTP Headers) and would consequentially also only include Information which specifically was declared valid in the response.

Sources/Apodini/Metadata/MetadataWithDelegatingHandler.swift
bauer-andreas: I think the DelegatingHandlerInitializer Metadata are special in the way that you may need to pass in some initializer arguments, which wouldn't work that easy when using a `MetadataBlock`.
Therefore I would leave those protocols in for now. Maybe we can remove one of those in the future again I we see that we don't really need it.

Sources/Apodini/Information/Information.swift
bauer-andreas: Removed with the latest refactoring.

Sources/Apodini/Information/DynamicInformationInstantiatable.swift
bauer-andreas: All of the docs I rewritten in the latest commit adhered to the DocC syntax 👍

Sources/Apodini/Information/Information.swift
bauer-andreas: So, the latest commit implements the concept as envisioned by @theMomax and me.

bauer-andreas:
Squashed and rebased onto latest `develop`.

bauer-andreas:
@theMomax implemented the changes we discussed today.

* Removed the `prepend` property of the `DelegatingHandlerContextKey` entries as they do not really have an effect, and cause issues when parsing metadata modifiers which add Initializers. In the future we may introduce some mechanism to define a order for modifiers spanning across the Component tree
* Add the `ensureInitializerTypeUniqueness` property to the entires of a `DelegatingHandlerContextKey` to enforce that a certain type of initializer must only instantiated once (on first occasion). This is to support Metadata which only wants to contribute one "global" DelegatingHandlerInitializer (and is e.g. important for ApodiniAuthorization)
* The `DelegatingHandlerInitializerVisitor` will now check for `Delegate`s on the Handler containing Handlers and parses the given Metadata. As discussed, at the current state, we can't support the scenario where DelegatingHandlers get transparently inserted between the Delegating Handler and the Delegated Handler (used to e.g. define really fine granular authorization requirements)
* Thus PR fixes an issue were a Handlers metadata would be parsed multiple times if it was annotated with a `resetGuard` Modifier.

Sources/Apodini/Information/Information.swift
```diff
@@ -5,38 +5,124 @@
...
+/// and make them accessible inside an ``Handler``.
+/// ``Information`` instances are stored in the ``InformationSet`` of a `Request` or ``Response``.
+///
+/// The following examples assume the existence of `ExampleInformation` (which is a ``InformationInstantiatable``)
+/// and `AnyExampleInformation` sharing the key `ExampleInformationKey`.
+/// Additionally we assume the custom defined subscript overload on the ``InformationSet`` `InformationSet[example: String]`.
+///
+/// ## Retrieving ``Information`` instances
+/// ```swift
+/// struct ExampleHandler: Handler {
```
theMomax, line 19: Nice docs!

Sources/Apodini/Information/InformationClass.swift
```diff
@@ -0,0 +1,54 @@
+//
+// Created by Andreas Bauer on 06.07.21.
+//
+
+/// A ```InformationClass``` is a way to group multiple, similar ``Information`` types
+/// which represent the same Metadata information and can be mapped to the same type of
+/// key and value.
+/// A ``InformationClass`` is typically defined for a specific ``InterfaceExporter`` or a group
```
theMomax, line 9: Not sure, but I think you have to add a blank line here to start the DocC Discussion section.

Sources/Apodini/Information/InformationClass.swift
```diff
@@ -0,0 +1,54 @@
...
+/// ``InterfaceExporter`` sharing the same middleware.
```
theMomax: a group **of** ``InterfaceExporter``**s**

Sources/Apodini/Information/InformationClass.swift
```diff
@@ -0,0 +1,54 @@
...
+/// The `entry` tuple matches the key and value types.
+///
+/// ## Declaring a ``Information`` to be part of a ``InformationClass``
+/// Now we want to declare the `AnyExampleInformation` to be part of the above-defined `ExampleInformationClass`.
+///
+/// ```swift
+/// // declaring protocol conformance
+/// extension AnyExampleInformation: ExampleInformationClass {}
+///
+/// // We declare the actual implementation of the class as an extension on the `InformationClass` itself.
```
theMomax, line 36: Double ticks

Sources/Apodini/Information/InformationClass.swift
```diff
@@ -0,0 +1,54 @@
...
+/// // This allows us to avoid any name conflicts on the `entry` property when conforming to multiple information classes.
+/// public extension ExampleInformationClass where Self == AnyExampleInformation {
+///     var entry: (key: String, value: String) {
+///         (key: self.key.key, value: self.value)
+///     }
+/// }
+/// ```
+///
+/// ## Filtering the ``InformationSet`` for a given ``InformationClass``
+/// An export could filter for a dedicated ``InformationClass`` like the following:
```
theMomax: export**er**?

Sources/Apodini/Modifier/DelegationFilterModifier.swift
```diff
@@ -11,11 +11,10 @@ import Foundation
 
 extension Component {
     /// Use a `DelegationFilter` to filter out `DelegatingHandlerInitializer`s for the contained sub-tree.
-    /// - Parameter prepend: If set to `true`, the modifier is prepended to all other calls to `reset` or `delegated`
-    ///                      instead of being appended as usual.
-    /// - Note: `prepend` should only be used if `I.Response` is `Self.Response` or `Self` is no `Handler`.
-    public func reset(using filter: DelegationFilter, prepend: Bool = false) -> DelegationFilterModifier<Self> {
-        DelegationFilterModifier(self, filter: AnyDelegateFilter(filter: filter), prepend: prepend)
+    /// - Parameter ensureInitializerTypeUniqueness: If set to true, it is ensure that the same ``DelegationFilter``,
+    ///     even though when inserted multiple times into the context, is only used a single time (the first time it got added).
```
theMomax: Fix grammar

Sources/Apodini/Modifier/DelegationModifier.swift
```diff
@@ -7,16 +7,18 @@
...
 
 extension Component {
     /// Use a `DelegatingHandlerInitializer` to create a fitting delegating `Handler` for each of the `Component`'s endpoints.
     /// All instances created by the `initializer` can delegate evaluations to their respective child-`Handler` using `Delegate`.
-    /// - Parameter prepend: If set to `true`, the modifier is prepended to all other calls to `delegated` instead of being appended as usual.
-    /// - Note: `prepend` should only be used if `I.Response` is `Self.Response` or `Self` is no `Handler`.
-    public func delegated<I: DelegatingHandlerInitializer>(by initializer: I, prepend: Bool = false) -> DelegationModifier<Self, I> {
-        DelegationModifier(self, initializer: initializer, prepend: prepend)
+    /// - Parameter ensureInitializerTypeUniqueness: If set to true, it is ensure that the same ``DelegatingHandlerInitializer``,
+    ///     even though when inserted multiple times into the context, is only used a single time (the first time it got added).
```
theMomax: Fix grammar

Sources/Apodini/Modifier/DelegationModifier.swift
```diff
@@ -88,47 +103,53 @@ public extension DelegatingHandlerInitializer {
...
+    /// Represents the entry type for the value of an ``DelegatingHandlerContextKey``
+    public struct Entry {
+        /// Every entry of the ``DelegatingHandlerContextKey`` is identified by a instance specific
+        /// `UUID` used to check if we already inserted into the ``Handler`` stack when parsing initializers.
+        private let uuid: UUID
+
+        /// The according ``AnyDelegatingHandlerInitializer`` used to instantiate the delegating ``Handler``.
+        let initializer: AnyDelegatingHandlerInitializer
+        /// If set to true, it is ensure that the same ``AnyDelegatingHandlerInitializer``, even though
+        /// when inserted multiple times into the context, is only used a single time (the first time it got inserted).
```
theMomax, line 137: What happens when you insert once, then the insertion gets filtered, and then you insert a second time?

Sources/Apodini/Response/Response.swift
```diff
@@ -11,37 +11,64 @@ import ApodiniUtils
 
 public struct Response<Content: Encodable>: ResponseTransformable {
     public static var nothing: Response<Content> {
-        Response<Content>(connectionEffect: .open)
+        Response<Content>(information: InformationSet(), connectionEffect: .open)
     }
     
     public static var end: Response<Content> {
-        Response<Content>(connectionEffect: .close)
+        Response<Content>(information: InformationSet(), connectionEffect: .close)
```
theMomax, line 18: Could you please also add functions for initializing `end` with associated `Information` and `Status`?

Sources/Apodini/SyntaxTreeVisitor/DelegatingHandlerInitializerVisitor.swift
```diff
@@ -0,0 +1,126 @@
...
+        self.visitor = visitor
+
+        self.queryInitializers()
+    }
+
+    func visit<H: Handler>(handler: H) throws {
+        preconditionTypeIsStruct(H.self, messagePrefix: "Delegating Handler")
+
+        // we only look at the Delegates for the "main" Handler (also to not get into an infinite loop)
+        // The other ones instantiated via DelegatingHandlerInitializer are covered below.
```
theMomax: Where "below" exactly? I really can't find where this would be handled. Note that a `DelegatingHandlerInitializer` might also construct a type that manually stacks `Delegate`s outside of its given `Handler` (i.e. the one you already analyzed).

Sources/ApodiniVaporSupport/HTTPStatus+ErrorType.swift
```diff
@@ -0,0 +1,29 @@
...
+// Created by Andreas Bauer on 03.07.21.
+//
+
+import Apodini
+import Vapor
+
+extension Vapor.HTTPStatus {
+    /// Creates a `Vapor``HTTPStatus` based on an `Apodini` `ErrorType`.
+    /// - Parameter status: The `Apodini` `ErrorType` that should be transformed in a `Vapor``HTTPStatus`
+    public init(_ error: Apodini.ErrorType) {
```
theMomax: I think it would make sense to make this an `ApodiniErrorCompliantOption`.

theMomax (approved):
The rework of `Information` looks very nice 💪

I only had some minor remarks here and there, except for one thing that might be a bug in the aggregation of metadata trough `Delegate`s. Once this is addressed, you can merge👍🏻

Sources/Apodini/Modifier/DelegationModifier.swift
theMomax: Depending on your level of confidence on this edge case, you might want to add a test-case😉

bauer-andreas: > What happens when you insert once, then the insertion gets filtered, and then you insert a second time?

The enforcement of the `ensureInitializerTypeUniqueness` is done through the `OrderedSet` (see the `Hashable` implementation of the ContextKey value). So a second value won't ever be added to the set (Remember, while we build the set incrementally in the Visitor, the context key value in the `ContextNode` will always contain all values ever added). Therefore the filter can always just apply to the "first" instance.

But good thought nonetheless. Agree it might be good to add a test case for that, just to make sure no regression introduces the described bug when deciding to e.g. exchanging the OrderedSet implementation.

Sources/ApodiniVaporSupport/HTTPStatus+ErrorType.swift
bauer-andreas: I think this is might be still a leftover back when this PR contained support for ApodiniError in the REST exporter (before you added it with your PR). If I remember correctly you already introduced a ApodiniErrorCompliantOption for the HTTPStatus.

Sources/Apodini/SyntaxTreeVisitor/DelegatingHandlerInitializerVisitor.swift
bauer-andreas: Not sure what the comment points to either 😅 
The PR currently ignores Delegates on Delegating Handlers instantiated via an initializer, as its a bit tricky to not fall into the case where we create a loop parsing the Handler passed to the `.instance` method. We could try to do a type based approach for that, to filter those out, but not sure how stable that would be.

Sources/Apodini/Response/Response.swift
bauer-andreas: yeah, sure.

Sources/Apodini/Information/Information.swift
bauer-andreas: yeah, got a bit invested into docs writing 😄

Sources/Apodini/Information/InformationClass.swift
bauer-andreas: This represents a comment inside `swift` code blocks. So single ticks are fine right?

Sources/Apodini/Modifier/DelegationModifier.swift
bauer-andreas: According to our discussion, with the latest commit, I adjusted the behavior, such that a unique Initializer is added nonetheless, if there was the same initializer, but filtered by a according DelegationFilter.

Sources/Apodini/SyntaxTreeVisitor/DelegatingHandlerInitializerVisitor.swift
bauer-andreas: The latest commits adds support for parsing Metadata defined on Delegates declared inside Handlers instantiated via `DelegatingHandlerInitializer`.
To avoid the infinite loop, I relied on a comparing of the Type against the last parsed Handler.

bauer-andreas:
@theMomax I incorporated your feedback and reiterated on the PR.

Specifically the most important points were to support the following use cases:
* DelegatingHandlerInitializers declared as `ensureInitializerTypeUniqueness` should still be inserted if the first instance in the context key value is filtered by a DelegationFilter.
* Metadata of Delegates declared on Handlers instantiated via a `DelegatingHandlerInitilizer` are now parsed correctly (similar as they are treated on the main `Handler` instance).

The PR added some more extensive testing to the GuardTests, ResponseTransformer tests and the DelegationTests, specifically checking the order of execution of those, not just if they were ever executed.
Doing so, I found that I broke the ordering of guards, and that ordering of Initializers defined in Metadata was also not treated accordingly.
Therefore this required some bigger rewrites.

Further, this PR added the following new metadata `Delegated(by:)`, `Reset(with:)`, `Guarded(by:)`, `ResetGuards` mapping their equivalent modifiers into the Metadata space.

As the PR changes a bit more than a few lines, I would request a re-review from you @theMomax. You may orient yourself on the newly added test cases, which should explain a lot of the complications, specifically the `DelegationTests.testExtensiveOrdering()` test case.

theMomax (approved):
LGTM👍🏻

PR #312, opened 2021-07-06
Update fluent version

ThePuscher:
# Update fluent version

With fluent 4.3.0 I get the error "'Content' requires that 'T' conform to 'Encodable'" in fluent > Sources > Fluent > Fluent+Paginate.swift:16 when trying to build my Apodini project. With fluent version 4.3.1, it works just fine. It seems to be an error within fluent.

## :bulb: Proposed solution
Update the required fluent version to 4.3.1

### Reviewer Nudging
Package.swift

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -34,7 +34,7 @@ let package = Package(
     ],
     dependencies: [
         .package(url: "https://github.com/vapor/vapor.git", from: "4.45.0"),
-        .package(url: "https://github.com/vapor/fluent.git", from: "4.3.0"),
+        .package(url: "https://github.com/vapor/fluent.git", from: "4.3.1"),
         .package(url: "https://github.com/vapor/fluent-kit.git", from: "1.13.0"),
         .package(url: "https://github.com/vapor/fluent-sqlite-driver.git", from: "4.1.0"),
         // Used by the `NotificationCenter` to send push notifications to `APNS`.


ThePuscher:
It's indeed the problem: https://github.com/vapor/fluent/releases/tag/4.3.1

PSchmiedmayer:
@ThePuscher Thanks for reporting the error and an identifying a suitable solution. We would need to apply the change to the _release_ branch as the _develop_ branch already contains breaking changes which would require us to tag a `0.3.0` release. To make your life easier and not migrate to `0.3.0` I would need to separately tag a `0.2.1` release on the _release_ branch. Therefore I will close the PR.

There is an even simpler solution to your problem: Make sure to run a `swift package update` command in your folder that contains the _Package.swift_ file. This should update the _Package.resolved_ file and will make sure that you use Fluent `4.3.1`. I recommend to push the _Package.resolved_ file to your repo for executables (as your web service) to make sure that all developers and the CI use the same Package version (and in this case your fluent version).

Alternatively you can adapt your own _Package.swift_ file to enforce Fluent `4.3.1`:
```swift
// swift-tools-version:5.4

import PackageDescription


let package = Package(
    name: "MyWebService",
    dependencies: [
        .package(url: "https://github.com/Apodini/Apodini.git", .upToNextMinor(from: "0.2.0")),
        .package(url: "https://github.com/vapor/fluent.git", from: "4.3.1")
    ],
    targets: [
        .executableTarget(
            name: "MyWebService",
            dependencies: [
                .product(name: "Apodini", package: "Apodini"),
                .product(name: "Fluent", package: "fluent")
            ]
        )
    ]
)
```
This is not the most elegant solution but would also work without the need to commit the _Package.resolved_ file.

PR #313, opened 2021-07-09
Codable Based Traversal

theMomax:
# Codable Based Traversal

Currently, the `Runtime` library is used to provide mutating access to all of our `Property`s on a `Handler` (or other object passed into a `Delegate`). This works fine, however the runtime performance of `Runtime` is quite bad (about the same as Mirror's). This is relevant, as Apodini requires mutating access to the `Handler`s properties at least **once upon initialization of a connection context**. That is, for each request-response pair, we need exactly **one** mutating access to all properties. For all streaming patterns, we also only need **one** mutating access per connection. This mutating access is the call to `Activatable.activate()`. Afterwards, all access is non-mutating (we currently make multiple accesses for injecting different types, such as `RequestInjectable`, or equivalences for `Environment` and `AnyObservableObject`s). While non-mutating access has a slightly better performance, it is still overhead that limits the throughput of an Apodini-webservice.

## :bulb: Proposed solution
Reflective programming is expensive at runtime. However, we also don't want to keep our simple interface for defining `Handler`s and not basically having the service-developer have to implement `Traversable` on every `Handler` themselves.

There are two solutions to this problem, both of which have pros and cons: The first is capturing an autoclosure of each `Handler` and initializing a new instance each and every time one is required. With this approach one could even achieve a solution that does not require usage of `Mirror` or `Runtime` at all. However, there are also a couple of downsides:
* The structure of the `Handler`s could change during runtime (especially with all our reusability features (e.g. `Binding`) this could easily be achieved) . We'd basically have to trust the developer that their `Handler`-closure always returns the same `Handler`. This is only really problematic for structural document-creation (e.g. OpenAPI).
* The second problem is that we have to trust on the developer that their `Handler`-closure never returns the same **instance**. Otherwise, those instances share knowledge resulting in everything from unexpected behavior to catastrophic information leaks. One might be able to validate that this is not the case at runtime.
* A `Handler` might be expensive to initialize.
* Initialization of a `Handler` might have unintended side-effects (e.g. when initialized for creating OpenAPI docs).

This concept can be seen in a very similar project: https://github.com/khanlou/Meridian.

The other solution (implemented in this PR) is usage of code generation. The problem here is that we don't want to require the developer take care of this using some command/script. Instead we want this process to be transparent in a way that no Swift-code is generated. Instead the compiler should generate the machine-code for traversing the `Handler` automatically. The only way to do this using the standard Swift toolchain is via `Codable`. We thus require `Handler`s to conform to `Codable`:

```swift
// MARK: FlatInstanceCoding Concept

/// Flat instance coding is a concept for providing fast (faster than what the `Runtime` library can do) mutating
/// access to a struct's properties by first encoding the properties into an array, then performing mutations on
/// that array, and finally decoding the struct from that array.
///
/// Flat instance coding works on three assumptions:
///     1. The order of the calls to `decode`/`encode` and the accompanying container-construction
///     are done in the same order for encoding and decoding.
///     2. Both encoding and decoding are non-failable.
///     3. The only properties that may exist on the encoded/decoded element as well as its recursive
///     children are ``DynamicProperty``, ``Properties`` and `_InstanceCodable` objects.
///     This is the requirement expressed via the ``PropertyIterable`` protocol. Encoding will fail
///     if any other (even `Codable`) object is encountered.
///
/// The whole object is encoded into an array of tuples containing the **instance** and its **name**.
/// The name is calculated from the coding-keys and the `namingStrategy` valid for the current scope.
/// The `namingStrategy` is ``Properties/defaultNamingStrategy`` by default, but is overridden
/// when in the context of a ``DynamicProperty`` or ``Properties`` element (which both provide their
/// own `namingStrategy`).
///
/// On this array `[(String, Any)]`, `Traversable` can be implemented in a very performant manner.
``` 

The advantage is the performance: decoding only has to be performed once at startup-time. Encoding once per connection-context. All the `Traversable` access is as performant as it could be as it is just iterating over an array.

I created a benchmark testing this concept in comparison to `Runtime`: https://gist.github.com/theMomax/7c9621c3b775923f90f50893a43fc32d

The downside is all `Handler`s have to conform to `Codable` (see release notes), meaning that you cannot use plain constants on `Handler`s anymore, but have to wrap them inside a `Binding.constant`.

## :gear: Release Notes 
All `Handler`s and other objects used within a `Delegate` must now conform to `PropertyIterable`. This requires every every (stored) element of the `Handler` to be an Apodini `Property` (This includes `Properties` and `DynamicProperty`) or a Delegate. If you currently use a non-`Property` element on a `Handler` (or other object used within a `Delegate`), wrap this element in a `Binding.constant`.

Furthermore, the callAsFunction on Delegate was renamed to „instance“.

## ➕ Additional Information

I had to rename the `Content` associatedtype on `Response` as it conflicted with `Component.Content` on the `Encodable` type (there are now default implementations on `Encodable` for both types).

### Related Issues
#11 discovered the need to have mutating access to properties in the first place.

### Testing
I wrote a new test for explicitly testing the capabilities of the FlatInstanceEncoder/-Decoder, but since it is used in `Delegate` is has very good coverage anyways.

### Reviewer Nudging
Start with `InstanceCodable.swift`. Get an idea of how the encoder/decoder work and what requirements this Codable conformance bings to `Handler`. Then look at the `Traversable` implementation for the decoder in `Traversable.swift` and the usage in `Delegate`. Then the new definition of `Handler` and the new protocol `HandlerDefiningComponent`, equals the old `Handler` and enables interfacing between the new `Handler` and `HandlerModifier`s. The rest is basically just refactoring and adaptions.

### DIFF
--- a/Sources/Apodini/Components/Component/AnyComponent.swift
+++ b/Sources/Apodini/Components/Component/AnyComponent.swift
@@ -21,7 +21,7 @@ public struct AnyComponent: Component, SyntaxTreeVisitable {
 }
 
 
-public struct AnyHandler: Handler, SyntaxTreeVisitable, VisitableHandler {
+public struct AnyHandler: _Handler, SyntaxTreeVisitable, VisitableHandler {
     public typealias Response = Never
     
     private let _accept: (SyntaxTreeVisitor) -> Void
--- a/Sources/Apodini/Components/Component/AnyComponent.swift
+++ b/Sources/Apodini/Components/Component/AnyComponent.swift
@@ -21,7 +21,7 @@ public struct AnyComponent: Component, SyntaxTreeVisitable {
 }
 
 
-public struct AnyHandler: _Handler, SyntaxTreeVisitable, VisitableHandler {
+public struct AnyHandler: HandlerDefiningComponent, SyntaxTreeVisitable, VisitableHandler {
     public typealias Response = Never
     
     private let _accept: (SyntaxTreeVisitor) -> Void
--- a/Sources/Apodini/Components/Handler/BaseHandler/Text.swift
+++ b/Sources/Apodini/Components/Handler/BaseHandler/Text.swift
@@ -6,10 +6,10 @@
 //
 
 public struct Text: Handler {
-    private let text: String
+    @Binding private var text: String
     
     public init(_ text: String) {
-        self.text = text
+        self._text = .constant(text)
     }
     
     public func handle() -> String {
--- a/Sources/Apodini/Components/Handler/Handler.swift
+++ b/Sources/Apodini/Components/Handler/Handler.swift
@@ -7,10 +7,10 @@
 
 import NIO
 
-public protocol Handler: PropertyIterable, _Handler { }
+public protocol Handler: PropertyIterable, HandlerDefiningComponent { }
 
-/// A `Handler` is a `Component` which defines an endpoint and can handle requests.
-public protocol _Handler: HandlerMetadataNamespace, Component {
+/// A `HandlerDefiningComponent` is a `Component` which defines an endpoint and knows a `Handler` that can handle requests.
+public protocol HandlerDefiningComponent: HandlerMetadataNamespace, Component {
     /// The type that is returned from the `handle()` method when the component handles a request. The return type of the `handle` method is encoded into the response send out to the client.
     associatedtype Response: ResponseTransformable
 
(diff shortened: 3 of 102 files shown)


theMomax:
In a discussion with @PSchmiedmayer, we decided - even though the concept works and provides a performance-benefit - the implied restrictions to the `Handler` syntax are not worth it. Instead, future work should focus on implementing the first approach described under **Proposed Solution** in the PR description. While this approach also has disadvantages, we do not think they are too bad. The first two problems can be solved by appropriate documentation and runtime validation in `DEBUG` mode. The second pair of problems basically have to be solved via a developer contract. There is already the precedence of SwiftUI, which essentially has exactly the same requirements/limitations for `View`s.

theMomax:
Please do NOT delete this branch!

PR #314, opened 2021-07-09
Apodini Authorization

bauer-andreas:
# Apodini Authorization

Apodini currently requires manual effort to do any kind of authentication or authorization.

## :bulb: Proposed solution

With this PR the collection Apodini framework components adds the ability to easily declare authentication and authorization
requirements on `Component`s.

A basic `Authorize` Metadata Declaration might then look like in the `Handler` shown below. 
Every authentication and authorization step consists of the following parts:
- The `Authenticatable` type representing the authenticated and authorized state.
- The `AuthenticationScheme` that is used to perform the user or protocol defined authentication scheme (while `BasicAuthenticationScheme` and `BearerAuthenticationScheme` are supported out of the box).
- A `AuthenticationVerifier` instance, which maps the output of the authentication scheme to the `Authenticatable` instance, and performs some initial verifications steps (e.g. password checks or signature checks).
- The Block of Authentication Requirements. These are evaluated against the `Authenticatable` instance. Predefined Authentication Requirements are `Allow`, `Deny` and `Verify`.

```swift
struct ExampleHandler: Handler {
    // ...
    var metadata: Metadata {
        Authorize(SomeAuthenticatable.self, using: SomeAuthenticationScheme(), verifiedBy: SomeAuthenticationVerifier()) {
            Verify(if: \.someProperty == "SomeValue")
            // ... `AuthorizationRequirement`s are listed here
        }
    }
}
```

The PR introduces out of the box support for **JSON Web Tokens** using the Vapoe `JWTKit`.

## :gear: Release Notes 

We introduce the new targets `ApodiniAuthorization` together with out of the box support for Basic Authentication (`ApodiniAuthorizationBasicScheme` target), Bearer Authentication (`ApodiniAuthorizationBearerScheme` target) and **Json Web Tokens** (`ApodiniAuthorizationJWT` target).

While the package provides the flexibility to support any arbitrary authentication and authorization mechanism, we want to specifically highlight JWT support. Refer to the documentation on how to define custom authentication schemes, verifiers and Authenticatables.

As the initial step, configure the `JWTSigner` in your web service as follows:
```swift
import Apodini
import ApodiniAuthorizationJWT

    var content: some Component {
        // ...
    }

        JWTSigner(.hs256(key: "secret"))
    }
}
```

For a more detailed overview of signer configuration have a look at the [JWTKit Documentation](https://github.com/vapor/jwt-kit).

The next step is to define your Token type, by conforming to `JWTAuthenticatable`:
```swift
struct SomeToken: JWTAuthenticatable {
    var exp: ExpirationClaim

    var isAdmin: Bool
}
```
The `JWTAuthenticatable` supports any JWT claims as defined by the [JWTKit](https://github.com/vapor/jwt-kit).

As the last step is to declare the `Authorize` Metadata on all the Apodini `Component`s you want to be access protected.
Below is a example `Handler` which does three things:
- Declares a `Authorize` Metadata which requires an authenticated and authorized `SomeToken` (authenticated via Bearer Authentication Scheme).
- Specifies the `Verify(notExpired:)` authorization requirement to verify the `ExpirationClaim`.
- Declares the `Authorized` property to get access to the `SomeToken` instance in the handle method.
```swift
struct ExampleHandler: Handler {
    var token = Authorized<SomeToken>()

    func handle() throws -> String {
        let tokenInstance = try token()
        // ...
        return ...
    }

    var metadata: Metadata {
        Authorize(SomeToken.self) {
            Verify(notExpired: \.exp)
        }
    }
}
```

This PR introduces the `ApodiniHTTPProtocol`, decoupling the HTTP Information instances from the VaporSupport package.

Some parts of the API shall gain `async` capabilities in the future (e.g. the `AuthenticationVerifier` to support database operations and other async verification processes). My plan would be to do that in a follow up PR (or wait if #317 is potentially merged first). I didn't want to get any EventLoopFuture ugliness into here and introduce some sort of hybrid.
But feel free to leave any feedback.

### Related PRs
- #310 introduce the necessary infrastructure to add DelegatingHandlerInitializer from Metadata

### Testing
This PR adds a new `ApodiniAuthorizationTests` target containing extensive testing for the introduced features.

### Reviewer Nudging

The code includes extensive documentation and examples. You might want to start by looking at the test cases, maybe specifically at the `JWTTests` which give real world examples.
Further, the main `AuthorizationMetadata` and `OptionalAuthorizationMetadata` provide overview of how each components interact with each other.


Package.swift
```diff
@@ -78,7 +86,9 @@ let package = Package(
         // Deploy
         .package(url: "https://github.com/vapor-community/vapor-aws-lambda-runtime.git", .upToNextMinor(from: "0.6.2")),
         .package(url: "https://github.com/soto-project/soto.git", from: "5.5.0"),
-        .package(url: "https://github.com/soto-project/soto-s3-file-transfer", from: "0.3.0")
+        .package(url: "https://github.com/soto-project/soto-s3-file-transfer", from: "0.3.0"),
+
+        .package(url: "https://github.com/vapor/jwt-kit.git", from: "4.0.0")
```
theMomax, line 92: maybe add a `// Authorization` so its separated from `// Deploy`

Sources/Apodini/Information/AnyInformation.swift
```diff
@@ -10,6 +10,17 @@ public protocol AnyInformation {
     /// Accepts a ``InformationSet`` instance which is used to collect the given ``AnyInformation``.
     /// - Parameter visitor: The ``InformationSet`` which should collect this instance.
     func collect(_ set: inout InformationSet)
+
+    /// Returns the untyped version of the ``AnyInformation`` instance. This method has no effect on ``Information``
+    /// instances and returns self. For ``InformationInstantiatable`` it returns the corresponding ``Information`` instance.
+    /// - Returns: The untyped ``AnyInformation``.
```
theMomax, line 20: Just a general comment. The correct DocC syntax would be to leave a blank line between the "Definition" and the "Discussion" like this:

```swift
/// Returns the untyped version of the ``AnyInformation`` instance.
/// 
/// This method has no effect on ``Information`` instances and returns self. For
/// ``InformationInstantiatable`` it returns the corresponding ``Information`` instance.
/// 
/// - Returns: The untyped ``AnyInformation``.
```

Sources/Apodini/Information/InformationSet.swift
```diff
@@ -5,21 +5,37 @@
...
+import ApodiniUtils
+
 /// An ``InformationSet`` is used to store ``Information`` instances, checked for uniqueness
 /// against their ``InformationKey``.
 public struct InformationSet {
     private var storage: [AnyHashable: AnyInformation] = [:]
+    /// Pretty much acts a typed cache for the above storage dict.
+    /// For every key contained in this dictionary, there is a according untyped value preset above.
+    /// The other way around doesn't hold.
+    private var capturedInstantiatables: [AnyHashable: AnyInformation] = [:]
```
theMomax, line 18: I think it would be a good idea to add a layer of abstraction here. I smell quite some bug potential🙃. Introduce a new wrapper around both `storage` and `capturedInstantiables` that abstracts the caching away please.

Sources/Apodini/Response/ApodiniError.swift
```diff
@@ -83,17 +83,42 @@ public struct ApodiniError: Error {
...
+    ) -> ApodiniError {
+        callAsFunction(reason: reason, description: description, information: information, options: options)
+    }
+
+    public func merge(with error: ApodiniError) -> ApodiniError {
+        precondition(type == error.type, "When merging ApodiniErrors, the error types must match. \(error.type) can't override \(type).")
+        return ApodiniError(
+            type: type,
+            reason: reason ?? error.reason,
+            description: description ?? error.description,
```
theMomax: Shouldn't `error` have precedence over `self` (at least that holds for `information.merge`, doesn't it)?

theMomax: One could also add an "original error" part to the reason/description...

Sources/ApodiniAuthorization/Authenticatable.swift
```diff
@@ -0,0 +1,6 @@
+//
+// Created by Andreas Bauer on 07.07.21.
+//
+
+/// Any type conforming to ```Authenticatable`` represents some state of a authenticatable instance.
```
theMomax: "an authenticatable"...and I am not sure what exactly you mean by that. Maybe add a discussion in the form of "An ``Authenticatable`` can be used together with ``?`` to do/create ?.

Sources/ApodiniAuthorization/AuthorizationErrorReason.swift
```diff
@@ -0,0 +1,34 @@
...
+    case authenticationRequired
+    /// Encountered malformed or invalid authentication information (error thrown inside ``AuthenticationScheme``)
+    case invalidAuthenticationRequest
+    /// Failed the authentication (error thrown inside ``AuthenticationVerifier``)
+    case failedAuthentication
+    /// The ``Authenticatable`` instance failed evaluation against the ``AuthorizationRequirement``s
+    case failedAuthorization
+    /// Represents some custom, potentially user defined error reason, to be used if the others don't fit.
+    /// A string based reason shall be supplied for debugging uses.
+    case custom(_ reason: String = "undefined")
```
theMomax: I'd remove the default "undefined".

Sources/ApodiniAuthorization/AuthorizationRequirements/AnyAuthorizationRequirement.swift
```diff
@@ -0,0 +1,30 @@
+//
+// Created by Andreas Bauer on 09.07.21.
+//
+
+/// Represents a type erase ``AuthorizationRequirement``
```
theMomax: erased

Sources/ApodiniAuthorization/AuthorizationRequirements/AnyAuthorizationRequirement.swift
```diff
@@ -0,0 +1,30 @@
...
+    /// - Returns: Returns a ``RequirementResult``.
+    /// - Throws: Throws an ``ApodiniError`` if necessary. While throwing is allowed, any expected errors
+    ///     should ideally be transformed into a ``RequirementResult`` using the ``RequirementResult/Cause/error`` cause.
+    func anyEvaluate<Element: Authenticatable>(for element: Element) throws -> RequirementResult
+}
+
+// MARK: AnyAuthorizationRequirement
+public extension AuthorizationRequirement {
+    /// Default implementation calling the typed evalute version.
+    func anyEvaluate<Element0: Authenticatable>(for element: Element0) throws -> RequirementResult {
```
theMomax: Why Element**0**? Just seems kinda weird...

Sources/ApodiniAuthorization/AuthorizationRequirements/AuthorizationCondition.swift
```diff
@@ -0,0 +1,114 @@
...
+///
+/// ```swift
+/// let condition: AuthorizationCondition<Example>
+///
+/// condition = \.someString == <expression>
+///
+/// condition = \.someString != <expression>
+/// ```
+///
+/// ## Comparing Operators
```
theMomax: Comparison

Sources/ApodiniAuthorization/AuthorizationRequirements/AuthorizationCondition.swift
```diff
@@ -0,0 +1,114 @@
...
+///
+/// condition = \.someInt >= <expression>
+///
+/// condition = \.someInt < <expression>
+///
+/// condition = \.someInt <= <expression>
+/// ```
+///
+/// ## Logical Operators
+/// The typical `&&`, `||` and `!` logical operators cna be used on ``AuthorizationCondition`` instances.
```
theMomax: can

Sources/ApodiniAuthorization/AuthorizationRequirements/AuthorizationRequirements.swift
```diff
@@ -0,0 +1,47 @@
...
+
+/// An ``AuthorizationRequirements`` wraps one or multiple ``AuthorizationRequirement`` instances.
+/// On Evaluation it will loop over the contained requirements until the first ``RequirementResult``
+/// is encountered which is not ``RequirementResult/undecided``.
+public struct AuthorizationRequirements<Element: Authenticatable>: AuthorizationRequirement {
+    let requirements: [AnyAuthorizationRequirement]
+
+    /// Initializes a new ``AuthorizationRequirements`` instance for a single ``AuthorizationRequirement``.
+    /// - Parameter requirement: The ``AuthorizationRequirement`` to be wrapped.
+    public init<Requirement: AuthorizationRequirement>(_ requirement: Requirement) where Requirement.Element == Element {
```
theMomax, line 17: Why does it only take one element? I expected an empty initializer accompanied by a
```swift
public func with<Requirement: AuthorizationRequirement>(_ requirement: Requirement) -> Self where Requirement.Element == Element
```

theMomax: Ah...you're using a resultBuilder...does it need to be public then?

Sources/ApodiniAuthorization/AuthorizedProperty.swift
```diff
@@ -0,0 +1,115 @@
...
+        delegate.environment(\.wrapped.stateContainer, stateContainer.inserting(instance))
+    }
+}
+
+// MARK: Environment
+
+private extension Application {
+    class WrappedContainer {
+        var stateContainer = AuthorizationStateContainer()
+    }
```
theMomax: ApodiniUtils defines `Box` for such purposes.

Sources/ApodiniAuthorization/AuthorizedProperty.swift
```diff
@@ -0,0 +1,115 @@
...
+
+    var wrapped: WrappedContainer {
+        WrappedContainer()
+    }
+}
+
+/// The ``SomeAuthorizationStateContainer`` type is a type erasing wrapper around an ``AuthorizationStateContainer``
```
theMomax: Is this an outdated comment?

Sources/ApodiniAuthorization/AuthorizedProperty.swift
```diff
@@ -0,0 +1,115 @@
...
+// MARK: Environment
+
+private extension Application {
+    class WrappedContainer {
+        var stateContainer = AuthorizationStateContainer()
+    }
+
+    var wrapped: WrappedContainer {
+        WrappedContainer()
+    }
```
theMomax, line 74: `Delegate.environment(_:_:)` only changes the **local** environment. You thus don't have to implement a storage, if you don't want to insert global default values or anything. If you would use those, I think the current implementation would even share data across different client-connections. Thus you should be able to just do the following:

```swift
private extension Application {
    var stateContainer: AuthorizationStateContainer {
        get {
            fatalError()
        }
        nonmutating set {
            fatalError()
        }
    }
}

```

Sources/ApodiniAuthorization/Delegates/AuthenticationScheme.swift
```diff
@@ -0,0 +1,57 @@
...
+    /// authentication challenge in the response on every failed authentication attempt.
+    ///
+    /// Every `ApodiniError` passed to this method will contain the according ``AuthorizationErrorReason``
+    /// in the error options. It can be queried like the following:
+    /// ```swift
+    /// let reason = apodiniError..option(for: .authorizationErrorReason)
+    /// ```
+    ///
+    /// An ``AuthenticationScheme`` might also define custom `ApodiniError` options which can be parsed here,
+    /// inorder to support wire format specific features. An error containing such custom error options
```
theMomax: in order

Sources/ApodiniAuthorization/Delegates/AuthenticationVerifier.swift
```diff
@@ -0,0 +1,28 @@
...
+///
+/// In an ``AuthenticationVerifier`` you can use any common ``Property`` similar as you can in a  ``Handler``.
+public protocol AuthenticationVerifier {
+    /// The input type received from an ``AuthenticationScheme``.
+    associatedtype AuthenticationInfo
+    /// The ``Authenticatable`` instantiated from the ``AuthenticationInfo``.
+    associatedtype Element: Authenticatable
+
+    /// This method uses the ``AuthenticationInfo`` result received from the respective ``AuthenticationScheme``
+    /// to instantiate the ``Element`` instance. The instantiating process might includes calls to databases
```
theMomax: include

Sources/ApodiniAuthorization/Delegates/Authenticator.swift
```diff
@@ -0,0 +1,130 @@
...
+            result = try requirements.evaluate(for: instance)
+        } catch {
+            throw error.apodiniError(options: .authorizationErrorReason(.failedAuthorization))
+        }
+
+        switch result {
+        case let .fulfilled(cause), let .undecided(cause): // undecided is a acceptance state as well!
+            logger.trace("Authorization on Handler \(H.self) succeeded with \(cause())")
+            if authenticatable.isAuthorized {
+                return try delegate().handle()
```
theMomax: Why don't you manually set the `instance` here as in L108 and the other cases? I am not sure what the effects are and why there should be a difference. I also think that if you don't override it, the `delegate` automatically uses the same `AuthorizationStateContainer` as  used by the local `authenticatable`.

Sources/ApodiniAuthorization/AuthorizationRequirements/RequirementResult.swift
```diff
@@ -0,0 +1,41 @@
...
+//
+
+/// A ``RequirementResult`` represents the result of an evaluation of an ``AuthenticationRequirement``.
+public enum RequirementResult {
+    /// Represents a positive ``RequirementResult``.
+    /// Captures the ``Cause`` leading to the result. The ``Cause`` is only captured for debug purposes.
+    case fulfilled(cause: @autoclosure () -> Cause = .unspecified)
+    /// Represents a neutral ``RequirementResult``.
+    /// Captures the ``Cause`` leading to the result. The ``Cause`` is only captured for debug purposes.
+    case undecided(cause: @autoclosure () -> Cause = .unspecified)
```
theMomax, line 22: `Authenticator` interprets this case as successful. I don't think that is obvious. Please add an according note to the docs.

theMomax (commented):
Looks pretty solid👍🏻 I mostly focused on the generic implementation (`Delegate`/`Metadata` related stuff) and less on the correctness of the middleware-specific implementation details.

I had some comments and questions here and there. I'll go over the test cases and complete the review once those are resolved.

Sources/Apodini/Information/InformationSet.swift
bauer-andreas: I thought about the same. Though I kept it this way as it avoid to introduce unnecessary mappings and inefficiency for the below Collection and Sequence conformance. As you can see, currently we can just map the according protocol 1:1 to the Collection and Sequence requirements.
After all those two properties are completely private and never see the outside of an `InformationSet`. And everything is thorough covered by test vectors.
So, I would argue against introducing another level of abstraction.

Sources/Apodini/Response/ApodiniError.swift
bauer-andreas: you are completely right. Great catch 👍

bauer-andreas: > One could also add an "original error" part to the reason/description...

How do you envision that?

Sources/ApodiniAuthorization/Authenticatable.swift
bauer-andreas: I'll try to clarify that by adding further documentation 👍

Sources/ApodiniAuthorization/AuthorizationRequirements/AnyAuthorizationRequirement.swift
bauer-andreas: `Element` is already defined in the `AuthorizationRequirement` and I went the classic way of just adding an index.
But I can try to envision some more creative naming.

Sources/ApodiniAuthorization/AuthorizationRequirements/AuthorizationRequirements.swift
bauer-andreas: Yes, it is required to be public as it is used as a default value for an initializer parameter in the `AuthorizationMetadata` and `OptionalAuthorizationMetadata`.

Sources/ApodiniAuthorization/AuthorizedProperty.swift
bauer-andreas: yes indeed 😁

Sources/ApodiniAuthorization/AuthorizedProperty.swift
bauer-andreas: The reason why I didn't just went with an `@EnvironemtObject` is that an `@Enviroment` property pulls out the value from the application key path if there isn't a value in the **local** environment.
This is therefore required to ensure that there is **always** a value present in the KeyPath, such that the Handler doesn't crash if something failed.

So using a fatalError in the `get` is definitely not possible, but I could probably use the `nonmutating set` to remove the need of the wrapper, as far as I understand that (as `.environment` requires an `WriteableKeyPath`)

> I think the current implementation would even share data across different client-connections

It doesn't. Its basically a copy-on-write. The first time the `@Environment` property is used it uses the statically, **empty** defined AuthorizationStateContainer from the extension. If an authenticator adds new state, it copies the value, modifies it accordingly (e.g. inserting the authorized Authenticatable instance) and inserts it into the **local** environment of the respective handler evaluation.

Sources/ApodiniAuthorization/Delegates/Authenticator.swift
bauer-andreas: The `evaluateRequirements(...)` method is not only called once an instance is fully authorized, but also at the beginning of `handle()` if the authenticator finds that there is already an authorized instance (of the same `Authenticatable` type).
See:
https://github.com/Apodini/Apodini/blob/37de2a6567d6e3190764c9d397728409ed9faef3/Sources/ApodiniAuthorization/Delegates/Authenticator.swift#L38-L45

If an authenticator finds that there is already an authorized instance, it skips its own authentication and authorization operations. Only if not turned off via `skipRequirementsForAuthorized` (which is false by default), it will evaluate the authorization requirements against the existing instance.
And that's the case what you commented above. If there is already an existing instance, and the requirements are fulfilled, there is no reason to change something in the environment, because after all nothing changed.

Sources/ApodiniAuthorization/AuthorizationRequirements/RequirementResult.swift
bauer-andreas: I'll add some docs 👍

Sources/Apodini/Information/InformationSet.swift
theMomax: If the caching mechanism is covered by regression tests, I am fine with leaving it as is.

Sources/Apodini/Response/ApodiniError.swift
theMomax: basically `reason: "\(reason) (original reason: \(error.reason))"` and the same for description.

Sources/ApodiniAuthorization/AuthorizationRequirements/AnyAuthorizationRequirement.swift
theMomax: Ok...if that's the classic way...I'd have made it an `E` ;)

bauer-andreas:
boschbruegge.ase.in.tum.de agent is out of space, therefore the last few bamboo builds are failing.

Sources/Apodini/Response/ApodiniError.swift
```diff
@@ -87,14 +107,26 @@ public struct ApodiniError: Error {
...
+            information: information.merge(with: error.information),
+            PropertyOptionSet(lhs: options, rhs: error.options)
         )
     }
+
+    private func preserveOriginalReasoning(new newMaybe: String?, previous previousMaybe: String?, _ name: String) -> String? {
+        guard let new = newMaybe, let previous = previousMaybe else {
+            return newMaybe ?? previousMaybe
+        }
+        return "\(new) (original \(name): \(previous)"
```
theMomax, line 131: I think that is the wrong way round. Usually, you would have the following usage:

```swift
struct SomeHandler: Handler {
    @Throws(reason: "Authorization failed") var authError

    var authorizedUser: Authorized<User>

    func handle() throws {
        do {
            let x = try authorizedUser()
        } catch {
            throw authError.merge(with: error.apodiniError)
        }
    }
}
```

With the current code you'd get: `Invalid Credentials (original reason: Authorization failed)`. At least that is how I'd use it. Of course you could argue that the call should be `error.apodiniError.merge(with: authError)`. but I'd consider this less intuitive. I guess it comes down to the term "merge" not specifying a precedence. I'd suggest something like: `authError.caused(by: error.apodiniError)`.

While you're at it, I think we could also make the argument just an error (or at least provide one overload that accepts an `Error`). As shown in the example above, this function will often be used in a `catch`, where the `error` isn't typed.

Sources/ApodiniAuthorization/AuthorizationRequirements/RequirementResult.swift
```diff
@@ -0,0 +1,51 @@
...
+public enum RequirementResult {
+    /// Represents a positive ``RequirementResult``.
+    /// Captures the ``Cause`` leading to the result. The ``Cause`` is only captured for debug purposes.
+    case fulfilled(cause: @autoclosure () -> Cause = .unspecified)
+    /// Represents a neutral ``RequirementResult``.
+    ///
+    /// When evaluating ``AuthorizationRequirement`` evaluation will continue with the next requirement
+    /// when encountering `undecided`. This is the fundamental difference to the other two cases.
+    /// Although, should the final result of some ``AuthorizationRequirement`` be undecided, it is nonetheless
+    /// and acceptance state. Everything which is not explicitly rejected is accepcted.
```
theMomax: "an" acceptance state

Sources/ApodiniVaporSupport/ApodiniError+AbortError.swift
```diff
@@ -24,38 +25,3 @@ extension ApodiniError: AbortError {
         HTTPHeaders(information)
     }
 }
-
-
-// MARK: HTTPResponseStatus Option
-
-extension HTTPResponseStatus: ApodiniErrorCompliantOption {
-    public static func `default`(for type: ErrorType) -> HTTPResponseStatus {
-        switch type {
```
theMomax, line 33: Was this deleted or moved somewhere else?

Sources/XCTApodini/XCTAssertRuntimeFailure.swift
```diff
@@ -13,12 +13,12 @@ import XCTest
 
 /// Asserts that an expression leads to a runtime failure.
 public func XCTAssertRuntimeFailure<T>(
-    _ expression: @escaping @autoclosure () -> T,
+    _ expression: @escaping @autoclosure () throws -> T,
```
theMomax, line 22: I find that a bit misleading, as one could think that throwing is also considered a `RuntimeFailure` and thus result in a fulfilled assertion. Please document accordingly.

Tests/ApodiniAuthorizationTests/JWTTests.swift
```diff
@@ -0,0 +1,269 @@
...
+            "Hello World"
+        }
+
+        var metadata: Metadata {
+            Authorize(ExampleJWTToken.self) {
+                Verify(notExpired: \.exp)
+                Verify(intendedAudience: \.aud, includes: "ExampleAudience")
+                Verify(notBefore: \.nbf)
+
+                Verify(issuer: \.iss, is: "https://other-option.org", "https://example.org")
```
theMomax, line 49: Love it❤️

Tests/ApodiniAuthorizationTests/JWTTests.swift
```diff
@@ -0,0 +1,269 @@
...
+    }
+
+    func testJWTTokenFailingEmailRequirement() throws {
+        // request tokenWithoutEmail
+        let tokenWithoutEmail: String = try XCTUnwrap(
+            exporter.request(on: tokenHandler, request: EmptyRequest(), with: app)
+                .typed(String.self)?.content
+        )
+
+        // test fullyCorrectToken
```
theMomax: I guess that should be `tokenWithoutEmail`?

theMomax (approved):
Amazing work! I think this is a huge step forward for Apodini👍🏻

I had a few notes and questions along the way. Once they are resolved, you're good to go!

Sources/ApodiniVaporSupport/ApodiniError+AbortError.swift
bauer-andreas: yes. this was moved into the `ApodiniHTTPProtocol` target, [see](https://github.com/Apodini/Apodini/blob/7ce97bed44164c15c0740818705cbbc2dc0abcad/Sources/ApodiniHTTPProtocol/ApodiniError/HTTPResponseStatusOption.swift).

The `ApdoniHTTPProtocol` is basically for those things which are only SwiftNIO based and don't have a dependence on Vapor. This is e.g. used for the Authentication Schemes which use the error option and the HTTP Information definitions. But the http authentication schemes don't require Vapor.

Tests/ApodiniAuthorizationTests/JWTTests.swift
bauer-andreas: good copy pasta catch 🍝

Sources/Apodini/Response/ApodiniError.swift
bauer-andreas: I agree, that making the precedence more explicit makes sense. The merge call wasn't really created with the intention for public use, but yes it is user facing and might be useful anyways. Will go for the `caused(by:)` approach with an Error overload.

bauer-andreas: One note, the `merge` call currently makes sense as we enforce that two Errors which get merged have the same Error type.
Also not sure if the `caused(by:)` really implies that stuff like options and information is actually **merged**.
Typically you would expect (e.g. like Java does it) that it just captures the "previous" error.

So would it maybe be better to keep the `merge` function name, but making the precedence explicit?
E.g. `merge(withRHS:)` or `merge(rhs:)`? (or lhs depending what makes more sense).

theMomax: Good point. Maybe detailed(by:)?

PR #317, opened 2021-07-13
Async/Await + AsyncSequence Refactoring

theMomax:
# Async/Await + AsyncSequence Refactoring

There are two problems:

The first is that the only option to do asynchronous processing in `Handler`s or `Guard`s is to use `EventLoopFuture`s, which is quite annoying.

The second is that the stateful processing pipeline used by exporters is based on [OpenCombine](https://github.com/OpenCombine/OpenCombine), which is not the most future-proof dependency (especially since Swift 5.5 introduced `AsyncSequence`, which can pretty much replace Combine for the most use-cases).

## :bulb: Proposed solution
Both problems can be solved by using the new concurrency primitives introduced with Swift 5.5.

The stateful processing pipeline for exporters was rewritten based on `AsyncSequence`. The stateless (one-off) pipeline is still based on `EventLoopFuture` as most middleware-frameworks haven't even released async/await support yet. Furthermore, usage of ELF should should be more performant until NIO3 is entirely rebuilt on async/await.

`Handler.handle()` is now `async`. One can still return `EventLoopFuture`s from `handle()`, even in addition to usage of async/await syntax.

For this to work, Apodini has to drop support for all Swift versions prior to 5.5 and all macOS versions prior to 12.0. 

## :gear: Release Notes 
* 🎉 `Handler.handle()` is now `async`!
* ⚠️You will need Swift 5.5 or higher to work with Apodini from now on!
* ⚠️ macOS versions prior to 12.0 are no longer supported!
* 🎉 `Guard` no longer returns an `EventLoopFuture<Void>`, but is an `async` void function.
* ⚠️ `SyncGuard` has been removed and all occurrences must be renamed to `Guard`.
* ⚠️ Since `Handler.handle()` is now an `async` context, it is no longer guaranteed to run on the `EventLoop` available via `@Environment(\.connection)`'s `eventLoop` property.
* ⚠️ `Delegate.callAsFunction()` was renamed to `Delegate.instance()`

### Related PRs
#307 introduced the OpenCombine-based stateful processing pipeline.
#265 and #295 previously attempted to add async/await support. 

### Testing
No new tests were added yet. The test checking `handle()` runs on the correct `EventLoop` was removed. Some tests testing custom `OpenCombine.Publisher` implementations were removed.

### Reviewer Nudging

Before you start the review, make sure you know what `Task { }` and `CheckedContinuation` do and how they interact with the `async`/`await` keywords!


Package.swift
```diff
@@ -1,12 +1,12 @@
...
+// swift-tools-version:5.5
 
 import PackageDescription
 
 
 let package = Package(
     name: "Apodini",
     platforms: [
-        .macOS(.v10_15)
+        .macOS(.v11)
```
bauer-andreas: Not sure if the `.v12` version is available in the toolchain packaged with Xcode Beta 3 (at least its [documented](https://developer.apple.com/documentation/swift_packages/supportedplatform/macosversion/3857793-v12)),
but in any case you should be able to pass it as a string.

`.macs("12.0")`

This should allow you to remove all the `@available`s in the other places.

theMomax: Yes, beta 3 solves that issue. It also ships `AsyncStream` which was another missing piece. I'll try to finalize this ASAP. The last obstacle will probably be the outdated CI setup.

Package.swift
```diff
@@ -108,17 +105,19 @@ let package = Package(
             ],
             exclude: [
                 "Components/ComponentBuilder.swift.gyb"
-            ]
+            ],
+            swiftSettings: [.unsafeFlags(["-Xfrontend", "-enable-experimental-concurrency"], nil)]
```
bauer-andreas, line 112: How will that impact usability of the Apodini as a dependency, see #290?

Sources/Apodini/Properties/Delegate.swift
```diff
@@ -47,7 +47,7 @@ public struct Delegate<D> {
     }
     
     /// Prepare the wrapped delegate `D` for usage.
-    public func callAsFunction() throws -> D {
+    public func instance() throws -> D {
```
bauer-andreas, line 51: Is there a reason why this got renamed?
This would also be a breaking change, which is currently not documented?

Tests/ApodiniDeployTests/InvocableHandlerTests.swift
```diff
@@ -40,7 +39,7 @@ private struct TestWebService: Apodini.WebService {
         }
     }
     
-    struct FInvoker: Handler {
+        struct FInvoker: Handler {
```
bauer-andreas: weird indent(?) same for the Greeter and Calculator structs below(?)

Tests/ApodiniTests/ModifierTests/OperationModifierTests.swift
```diff
@@ -22,7 +21,7 @@ final class OperationModifierTests: ApodiniTests {
         }
     }
 
-    struct TestWebService: WebService {
+        struct TestWebService: WebService {
```
bauer-andreas: ident?

theMomax:
@PSchmiedmayer what‘s the state on updating the CI? The PR requires Xcode beta 3 to run.

bauer-andreas:
https://github.com/Apodini/Apodini/blob/08d01558609cc07e70ce74f0a0b1bbafec68a127/.github/workflows/build-and-test.yml#L92
try replacing that with  `image: swiftlang/swift:nightly-5.5-${{ matrix.linux }}`, maybe that already works.

You probably want to remove the `latest-nightly-swift-version-linux` job, as the tag used there is 2 months old: https://hub.docker.com/layers/swiftlang/swift/nightly-focal/images/sha256-bc056facee1a6868a2a5ed247acf6946a6ea57bf74acd37e1cb2f3ec46d518ce?context=explore
https://github.com/Apodini/Apodini/blob/08d01558609cc07e70ce74f0a0b1bbafec68a127/.github/workflows/build-and-test.yml#L37

For the macOS builds its probably a bit more tricky, depending if the macOS runners ship with Xcode 13 beta installed: 
https://github.com/Apodini/Apodini/blob/08d01558609cc07e70ce74f0a0b1bbafec68a127/.github/workflows/build-and-test.yml#L19-L23

Looking at the setup-xcode action it only takes what is installed on the system: https://github.com/maxim-lobanov/setup-xcode/blob/cb46b9b93df4cde677a1a3667ef111c2e7f901ef/src/xcode-selector.ts#L8-L14

Nonetheless you could try to specify `"13.0"` as the Xcode-version (according to the [docs](https://github.com/maxim-lobanov/setup-xcode)), worst case it doesn't work but will give you an overview of all installed Xcode versions on the runners.

Seeminlgy, GitHub runners included Xcode betas in the past as well: see https://gist.github.com/fabianfett/fd811d7921eb856bb100c5c15565077f

PSchmiedmayer:
@theMomax @Supereg I am currently in contact with the Admins to see if we can get access to macOS 12 runner. They promised me to get back to me by the end of the day.
The Linux builds should be unaffected by the macOS build agents. Using the latest Swift 5.5 images should allow you to use async await on non Darwin platforms.

Sources/Apodini/Semantic Model Builder/KnowledgeSources/DelegateFactory.swift
```diff
@@ -0,0 +1,29 @@
...
+    
+    public init<B>(_ blackboard: B) throws where B: Blackboard {
+        self.delegate = Delegate(blackboard[EndpointSource<H>].handler, .required)
+    }
+    
+    /// Creates one instance of the ``Delegate``.
+    ///
+    /// - Note: Use a new instance for each client-connection! Otherwise data may be shared between
+    /// all clients using the same instance.
+    public func instance() -> Delegate<H> {
```
bauer-andreas, line 24: Maybe a bit picky (by a lot), but I think it might be a bit weird having the same method name `instance` use twice in similar context for a) in the DelegateFatory to return a `Delegate` instance and b) in the `Delegate` use to return the delegated `Handler` instance.
This creates the possibility of endpoint[DelegateFactory<H>.self].instance().instance().
Just something I stumbled upon when reading over the code where it got confused. Maybe its just me though 😁

Sources/ApodiniExtension/AsyncSequenceHelpers/AsyncJustSequence.swift
```diff
@@ -0,0 +1,45 @@
...
+//
+// SPDX-FileCopyrightText: 2019-2021 Paul Schmiedmayer and the Apodini project authors (see CONTRIBUTORS.md) <paul.schmiedmayer@tum.de>
+//
+// SPDX-License-Identifier: MIT
+//
+
+
+import _Concurrency
+
+public struct Just<Element>: AsyncSequence {
```
bauer-andreas, line 13: Docs 🙃

Sources/ApodiniExtension/AsyncSequenceHelpers/AsyncMergeSequence.swift
```diff
@@ -0,0 +1,232 @@
...
+// SPDX-FileCopyrightText: 2019-2021 Paul Schmiedmayer and the Apodini project authors (see CONTRIBUTORS.md) <paul.schmiedmayer@tum.de>
+//
+// SPDX-License-Identifier: MIT
+//
+
+
+import Foundation
+import _Concurrency
+
+public struct AsyncMergeSequence<Base, Other>: AsyncSequence where Base: AsyncSequence, Other: AsyncSequence, Base.Element == Other.Element {
```
bauer-andreas, line 15: Maybe some short docs defining this particular AsyncSequence. Maybe including some words about what promises are made or not made in terms of sequence, as far as I see there are none?

.github/workflows/test-webservice.yml
```diff
@@ -26,7 +26,7 @@ defaults:
 jobs:
   macos:
     name: macOS ${{ matrix.configuration }}
-    runs-on: macos-11
+    runs-on: macos-12
```
bauer-andreas: Not sure, but I don't think there are any macOS-12 runner pools yet.

.github/workflows/test-webservice.yml
```diff
@@ -35,7 +35,7 @@ jobs:
     - uses: actions/checkout@v2
     - uses: maxim-lobanov/setup-xcode@v1.2.3
       with:
-        xcode-version: latest
+        xcode-version: '13.0'
```
bauer-andreas: Looking at the output of the workflow, the GitHub runners currently ship with Xcode build 13A5192i which should be beta 3 if I'm not wrong.
So the toolchain itself is up to date, but it probably fails at the OS version, as there seemingly aren't any macOS 12 runners ://

bauer-andreas (approved):
Great work 🚀 This definitely future proofs Apodini by another step.

I only had some few comments/discussion points 👍

Sources/Apodini/Properties/Delegate.swift
theMomax: Yeah, I forgot to document that in the PR description. I renamed it as I didn't particularly like the callAsFunction syntax when combined with e.g the `.environment(_:_:)` chaining:

```swift
myHandlerDelegate
    .environment(\.some, x)()
    .handle()
```
vs
```swift
myHandlerDelegate
    .environment(\.some, x)
    .instance()
    .handle()
```

I'll adapt the release notes.

Tests/ApodiniDeployTests/InvocableHandlerTests.swift
theMomax: Weird indeed🤷‍♂️

Sources/Apodini/Semantic Model Builder/KnowledgeSources/DelegateFactory.swift
theMomax: I kinda see your point, but then I actually don't. `endpoint[DelegateFactory.self].instance().instance()` should never occur (outside of tests) as there should be some other stuff going on in between instantiating a `Delegate` and preparing its wrapped child.
If you have a better suggestion for either of the two let me know.

Sources/ApodiniExtension/AsyncSequenceHelpers/AsyncJustSequence.swift
theMomax: Thanks!

Sources/ApodiniExtension/AsyncSequenceHelpers/AsyncMergeSequence.swift
theMomax: Yeah, there should be at least a short description. I put the real discussion on the `AsyncSequence` extension (`.merge(with:)`). That's also how it is done in the standard libs' docs.

Package.swift
theMomax: Well...without the flags you get runtime crashes, so I guess we don't really have a choice. Once Swift 5.5 is released
we should be able to remove them again. Until then the project requires an experimental setup anyways.

.github/workflows/test-webservice.yml
theMomax: See https://github.com/Apodini/Apodini/pull/317#discussion_r672958871

.github/workflows/test-webservice.yml
theMomax: Yeah, that's correct. There are no macOS 12 runners yet, but Apodini is enlisted to be among the first testers once they are launched. I put the macOS version number to 12 for two reasons:
* We'd notice if they launch macOS 12 runners before we get our own runners working.
* When I put it to 11 with Xcode 13, the CI even manages to compile the whole project (even though the macOS(.v12) SPM tag clearly states that macOS 12 is required). It then fails at startup because the `_Concurrency` library is missing.  That's just confusing. As is we clearly see the runners are not fit for running the project.

Sources/Apodini/Properties/Delegate.swift
bauer-andreas: Yeah noticed that as well, that nesting is always a bit weird.
Just not sure of the name, but on the other hand I also don't have a better suggestion on hand.

But yeah, document it and I'm fine 👍

Sources/Apodini/Semantic Model Builder/KnowledgeSources/DelegateFactory.swift
bauer-andreas: After all, I do not have a super strong opinion on that. And as you said, the use cases where each of those methods are used, are completely separate. Fair point. 👍

philippzagar (approved):

Issue #318, opened 2021-07-13
[MINOR FEATURE] Improve the `AutomaticCommunicationalPattern` Knowledge Source 


PSchmiedmayer:
**Problem**
Currently a `Handler` like is detected as using the `.bidirectionalStream` by the `AutomaticCommunicationalPattern` based on the fact that it returns a `Response<...>` response of the `handle` function even though it might just use the `Response<...>` as a way to return additional `Information` to the client.

**Reproducible Example**
The following `Handler` is detected as `.bidirectionalStream`:
```swift
struct BlobGreeter: Handler {
    @Parameter(.http(.path)) var name: String
    
    @Parameter(.http(.query)) var greeting: String?
    
    
    func handle() -> Apodini.Response<Blob> {
        Response.send(
            Blob(Data("\(greeting ?? "Hello"), \(name)!".utf8), type: .text(.html)),
            information: [AnyHTTPInformation(key: "Test", rawValue: "Test")]
        )
    }
}
```
You can let Apodini know about the correct pattern using the
```swift
var metadata: Metadata {
    Pattern(.requestResponse)
}
```
metadata.

**Proposed Solution**
The expected behavior would be to classify the Handler shown above as a `.requestResponse` pattern as it does not include a `@State`, `@ObservedObject`, or `@Environment` value that could be observed.
The `AutomaticCommunicationalPattern` should be extended by also checking the available properties in the the Handler and any delegating handlers it uses to detect if `@State`, `@ObservedObject`, or `@Environment` (using an observable object) could be used to trigger a web service side streaming event.

theMomax:
I would consider this more of a missing feature than a bug. The problem is that even if an endpoint uses no `@State`, `AnyObservedObject` and does not inspect `@Environment(\.connection)` (giving it access to the `ConnectionState`), it could still be intended to be a bidirectional streaming endpoint as long as it returns a `Response`. This endpoint would violate all our design guidelines, but it is possible. The intention behind `AutomaticCommunicationalPattern` was to only restrict the pattern if it is **certain**. I.e. when writing an `InterfaceExporter` and the `AutomaticCommunicationPattern` is `.requestResponse`, I can safely assume the first evaluation of the `Handler` will result in a `Response` with `ConnectionEffect.close`.

I see two possible options:

* We change the strategy behind the automatic detection to be a _best guess_ (even if that is not certain) and use the heuristics you suggested. In that case we'd have to think about how this affects handling of evaluation in exporters.
[...]

PR #319, opened 2021-07-13
Fix Blob Responses Containing Information for the REST and HTTP Exporters

PSchmiedmayer:
# Fix Blob Responses containing Information for the REST and HTTP exporters

The Information carried with the `MimeType` of the `Blob` response type was not translated into HTTP header information in the REST and HTTP exporters. In addition any information passed down to the REST exporter when returning a `Blob` was lost in the exporting process.

## :bulb: Proposed solution
This PRs adds and tests the functionality that Information attached to a `Response<Blob>` as well as the `Blob`s MimeType is correctly passed to the HTTP response in the `HTTP` and the `REST` exporters. 

## :gear: Release Notes 
Adds a `.custom(type: String, subtype: String, parameters: [String: String])` `MimeType` that can be used to express `MimeType`s outside of the ones defined in `Apodini`.

### Related Issues
While creating the PR I have encountered an interesting behavior in the automatic detection of the communication pattern that is documented in #318.

### Testing
Added tests that cover the HTTP and REST exporter and test the behavior with `Blob`s and `Information`.

### DIFF
--- a/Tests/ApodiniTests/RESTInterfaceExporterTests.swift
+++ b/Tests/ApodiniTests/RESTInterfaceExporterTests.swift
@@ -383,7 +383,7 @@ class RESTInterfaceExporterTests: ApodiniTests {
     }
     
     func testRESTInformation() throws {
-        struct BlobInformationHandler: Handler {
+        struct InformationHandler: Handler {
             func handle() -> Apodini.Response<String> {
                 Response.send(
                     "Paul",
@@ -395,7 +395,7 @@ class RESTInterfaceExporterTests: ApodiniTests {
         
         struct TestWebService: WebService {
             var content: some Component {
-                BlobInformationHandler()
+                InformationHandler()
             }
             
             var configuration: Configuration {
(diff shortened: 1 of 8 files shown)


Tests/ApodiniTests/RESTInterfaceExporterTests.swift
```diff
@@ -381,4 +381,84 @@ class RESTInterfaceExporterTests: ApodiniTests {
         XCTAssertEqual(restoredHeaders.first(name: .authorization), value)
         XCTAssertEqual(restoredHeaders.first(name: .eTag), "W/\"someTag\"")
     }
+    
+    func testRESTInformation() throws {
+        struct BlobInformationHandler: Handler {
```
theMomax: I guess this should be `(NonBlob)InformationHandler`?

theMomax (approved):
LGTM👍🏻 Just fix that one Handler-name please.

bauer-andreas (approved):
Nothing to add 🚀

Tests/ApodiniTests/RESTInterfaceExporterTests.swift
PSchmiedmayer: Good catch, will be changed to `InformationHandler` 👍

PR #323, opened 2021-07-19
Simplify the Database Configuration

PSchmiedmayer:
# Simplify the Database Configuration

ApodiniDatabase currently provides a wrapper around some possible database configurations that limits the about of possibilities and introduces unnecessary dependences.

## :bulb: Proposed solution
Use the build in FluentKit mechanisms for the database configuration.

## :gear: Release Notes 
Introduces a breaking change for the `DatabaseConfiguration` adding a database databaseID that is passed to FluentKit. E.g.:
```swift
 DatabaseConfiguration(.sqlite(.memory))
```
is changed to
```swift
 DatabaseConfiguration(.sqlite(.memory), as: .sqlite)
```
[Check out the Fluent documentation for more information](https://docs.vapor.codes/4.0/fluent/overview/).


Package.swift
```diff
@@ -163,12 +159,7 @@ let package = Package(
             dependencies: [
                 .target(name: "Apodini"),
                 .target(name: "ApodiniVaporSupport"),
-                .product(name: "Fluent", package: "fluent"),
-                .product(name: "FluentKit", package: "fluent-kit"),
-                .product(name: "FluentMongoDriver", package: "fluent-mongo-driver"),
-                .product(name: "FluentSQLiteDriver", package: "fluent-sqlite-driver"),
```
bauer-andreas, line 177: Removing the SQLLite drivers should allow us to remove the sqlite dependency installs currently happening in the CI right?

bauer-andreas (approved):
Great change. The PR introduces some future proofing for the `ApdoiniDatabase` package.
Also from a dependency point of view it is only a win in general to no longer depend on the individual drivers as direct dependencies.

I got only a small comment in regards to the CI setup, which could be simplified by removing the install steps for the sqllitedev package.

Package.swift
PSchmiedmayer: Sadly no. We still use the in memory database SQLite driver for the unit tests and it therefore it is still required to have the dependencies installed in the CI environment.

bauer-andreas: Maybe related to some missing understanding of fluent internals, but we can use the sqlite driver without specifying it as a dependency in our `Package.swift`? I mean, it compiles, so the answer is probably yes 🙃

PSchmiedmayer: It is still defined as a package dependency (line 46 in Package.swift) but just not a dependency in the ApodiniDatabase target and only in the ApodiniTests target. The dependency is removed here but we still use it in the other test target.

bauer-andreas: Ah, I completely confused the dependencies section 😅 Thanks for the clarification.

PR #326, opened 2021-07-26
Refactor deployment provider structure

hendesi:
# Refactor deployment provider structure (2nd Try)

**Find the detailed proposal [here](https://confluence.ase.in.tum.de/pages/viewpage.action?spaceKey=~ga96xoz&title=Deployment+Structure+Refactoring+Proposal)**

Currently when exporting the structure of a web service's structure, the complete structure is exported and persisted. This might lead to unnecessary overload and does not allow for any customisation or user-specific input. When starting the system on a deployment node, the deployment provider needs to rely on specific environment variables, which is not an optimal design.     

## :bulb: Proposed solution
This PR proposes a refactoring removing the need to export the complete web service structure by moving the generation of the deployment structure to a web service subcommand and the startup of the system on a deployment node to a specific subcommand reducing the reliance of specific environment variables and providing a clear and evolvable interface exposed as part of the web service CLI.

This PR adds two new CLI Commands. The first is `deploy`. This just serves a container for all deployment-related cli commands. The second `export-ws-structure` is a subcommand of `deploy` and handles the retrieval of the web service's structure. See __Additional Information__ on how the commands are added.

This PR expands the current `DeploymentProviderRuntime` protocol by a new property 
```
static var exportCommand: StructureExporter.Type { get }
static var startupCommand: DeploymentStartupCommand.Type { get }
```

The first command `exportCommand` defines the strategy on how the structure of the deployed system is retrieved. This is done by the `StructureExporter` protocol. It conforms to `ParsableCommand`, this means that it is automatically added as a subcommand to `export-ws-structure` when the runtime is set in the deploy configuration. For example, the custom retrieval for the local host deployment might be called by running `swift run TestWebService deploy export-ws-structure local`. The protocol requires the following properties:
```
public protocol StructureExporter: ParsableCommand {
    /// The filePath to which the structure is persisted to
    var filePath: String { get }
    /// The id of the deployment provider that calle
    var identifier: String { get }
    /// Specifies how the id of the deployment node should be computed
    var nodeIdProvider: (Set<CollectedEndpointInfo>) -> String { get }

    func retrieveStructure(
        _ endpoints: Set<CollectedEndpointInfo>,
        config: DeploymentConfig,
        app: Application
    ) throws -> AnyDeployedSystem
}
```

A struct that is implementing this protocol can added arbitrary additional cli arguments and options. The protocol's function `retrieveStructure` is called from `ApodiniDeployInterfaceExporter` after all endpoints have been exported. The service automatically quits after the call. Implementing the function allows to export all information that is really needed by the deployment provider. Nothing more, nothing less. It grants access to all relevant properties.
The return type of the function needs to conform to `AnyDeployedSystem`. A runtime can specify a custom deployed system in its common package and return it from this function. 
Since `StructureExporter` conforms to ParsableCommand, it can and should specify the `run` method. This should follow this following structure: 
```
 public func run() throws {
       let app = Application()

        app.storage.set(DeploymentStructureExporterStorageKey.self, to: self)
        try Service.start(app: app, webService: Service())
}
```
As said, when for example `swift run TestWebService deploy export-ws-structure local` is called, the web service is started and the implementation of `retrieveStructure` is called, the result is saved to disk and the web service is shutdown afterwards.

The second property `startup` that is now required needs to conform to `DeploymentStartupCommand`. 
```
public protocol DeploymentStartupCommand: ParsableCommand {
    /// The file path of the deployment structure json.
    var filePath: String { get }
    /// The id of the deployment node
    var nodeId: String { get }
    var deployedSystem: AnyDeployedSystem.Type { get }
}
``` 
Like `StructureExporter` it also conforms to `ParsableCommand` and is therefore automatically added as a subcommand to the `startup` command. The first two properties are required to be able to initialise the runtime from the `ApodiniDeployInterfaceExporter`. The last property can be set if the deployment provider uses a custom type of `AnyDeployedSystem`. During the startup process the json at the `filePath` is decoded to the specified type. Similar to the structure exporter a struct implementing this protocol should implement the `run` method. It should also follow the same structure using the `DeploymentStartUpStorageKey` storage key. 
With this PR, the `LifecycleHandler` is extended by `func filter(_ endpoints: [AnyEndpoint], app: Application)`. This , for example, allows to disable some endpoints before exporting them. Just create a custom life cycle handler and add it to the app instance the in the `run` method. 

## :gear: Release Notes 
- Refactored logic of structure retrieval. Implement `exportCommand` and `startupCommand` to use the new logic.
- A configuration can now define a `command` that is automatically added to the Apodini CLI if no custom configuration is set

This PR adds an `ParsableCommand.Type` to the `Configuration` protocol, with the default implementation being an empty, invisible command. All configurations that are explicitly define a command have that command automatically added as subcommands to the CLI of the web service as long as no custom `CommandConfiguration` is provided. The reason for this in the first place was, that commands such as `deploy export-ws-structure local` or `deploy startup local` should not have to be manually added as subcommands by the user, just to be able to run the deployment specific actions. This is the same with the subcommands for each deployment provider: By making    
`DeploymentProviderRuntime` require a `exportCommand`, the need for manually declaration is removed. 

As long as the web service has e.g. `ApodiniDeploy(runtimes: [LocalhostRuntime<Self>.self])` in its configuration, all deployment-related subcommands are automatically set. If the `CommandConfiguration` has been override by the user, all needed commands need to be added manually.

### Related PRs
This closes the current refactoring pr #309. 

### Testing
I adjusted the existing test cases to fit the new exporting logic.

### Reviewer Nudging
It's a good idea to look at the `StructureExporter` and `DeploymentStartupCommand` and how the localhost deployment provider implements it and its logic.

### TODO
- [x] Add startup cli and possibility to pass custom actions
- [x] Remove old code relicts
- [x] Add test cases


Package.swift
```diff
@@ -391,6 +391,7 @@ let package = Package(
             dependencies: [
                 .target(name: "Apodini"),
                 .target(name: "ApodiniUtils"),
+                .target(name: "ApodiniOpenAPI"),
```
PSchmiedmayer: As far as I can see this can be removed. You can also remove all `import OpenAPIKit` mentions outside of the AWS and Localhost packages (e.g. CollectedEndpointInfo.swift and ApodiniDeployInterfaceExporter+ExportWebServiceStructure.swift)

Sources/Apodini/Configurations/Configuration.swift
```diff
@@ -14,6 +14,16 @@ public protocol Configuration {
...
+    
+    /// A default CLI command that can be defined by the configuration.
+    /// This command is automatically integrated into the Apodini CLI,
+    /// if the `CommandConfiguration` has not been overridden.
+    var command: ParsableCommand.Type { get }
+    
+    // swiftlint:disable identifier_name
+    /// *For internal use only:* An array of the `command` of the configuration.
+    /// Used to allow iteration over the commands of a `ConfigurationBuilder`.
+    var _commands: [ParsableCommand.Type] { get }
```
PSchmiedmayer, line 26: I don't think this has to be part of the protocol requirement. Can be removed here.

Sources/Apodini/Configurations/Configuration.swift
```diff
@@ -22,6 +32,19 @@ public protocol ConfigurationCollection {
     @ConfigurationBuilder var configuration: Configuration { get }
 }
 
+extension Configuration {
+    // swiftlint:disable identifier_name
+    /// *For internal use only:* An array of the `command` of the configuration.
+    /// Used to allow iteration over the commands of a `ConfigurationBuilder`.
+    public var _commands: [ParsableCommand.Type] {
```
PSchmiedmayer, line 39: Can be made internal as far as I can see.

Sources/Apodini/Configurations/Configuration.swift
```diff
@@ -37,11 +60,25 @@ public struct EmptyConfiguration: Configuration {
...
+
 
 extension Array: Configuration where Element == Configuration {
     public func configure(_ app: Application) {
         forEach {
             $0.configure(app)
         }
     }
+    // swiftlint:disable identifier_name
+    public var _commands: [ParsableCommand.Type] {
```
PSchmiedmayer, line 79: Can be made internal as far as I can see.

Sources/Apodini/WebService.swift
```diff
@@ -34,15 +34,19 @@ extension WebService {
...
     /// - Parameters:
     ///    - waitForCompletion: Indicates whether the `Application` is launched or just booted. Defaults to true, meaning the `Application` is run
-    ///    - webService: The instanciated `WebService` by the Swift ArgumentParser containing CLI arguments.  If `WebService` isn't already instantiated by the Swift ArgumentParser, automatically create a default instance
+    ///    - app: The instanciated `Application` that will be used to boot and start up the web service. Passes a default plain application, if nothing is specified.
+    ///    - webService: The instanciated `WebService` by the Swift ArgumentParser containing CLI arguments.  If `WebService` isn't already instanciated by the Swift ArgumentParser, automatically create a default instance
     /// - Returns: The application on which the `WebService` is operating on
     @discardableResult
-    static func start(waitForCompletion: Bool = true, webService: Self = Self()) throws -> Application {
-        let app = Application()
+    public static func start(waitForCompletion: Bool = true, app: Application = Application(), webService: Self = Self()) throws -> Application {
```
PSchmiedmayer: The second start function can now be removed. You can just replace the all to it blow with the content of the other function and you will probably need to add some `try` statements to the tests.

Sources/ApodiniDeploy/ApodiniDeployCommand.swift
```diff
@@ -0,0 +1,38 @@
...
+/// This command is added to the Apodini CLI by default. If you override `CommandConfiguration` of your web service
+/// and want to use a deployment provider, please make sure to add this manually as a `subcommand`.
+public struct ApodiniDeployCommand: ParsableCommand {
+    public static var configuration = CommandConfiguration(
+        commandName: "deploy",
+        abstract: "Apodini deployment provider",
+        discussion: """
+             Offers utilities for the deployment of an Apodini web service
+             """,
+        version: "0.0.1",
```
PSchmiedmayer: Probably makes sense to match the Apodini version here.

Sources/ApodiniDeploy/ApodiniDeployCommand.swift
```diff
@@ -0,0 +1,38 @@
...
+        subcommands: []
+    )
+    
+    public init() {}
+    
+    public func run() throws {
+        print("deploy")
```
PSchmiedmayer: I would add a more descriptive message here, e.g. print the help message and exit. I think the help message also includes the subcommands, e.g. something like:

```swift
ApodiniDeployCommand.helpMessage()
ApodiniDeployCommand.exit(withError: ...)
```

Sources/ApodiniDeployBuildSupport/DeployedSystem.swift
```diff
@@ -37,74 +59,112 @@ public struct DeployedSystem: Codable {
...
-    
-    
-    public func readUserInfo<T: Decodable>(as _: T.Type) -> T? {
-        try? T(decodingJSON: userInfo)
-    }
+//    
+//    
+//    public func readUserInfo<T: Decodable>(as _: T.Type) -> T? {
+//        try? T(decodingJSON: userInfo)
+//    }
```
PSchmiedmayer:
```suggestion
```

Sources/DeploymentTargetLocalhostRuntime/LocalhostRuntimeSupport.swift
```diff
@@ -52,4 +56,86 @@ public class LocalhostRuntime: DeploymentProviderRuntime {
...
+}
+
+public struct LocalhostStartupCommand<Service: WebService>: DeploymentStartupCommand {
+    public static var configuration: CommandConfiguration {
+        CommandConfiguration(commandName: "local",
+                             abstract: "Start a web service - Localhost",
+                             discussion: """
+                                    Starts up an Apodini web service for the localhost deployment
+                                  """,
+                             version: "0.0.1")
```
PSchmiedmayer:
```suggestion
        CommandConfiguration(
            commandName: "local",
            abstract: "Start a web service - Localhost",
            discussion: "Starts up an Apodini web service for the localhost deployment",
            version: "0.0.1"
         )
```

PSchmiedmayer: The same formatting also applies to the rest of the PR, e.g. above :)

Sources/DeploymentTargetLocalhostRuntime/LocalhostRuntimeSupport.swift
```diff
@@ -52,4 +56,86 @@ public class LocalhostRuntime: DeploymentProviderRuntime {
...
+    public static var exportCommand: StructureExporter.Type {
+        LocalhostStructureExporterCommand<Service>.self
+    }
+    
+    public static var startupCommand: DeploymentStartupCommand.Type {
+        LocalhostStartupCommand<Service>.self
+    }
+}
+
+public struct LocalhostStructureExporterCommand<Service: WebService>: StructureExporter {
```
PSchmiedmayer: I think the commands can be made internal (and e.g. move them to separate files) or private if they stay in the file here.

Sources/DeploymentTargetAWSLambdaRuntime/LambdaRuntime.swift
```diff
@@ -52,4 +57,101 @@ public class LambdaRuntime: DeploymentProviderRuntime {
...
+        LambdaStructureExporterCommand<Service>.self
+    }
+    
+    public static var startupCommand: DeploymentStartupCommand.Type {
+        LambdaStartupCommand<Service>.self
+    }
+}
+
+
+public struct LambdaStructureExporterCommand<Service: WebService>: StructureExporter {
```
PSchmiedmayer: The subcommands and their properties can be made private or moved to an other file (maybe cleaner) and can be made internal.

Sources/Apodini/WebService.swift
PSchmiedmayer: Maybe consider replacing the `waitForCompletion` with an enum that has a bit clearer semantic meeting. I am not sure if the current naming is that great wither. Isn't the `run` method the one that is blocking and the the `waitForCompletion` call only indicates if the `LifecycleHandler` are called with?
- `.run`: Starts up the complete web service and blocks the execution until the web service exits itself or gets killed
- `.start`: Starts up the complete web service but doesn't block so the `run()` method is never called and the web service just exists (probably what we want for quite some things such as the exporters so they don't have to call `exit()`

With the possibility to add additional elements in the future.

Sources/DeploymentTargetAWSLambdaCommon/LambdaCommon.swift
```diff
@@ -70,3 +71,28 @@ public extension AnyOption where OuterNS == DeploymentOptionsNamespace {
...
+    
+    public var deploymentProviderId: DeploymentProviderID
+    
+    public var openApiDocument: OpenAPI.Document
+    
+    public init<T: Encodable>(
+        deploymentProviderId: DeploymentProviderID,
+        nodes: Set<DeployedSystemNode>,
+        userInfo: T?,
+        userInfoType: T.Type = T.self,
```
PSchmiedmayer: Would it be possible to make the `AnyDeployedSystem` have an associated type that is Codable so you don't have to implement the custom encoding there? E.g. a default implementation of `AnyDeployedSystem` just decodes the associated type dynamically.

Sources/ApodiniDeployRuntimeSupport/ExportStructureCommand.swift
```diff
@@ -0,0 +1,86 @@
...
+        version: "0.0.1",
+        subcommands: []
+    )
+
+    public init() {}
+    
+    public func run() throws {
+        fatalError(
+            "Error: Should not be called directly. Use the startup command in the DeploymentProviderRuntime protocol."
+        )
```
PSchmiedmayer: Similar to my comment in the export command we should print out the help here and then exit with an appropriate error.

Sources/ApodiniDeployRuntimeSupport/ExportStructureCommand.swift
```diff
@@ -0,0 +1,86 @@
...
+        
+        public init() {}
+    }
+    
+    public static var configuration = CommandConfiguration(
+        commandName: "startup",
+        abstract: "Startup command for a node of a deployment system",
+        discussion: """
+                    Starts up a node a deployment system
+                    """,
```
PSchmiedmayer:
```suggestion
        discussion: "Starts up a node a deployment system",
```

Sources/ApodiniDeployRuntimeSupport/ExportStructureCommand.swift
```diff
@@ -0,0 +1,86 @@
...
+        
+        public init() {}
+    }
+    
+    public static var configuration = CommandConfiguration(
+        commandName: "export-ws-structure",
+        abstract: "Export web service structure",
+        discussion: """
+                    Exports an Apodini web service structure
+                    """,
```
PSchmiedmayer:
```suggestion
        discussion: "Exports an Apodini web service structure",
```

Sources/ApodiniDeployRuntimeSupport/ExportStructureCommand.swift
```diff
@@ -0,0 +1,86 @@
...
+/// A deployment provider may define a subcommand to this to account for custom options.
+/// **Note**: Should not be called by the user directly
+public struct ExportStructureCommand: ParsableCommand {
+    /// Often used options that can be imported into a subcommand.
+    public struct ExportOptions: ParsableArguments {
+        @Argument(help: "The location of the json file")
+        public var filePath: String = "service-structure.json"
+        
+        @Option(help: "The identifier of the deployment provider")
+        public var identifier: String = "de.lukaskollmer.ApodiniDeploymentProvider.Localhost"
```
PSchmiedmayer: Localhost doesn't seem correct here. I would go for a more neutral `de.tum.in.ase.apodini.deploy` or something like this.

Sources/ApodiniDeployBuildSupport/DeployedSystem.swift
```diff
@@ -37,74 +59,112 @@ public struct DeployedSystem: Codable {
...
 }
 
+/// A node within the deployed system
+public struct DeployedSystemNode: Codable, Identifiable, Hashable, Equatable {
+    /// ID of this node
+    public let id: String
+    /// exported handler ids
+    public let exportedEndpoints: Set<ExportedEndpoint>
+    /// Additional deployment provider specific data
+    public private(set) var userInfo: Data?
```
PSchmiedmayer, line 68: Maybe it makes sense to make this generic over a specific `UserInfo` type that is Codable?

Sources/ApodiniDeployBuildSupport/DeployedSystem.swift
```diff
@@ -9,26 +9,48 @@
...
+///     2. Return the type in your implementation of `retrieveStructure` of `StructureExporter`.
+///     3. In your deployment provider, specify the decode type of `retrieveSystemStructure` with your type.
+public protocol AnyDeployedSystem: Codable {
+    var deploymentProviderId: DeploymentProviderID { get }
+    
+    /// The nodes the system consists of
+    var nodes: Set<DeployedSystemNode> { get }
+    
+    /// Additional, deployment provider specific data
+    var userInfo: Data { get set }
```
PSchmiedmayer: Maybe it makes sense to add an associated type (or two if you also make the `DeployedSystemNode` generic over a type so you can access the userInfo directly after it is decoded and you don't need to introduce a typed additional property in the instantiations of `AnyDeployedSystem`.

Sources/ApodiniDeploy/ApodiniDeployInterfaceExporter.swift
```diff
@@ -145,54 +133,50 @@ class ApodiniDeployInterfaceExporter: LegacyInterfaceExporter {
...
+        
+        let configUrl: URL
+        let currentNodeId: String
+        
+        let deployedSystem: AnyDeployedSystem
+        
+        // The aws deployment provider still uses env variables,
+        // so first check if there are any present
+        if env[WellKnownEnvironmentVariables.fileUrl] != nil,
+           env[WellKnownEnvironmentVariables.currentNodeId] != nil {
```
PSchmiedmayer: What speaks agains moving them to command line arguments in the startup command? As far as I can see they should be usable in that way and are only used on startup as far as I can see.

Sources/ApodiniDeploy/ApodiniDeployInterfaceExporter+ExportWebServiceStructure.swift
```diff
@@ -10,41 +10,28 @@ import Foundation
...
-        try webServiceStructure.writeJSON(
-            to: outputUrl,
+        let config = DeploymentConfig(defaultGrouping: self.exporterConfiguration.config.defaultGrouping, deploymentGroups: allDeploymentGroups)
+        
+        let deployedSystem = try structureExporter.retrieveStructure(Set(self.collectedEndpoints), config: config, app: self.app)
+        try deployedSystem.writeJSON(
+            to: URL(fileURLWithPath: structureExporter.filePath),
             encoderOutputFormatting: [.prettyPrinted, .withoutEscapingSlashes]
         )
+        exit(EXIT_SUCCESS)
```
PSchmiedmayer: As noted in my comment on the `WebService` type it might make sense to move the exit call in a structured and predicable way to the `WebService` start function if e.g. any other logic would have to be executed after the structure exporter and the server doesn't start up If e.g. a developer forgets to add the exit call to their structured exporter.

PSchmiedmayer (changes requested):
Hi @hendesi 
I really like the changes with this PR. As promised in the meeting I took a look at the PR and provide you some comments here and there. Let me know if you have any questions but I think this PR is going in the right direction 💪

@lukaskollmer Will definitely be able to provide more detailed feedback on the deployment provider logic itself 👍

Sources/ApodiniDeploy/ApodiniDeployInterfaceExporter+ExportWebServiceStructure.swift
hendesi: Not sure if I understand your comments correctly, but this method `exportDeployedSystemIfNeeded()` is only executed if the web service is started by the `export-ws-structure [subcommand]` command. For this we don't need any other logic to be executed after the structure exporter has exported the system nor should a developer manually add exit calls anywhere. 
The developer should not call `exportDeployedSystemIfNeeded()` directly, but just needs to define how the system is exported and override `run` to start web service. He does not need to call exit by himself.
So why would he need to pass here any flags how long the web service runs, if the only purpose of this is to export the structure anyway?
While I quite get the intention to provide some generic and nicer way to give a clearer semantic meaning than `waitForCompletion`, I don't think this should be in the scope of the refactoring and honestly I can't see any use cases for this at least for now. 
As said, maybe I misunderstood you, so please clarify if so 👍

Sources/ApodiniDeploy/ApodiniDeployInterfaceExporter.swift
hendesi: I reason i added this was that as far as I remember there were some down sides to not using environment variables when we talked about the refactoring a couple of weeks ago. Maybe @lukaskollmer has some input on that. I'm still no super familiar with the aws deployment provider.

Sources/Apodini/WebService.swift
hendesi: See my [comment](https://github.com/Apodini/Apodini/pull/326#discussion_r686803378) here

Sources/DeploymentTargetAWSLambdaRuntime/LambdaRuntime.swift
hendesi: The intension to make the subcommands public was that if the user/ developer manually overrides the `CommandConfiguration` in its main file, but still wants to support deployment, he needs to add the commands himself. But I agree, the commands could be moved to separate files.

Sources/ApodiniDeployBuildSupport/DeployedSystem.swift
hendesi: I'm going to answer here for all comments that are concerned with the `userInfo`.

Instead of introducing a generic type, I removed `userInfo` completely from `DeployedSystem`. The reason for making deployed system a protocol was to allow deployment providers to subclass it and add abritrary properties that are specific for their use case. Therefore, I created a `LocalhostDeployedSystem` and `LambdaDeployedSystem` that contain the needed properties each. I think this makes it easier for new deployment providers to just provide a struct that has all properties that are needed as it does not constrain the additional info to just one type.

Sources/DeploymentTargetAWSLambdaRuntime/LambdaRuntime.swift
PSchmiedmayer: I see, makes sense ... maybe we should add some documentation to all public declarations so a user is informed about the general purpose of the command and if/how the developer would use it if it is manually needed 👍

Sources/ApodiniDeploy/ApodiniDeployInterfaceExporter.swift
PSchmiedmayer: Ah, I remember that this was due to the fact that this would require multiple version of the web service being uploaded to the S3 storage as the command line arguments are fixed and can not be changed for each Lambda Startup with a `.zip` file, only Environment variables can be changed.

As far as I remember we just considered adding a small shell script or something that passes these environment variables to the command line arguments or explore if it is possible to directly reference environment variables in the startup command for Lambda functions to eliminate them here and make it easier to test this locally as you only have to rely on the CLI interface and do not have to also remember and set environment variables too.

Sources/ApodiniDeploy/ApodiniDeployInterfaceExporter+ExportWebServiceStructure.swift
PSchmiedmayer: The intention behind the refactoring was/is the following: I think it makes is easier to reason about the behavior of the web service if we reduce or even completely eliminate the `exit(...)` calls across the code base and focus on a clear and followable startup + shutdown routine in the WebService. Your example has shown why this is needed. I should have been clearer what I have meant regarding a "developer" and the exit calls: E.g. some other Subcommand (e.g. database reverts, database migrations, generating a migration guide like Eldi is doing) will all have to manually put exit calls in their exporters or custom configuration logic.
Bundling this in the start command and therefore insuring that there are two (and could even be extended to more in the future) startup routines as detailed in my comments above would make this clearer for implementing subcommands that don't rely on starting up the web service completely.
Therefore they would call `try Service.start(options: .startOnly, app: app, webService: Service())` and all exporters and everything would be finished and done and the web service exits controlled after all the work without booting the web service and/or going in a run loop to process requests.

As this surfaces with this PR I would use the "see something, do something" approach and try to make it nice right awai instead of ending up with multiple dependencies implementing this themselves and having to introduce a change that they have to adapt to eventually later on.

hendesi: Thanks for clarifying!
I agree, this definitely makes sense. I will have a look at this. 👍

hendesi: @PSchmiedmayer 88d6573 adds `WebServiceExecutionMode`. Let me know what you think about this approach.

Sources/ApodiniDeploy/ApodiniDeployInterfaceExporter.swift
lukaskollmer: @PSchmiedmayer you're right, the lambdas require these values be passed via the environment if there is to be only one zip file.
You're also right in that it should be possible to, as part of the lambdas' `boostrap` program, read these from the environment and set them as CLI params; I'll see if I can implement that into this PR, that shouldn't be too much work.

PSchmiedmayer: Great to hear @lukaskollmer, thanks!

Sources/ApodiniDeploy/ApodiniDeployInterfaceExporter+ExportWebServiceStructure.swift
PSchmiedmayer: @hendesi Yes, I think this reflects my vision. I had two comments regarding the changes that you can find below

lukaskollmer (approved):
Alright, first of all sorry for the delay in reviewing this.

I had a look at the changes and I think this is a pretty nice improvement over the current implementation in develop, particularly the more command-based structure.

@hendesi I made some changes to your code, wrt the way lifecycle handlers can filter endpoints. The previous implementation was somewhat confusing, in that it would run the first lifecycle handler twice, re-filter all endpoints for each interface exporter (even though it would always use the previous filtering's output as the next iteration's input, meaning that (assuming the lifecycle handlers filter deterministically) all iterations after the first were unnecessary since the output would always equal the input).

The new implementation changes this, by:
- renaming `filter` to `map`, which better describes the function's intended use case
- adding an `interfaceExporter` parameter to `map`, so that a lifecycle handler can make different mapping decisions based on the used interface exporter
- calling map once per interface exporter, with the full set of collected endpoints as input (instead of using the previous call's output as input)

@PSchmiedmayer I also changed the lambda integration to use the bootstrap script to read the apodini-specific environment variables and pass them as CLI parameters. This means that the ApodiniDeploy-specific parts of Apodini now don't contain any environment-value-dependent code paths anymore (apart from things like reading credentials, and the tests)

I'm not sure if this can be merged as-is, there's some tests that are still failing for me (primarily the NegativeCompileTests), but since all deployment-related tests work fine I'm not sure whether these failures are related to this PR or just to the fact that I'm running this on a somewhat unstable beta-OS.

---

One additional thing I noticed, and this doesn't really fall w/in the scope of this PR but I just wanted to bring it up anyway, is that having a distinction between `LifecycleHandler.willBoot` and `LifecycleHandler.didBoot` seems pretty pointless to me.
The current implementation of `Application.boot()` is literally just https://github.com/Apodini/Apodini/blob/d68e398166acc43308fecd5ed0529a7bca66dc9e/Sources/Apodini/Application/Application.swift#L192-L200

Since there's nothing that happens between invoking all `willBoot` handlers and invoking all `didBoot` handlers, I'm not really sure why there is a distinction between the two to begin with.

Also, I think that `LifecycleHandler.shutdown` should be allowed to throw, considering that `willBoot` and `didBoot` are.

PSchmiedmayer:
@lukaskollmer Thanks for the detailed review and improvements!

[I like the new implementation changes you mentioned above.](https://github.com/Apodini/Apodini/pull/326#pullrequestreview-730130960)

Also thanks for improving the AWS deployment provider to complete rely on command line arguments and adding the startup script.

Regarding the `NegativeCompileTests`: This is a current issue after updating the CI to beta 5. It also blocks the following PR: #331 ... I already discussed this with @Supereg, he will take a look at this after handing in his bachelor's thesis on Monday 👍

I also agree with all the proposed changes to the `LifecycleHandler`s, combining `willBoot` and `didBoot` makes sense and making `shutdown` throwing also sounds good.

bauer-andreas:
> Regarding the NegativeCompileTests: This is a current issue after updating the CI to beta 5. It also blocks the following PR: #331 ... I already discussed this with @Supereg, he will take a look at this after handing in his bachelor's thesis on Monday 👍

@PSchmiedmayer if it is blocking this PR and similarly your PR Paul its just fine with me, if we disable those tests until I can work on them. As long as nobody completely rewrites the protocol hierarchy for the Metadata system, the potential for introducing a regression should be low 😄.
Just disable the XCTest case inside here: https://github.com/Apodini/Apodini/blob/d68e398166acc43308fecd5ed0529a7bca66dc9e/Tests/NegativeCompileTestsRunner/NegativeTestRunner.swift#L13

hendesi:
@lukaskollmer Many thanks for the review and the adjustments you made to this branch! I implemented the few changes you proposed regarding the lifecycle methods.

@PSchmiedmayer I think beside the failing negative compile, all of your requested changes have been addressed. Would be great if you could give this another look to see if there are any open issues? 👍

Package.swift
```diff
@@ -463,6 +462,7 @@ let package = Package(
...
```
PSchmiedmayer: As far as I can see this is still not needed for the `ApodiniDeployBuildSupport`.

Package.swift
```diff
@@ -463,6 +462,7 @@ let package = Package(
...
                 .product(name: "Logging", package: "swift-log"),
                 .product(name: "Runtime", package: "Runtime"),
                 .product(name: "OpenAPIKit", package: "OpenAPIKit")
```
PSchmiedmayer: I think we can remove this dependency here.

Package.swift
```diff
@@ -451,7 +451,6 @@ let package = Package(
                 .target(name: "Apodini"),
                 .target(name: "ApodiniUtils"),
                 .target(name: "ApodiniVaporSupport"),
-                .target(name: "ApodiniOpenAPI"),
                 .target(name: "ApodiniDeployBuildSupport"),
                 .target(name: "ApodiniDeployRuntimeSupport"),
                 .product(name: "OpenAPIKit", package: "OpenAPIKit")
```
PSchmiedmayer: I think we can remove this dependency here. Please check what dependencies are really needed for the deployment related packages.

Sources/Apodini/Configurations/Configuration.swift
PSchmiedmayer: As far as I can see the comment has not been addressed yet. Is there a clear requirement to make this part of the protocol requirement. Also applies to my comment here: https://github.com/Apodini/Apodini/pull/326/files#r686172819

Sources/Apodini/WebService.swift
```diff
@@ -32,51 +32,54 @@ public extension WebService {
...
     
-
-     /// This function is provided to start up an Apodini `WebService`. The `app` parameter can be injected for testing purposes only. Use `WebService.start()` to startup an Apodini `WebService`.
-     /// - Parameters:
-     ///    - app: The app instance that should be injected in the Apodini `WebService`
-     ///    - webService: The instanciated `WebService` by the Swift ArgumentParser containing CLI arguments.  If `WebService` isn't already instanciated by the Swift ArgumentParser, automatically create a default instance
-    static func start(app: Application, webService: Self = Self()) {
+    
+    /// Start up a web service using the specified application. Does not boot or run the web service. Intended primarily for testing purposes.
+    func start(app: Application) {
```
PSchmiedmayer, line 73: As far as I can see this method can be removed and merged with the other `start` method above.

Sources/ApodiniDeployRuntimeSupport/ExportStructureCommand.swift
```diff
@@ -0,0 +1,84 @@
...
+        discussion: "Starts up a node a deployment system",
+        version: "0.3.0"
+    )
+
+    public init() {}
+    
+    public func run() throws {
+        StartupCommand.helpMessage(columns: nil)
+        StartupCommand.exit(withError:
+                                    ApodiniDeployRuntimeSupportError(message: "Calling this command directly is not supported.")
```
PSchmiedmayer: This intention looks a bit strange, I would indent it +4 spaces from the previous line.

Sources/ApodiniDeployRuntimeSupport/ExportStructureCommand.swift
```diff
@@ -0,0 +1,84 @@
...
+        commandName: "export-ws-structure",
+        abstract: "Export web service structure",
+        discussion: "Exports an Apodini web service structure",
+        version: "0.3.0"
+    )
+    
+    public func run() throws {
+        ExportStructureCommand.helpMessage(columns: nil)
+        ExportStructureCommand.exit(withError:
+                                    ApodiniDeployRuntimeSupportError(message: "Calling this command directly is not supported.")
```
PSchmiedmayer: This intention looks a bit strange, I would indent it +4 spaces from the previous line.

PSchmiedmayer (approved):
@hendesi [As requested](https://github.com/Apodini/Apodini/pull/326#issuecomment-899036767) I have taken a second look at the changes. I only had a few comments that should be resolved but apart from this I approve the changes, thanks for all the work that went into the PR and also a huge thanks to @lukaskollmer for the help! 💪

#331 Will also introduce a `XCSkip` statement for the Negative compile tests as suggested by @Supereg ([comment](https://github.com/Apodini/Apodini/pull/326#issuecomment-899036636)) and should therefor also unblock this PR.

Sources/Apodini/WebService.swift
lukaskollmer: I'd be careful about that. @hendesi did merge them as part of this PR, but I reverted this to the previous structure (having two of them) because the tests sometimes (i.e. essentially always) invoked the wrong overload (and ended up actually starting a web service, instead of just populating an `Application` object, thus causing the tests to hang; you can see this if you look into the github actions history, there's a bunch of test runs that timed out after 6 hours of inactivity). I wanted to make it more difficult to accidentally call the wrong overload, that's why it's a separate function again. Also, this one is internal whereas the first one is public, because we probably don't want to expose the entry point intended for testing.

Some additional thoughts on this:
- We essentially have two separate versions of the `start` function, one that may be blocking (because it actually starts a web service, this is the first one), and one that is always non-blocking (because it only populates the `Application`, this is the second one)
- I agree that this is not an ideal solution, but I'm not really sure what would be the best way to fix this.
- For example, I'd also like to get rid of the `webService` parameter, since we can simply instantiate a web service directly in the static function. I checked this and there is no single call site in all of Apodini that actually uses the `webService` parameter (edit: with the exception of the `run` function required by SwiftArgumentParser, but i feel like that shouldn't require introducing a publicly exposed parameter here)
- Were they to be combined into one function, there'd probably need to be a distinction as to whether the function is being invoked as part of a unit test or not. This matters because in the case of running tests, the start function would get invoked many times. Which means that it cannot perform any operations that cannot be performed multiple times. This is currently the case w/ the public overload because that initialises the logging system, which will crash if you attempt to initialise it multiple times. The second overload intentionally does not initialise the logging system.
- I'm also not sure if `startup, `boot`, and `run` are necessarily the best words to use here. For me at least, "boot" would be something that occurs as part of a "startup" operation, rather than the other way around (as is currently the case). But then again I don't really have any better ideas as to what this could be called, or how it could be implemented.
- An alternative I considered would be to move this to the `Application` object, so that you'd have `app.start(with: MyWebService())` rather than `MyWebService().start(app: app)`. But that wouldn't be as nice to use as the current API

PSchmiedmayer: @lukaskollmer Thanks for the detailed overview. I agree with the assessments you made in your comments and for the sake of closing this PR I agree with keeping the two start functions as they are for now. 
It would be great if you can transform your comment in an issue in this repo so we don't forget about this and then improve this in a subsequent PR that is tailored on improving this part of the tests and startup routine 👍

Sources/Apodini/Configurations/Configuration.swift
hendesi: @PSchmiedmayer As far as I tested, this needs to stay in the Configuration protocol  and public, otherwise the commands won't get added correctly to the cli, but just the default `EmptyCommand`

Sources/Apodini/Configurations/Configuration.swift
hendesi: @PSchmiedmayer As far as I tested, this needs to stay in the Configuration protocol  and public, otherwise the commands won't get added correctly to the cli, but just the default `EmptyCommand`

PSchmiedmayer: @hendesi You are right, I have overlooked the dynamic calling behavior of protocol extensions. You are right that is is required to have that element in the protocol requirements. You can discard my comments about that 👍

Sources/Apodini/WebService.swift
PSchmiedmayer: Created #337 to document this issue.

PR #328, opened 2021-07-30
Advanced `Delegate` example for TestWebService

theMomax:
# Advanced `Delegate` example for TestWebService

## :recycle: Current situation & Problem & Solution
TestWebService did not feature extensive use of `Delegate`. I thought adding another service would do no harm and I could need an example for my thesis anyways.

During the implementation I noticed that the `.environment(_:_:)` functions of `Delegate` required a `WritableKeyPath`, which was unnecessary as they only override the _local_ environment. Thus I also changed that to just `KeyPath`.

I added a _Weather_ subsystem to the TestWebService. It features two endpoints that return weather-information for a specific date and location. The _Weather_ subsystem's _trip_ subsystem uses an advanced `Delegate` based modifier which repeatedly evaluates the normal weather-endpoints to generate the weather-information for a whole road trip.

## :gear: Release Notes 
* `Delegate.environment(_:_:)` and `Delegate.setObservable(_:to:)` no longer require `WritableKeyPath`s, but normal `KeyPath`s instead. Any preexisting code should work without adaptions.

### Testing
No additional tests.

### Reviewer Nudging
Should be easy😅. Just don't take the logic in the route- and temperature-service for too serious!


Sources/Apodini/Properties/Delegate.swift
```diff
@@ -153,7 +153,7 @@ extension Delegate {
 extension Delegate {
     /// Change a `delegate`'s `ObservedObject` to observe another `value`.
     @discardableResult
-    public func setObservable<V: ObservableObject>(_ keypath: WritableKeyPath<D, ObservedObject<V>>, to value: V) -> Delegate {
+    public func setObservable<V: ObservableObject>(_ keypath: KeyPath<D, ObservedObject<V>>, to value: V) -> Delegate {
```
bauer-andreas, line 156: By migrating to requiring writable key paths anymore, you can remove the explicit setter in the AuthoirzedProperty of ApodiniAuthorization:

https://github.com/Apodini/Apodini/blob/9029e6899717fce389280f26e5c52a84f17a7444/Sources/ApodiniAuthorization/AuthorizedProperty.swift#L71-L73

bauer-andreas (approved):
looks good to me. Though, I assumed that the example would make sense somehow lol

Just a small comment, pointing on some code cleanup which can be removed now.

philippzagar (approved):
LGTM!

Sources/Apodini/Properties/Delegate.swift
theMomax: Removed in 54f20eec9d74639a56649865425b9b7b4666dce3 👍🏻

PR #330, opened 2021-08-03
Apodini Observe (for logging)

philippzagar:
# Apodini Observe (for logging)

At the moment, logs can be created in a `Handler` via the `Logger` in the `Environment`. An example could look like this:
```swift
struct Test: Handler {
    @Environment(\.logger) var logger

    func handle() -> String {
        logger.info("Hello world")
        return "Hello! 👋"
    }
}
```

Furthermore, log messages are generated at startup (eg. to print out the current working directory or an OpenAPI specification).
This approach works reasonably well for development purposes since the logs are directed to stdout and not further processed. 
However, in production systems, this approach has some problems:
- The to be used logging backend (so the `LogHandler` which actually logs the data since [swift-log](https://github.com/apple/swift-log) only provides an API) is fixed to direct the output to stdout and can't be configured by the developer of the `WebService`. In a production system, the logs should be directed to something like Logstash (eg. with the ELK stack)
- The `logLevel`, so what kind of logs to actually collect, is only statically configured to `info` or `debug`
- The `Logger`doesn't automatically include any metadata that could be useful to track a fault in the code (eg. parameter values, etc.)

## :bulb: Proposed solution
We introduce `ApodiniObserve`, a package that allows to easily instrument the code and collect logging information in Apodini. It dynamically aggregates metadata from the respective `Connection`, `Request` as well as from the `BlackBoard` and provides this metadata to the `Logger`. The logging data then can be sent to an arbitrary logging backend (in our case the [swift-log-elk](https://github.com/Apodini/swift-log-elk) backend which was also written by us).

The main building block is the `ConfiguredLogger`, a `DynamicProperty` that provides preconfigured Logging services in a `Handler`. A logger provided by the `ConfiguredLogger` wrapper is preconfigured and automatically contains certain metadata like:
- `Endpoint` data
- `Exporter` type
- `Connection` data
- `Information` (of the `Connection`)
- `Request` data like the actual parameter values (capped at a certain size), parameterNamespace, operation type, service type etc.
- Many more...

To aggregate the metadata properly, we made some changes to `Delegate` and introduced an `InitialStorage` (to store logging metadata) that is available before the `Delegate` is activated.

An example Logging scenario could look like this:
```swift
    @Environment(\.logger)
    var logger
    
    @Environment(\.eventLoopGroup)
    var eventLoopGroup

        REST {
            OpenAPI()
        }
        
        LoggerConfiguration(
            logHandlers: LogstashLogHandler.init, StreamLogHandler.standardOutput,
            logLevel: .debug) {
            LogstashLogHandler.setup(
                hostname: "0.0.0.0",
                port: 31311,
                useHTTPS: false,
                eventLoopGroup: eventLoopGroup,
                backgroundActivityLogger: logger,
                uploadInterval: TimeAmount.seconds(5),
                logStorageSize: 500_000,
                maximumTotalLogStorageSize: 4_000_000
            )
        }
    }

    var content: some Component {
        Greeter()
    }
}

struct Greeter: Handler {
    @ConfiguredLogger var logger

    func handle() -> String {
        logger.info("Hello world")
        return "Hello! 👋"
    }
}
```

This would result in the following log entry in ElasticSearch/Kibana (I omitted the parameters in the 'Handler'): https://gist.github.com/philippzagar/9b51bded696a4b1e623801ca52c5ae9d (I shared it as a gist to don't bloat the PR)

Note that `ApodiniObserve` is not depending on specific logging backends like [swift-log-elk](https://github.com/Apodini/swift-log-elk). Therefore, only the actually implemented `WebService` has to depend on [swift-log-elk](https://github.com/Apodini/swift-log-elk). However, this requires access to properties of the `Application` (eg. EventLoopGroup, Logger) within the `Configuration`of the `WebService` to configure the logging backend which results in the need to inject the `Application` to the `WebService`

Something like that is now possible:
```swift
    @Environment(\.logger)
    var logger
    
    @Environment(\.eventLoopGroup)
    var eventLoopGroup

        LoggerConfiguration(logger, eventLoopGroup)       // Just an example 
    }
}
```

This PR also fixes a bug, where the [Swift argument parser](https://github.com/apple/swift-argument-parser) doesn't copy the values of property wrapped variables in the `WebService` to the reinstanciated `WebService` after the arguments are parsed and inserted into the respective properties. To overcome this, we save the data of the property wrapped variables in the `WebService` before parsing the arguments and then restoring them to the reinstanciated `WebService`. This is done via a `Mirror` and we were able to circumvent using the [Runtime framework](https://github.com/wickwirew/Runtime) via some class type tricks (via a `Box` approach to wrap values inside a class).

Also, huge thanks to @theMomax for having the patience to explain the best approaches to collect logging metadata after the changes introduced in #307 :)

## :gear: Release Notes 
This PR doesn't introduce breaking changes to Apodini, just fixed a bug regarding the persistence of property wrapped variables in the `WebService` during Argument Parsing with the [Swift argument parser](https://github.com/apple/swift-argument-parser). Also introduces `InitialStorage` to `Delegate`s and allows accessing the `Environment` directly in the `WebService`.

### Related PRs
It took several iterations to get to a final solution for collecting logging metadata (especially after #307 which basically resulted in a complete rewrite of `ApodiniObserve`). Please take a look at https://github.com/Apodini/Apodini/tree/feature-logging-new or https://github.com/Apodini/Apodini/tree/feature/logging for past implementations.

Also, `ApodiniObserve` is designed to use the [swift-log-elk](https://github.com/Apodini/swift-log-elk) logging backend (as shown above). Feels free to check this repository out as well.

### Tests
End-to-End tests will follow once Apodini Observe has the ability to collect metrics

### Reviewer Nudging
Start by taking a look at the `ConfiguredLogger` to understand the basic principles and what data is aggregated and how. Then check out the `LoggerExporter` and the changes to `Delegate`. Also take a look at the `WebService` and `Environment` where changes were made to allow `@Environment` in the `WebService` and cache the values of property wrapped variables in the `WebService`.


philippzagar:
It seems that an ApodiniDeploy test case fails in a certain configuration:

```
error	09-Aug-2021 01:08:14	/opt/bamboo-agent-home/xml-data/build-dir/APODINI-APODINI65-MIRBT/Tests/ApodiniDeployTests/LambdaDeploymentProviderTests.swift:208: error: -[ApodiniDeployTests.LambdaDeploymentProviderTests testLambdaDeploymentProvider] : XCTAssertEqual failed: ("200") is not equal to ("500")
error	09-Aug-2021 01:08:14	/opt/bamboo-agent-home/xml-data/build-dir/APODINI-APODINI65-MIRBT/Tests/ApodiniDeployTests/LambdaDeploymentProviderTests.swift:315: error: -[ApodiniDeployTests.LambdaDeploymentProviderTests testLambdaDeploymentProvider] : failed - request to '/v1/aws_rand?lowerBound=7&upperBound=520' failed.: The data couldn’t be read because it is missing.
error	09-Aug-2021 01:08:21	/opt/bamboo-agent-home/xml-data/build-dir/APODINI-APODINI65-MIRBT/Tests/ApodiniDeployTests/LambdaDeploymentProviderTests.swift:214: error: -[ApodiniDeployTests.LambdaDeploymentProviderTests testLambdaDeploymentProvider] : Asynchronous wait failed: Exceeded timeout of 10 seconds, with unfulfilled expectations: "Test /v1/aws_rand".
error	09-Aug-2021 01:08:25	Test Case '-[ApodiniDeployTests.LambdaDeploymentProviderTests testLambdaDeploymentProvider]' failed (307.503 seconds).
error	09-Aug-2021 01:08:25	Test Suite 'LambdaDeploymentProviderTests' failed at 2021-08-09 01:08:25.185.
error	09-Aug-2021 01:08:25		 Executed 1 test, with 3 failures (0 unexpected) in 307.503 (307.503) seconds
```

I'm not quite sure if the PR caused the problem since I didn't change anything in `ApodiniDeploy`. I'll look into that.

philippzagar:
> It seems that an ApodiniDeploy test case fails in a certain configuration:
> 
> ```
> error	09-Aug-2021 01:08:14	/opt/bamboo-agent-home/xml-data/build-dir/APODINI-APODINI65-MIRBT/Tests/ApodiniDeployTests/LambdaDeploymentProviderTests.swift:208: error: -[ApodiniDeployTests.LambdaDeploymentProviderTests testLambdaDeploymentProvider] : XCTAssertEqual failed: ("200") is not equal to ("500")
> error	09-Aug-2021 01:08:14	/opt/bamboo-agent-home/xml-data/build-dir/APODINI-APODINI65-MIRBT/Tests/ApodiniDeployTests/LambdaDeploymentProviderTests.swift:315: error: -[ApodiniDeployTests.LambdaDeploymentProviderTests testLambdaDeploymentProvider] : failed - request to '/v1/aws_rand?lowerBound=7&upperBound=520' failed.: The data couldn’t be read because it is missing.
> error	09-Aug-2021 01:08:21	/opt/bamboo-agent-home/xml-data/build-dir/APODINI-APODINI65-MIRBT/Tests/ApodiniDeployTests/LambdaDeploymentProviderTests.swift:214: error: -[ApodiniDeployTests.LambdaDeploymentProviderTests testLambdaDeploymentProvider] : Asynchronous wait failed: Exceeded timeout of 10 seconds, with unfulfilled expectations: "Test /v1/aws_rand".
> error	09-Aug-2021 01:08:25	Test Case '-[ApodiniDeployTests.LambdaDeploymentProviderTests testLambdaDeploymentProvider]' failed (307.503 seconds).
> error	09-Aug-2021 01:08:25	Test Suite 'LambdaDeploymentProviderTests' failed at 2021-08-09 01:08:25.185.
> error	09-Aug-2021 01:08:25		 Executed 1 test, with 3 failures (0 unexpected) in 307.503 (307.503) seconds
> ```
> 
> I'm not quite sure if the PR caused the problem since I didn't change anything in `ApodiniDeploy`. I'll look into that.

So I noticed that this test case already went through once (https://bamboo.ase.in.tum.de/browse/APODINI-APODINI65-28), I just did some Doc improvements in between. Therefore I guess the error lies in the test case itself (of ApodiniDeploy).

Sources/ApodiniExtension/HTTPInformation+LoggingMetadata.swift
```diff
@@ -0,0 +1,74 @@
...
+// SPDX-FileCopyrightText: 2019-2021 Paul Schmiedmayer and the Apodini project authors (see CONTRIBUTORS.md) <paul.schmiedmayer@tum.de>
+//
+// SPDX-License-Identifier: MIT
+//    
+
+import Foundation
+import Apodini
+import Logging
+
+public struct LoggingMetadataInformation: Information {
```
theMomax: Could you please move this file out of ApodiniExtension into ApodiniObserve? It is purely related to logging and not part of any ApodiniExtension APIs.

theMomax (commented):
I only looked at the general structure and where things are implemented. That part looks good to me (except for the one file I commented on). I'd suggest someone else (@Supereg @eldcn @hendesi) do the real review as I too contributed to this PR.

Sources/ApodiniExtension/HTTPInformation+LoggingMetadata.swift
philippzagar: Well the problem is that then `ApodiniGRPC`, `ApodiniWebSocket` etc. would all need to depend on `ApodiniObserve`. That's not the goal of `ApodiniObserve`, maybe we will even break it out to a separate repo sometime.

Since I didn't want to place the file in `Apodini` itself, i did it in `ApodiniExtension`. Do you have other suggestions? :)

theMomax: I'd suggest to not think in terms of use-cases but in terms of features:
* introduce a new protocol `StringKeyedEncodableInformation` (in ApodiniExtension), which requires a string key plus an encodable value
* introduce a new struct `TextInformation` (in ApodiniExtension), which just has a string key and string value
* conform existing HTTPInformation types and `TextInformation` to `StringKeyedEncodableInformation` (in ApodiniExtension)
* move `LoggingMetadataInformation` to ApodiniObserve and make it initializable from a `StringKeyedEncodableInformation`

Would that work?

philippzagar: Yep that should work, I'll take a closer look in the evening

Sources/Apodini/Semantic Model Builder/InterfaceExporter/InterfaceExporter.swift
```diff
@@ -69,6 +69,19 @@ public extension InterfaceExporter {
     }
 }
 
+public struct ExporterTypeMetadata: EnvironmentAccessible {
```
bauer-andreas: Not sure if it is just me, but I see the potential to confuse the term Metadata with a) the Apodini Metadata system and b) Logger Metadata.

Would it maybe make sense to spend some more characters to make that explicit? For example using `ExporterTypeLoggerMetadata`?

Sources/Apodini/Semantic Model Builder/Model/EndpointParameter.swift
```diff
@@ -153,18 +153,25 @@ public struct EndpointParameter<Type: Codable>: _AnyEndpointParameter, Identifia
...
         self.defaultValue = defaultValue
-
-        // If somebody wants to make this more fancy, one could add options into the @Parameter initializer
-        var description = "@Parameter var \(name): \(Type.self)"
+        
+        let httpOption = options.option(for: PropertyOptionKey.http)
+        var description = "@Parameter"
+        
+        if let httpOption = httpOption {
+            description += "(.http(.\(httpOption))"
```
bauer-andreas: How hard would it be to not just hardcode the http option into the response, but any arbitrary option?
I assume the biggest problem would be to retrieve the property name?

Sources/ApodiniHTTPProtocol/Information/Expires.swift
```diff
@@ -10,7 +10,7 @@ import Foundation
 import Apodini
 
 
-// MARK: ETag
+// MARK: Expires
```
bauer-andreas, line 13: Good catch 😄

Sources/ApodiniObserve/LoggerExporter.swift
```diff
@@ -0,0 +1,86 @@
...
+            let responseType: ResponseType
+            let serviceType: ServiceType
+        }
+        
+        public var value: BlackboardMetadata
+    }
+    
+    // Not sure if those are maybe needed?
+    let app: Apodini.Application
+    let exporterConfiguration: LoggerConfiguration
```
bauer-andreas: are you sure now? or what's the point of the comment? 🙃

Sources/ApodiniObserve/ConfiguredLogger.swift
```diff
@@ -0,0 +1,232 @@
...
+    
+    /// The ``Storage`` of the ``Application``
+    @Environment(\.storage)
+    var storage: Storage
+    
+    /// The ``Logger`` of the ``Application``
+    @Environment(\.logger)
+    var logger: Logger
+    
+    /// Metadata from the ``Blackboard`` that is injected into the environment of the ``Handler``via a ``Delegate``
```
bauer-andreas: nit

```suggestion
    /// Metadata from the ``Blackboard`` that is injected into the environment of the ``Handler`` via a ``Delegate``
```

Sources/ApodiniObserve/ConfiguredLogger.swift
```diff
@@ -0,0 +1,232 @@
...
+            else {
+                #if DEBUG
+                builtLogger?.logLevel = .debug
+                #else
+                builtLogger?.logLevel = .info
+                #endif
+            }
+        } else {
+            // Not pretty, but otherwise ApodiniObserve would need to depend on ApodiniWebsocket
+            if String(describing: exporterTypeMetadata.exporterType) == "WebSocketInterfaceExporter" {
```
bauer-andreas: Are there plans to solve this issue with a follow up PR, creating a generic way of doing those things?

Hardcoding the type name is pretty fragile and not really robust e.g. easy to break with a potential refactoring.
For now, do we have at least some regression testing that it is ensured that this doesn't get silently out of date and breaks functionality?

Sources/ApodiniObserve/ConfiguredLogger.swift
```diff
@@ -0,0 +1,232 @@
...
+                // Write connection state
+                builtLogger?[metadataKey: "connectionState"] = .string(connection.state.rawValue)
+                
+                // Write request metadata
+                builtLogger?[metadataKey: "request"] = .dictionary(self.getRequestMetadata(from: connection.request))
+            }
+        }
+        
+        return builtLogger!
```
bauer-andreas: Using a if let or guard let binding you should probably be able to get rid of the force unwrap?

Sources/ApodiniObserve/ConfiguredLogger.swift
```diff
@@ -0,0 +1,232 @@
...
+    @Environment(\.storage)
+    var storage: Storage
+    
+    /// The ``Logger`` of the ``Application``
+    @Environment(\.logger)
+    var logger: Logger
+    
+    /// Metadata from the ``Blackboard`` that is injected into the environment of the ``Handler``via a ``Delegate``
+    @Environment(\LoggerExporter.BlackboardMetadata.value)
+    var blackboardMetadata
```
bauer-andreas: CC: @theMomax 

Just a though. This environment property requires that in every case the user properly configured the LoggerExporter in their web service.
If the user forgot to do that this would crash a runtime with some generic error that the keyPath couldn't be found in the environment.
I would propose to provide a proper error message to the user, notifying them that they forgot to configure a logger in their web service (e.g. particularly useful when using off-the-shelve components).
There is straightforward way to check existence of an environment keyPath. In my Authorization PR I did it by declaring the keyPath property as an extension to the `Application`. In those cases, the `@Environment` property will call the getter of the property on the Application if it doesn't find an according value in the environment.
There you could fatal out with a proper runtime error message, maybe?

@theMomax not sure if we should maybe think about a more straightforward way of dynamically checking existence of an environment value(?). This is now the second time that we have this use case I think.

Side note: Am I right with the following: as soon as you use a `@ConfiguredLogger` you are required to have an according LoggerConfiguration in your web service right? Thinking of the use case of using off-the-shelve/third-party components, which uses a `@ConfiguredLogger` (e.g. to provide maximum compatibility), would it make sense to somehow provide a default logger, such that users which don't necessarily have a dedicated Logger configuration can still use such prebuilt components?

Sources/ApodiniObserve/ConfiguredLogger.swift
```diff
@@ -0,0 +1,232 @@
...
+                } else if let expires = anyHTTPInformation.typed(Expires.self) {
+                    partialResult[Expires.header] = .string(expires.rawValue)
+                } else if let redirectTo = anyHTTPInformation.typed(RedirectTo.self) {
+                    partialResult[RedirectTo.header] = .string(redirectTo.rawValue)
+                } else {
+                    partialResult[anyHTTPInformation.key.key] = .string(anyHTTPInformation.value)
+                }
+            }
+        }
+    }
```
bauer-andreas: I don't really like, that we have an explicit dependency on `ApodiniHTTPProtocol`. As far as I can see, this is the only reason why this dependency exists?

First of all, is there any value in saving e.g. the Cookie Information as a `.dictionary` and don't just use the `.string` raw value format? Afaik this is the only reason currently for enumerating every currently supported HTTPInformation type (Etag, Redirect, etc). I mean, any other header is currently also just passed as the rawValue representation, and any future HTTPInformation is probably treated the same unless someone **manually** adjust this implementation right here.
So for one, I would propose to not look at the possible `InformationInstantiables` of an Information type but instead always just use the value/rawValue of the "base" `Information` type. In the example of `AnyHTTPInformation` one would just use the string header and the string value. Specifically, as any third party can introduce their own typed `HTTPInformation` implementation, I don't think it make sense to try to maintain an enumeration of those to somehow provide more "detailed" Metadata?

Secondly, this currently only covers `AnyHTTPInformation` Information and ignores any other potential Information instances. I think we should be able to use the `InformationClass` feature to provide a more generic implementation for this (refer to e.g. `HTTPHeaderInformationClass` for a fully working example).
Would something like the following work:
1. Create a `StringKeyedStringInformationClass` e.g. 
```swift
public protocol StringKeyedStringInformationClass: InformationClass {
    var entry: (key: String, value: String) { get }
}
```
Maybe you could realize some kind of `StringKeyDecodableInformationClass` reusing your `IntermediateRepresentation`. You just need to keep it type erased as we later use the information class type to filter for those in an array loop (refer to step 3 or e.g. see how it is done with the `HTTPHeaderInformationClass` in the `HTTPHeaders+InformationKey.swift` file).
The `StringKeyedStringInformationClass` could maybe be stilled introduced to be a more specific version of a `StringKeyDecodableInformationClass`(?).
2. Ideally, the above introduced InformationClass is placed inside Apodini as a "default" InformationClass. Such that the `ApodiniHTTPProtocol` package can use it without requiring a dependence between it and `ApodiniAuthorize`.
3. And finally, `ApodiniObserve` would use a logic like the following to collect all the information (replacing the code above):
```swift
 informationSet.reduce(into: [:]) { partialResult, info in
     if let info = info as? StringKeyedStringInformationClass {
         partialResult[info.entry.key] = .string(info.entry.value)
     }
 }
```


Would that make sense to you? Could that even work?

Sources/ApodiniExtension/HTTPInformation+LoggingMetadata.swift
bauer-andreas: Instead of creating weird generic defaults in ApodiniExtension which only get used for a very specific use case, can we do the following:

* Move the `LoggingMetadataInformation` into `ApodiniObserve`. It is specific to this package in is only every used exactly by this package.
* **don't** create an explicit dependency to `ApodiniObserve ` in the Rest and gRPC exporters but use `canImport(ApodiniObserve)` in those exporters to add the `LoggingMetadataInformation` to the InformationSet.

In the example of gRPC one could rewrite the `getLoggingMetadataInformation` function like the following: 

```swift
 static func getLoggingMetadataInformation(_ message: GRPCMessage) -> [AnyInformation] {
     #if canImport(ApodiniObserve) 
         [
             LoggingMetadataInformation(key: .init("data"), metadataValue: message.data.count <= 32_768 ? .string(message.data.base64EncodedString()) : .string("\(message.data.base64EncodedString().prefix(32_715))... (Further bytes omitted since data too large!)")),
             LoggingMetadataInformation(key: .init("dataLength"), metadataValue: .string(message.length.description)),
             LoggingMetadataInformation(key: .init("compression"), metadataValue: .string(message.compressed.description)),
             LoggingMetadataInformation(key: .init("didCollectAllFragments"), metadataValue: .string(message.didCollectAllFragments.description))
         ]
     #else
         []
     #endif
 }
```

At least this is how I expect `canImport` to work, right? (help @theMomax).

If this shouldn't be possible, I would advocate for creating a separate `ApodiniLoggingSupport` package (which could potentially be reused by other Logging facility than ApodiniObserve) which just contains the Information definitions.

bauer-andreas (commented):
Great work 🚀

I made some infrastructural style comments which probably need some further discussion, but apart from that, I really think this is a great addition to the toolset of Apodini 👍

Sources/ApodiniObserve/ConfiguredLogger.swift
theMomax: Good point! I think there are two possible strategies:
* We make `Optional` `Traversable` (in SwiftUI it is). This way a standard component that wants to support a logger could just declare it optional and take an instance in its initializer or use nil as a default. This is currently also possible with `Properties`, but way more ugly. With this approach the service developer is in full control and everything is explicit.
* We promote the path you chose for Authorization by creating such a `_myEnvironment.isPresent`. This would result in more automatism. This could also result in misleading/unexpected scenarios. An additional disadvantage here is that we promote a culture were libraries just setup a default for everything, even though that might never be used. This in turn could lead to bad startup-times in the long term.

I'd suggest to go for the first option, even though that might result in a couple of more lines to be changed when e.g. adding a logger. However, I think we should discuss that in a bigger group! The good thing is neither of the options are difficult to implement.

Sources/ApodiniExtension/HTTPInformation+LoggingMetadata.swift
theMomax: I also thought about such approach @Supereg and I think that should work as you described. However, I think for the `Information` API to be successful in the long run we have to have common "weird" interfaces that new information can conform to and automatically gain compatibility with most features.

If I were to implement a new `Information`, I'd much prefer a "Just conform to either `StringKeyedEncodableInformation` or <insert-other-future-for-example-integer-based-protocol> and you should be fine" strategy over a "Here is a list of 10 common use-cases you basically want to support with your `Information` and they all define custom protocols you have to conform to, so have fun implementing all of them".

bauer-andreas: > If I were to implement a new Information, I'd much prefer a "Just conform to either StringKeyedEncodableInformation or and you should be fine" strategy over a "Here is a list of 10 common use-cases you basically want to support with your Information and they all define custom protocols you have to conform to, so have fun implementing all of them".

I agree with you in the general case. I think I proposed a similar approach in https://github.com/Apodini/Apodini/pull/330#discussion_r685154826 building on top of the InformationClass feature.
In any case, such a feature should be built around the `InformationClass` feature, providing some default and generic InformationClasses in the main Apodini package.

But I don't think this applies to this case here. `LoggingMetadataInformation` is a specific Information with a specific use case which is not just a string key-value pair but has dedicated **semantics**. While it might conform to a `StringKeyedCodableInformationClass` or to a more specific `StringKeyedStringInformationClass` the information itself can't be a generic `TextInformation`. It transports logging information, and I my opinion this should be made explicit.

In every case, the Information instance should always be specific to the use case. However `InformationClasses` can be used to facilitate reuse and genericness.
E.g. the `AnyHTTPInformation` could also just be a `TextInformation` as its simply a string key-value-pair. But a `AnyHTTPInformation` carries important context about the origin and use of that information. Its pretty much the same think, in my view.

Whats your opinion on that?

Sources/ApodiniObserve/ConfiguredLogger.swift
bauer-andreas: > This in turn could lead to bad startup-times in the long term.

I don't quite understand that argument. But I would probably in any case advocate for option 1. Seems more straight forward and natural to me. I think this would also work quite well for the `Authorized` property and the `@ConfiguredLogger` propertyWrapper. Both of those can declare their `@Environment` optional and provide according failure logic in those events.

I don't want to block Philipps work though. Should the result of those discussions be implemented in a separate PR? 

Whats your take on my second point, about having some sort of default logger in the case where no Logger configuration is present. Does that even make sense? I mean the propertyWrapper is called `@ConfiguredLogger` and not just `@Logger`. Would it maybe make sense to create a `@Logger` which falls back to the default Application logger if nothing is configured? Maybe I'm missing some of the vision behind the ApodiniObserve package?

Sources/ApodiniExtension/HTTPInformation+LoggingMetadata.swift
theMomax: I think I'm not too far from your position on this. The concern I have with use-case specific `Information` or `InformationClass`es is that the universe of use-cases has infinite dimensions. As far as I understood your concern with semantics is that e.g. a password HTTP header (which would conform to `StringKeyedCodableInformationClass`) should not be logged. And I totally agree. However, once we make the `InterfaceExporter` responsible for this differentiation, the number of decisions and implementations that have to be made explodes. And I fear that this would end up with most exporter - use case pairs to just have no integration whatsoever.

Maybe we should pursue an attribute-based approach. I.e. the `StringKeyedCodableInformationClass` has requires a set of attributes to be defined by the exporter. The attributes that exist are defined in ApodiniExtension and use either neutral or the exporters' language. E.g. instead of having attributes "suitableForLogging", "requiredForCacheEquality" and "requiredForAuthentication" , we'd have "containsCredentials". What do you think @Supereg ?

Sources/ApodiniObserve/ConfiguredLogger.swift
theMomax: > I don't want to block Philipps work though. Should the result of those discussions be implemented in a separate PR?

Yes, definitely!

> Whats your take on my second point, about having some sort of default logger in the case where no Logger configuration is present. Does that even make sense? I mean the propertyWrapper is called @ConfiguredLogger and not just @Logger. Would it maybe make sense to create a @Logger which falls back to the default Application logger if nothing is configured? Maybe I'm missing some of the vision behind the ApodiniObserve package?

I think the `Configured` prefix is just to avoid naming collision with the normal `Logger` type. (Maybe `ApodiniLogger` or `AutoLogger` would be a better fit, though.) I don't think a default logger configuration makes sense. With the solution to the first problem (making `Optional` `Traversable`), we don't need a default logger, because that enables standard components to just not use a logger if not explicitly enabled by the service developer. We could add a default logger based on a `canImport()`, but I don't really see the point in adding such magic, as the consequence of not having a default logger is basically the following:
1. Service Developer forgets to add a logger configuration but uses `@ConfiguredLogger` somewhere in the code.
2. They get a fatal error while testing their change
3. They do a quick online search
4. They add the logger configuration item and gained a better understanding of the framework from the beginning

Idk...I just dislike frameworks that do everything automatically because I have no clue what is going on until I need to configure something. And when I need to do so, I first have to search for where I can even add the configuration, because the basic configuration wasn't part of the getting started guide - because the default configuration is just added somewhere behind the curtain if not explicitly stated otherwise.

Sources/ApodiniExtension/HTTPInformation+LoggingMetadata.swift
bauer-andreas: I think we are pretty much on the same page, but talk past each other.

Maybe to summarize my views.

* Introduce a `StringKeyedCodableInformationClass` or a `StringKeyedStringInformationClass` protocol (maybe the `StringKeyedStringInformationClass` could even just be a more specializes version of the codable one using protocol inheritance). The only difference to your original idea is that it builds upon the `InformationClass` protocol/feature, giving the whole thing a name.
The `LoggingMetadataInformation` would be part of the `StringKeyedCodableInformationClass`.
The existing `AnyHTTPInformation` would also be part of the `StringKeyedStringInformationClass` (again, the `HTTPHeaderInformationClass` protocol might just inherit from the `StringKeyedStringInformationClass`)!
* I'm opposed to the idea of introducing a generic `TextInformation` (from your original comment). What information does it transport? Where should the information be used? I think any `Information` **instance** should always be use-case-specific (e.g. `LoggingMetadataInformation` or `AnyHTTPInformation`). We should solely rely on `InformationClass`es for the generalization (by providing some standard ones in the Apodini[Extension] target, like the ones from the previous point). That was my main point from the previous comment.
* I like the idea of marking certain Information to contain sensitive data. I haven't thought about that yet. But particularly a HTTPInformation might transport sensitive information (e.g. the Authorization header). Adding an optional property to the `InformationClass` protocol probably makes the most sense(?)  
Do we have other areas where sensitive data might accidentally be logged?
* `LoggingMetadataInformation` should either be moved into the `ApodiniObserve` target and use `canImport(...)` in the exporters or move it into a separate target which can be imported from the exporters (resembles https://github.com/Apodini/Apodini/pull/330#discussion_r685161204). This again resembles my point of view, that a `Information` **instance** is always create with a predefined context. Therefore, creation of such instances can't be done in a generic way. Only querying can be (InformationClasses).

Sources/ApodiniObserve/ConfiguredLogger.swift
bauer-andreas: I agree with your points made. Specifically summarized with:

> With the solution to the first problem (making Optional Traversable), we don't need a default logger, because that enables standard components to just not use a logger if not explicitly enabled by the service developer.

Maybe as a note to @philippzagar. Should we incorporate that ability to optionally call I logger, I would love to see the logging framework be incorporated into ApodiniAuthorization. Specifically in the `Authenticator` and the `AuthorizationRequirementsChecker`.
Those two handlers log the result of the AuthorizationRequirements evaluation wich might be important for debugging reasons, to be able to track why a given request was rejected.

E.g. logging statements in 
https://github.com/Apodini/Apodini/blob/251bbba60b1023443b6973d80ba42c4155fe143a/Sources/ApodiniAuthorization/Delegates/Authenticator.swift#L120

or  

https://github.com/Apodini/Apodini/blob/251bbba60b1023443b6973d80ba42c4155fe143a/Sources/ApodiniAuthorization/Delegates/AuthorizationRequirementsChecker.swift#L59

Sources/ApodiniExtension/HTTPInformation+LoggingMetadata.swift
theMomax: I think the problem I see is that the logger might want to log things that are e.g. normal HTTP headers. They are not explicitly there to be logged, but could be used in a different context too. With your setup, I guess either the logger just queries for LoggingMetadataInformation, or it queries for the broader StringKeyedCodableInformationClass and gets both types (logging and HTTP headers)?

Sources/Apodini/Semantic Model Builder/InterfaceExporter/InterfaceExporter.swift
philippzagar: Yep I completely agree with you, I think your proposed name `ExporterTypeLoggerMetadata ` would fit pretty well

Sources/ApodiniObserve/LoggerExporter.swift
philippzagar: Well, for now, they are not really needed. But all `Exporter`s are structured like that (application, exporterConfiguration) (maybe it would even make sense to include that in the `InterfaceExporter` protocol sometime), therefore I would leave the code in there.

philippzagar: And, tbh, I just forgot to delete the comment :D

bauer-andreas: I'm totally fine with leaving them in. The comment just seemed like it was forgotten 😁

Sources/ApodiniExtension/HTTPInformation+LoggingMetadata.swift
bauer-andreas: > I think the problem I see is that the logger might want to log things that are e.g. normal HTTP headers. They are not explicitly there to be logged, but could be used in a different context too. With your setup, I guess either the logger just queries for LoggingMetadataInformation, or it queries for the broader StringKeyedCodableInformationClass and gets both types (logging and HTTP headers)?

I'm not sure what your point is exactly? With my setup you can query for any `StringKeyedCodableInformationClass`, you can query for only the `HTTPHeaderInformationClass`, there should maybe be also a `LoggingMetadataInformationClass` (or just query for `LoggingMetadataInformation`), and of course you can do any sort of filtering by combining those: querying for `StringKeyedCodableInformationClass` but excluding `HTTPHeaderInformationClass` to get any key-value style Information instances which are not HTTP headers.

theMomax: Ok, if that is possible, that is probably the best solution we have right now 👍🏻. You're the specialist on that topic😁

Sources/ApodiniObserve/ConfiguredLogger.swift
philippzagar: I absolutely agree with you, that was just meant to be a temporary solution.

I could use something like `#if canImport(ApodiniWebSocket)`, but the `WebSocketInterfaceExporter` type isn't public anymore, therefore I would have to use the public `WebSocket`struct and attach `InterfaceExporter` to it, also not very pretty, but at least a little better. 
I also tried to just save the type of the `Exporter` to the the `WebSocket.ExporterConfiguration` (as a static property), but somehow this results in quite weird and ever-changing errors when accessing the property from `ApodiniObserve` (see below)

```swift
#if canImport(ApodiniWebSocket)
            if exporterTypeMetadata.exporterType == ApodiniWebSocket.WebSocket.ExporterConfiguration.exporterType {
                // Write connection state
                builtLogger?[metadataKey: "connectionState"] = .string(connection.state.rawValue)
                
                // Write request metadata
                builtLogger?[metadataKey: "request"] = .dictionary(self.getRequestMetadata(from: connection.request))
            }
#endif
```
Do you have any other suggestions for me on how to easily achieve smt. like that? :)

Sources/ApodiniObserve/ConfiguredLogger.swift
philippzagar: Yep I can do that, but this property cannot be `nil` anyways.

Sources/Apodini/Semantic Model Builder/Model/EndpointParameter.swift
philippzagar: Yep exactly. I took a quick look into that too but then decided to leave it be.

philippzagar: Of course i could just go by the available `PropertyOptionKey`s, so `.http`, `.mutability` etc. Not very future-proof, but yeah. Would that be acceptable to you?

bauer-andreas: I'm honestly not sure what's best. Also don't have a strong opinion on that.
Maybe just concatenate the string version of all the option values, hoping that the values themselves are self explanatory to indicate the property key?

Not sure if it makes sense to blindly add options to the description at all? Maybe for a debugDescription? But the `description` is currently used in OpenAPI as a default (currently not overwritteable). Are the stringified options relevant at all?
I'm really not sure, just wanted to start the discussion, as you seemingly made some thoughts around that topic. But I'm also fine to leave it as you realized it.

Sources/ApodiniObserve/ConfiguredLogger.swift
bauer-andreas: Hm I see. I'm okay with the current approach if it stays temporary.

What I had in mind was to think the other way round. Providing a generalized way of doing such "updates" to the logger. I could see two possible approaches:
* ApodiniObserve provides some sort of hooks which are executed on such occasions. This could be realized with some sort of global storage maybe. The WebSocket Exporter would register such hooks with ApodiniObserve and those would then be executed on the right place to update the Loggers metadata. Thinking that thought to the end, it might not be the best solution to this exact problem, as it might actually be a problem for the WebSocket exporter to compute the updated values.
* Thinking about it. We really only need to update the `connectionState` and `request` metadataKeys. Why? Because the WebSocket exporter uses a `bidirectionalStream` CommunicationalPattern when exporting its endpoints. And its not only the WebSocket Exporter, I assume the update of those logger metadata entires would need to be also done for the plain HTTP exporter if it uses the `bidirectionalStream` CommunicationalPattern (The same thing probably also applies to the `clientSideStream` pattern 🤔). @theMomax can we - specifically the `ConfiguredLogger` property - detect which CommunicationalPattern was used by the Exporter to export the currently executed Handler? We have the `CommunicationalPattern` Metadata, but this only represents a request to the exporter which pattern should be used and doesn't necessarily result in that pattern being exported, right?  
Thinking about it, we really only want to update those two keys if they changed right? For the `connectionState` it is probably relative simple, as overwriting a single value on each logger access isn't probably that huge of an efficiency problem. But can we efficiently detect if the request changed, to not compute the whole metadata value on every logger access?

Sources/ApodiniObserve/ConfiguredLogger.swift
```diff
@@ -0,0 +1,232 @@
...
+                #endif
+            }
+        } else {
+            // Not pretty, but otherwise ApodiniObserve would need to depend on ApodiniWebsocket
+            if String(describing: exporterTypeMetadata.exporterType) == "WebSocketInterfaceExporter" {
+                // Write connection state
+                builtLogger?[metadataKey: "connectionState"] = .string(connection.state.rawValue)
+                
+                // Write request metadata
+                builtLogger?[metadataKey: "request"] = .dictionary(self.getRequestMetadata(from: connection.request))
```
bauer-andreas: Just noted while looking over it.
Is it correct that in successive updates to the `request` metadataKey, the information is left out?
Above it is initialized like the following:
```swift
builtLogger?[metadataKey: "request"] = .dictionary(self.getRequestMetadata(from: request)
                                                                 .merging(self.getRawRequestMetadata(from: connection.information)) { _, new in new })
```

Sources/ApodiniObserve/ConfiguredLogger.swift
philippzagar: Mh I just checked a possible way to get access to the `CommunicationalPattern` within the `ConfiguredLogger` (or soon the `ApodiniLogger`, great suggestion!)
Couldn't we just use (in case of the `HTTP` exporter) the `VaporEndpointKnowledge` or `ProtocolAgnosticEndpointKnowledge` `KnowledgeSource`s in a similar way like in the `LoggerExporter`? That should work, but I have to take a closer look into it. However, that would require a dependency to `ApodiniHTTP`, something that you wanted to avoid (maybe try a `#canImport` again?).
Great catch nonetheless, I hadn't even thought about the (new) `HTTP` exporter and the different CommunicationalPatterns.

As for the `WebSocketInterfaceExporter`, no idea how to approach that (@theMomax ?)

philippzagar: Also, i think there's not really an efficient way to detect a change to the request (eg. we're aggregating some logging metadata data from Vapor, I don't think there's even a possibility to detect changes without evaluation then, right?)

theMomax: Just inspect the `blackboard[CommunicationalPattern.self]`. Don't differentiate based on the exporter. If it is `requestResponse` or `serviceSideStreaming` the connection should always be close and you should only ever have one request/evaluation.

Sources/Apodini/Semantic Model Builder/Model/EndpointParameter.swift
philippzagar: I'm not really a friend of just concatenate the string version of all the option values, seems just off and not really helpful to me, but if you think that's a good idea, I can definitely do that.

Another thought: Add a function (more swifty way: computed property) to `PropertyOptionKey` that returns the runtime type of `Option: PropertyOption` via `String(describing: ...)`. Would that be acceptable to you? (However I somehow guess we don't want to implement "the easy way out" :D)

And yep, probably makes more sense to do that in a debugDescription.

Sources/ApodiniObserve/ConfiguredLogger.swift
philippzagar: Ah yeah, looking into the `ProtocolAgnosticEndpointKnowledge`more closely, I came to that conclusion too. Thanks Max!

Sources/ApodiniObserve/ConfiguredLogger.swift
philippzagar: Yep you're right, i somehow missed that while CMD+C and CMD+V :D
We need to reevaluate the RawRequestMetadata as well since eg. the raw HTTPbody could change

Sources/Apodini/Semantic Model Builder/Model/EndpointParameter.swift
bauer-andreas: > Another thought: Add a function (more swifty way: computed property) to PropertyOptionKey that returns the runtime type of Option: PropertyOption via String(describing: ...). Would that be acceptable to you? (However I somehow guess we don't want to implement "the easy way out" :D)


Does that work? Like is there a place were we can just grab the swift string representation of _something_ and get the expected representation? Cause the type of the http option (function!) is just `AnyPropertyOption<ParameterOptionNameSpace>`. And the type of the http option key is `PropertyOptionKey<ParameterOptionNameSpace, HTTPParameterMode>, so not sure if they deliver any meaningful additional information?

After all, I'm not 100% sure what you mean exactly, maybe a small example? I mean if it works I'm all in 👍 


> I'm not really a friend of just concatenate the string version of all the option values, seems just off and not really helpful to me, but if you think that's a good idea, I can definitely do that.

I'm not really sure if that is a good idea either 😅 it's just one way I could think of to bring in all the options without the need to hardcode them.

Sources/ApodiniObserve/ConfiguredLogger.swift
bauer-andreas: So we found the perfect solution? Do I understand that correctly?

philippzagar: Yep, Max gave me the final hint

Sources/Apodini/Semantic Model Builder/Model/EndpointParameter.swift
philippzagar: I would have done something like that:
```swift
public class AnyPropertyOptionKey {
    var propertyOptionType: String {
        fatalError("not implemented")
    }
}

public class PropertyOptionKey<PropertyNameSpace, Option: PropertyOption>: AnyPropertyOptionKey {
    override var propertyOptionType: String {
        String(describing: Option.self)
    }
}
```

In case of `.http`, this results in the output "HTTPParameterMode", which is basically what we wanted, right?

philippzagar: After making `PropertyOptionSet` iterable, we can do smt. like that: 
```swift
self.options.forEach({ key, value in
    print("\(key.propertyOptionType) = \(value)")
})
```

Resulting in:
"HTTPParameterMode = body
HTTPParameterMode = body"

for 

```swift
@Parameter(.http(.body))
var name: String? = "World"

@Parameter(.http(.body))
var newMove: CreateMoveMediator
```

bauer-andreas: > After making `PropertyOptionSet` iterable, we can do smt. like that:
> 
> ```swift
> self.options.forEach({ key, value in
>     print("\(key.propertyOptionType) = \(value)")
> })
> ```
> 
> Resulting in:
> "HTTPParameterMode = body
> HTTPParameterMode = body"
> 
> for
> 
> ```swift
> @Parameter(.http(.body))
> var name: String? = "World"
> 
> @Parameter(.http(.body))
> var newMove: CreateMoveMediator
> ```

I like that 👍 🚀

Sources/ApodiniObserve/ConfiguredLogger.swift
philippzagar: So, I'll revolve this conversation for now and we will pick it up again sometime after this PR is merged and both of you have submitted your thesis :)

Sources/ApodiniExtension/HTTPInformation+LoggingMetadata.swift
philippzagar: @Supereg Please take a look at how I implemented it.
I had to introduce a separate target `ApodiniLoggingSupport` as discussed, since I couldn't get `canImport(...)` to work (no clue why, I got ever-changing error messages). The target contains all "log helpers", which would allow for easy extensibility.
 I'm not quite sure how to express the inheritance from `StringKeyedCodableInformationClass ` to `StringKeyedStringInformationClass` without associated types etc. Could you give me some pointers here?
I also made the `Authorization` header type `sensitive`, but the default of the `InformationClass` is false (since probably most information headers don't contain sensitive data).

philippzagar:
> It seems that an ApodiniDeploy test case fails in a certain configuration:
> 
> ```
> error	09-Aug-2021 01:08:14	/opt/bamboo-agent-home/xml-data/build-dir/APODINI-APODINI65-MIRBT/Tests/ApodiniDeployTests/LambdaDeploymentProviderTests.swift:208: error: -[ApodiniDeployTests.LambdaDeploymentProviderTests testLambdaDeploymentProvider] : XCTAssertEqual failed: ("200") is not equal to ("500")
> error	09-Aug-2021 01:08:14	/opt/bamboo-agent-home/xml-data/build-dir/APODINI-APODINI65-MIRBT/Tests/ApodiniDeployTests/LambdaDeploymentProviderTests.swift:315: error: -[ApodiniDeployTests.LambdaDeploymentProviderTests testLambdaDeploymentProvider] : failed - request to '/v1/aws_rand?lowerBound=7&upperBound=520' failed.: The data couldn’t be read because it is missing.
> error	09-Aug-2021 01:08:21	/opt/bamboo-agent-home/xml-data/build-dir/APODINI-APODINI65-MIRBT/Tests/ApodiniDeployTests/LambdaDeploymentProviderTests.swift:214: error: -[ApodiniDeployTests.LambdaDeploymentProviderTests testLambdaDeploymentProvider] : Asynchronous wait failed: Exceeded timeout of 10 seconds, with unfulfilled expectations: "Test /v1/aws_rand".
> error	09-Aug-2021 01:08:25	Test Case '-[ApodiniDeployTests.LambdaDeploymentProviderTests testLambdaDeploymentProvider]' failed (307.503 seconds).
> error	09-Aug-2021 01:08:25	Test Suite 'LambdaDeploymentProviderTests' failed at 2021-08-09 01:08:25.185.
> error	09-Aug-2021 01:08:25		 Executed 1 test, with 3 failures (0 unexpected) in 307.503 (307.503) seconds
> ```
> 
> I'm not quite sure if the PR caused the problem since I didn't change anything in `ApodiniDeploy`. I'll look into that.

I talked to Paul regarding this issue, it's a known problem with the available runners. Should be fixed soon.

PSchmiedmayer:
@philippzagar The CI setup is fixed with the latest commits on develop and some updates to the CI Macs to Xcode 13 beta 5. Merging develop into this branch should ideally resolve all the [issues you mentioned in your comment](https://github.com/Apodini/Apodini/pull/330#issuecomment-897645377).

Sources/Apodini/Properties/Options/PropertyOptionKey.swift
```diff
@@ -27,11 +27,20 @@ public class AnyPropertyOptionKey: Equatable, Hashable {
...
+        fatalError("AnyPropertyOptionKey.propertyOptionType not implemented!")
+    }
 }
 
 
 /// A `PropertyOptionKey` can be associated with a `PropertyNameSpace` and and store an `Option` that is associated with the `PropertyOptionKey` within the `PropertyNameSpace`.
 public class PropertyOptionKey<PropertyNameSpace, Option: PropertyOption>: AnyPropertyOptionKey {
+    override var propertyOptionType: String {
+        let test = String(describing: Option.self)
+        return test
```
bauer-andreas: Can be inlined(?) and test seems like a leftover property name?

Sources/ApodiniExtension/HTTPInformation+LoggingMetadata.swift
```diff
@@ -10,6 +10,7 @@ import Foundation
 import Apodini
 import Logging
 
+/*
```
bauer-andreas: Forgotten source file.

Sources/ApodiniHTTPProtocol/Information/Authorization.swift
```diff
@@ -17,10 +17,13 @@ public struct Authorization: HTTPInformation {
...
     public let value: Value
 
-
     public var rawValue: String {
         value.rawValue
     }
+    
+    public var sensitive: Bool {
+        true
+    }
```
bauer-andreas: If I'm not wrong, I think this doesn't work as expected (and it might be worthy to add some test vectors checking that Authorization Headers aren't included in logging messages?).

Note the difference between `Information` and `InformationInstantiatable`.
`AnyHTTPInformation` is the type conforming to `Information`. Those instances are created for every HTTP header in a request. `AnyHTTPInformation` is also the type which conforms to the `InformationClass` protocol (transitively via `HTTPHeaderInformationClass`).
`Authorization`, on the other hand, conforms to `InformationInstantiatable` (transitively via `HTTPInformation`). `InformationInstantiatable` are used to be **on-the-fly** created from the according `Information` type to receive a typed version of that. `InformationInstantiatable` are solely used by end-users (we also cached those **if available** but only to perform reduction logics e.g. for the WWWAuthenticate).

But only having a `Information` we don't necessarily know an according `InformationInstantiatable`. The other way always holds.

So to implement the `sensitive` property you first of all probably need to declare it on `AnyHTTPInformation` and probably hardcode a list of known sensitive HTTP Headers (you can use e.g. `Authorization.header` for less magic constants).

Sources/ApodiniObserve/Logs/ApodiniLogger.swift
```diff
@@ -122,17 +115,26 @@ public struct ConfiguredLogger: DynamicProperty {
                 #endif
             }
         } else {
-            // Not pretty, but otherwise ApodiniObserve would need to depend on ApodiniWebsocket
-            if String(describing: exporterTypeMetadata.exporterType) == "WebSocketInterfaceExporter" {
-                // Write connection state
-                builtLogger?[metadataKey: "connectionState"] = .string(connection.state.rawValue)
+            // Connection stays open since these communicational patterns allow for any amount of client messages
+            switch self.blackboardMetadata.communicationalPattern {
```
bauer-andreas, line 119: Turned out great 🚀

bauer-andreas (approved):
Looks great 👍 🔥 

I like how the realization of those pre-defined InformationClasses turned out.

Just had one bigger comment in regard to the `sensitive` property, which to my eye will not work as expected.
Will nonetheless approve the PR, though with the request to fix that ☺️

Sources/ApodiniExtension/HTTPInformation+LoggingMetadata.swift
philippzagar: Thanks, completely missed that

Sources/ApodiniHTTPProtocol/Information/Authorization.swift
philippzagar: Yep you're absolutely right, that didn't work out as expected. As mentioned, I will add end-to-end tests in a separate PR :)

If we declare the `sensitive` property on `AnyHTTPInformation`, `ApodiniObserve` would have to depend on `ApodiniHTTPProtocol` again, a path we chose not to follow (since we would need to query after it in `ApodiniObserve`). 

Or am I missing something?

bauer-andreas: > Yep you're absolutely right, that didn't work out as expected. As mentioned, I will add end-to-end tests in a separate PR :)

I think it would be enough if you add it onto the exiting InformationTests to add some asserts here and there for the sensitive property(?)

> If we declare the sensitive property on AnyHTTPInformation, ApodiniObserve would have to depend on ApodiniHTTPProtocol again, a path we chose not to follow (since we would need to query after it in ApodiniObserve).

Nah, the `sensitive` property is declared in the `InformationClass` protocol right? And `ApodiniObserve` accesses the `sensitive` property over the `InformationClass` protocol?

As described above, `AnyHTTPInformation` is the type which conforms to `InformationClass`, so adding the `sensitive` property onto that is the way to go(?)

like
```swift
// adding to AnyHTTPInformation
var sensitive: Bool {
  header == Authorization.header
}
```

philippzagar: Yep, that makes a lot of sense.
My initial approach was to remove the `sensitive` property from `InformationClass` completely, but if we leave it in there, my initial problem goes away :)

philippzagar:
I didn't write test cases for this PR since the simultaneous open PR #334 changed up some things on how the metadata of the logger is aggregated. Therefore, it wouldn't make much sense to write instantly outdated test cases.
Therefore, @PSchmiedmayer and I agreed to merge this PR in its current form, with the promise that #334 will bring back the code coverage to its "normal" percentage.

PR #331, opened 2021-08-09
Remove unsafe Swift flags

PSchmiedmayer:
# Remove unsafe Swift flags

Apodini currently uses unsafe Swift flags in the Package.swift file which prevents us to tag a release.

## :bulb: Proposed solution
Removes the unsafe Swift flags.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -119,8 +119,7 @@ let package = Package(
             ],
             exclude: [
                 "Components/ComponentBuilder.swift.gyb"
-            ],
-            swiftSettings: [.unsafeFlags(["-Xfrontend", "-enable-experimental-concurrency"], nil)]
+            ]
         ),
         
         .target(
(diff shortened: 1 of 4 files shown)


Package.swift
```diff
@@ -119,8 +119,7 @@ let package = Package(
             ],
             exclude: [
                 "Components/ComponentBuilder.swift.gyb"
-            ],
-            swiftSettings: [.unsafeFlags(["-Xfrontend", "-enable-experimental-concurrency"], nil)]
```
bauer-andreas, line 123: Are those still required to be passed manually now?

bauer-andreas (approved):

Package.swift
PSchmiedmayer: No, you don't have to manually pass in the flags to the Swift compiler. All the functionality hidden behind the flags is now be part of the toolchain without the need for any feature flags.

PSchmiedmayer:
@Supereg It took some time to update the CI to the latest Xcode version and macOS version and all tests except the `ApodiniNegativeCompileTests` seem to pass at the moment. This seems to be related to a new Version of the Swift compiler. which results in different error messages. What would be the best way to approach this? Seems like the "failed to produce diagnostic for expression; please file a bug report" also appear here as well.

PR #332, opened 2021-08-11
Make Authorised a property wrapper

bauer-andreas:
# Make Authorised a property wrapper

Explained in https://github.com/Apodini/ApodiniXpenseExample/pull/1#discussion_r686726564 in great detail, a simple `@Authorized var user: ExampleUser` approach cannot be used for the Authorized property, as the [[SE-0310]](https://github.com/kavon/swift-evolution/blob/main/proposals/0310-effectful-readonly-properties.md#extensions-considered) proposal doesn't include support for the `wrappedValue` of property wrappers.
Therefore a plain property based solution was originally chosen in #314.
@PSchmiedmayer noted in the above linked discussion that nonetheless a property wrapper based solution would be preferable from a consistency standpoint.

## :bulb: Proposed solution

This PR refactors the `Authorized` property to an property wrapper, which exposes a wrapper around the Authencitatable instance in its `wrappedValue`.

Should Swift ever get support for throwing `wrappedValues` one would change the implementation, such that `wrappedValue` returns the user instance directly. Attributes like `isAuthorized` could be accessed via the projected value of the property wrapper.

## :gear: Release Notes 

Instead of writing ...
```swift
var authorizedUser = Authorized<ExampleUser>()
// ...

// `isAuthorized` access
authorizedUser.isAuthorized

// user instance access
try authorizedUser()
```

... the declaration was changed to be based in a property wrapper with a `wrappedValue` exposing the same interface as the original Authorized property:

```swift
@Authorized(ExampleUser.self) var authorizedUser
// ...

// `isAuthorized` access
authorizedUser.isAuthorized

// user instance access
try authorizedUser()
```

### Related PRs
- #314 
- https://github.com/Apodini/ApodiniXpenseExample/pull/1

### Testing
No further testing was added.

### Reviewer Nudging
I hope this will be a quick and easy review.

### DIFF
--- a/Sources/ApodiniAuthorization/AuthorizedProperty.swift
+++ b/Sources/ApodiniAuthorization/AuthorizedProperty.swift
@@ -8,16 +8,16 @@
 
 import Apodini
 
-/// The ``Authorized`` `DynamicProperty` can be used to access the ``Authenticatable``
-/// instance created through a authorization Metadata.
+/// The ``Authorized`` property wrapper can be used to access the ``Authenticatable``
+/// instance created through an authorization Metadata.
 ///
 /// Given the example of an `ExampleUser` ``Authenticatable``, the following code may be used to
 /// access the authenticated and authorized instance.
 /// ```swift
 /// struct ExampleHandler: Handler {
-///     var authorizedUser = Authorized<ExampleUser>()
+///     @Authorized(ExampleUser.self) var authorizedUser
 ///
-///     func handle() -> String {
+///     func handle() throws -> String {
 ///         // you might want to use `authorizedUser.isAuthorized` if using optional authorization
 ///
 ///         let user = try authorizedUser()
@@ -26,7 +26,15 @@ import Apodini
 ///     }
 /// }
 /// ```
+@propertyWrapper
 public struct Authorized<Element: Authenticatable>: DynamicProperty {
+    public var wrappedValue = AuthorizedAuthenticatable<Element>()
+
+    public init(_: Element.Type = Element.self) {}
+}
+
+/// The wrapped value of the ``Authorized`` property wrapper. See according docs.
+public struct AuthorizedAuthenticatable<Element: Authenticatable>: DynamicProperty {
     @Environment(\.authorizationStateContainer)
     private var stateContainer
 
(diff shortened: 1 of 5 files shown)


Sources/ApodiniAuthorization/AuthorizedProperty.swift
```diff
@@ -8,16 +8,16 @@
...
-/// instance created through a authorization Metadata.
+/// The ``Authorized`` property wrapper can be used to access the ``Authenticatable``
+/// instance created through an authorization Metadata.
 ///
 /// Given the example of an `ExampleUser` ``Authenticatable``, the following code may be used to
 /// access the authenticated and authorized instance.
 /// ```swift
 /// struct ExampleHandler: Handler {
-///     var authorizedUser = Authorized<ExampleUser>()
+///     @Authorized(ExampleUser.self) var authorizedUser
```
PSchmiedmayer, line 18: Maybe even `@Authorized var authorizedUser: ExampleUser` looks more natural? As far as I can see the default value in initializer would enables this?

PSchmiedmayer (approved):
Looks very nice! I only had a comment regarding the example and usage of the property wrapper (not pass the type to the property wrapper but use the type annotation of the wrapped variable). Even though this might be a bit confusing as you expect one type and get an other ... what do you think?

Sources/ApodiniAuthorization/AuthorizedProperty.swift
bauer-andreas: Isn't that the whole problem that `wrappedValue` can't be throwing? Afaik - and maybe its only me not knowing about the internals of a property wrapper well enough - but the type of the property on which the property wrapper is used on must match the type of the `wrappedValue`, right?
As we have the type `AuthorizedAuthenticatable<Element>` of the `wrappedValue` the propertyWrapper forces the type of the property to `AuthorizedAuthenticatable<Element>`.

Like the following two would be equivalent:
```swift
@Authorized(ExampleUser.self) var authorizedUser
@Authorized(ExampleUser.self) var authorizedUser: AuthorizedAuthenticatable<ExampleUser>

// and probably that works also due to the default init param
@Authorized var authorizedUser: AuthorizedAuthenticatable<ExampleUser>
```

But I'm not sure how your approach would work? Can the type of the `init(wrappedValue:)` be different to the type of the `wrappedValue`?

PSchmiedmayer: You are right, this is my mistake. I have missed this here, my bad 🙈

PR #338, opened 2021-08-18
Migrating DeploymentOptions to Metadata System

bauer-andreas:
# Migrating DeploymentOptions to Metadata System

Currently, ApodiniDeploy introduces its own specification language for deployment options.
Further, it also introduced its own Options system as, previous to #326, it required to be able to export all deployment options.

## :bulb: Proposed solution

This PR proposed the following two migrations:
* Use the Metadata system #280 for specification of deployment options
* As exportability is not required, the separate Options API is removed and we use the standard Apodini Options API as a backend.

## :gear: Release Notes 

Deployment Options were migrated to the Metadata system.

### HandlerWithDeploymentOptions

The `HandlerWithDeploymentOptions` protocol was removed and deployment options are now declared inside the Metadata Blocks of components.

You migrate from ...
```swift
struct SomeHandler: Handler, HandlerWithDeploymentOptions {
  // ...
  static var deploymentOptions: [AnyDeploymentOption] {
    [
      AnyDeploymentOption.memory(.mb(150)).when(\Self.handlerId == .main),
      AnyDeploymentOption.memory(.mb(180)).when(\Self.handlerId == .other),
      AnyDeploymentOption.timeout(.seconds(12))
    ]
  }
}
```
... to ...
```swift 
struct SomeHandler: Handler, HandlerWithDeploymentOptions {
  // ...
  var metadata: Metadata {
    if handlerId == .main {
      Memory(.mb(150))
    } else if handlerId == .other {
      Memory(.mb(180))
    }

    Timeout(.seconds(12))
  }
}
```

### Deployment Options Modifier

The `deploymentOptions()` Modifier was removed. Use the `metadata` modifier instead and specify Metadata as shown above.

### Related PRs
- Deployment Provider refactoring #326 
- Metadata DSL introduction #280 

### Testing
No further tests were added.

### Reviewer Nudging
--


bauer-andreas:
@PSchmiedmayer codecov coverage fails for this PR. As this is partly due to the current ApodiniDeploy coverage I would propose to use admin powers for merging once approved.

PSchmiedmayer:
@Supereg Sure, we had a similar situation with the last Apodini Deploy PR #326 👍

[Looking at the diff in the test coverage](https://app.codecov.io/gh/Apodini/Apodini/compare/338/diff#D1) I think that we can even bring this PR down to a threshold that would be acceptable by CodeCov.
Looks like some of the functions, especially in `Sources/ApodiniUtils/Swift.swift` are no longer needed and are therefore currently not tested and could be removed. This can further increasing the test coverage of the Apodini Package in combination with one or two useful additional unit tests.

Nevertheless it might be that we don't reach that threshold and it this is the case I am happy to merge the PR with my admin privileges 🚀

Sources/ApodiniDeployRuntimeSupport/StructureExporter.swift
```diff
@@ -53,12 +53,33 @@ extension StructureExporter {
...
+        try retrieveDefaultDeployedSystem(endpoints, config: config, app: app, deriveUserData: nil)
+    }
+
+    /// Computes the default deployed system where all endpoints are matched to deployment groups if possible.
+    public func retrieveDefaultDeployedSystem<T: Codable>(
+        _ endpoints: Set<CollectedEndpointInfo>,
+        config: DeploymentConfig,
+        app: Application,
+        deriveUserData: @escaping (CollectedEndpointInfo) -> T
+    ) throws -> DeployedSystem {
```
hendesi, line 72: Maybe I'm missing something, but was exactly is the purpose of this method and in particular the `deriveUserData` parameter? I'd assume it's to manually decide how to handle/ decode the user data in a custom type, similar to what was talked in rocket chat, is this correct?  
But this parameter does only work for whoever how wants to use the default deployed system, or am i getting this wrong? Anybody relying on a custom implementation cannot use this. 
Would it be possible to add this to the `StructureExporter` protocol with a default `nil` implementation similar to `nodeIdProvider`, then everybody implementing this protocol could decide if he wants to use it or not

Sources/ApodiniDeployBuildSupport/CollectedEndpointInfo.swift
```diff
@@ -18,7 +18,7 @@ public enum WellKnownEnvironmentVariables {
     /// This environment variable is only set of the web service is running as part of a managed deployment.
     public static let currentNodeId = "AD_CURRENT_NODE_ID"
     
-    /// Key for an environment variable specifying the execution mode of ApodiniDeploy, ether dump the WebService's model structur or launch the WebService with custom config
+    /// Key for an environment variable specifying the execution mode of ApodiniDeploy, ether dump the WebService's model structure or launch the WebService with custom config
     public static let executionMode = "AD_EXECUTION_MODE"
```
hendesi, line 23: I just noticed that i have missed this in the refactoring of deployment structure. All the static properties that serve as keys/ values for deployment related environment variables shouldn't be needed anymore and can be removed I think. But not sure if this has to be done with this PR.

hendesi (commented):
The changes look really good to me 👍 Makes the hole deployment options adding a lot cleaner imo. I just have one major point that needs some clarification.

Sources/ApodiniDeployBuildSupport/CollectedEndpointInfo.swift
bauer-andreas: Just removing those two properties? I think I can do that 👍

bauer-andreas: Just checked, both of those properties are used within the `AWSIntegration.swift` file(?)

So I would propose, if you still think that those can be removed, that you do that in a separate PR yourself, if that's okay.

Sources/ApodiniDeployRuntimeSupport/StructureExporter.swift
bauer-andreas: The `retrieveDefaultDeployedSystem` of the `StructureExporter` creates `ExportedEndpoint`s from `CollectedEndpointInfo`s.
`CollectedEndpointInfo` contains deployment options, the `ExportedEndpoint` contains a `Data` property which, similar to `DeployedSystemNode`, is en and decoded from a `Codable` type provided by the deployment provider.
This basically realizes the "context" thingy we talked about.
In order, to not be required to write your own `retrieveDeployedSystem` if you want to interpret deployment options and store them in your `userData`, we provide the closure which will then create such a userData instance from the provided `CollectedEndpointInfo`.

bauer-andreas: And yes, good catch. Completely forgot to add it to the protocol itself 👍

bauer-andreas: > Would it be possible to add this to the StructureExporter protocol with a default nil implementation similar to nodeIdProvider, then everybody implementing this protocol could decide if he wants to use it or not

> And yes, good catch. Completely forgot to add it to the protocol itself 👍

Nevermind. This method is not there to be implemented, it is there to be used in your own `retrieveDeploymentStructure`. Similarly, as both `LambdaStructureExporterCommand` and `LocalhostStructureExporterCommand` call the `retrieveDefaultDeployedSystem` without the closure in their implementation.

bauer-andreas: > Anybody relying on a custom implementation cannot use this.

Anybody doing a custom implementation can just instantiate a `ExportedEndpoint` with according userData themselves.

hendesi: Yes, now i get it. Thanks for clarifying 👍

Sources/ApodiniDeployBuildSupport/CollectedEndpointInfo.swift
hendesi: Ah, I see. I guess, i mixed it up with [WellKnownEnvironmentVariables](https://github.com/Apodini/Apodini/blob/a990489b5b392ad34d66a18da14ed3a2335d6c4c/Sources/ApodiniDeployBuildSupport/CollectedEndpointInfo.swift). 

Sure, never mind. I'll just clean them up in a later pr. 👍

hendesi (approved):
LGTM 👍

Issue #329, opened 2021-08-03
[MAJOR FEATURE] Configurable Web Service Structure


theMomax:
**Describe your use case**
We currently have separate REST and HTTP exporters. Both do pretty much the same, except REST incorporates relationship-information.

**Describe the problem**
Both exporters have to be maintained separately. Maintenance would become even more of a problem if we wanted to support relationship-information for further middleware.

**Propose a solution**
The `Blackboard` pattern allows `InterfaceExporter`s to override `KnowledgeSource` instances on the `Blackboard`. While that allows them to manipulate the information stored on the `Blackboard`s, it does not allow them to manipulate the structure of the web service as it is used by the `SemanticModelBuilder`. E.g. an `InterfaceExporter` could delete or add an endpoint to the knowledge base by overriding the `Blackboards` `KnowledgeSource`, however the `SemanticModelBuilder` wouldn't notice and thus not call the `InterfaceExporter.export()` function for this added endpoint. Similarly, `InterfaceExporter`s could override the `EndpointSource` `KnowledgeSource` to replace/alter/wrap the `Handler`. Again, `SemanticModelBuilder` wouldn't notice and call the `InterfaceExporter.export()` with the old `Handler.Type`.

If we were to allow all that, we could build a `Configuration` (that is also an `InterfaceExporter`) that doesn't export any middleware-integration itself, but only extends the web service's structure with relationship-information.

Example:

```swift
var configuration: Configuration {
    HTTP() // results in old HTTP behavior
    WebSocket() // results in old WebSocket behavior
    ApplyRelationshipInformation() // transforms the web service
    HTTP() // results in old REST behavior
    WebSocket() // results in new WebSocket + Relationship behavior
}
```

**Describe alternatives you've considered**
This approach somehow breaks with the concept that `InterfaceExporter`s define their own view on the exact same web service. I also don't know how that could make sense with e.g. GraphQL, where relationship information is such an integral element of the middleware itself. It is an idea that should be discussed thoroughly before taken any further. We could of course continue to focus making maintenance of `InterfaceExporter`s easier by increasing potential for reuse.

theMomax:
[...]

PR #339, opened 2021-08-21
Allow for more detailed OpenAPI web service documentation through WebService Metadata

### LINKED ISSUE #232: [MAJOR FEATURE OpenAPIExporter] Enable developer to provide meta/descriptive information to enhance OAS
**Additional context**
The idea of a "Description/Meta DSL (:crystal_ball:)" has been floating around in our heads for quite some time (but we didn't manage to determine a catchy name yet). There might be use cases for different exporters and @Supereg already translated the idea in a [vague sketch](https://gist.github.com/Supereg/124e6beb555719839dba58e420f74f5d), which is WIP.
Regarding the OpenAPIExporter, I want to capture some of the use cases I could think of, in this issue.

**Describe your use case**
When developers want to use OAS for documenting their RESTful web application and/or generating client side code they might appreciate a descriptive OAS (with some prose where needed). In other words, solely adding paths, endpoints, and types to an OpenAPI document does not exploit the full potential of OAS. In fact, many `description` and `example` fields should help to provide more information on certain sections in the specification. These fields will not only enrich the representation in Swagger-UI by additional hints and texts, but also lead to more specific type and function names (instead of created defaults) as well as in-code documentation within the generated client code as this is nourished by information that I would categorize as meta information.

**Describe the problem**
As of now, I would say the minimum "description support" is implemented via `TagModfier`(#210), `DescriptionModifier` (#117) and `OpenAPIConfiguration` (#165). 

**Propose a solution**
OAS use cases for Meta DSL:
1. [`Info Object`](https://swagger.io/specification/#info-object): literally "_provides metadata about the API_", in terms of `title`, `description`, `termsOfService`, `contact`, `license`, `version` (whereof `title` and `version` are currently covered by `OpenAPIConfiguration`).

Adding more meta information on API level would translate to this OAS document and Swagger-UI representation respectively:
image

2. [`Schema Object`](https://swagger.io/specification/#schema-object): allows the definition of input and output types. One property of the schema object is `example`: "_A free-form property to include an example of an instance for this schema_". Additionally, a `description` field can be added on object and/or property level.

Adding more meta information on type level would translate to this OAS document and Swagger-UI representation respectively:
image

3.  [`Path Item Object`](https://swagger.io/specification/#path-item-object), [`Operation Object`](https://swagger.io/specification/#operation-object), [`Parameter Object`](https://swagger.io/specification/#parameter-object): each providing the possibility of adding `description` (on operation level: this is covered by `DescriptionModifier` applied on `Apodini.Handler`, on parameter level: the declaration is used as description, i.e., `@Parameter var gameId: String`). Additionally, the path item object and operation object can hold a `summary`, applied to all operations in this path or representing a short summary of what the operation does, respectively.

In the screenshot below, only the field "summary" (right beneath "put") was added compared to the current support of the `OpenAPIExporter`. Nevertheless, I could imaging that the parameter descriptions can be made customizable.
image

I think there are many possibilities to enrich the OAS with meta information, which implies that one should prioritize the ones adding the most value. This is especially advisable in the case of `description` fields, which should not lead to a jumble of description level  within the Meta DSL, making it bloated and infeasible.

4. [`Tag Item Object`](https://swagger.io/specification/#tag-object): Currently, developers can "tag" endpoints by using the `TagModifier` on `Apodini.Handler`. This leads to subsections in Swagger-UI (as seen in the screenshot) as well as separate service files being generated instead of one big "defaultService" holding all the different functions.

An enhancement would be, that developers can specify the tags and an according description by the means of the Meta DSL. 
Having a collection tag names and descriptions would allow for a summarizing `tags` section in the OpenAPI document and descriptions on the level of tag-subsections (such as `api-action`, `api-game`, `api-player` in the screenshot example)
image

bauer-andreas:
# Allow for more detailed OpenAPI web service documentation through WebService Metadata

Currently the user has no way of supplying information for the uncovered files in the `Info Object` and the `OpenAPI Object` of the generated OpenAPI specification.

## :bulb: Proposed solution

This PR adds the new **WebService** Metadata to cover the missing fields in the OpenAPI specification.
Further, some minor infrastructure changes were done to access the web service `Context` from the `WebServiceModel`.

## :gear: Release Notes 

The following new **WebService** Metadata is available in the `Apodini` package:
* `Description(:)` (the web service version)
* `Version(prefix:major:minor:patch:)`

The `Version` Metadata replaces the version property on the `WebService`. Please migrate accordingly.

The following new **WebService** Metadata is available in the `ApodiniOpenAPI` package:
* `Contact(name:url:email:vendorExtensions:)`
* `License(name:url:vendorExtensions:)`
* `TermsOfService(url:)`
* `ExternalDocumentation(description:url:vendorExtensions:)`
* `TagDescription(name:description:externalDocs:vendorExtensions:)`
* `TagDescriptions` Restricted Metadata Block to declare `TagDescription`

### Related PRs
- Partly resolves #232 

### Testing
New test was added to cover those new Metadata.

### Reviewer Nudging

Look at the tests to see how the Metadata is used:

https://github.com/Apodini/Apodini/blob/5e92140f1f3abe67ffd6ee4edbc153596b9ea91a/Tests/ApodiniTests/OpenAPITests/OpenAPIWebServiceMetadataTests.swift#L22-L44


bauer-andreas:
The PR is read for review. The Bamboo CI is currently failing as of the AWS Deploy issue described in rocket chat.

Sources/Apodini/Metadata/BuiltinMetadata/DescriptionMetadata.swift
```diff
@@ -69,20 +102,11 @@ public struct ContentDescriptionMetadata: ContentMetadataDefinition {
 }
 
 
-extension Component {
```
theMomax, line 72: Why can't we have a `ComponentDescriptionMetadata` next to `HandlerDescriptionMetadata`? I think it would make sense to allow the description to be declared anywhere. The `Handler` version should override the `Component` version.

Sources/Apodini/WebService.swift
```diff
@@ -128,28 +125,24 @@ extension WebService {
...
         visitor.finishParsing()
     }
     
-    func visit(_ visitor: SyntaxTreeVisitor) {
+    func accept(_ visitor: SyntaxTreeVisitor) {
         metadata.collectMetadata(visitor)
 
-        visitor.addContext(APIVersionContextKey.self, value: version, scope: .environment)
+        let version = visitor.currentNode.peekValue(for: APIVersionContextKey.self)
         visitor.addContext(PathComponentContextKey.self, value: [version], scope: .environment)
```
theMomax, line 142: I don't like how the `peekValue(for:)` so "far away" from the `SyntaxTreeVisitor`.  I could imagine solving this on a `Blackboard` basis in a cleaner way, but that would require some work. I suggest to leave it as is right now, but I'll make a note to #329. We should think about having default `Configuration`s on a `WebService`. This way we could have a `Configuration` that manipulates the paths of all endpoints by prepending the `Version` element. I always thought it should be possible not to include the path should by adding a `Configuration`. However, before we can implement #329 and this use case, I think a more sophisticated for ordering the execution of `Configuration`s should be implemented than just doing so _in order_.

theMomax (approved):
LGTM! I think there was only one real concern, the other is not in the scope of this PR. Good job👍🏻

Sources/Apodini/Metadata/BuiltinMetadata/DescriptionMetadata.swift
bauer-andreas: The original `Description` Metadata was migrated from am `ComponentMetadataDefinition` to be a `HandlerMetadataDefinition`.
The purpose of the `Description` Metadata is to add a **detailed** description (= **documentation**) to a Handler. 
This is how it is currently used in e.g. the OpenAPI exporter.
Therefore, there is no use of creating description declarations which apply to multiple endpoints, as each description should be mindfully written for a specific endpoint.
That all was the rationale behind the referenced change.

What you described was the previous behavior.

Sources/Apodini/WebService.swift
bauer-andreas: As a short term solution: An idea was to move all that logic (beginning and including the `metadata.collect...`) into the SyntaxTreeVisitor vistit method. However then the PathComponentContextKey would be added **outside** of the WebService, which then to me fells a bit misplaced.

What I would like to see in the future, could be a pattern based solution. Make the "path prefix" configurable by the user, and add some sort of version "pattern" (this can be string but also type based) which can be inserted by the user (defining a sensible default). This could also allow prefixes like `"/someservice-v2/..."`.
The actual insertion of the version is then done when parsing all the PathComponentContextKeys.

Just some thoughts coming up. I agree that there needs to be a better solution

Issue #233, opened 2021-02-08
[MAJOR FEATURE OpenAPIExporter] Add security scheme object to OAS


lschlesinger:
**Describe your use case**
In case the REST API implements security mechanisms, the OpenAPIExporter should properly document the applied security schema(s) for every endpoint.

**Current state**
As Apodini DSL does not yet support the definition of authentication and other security mechanisms. The generated OAS does not contain a field `security`, neither on document (whole API) nor on operation (endpoint) level. 

**Propose an approach**
Simultaneously to integrating authentication and security mechanisms to the exported REST API, the OpenAPIExporter should map such concepts provided in the Apodini DSL to swift types that can be translated in OAS. Such swift types are already implemented in the [OpenAPIKit](https://github.com/mattpolzin/OpenAPIKit/) library: `OpenAPI.SecurityScheme`, `OpenAPI.OauthFlow`, `OpenAPI.Docuement.SecurityRequirement`, as shown [here](https://github.com/mattpolzin/OpenAPIKit/blob/master/documentation/specification_coverage.md#security-scheme-object-openapisecurityscheme).

**Additional context**
Some information retrieved from the [official documentation](https://swagger.io/specification/):
According to the documentation the `security` field in an OpenAPI document represents a "_declaration of which security mechanisms can be used across the API_", as array of [**Security Requirement Objects**](https://swagger.io/specification/#security-requirement-object). In other words, it is a summary of all security mechanisms used on [**Operation Object**](https://swagger.io/specification/#operation-object)s throughout the API.
A security requirement object lists required security schemas to execute this operation, encapsulated in **[Security Scheme Object](https://swagger.io/specification/#security-scheme-object)**. Supported schemes are HTTP authentication, an API key (either as a header, a cookie parameter or as a query parameter), OAuth2's common flows (implicit, password, client credentials and authorization code) as defined in RFC6749, and OpenID Connect Discovery.

bauer-andreas:
Solved by #340. As described in the PR, due to different modelling strategies of security requirements between OpenAPI and ApodiniAuthorization the mapping is not perfect.

PR #340, opened 2021-08-21
Adding Metadata support to add OpenAPI Security schemes and requirements

bauer-andreas:
# Adding Metadata support to add OpenAPI Security schemes and requirements

As of now, the security scheme and security requirement object are completely unsupported in `ApodiniOpenAPI` (see #233).

## :bulb: Proposed solution

This PR introduces the `Security` Metadata which is used to generate the `Security Scheme Object` and the `Security Requirement Object`.
Further, ApodiniAuthorization, specifically the `BasicAuthenticationScheme` and `BearerAuthenticationScheme` implementations add automatic support for the Security Metadata. Therefore, if the user uses those out-of-the-box components, they do need to do anything to achieve a OpenAPI documentation of those security schemes.

## :gear: Release Notes 

`ApodiniOpenAPI` can now document security schemes and security requirements.
If you use `ApodiniAuthorization`, particularly the `BasicAuthenticationScheme` or `BearerAuthenticationScheme` (this includes `ApodiniAuthorizationJWT`), you do not need to do anything and authentication requirements are documented automatically.
You may want to look into the initialisers of `BasicAuthenticationScheme` and `BearerAuthenticationScheme` if you want to modify naming, add documentation to the security scheme or adding vendor extensions.

To add security schemes manually, look at the `Security` Metadata. Here are a few examples:
* Adding API Key from a query `@Parameter`: `Security(name: "api_key", .apiKey(at: $myApiKeyParameter))`
* Adding arbitrary API Key: `Security(name: "api_key", .apiKey(name: "X-Api-Key-Header", location. header))`
* Documenting HTTP Digest auth: `Security(name: "digest", .http(scheme: "digest"))`
* OpenID Connect URL to OAuth2 requirements: `Security(name: "my_auth", .openIdConnect(url: ...))`
* OAuth2 flows: `Security(name: "my_oauth", .oauth2(flows: ...))`

Edit: addressed (see https://github.com/Apodini/Apodini/pull/340#issuecomment-904837315).

Swift Package Manager does not support optional dependencies. I tried to use `#if canImport` on those places where `ApodiniAuthorization` packages define the `Security` Metadata (defined in `ApodiniOpenAPI`, which also expose `OpenAPIKit` types), however this typically resulted in the compiler failing to properly order its compilation order in respect to those "optional dependencies".
Therefore, to me the only viable solution as of now, is to make `ApodiniAuthorization` dependent on the `ApodiniOpenAPI` package (though with `@_implementationOnly`). Introducing an intermediate package (e.g. `ApodiniOpenAPIMetadata`) didn't make much sense to me, as this package would also need to expose `OpenAPIKit`, which wouldn't improve the situation really.
Any ideas are welcome.

### Related PRs
- solves #233 

### Testing
Additional test vectors were added.

### Reviewer Nudging
Maybe start looking at the tests, look at the explanations in #233 to get a rough understanding of the OpenAPI spec.
Read the additional information section above.


bauer-andreas:
Again, CI is currently blocked by the issues around AWS ApodiniDeploy tests, documented in Rocket chat (@PSchmiedmayer).
I nonetheless request review on this PR, thanks :)

Sources/ApodiniAuthorization/Delegates/AuthorizationRequirementsChecker.swift
```diff
@@ -38,7 +39,7 @@ struct AuthorizationRequirementsChecker<H: Handler, Element: Authenticatable>: H
         if !authenticatable.isAuthorized {
             switch type {
             case .optional:
-                return await try delegate.instance().handle()
+                return try await delegate.instance().handle()
```
philippzagar: Nice that this is fixed now, I think i was getting some compiler warnings because of that :D

Sources/ApodiniOpenAPI/Document/OpenAPIPathsObjectBuilder.swift
```diff
@@ -89,14 +89,51 @@ private extension OpenAPIPathsObjectBuilder {
...
+
+
+        let securitySchemes = endpoint[Context.self].get(valueFor: SecurityMetadata.self).mapToOpenAPISecurity(on: endpoint)
+
+        var securityArray: [OpenAPIKit.OpenAPI.SecurityRequirement] = []
+        var requiredSecurityRequirementIndex: Int?
+
+        var requiresAuthentication = false
+
+        // see https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#security-requirement-object
```
philippzagar, line 103: Love the references to the OpenAPI Docs!

philippzagar (approved):
To my limited understanding of ApodiniOpenAPI and ApodiniAuth, this looks fine to me!

I also struggled with `#if canImport` in ApodiniObserve, my solution was (as you proposed) to create a new target. However, as you pointed out, a similar approach wouldn't make much sense here, therefore I think the dependency is fine. If @hendesi has another idea, I would be glad to change my mind

Sources/ApodiniAuthorization/Delegates/AuthorizationRequirementsChecker.swift
bauer-andreas: Seems like I somehow messed that up back then 😅

bauer-andreas:
I've done my best to address the above-described issue of ApodiniAuthorization depending on OpenAPI.

I have introduced a dedicated target for the definition of the OpenAPI Security Metadata which does not depend on any OpenAPIKit dependencies. It introduces some more generalised type representations and resulted in some minor duplications of OpenAPIKit models.
Further some functionality, like the definition of OAuth2 Flows is only available in the ApodiniOpenAPI package (via extensions). Maintaining our own version of the OAuthFlow type would have resulted in major code duplication. Since we currently do not need to define OAuth flows in ApodiniAuthorization, I think that is an acceptable solution.

PR #341, opened 2021-08-22
Update Deployment providers to cli changes + Test cases

hendesi:
# Update Deployment providers to cli changes + Test cases

Among others #326 introduced the possibility to automatically add subcommands defined by a `Configuration` to the web service CLI. This brings the problem that `Self()` does not initialise any options or arguments that may have been defined and therefore crashes at startup.

This was fixed by #343, but fixed lacked some in-deep test coverage as well as provide sufficient support for the deployment providers. 

## :bulb: Proposed solution
This PR adds `ParsableArgumentsTests` that cover the automatic command adding including potential nested sub commands for web services that contain arguments or options.

It also adds support for the deployment providers. This is done via an additional argument that is added to all relevant deployment provider CLIs:
```
@Argument(parsing: .unconditionalRemaining, help:"CLI arguments of the web service")
var webServiceArguments: [String] = []
```
Using the parsing option `.unconditionalRemaining` allows to group arbitrary following arguments together in a way that they can be pass straight to any deployment related subcommands, e.g. deploy startup or deploy export-ws-structure. See the [docs](https://github.com/apple/swift-argument-parser/blob/main/Documentation/02%20Arguments%2C%20Options%2C%20and%20Flags.md)  for more info.

## :gear: Release Notes 
- Applies the fix of #343 to the deployment providers and adds test cases for it.

### Related PRs
Adds tests for the fix in #343 and adapt it to the deployment providers

### Testing
Adds `ParsableArgumentsTests` to cover the changes of #343

### Reviewer Nudging
Should be straight-forward


bauer-andreas (approved):
To my limited understanding of the matter, this fix looks good to me.

hendesi:
@Supereg Thanks for the quick review!
Unfortunately there is an issue with the current fix as I pointed out in the edit in my first comment. I wasn't able to find a solution for this that keeps the subcommand inference and at the same time allows custom arguments and options. 
To provide at least some fix as this is an issue on release, I changed it such that the automatic command adding is deactivated if options/arguments are added.

@PSchmiedmayer Are you fine with this temporal solution or do you have any input on this by any chance?

Sources/Apodini/WebService.swift
```diff
@@ -170,3 +176,8 @@ public enum WebServiceExecutionMode {
     /// It exits afterwards.
     case boot
 }
+
+private protocol CLIParsable {}
+extension Option: CLIParsable {}
+extension Argument: CLIParsable {}
+extension OptionGroup: CLIParsable {}
```
philippzagar: What about Flag?

philippzagar: Be sure to check the documentation of the ArgumentParser, maybe there are even more CLI types that I'm not aware of

philippzagar (approved):
Generally looks fine to me, but it's a bit of a hacky solution :D
However, I guess as long as we don't run into other problems, this approach is fine.

Sources/Apodini/WebService.swift
PSchmiedmayer: I agree with @philippzagar, we are currently missing a few types and there might even be more types added in the future. I would go with a different approach as detailed above and remove this protocol 👍

Sources/Apodini/WebService.swift
```diff
@@ -78,9 +79,14 @@ extension WebService {
...
-    /// The command configuration of the `ParsableCommand`
+    /// The command configuration of the `ParsableCommand`.
+    /// The default value depends if there are arguments or options specified in the web service.
+    /// If so, it contains the default empty configuration.
+    /// If not, it automatically adds the commands of the specified `Configurations` as sub command.
     public static var configuration: CommandConfiguration {
-        CommandConfiguration(subcommands: Self().configuration._commands)
+        let mirror = Mirror(reflecting: Self())
+        let hasNoRelevantChildren = !mirror.children.contains { $0.value is CLIParsable }
+        return CommandConfiguration(subcommands: hasNoRelevantChildren ? Self().configuration._commands : [])
```
PSchmiedmayer: I took a short look at the problem and I still think the solution you proposed here is problematic @hendesi:
As noted by @philippzagar we are missing some possible property wrappers and I think this already highlights the problem with the approach. The Swift argument parser provides a `parseOrExit` method that can be loaded from the globally available `CommandLine.arguments`.

**Update:**
@hendesi @philippzagar @Supereg I have transformed my long comment here into a separate PR that you can find here: #343 ... I think it addresses the shortcomings and ideas that were mentioned in this comment. 

@hendesi There are still a few open things that are needed to actually fix the problem of being able to pass arguments from a deployment provider to a web service that should be addressed in this PR.
In addition I would ask you to add some unit tests as part of this restructuring that make sure that we don't run into the same problems again and regressions on the command line parsing front are reduced 👍

PSchmiedmayer (changes requested):
This is an unfortunate miss, all example projects that we have fail with `0.4.X` as far as I can see as they all use some subcommands.

I have taken a look at the issue and have a few ideas about your thoughts mentioned above. I noted this down in the review comments.

@hendesi Please add several unit tests that automatically test this behaviors with and without command line arguments as well as some deployment related subcommands to make sure that we don't have such a regression in the future.

Sources/Apodini/WebService.swift
PSchmiedmayer: @hendesi I have updated the comment above with a longer explanation and have pushed my ideas to an other branch. I will update the comment here after creating the other PR.

PSchmiedmayer (approved):
@hendesi Thanks for the improvements and the additional test cases. The additions look good to me. 

The same appraoach can probably also be applied to the `migrator` subcommand in #345 (CC @eldcn)

PR #343, opened 2021-08-23
Fix Crash When Adding CLI Options to a Web Service

PSchmiedmayer:
# Fix Crash When Adding CLI Options to a Web Service

Among others, #326 introduced the possibility to automatically add subcommands defined by a `Configuration` to the web service CLI. This was done by:
```
public static var configuration: CommandConfiguration {
        CommandConfiguration(subcommands: Self().configuration._commands)
}
```
This brings the problem that `Self()` does not initialize any options or arguments that may have been defined, and therefore, crashes at startup.

## :bulb: Proposed solution
Create a more sophisticated parsing of the arguments that follows the following approach. While taking a look at the solution in #341, I came up with a few ideas. Instead of leaving them in a comment there, I added them to the code base and created this PR. 
My idea would be to merge this PR sooner than later and tag a new release and then let @hendesi add some test cases in the #341 PR and fix the issues with the deployment providers in the #341 PR as well.

### Parsing arguments and still getting the configuration even if arguments are present: 

The problem is that the argument parser parses the configuration multiple times and recursively, and therefore we need some hacky caching mechanism is needed:

On the first call of the `configuration` property, we use the Swift argument parser provided `parseOrExit` method that can be loaded from the globally available `CommandLine.arguments`.
As the `configuration` is accessed during the execution of `parseOrExit` recursively and multiple times, we just return an empty configuration for this case using the `_firstParse` check. 

Once we have created the configuration, we store it in the `_configuration` to avoid doing this multiple times and providing access the subcommands to the argument parser.

 ### A deployment provider related note:

In addition: A deployment provider **must** can pass any arbitrary arguments to a sort of "passthrough arguments" argument that is just piped through to the web service. E.g., if the web service as an argument that must be set the arguments are piped through to the web service. 

I think we should merge this PR as soon as possible, so adding this to the deployment providers should be done later in PR #341, but IMO this should also be done sooner than later.
This PR already adds a `parseAutomaticSubcommands` that already handles the possible presence of more arguments and a subcommand being in the `arguments` array:: 
```swift
// ...


let arguments = Array(CommandLine.arguments.dropFirst())

func parseAutomaticSubcommands(arguments: [String]) -> [ParsableCommand.Type] {
    do {
        print(arguments)
        return try Self.parse(arguments).configuration._commands
    } catch {
        if arguments.isEmpty {
            return []
        } else {
            let newArguments = Array(arguments.dropLast())
            return parseAutomaticSubcommands(arguments: newArguments)
        }
    }
}

let subcommands = parseAutomaticSubcommands(arguments: arguments)

// ...
```
We drop arguments until we reach the arguments of the web service. If we can't parse it, we return no subcommands, and an appropriate error message will be thrown anyways. 

The subcommands would then need to use `@OptionGroup` as detailed in the Swift argument parser docs to also catch all arguments of the web service to not produce any error messages. The good thing is that all subcommands are generic over the `WebService` type anyways, so this should work out fine: https://github.com/apple/swift-argument-parser/blob/main/Documentation/03%20Commands%20and%20Subcommands.md

A subcommand configuration with a small subcommand looking like this: 
```
struct TestSubcommandConfiguration<W: WebService>: Configuration {
    func configure(_ app: Application) {
        print("👋")
    }
    
    struct TestCommand: ParsableCommand {
        public static var configuration: CommandConfiguration {
            CommandConfiguration(commandName: "test")
        }
        
        @OptionGroup var webServiceOptions: W
        @Option var test: Int
        
        func run() throws {
            print("Test: \(test)")
        }
    }
    
    var command: ParsableCommand.Type {
        TestCommand.self
    }
}
```
and added to a web service like this:
```swift

@main
struct MyWebService: WebService {
    @Option var port: Int = 80
    
    
    var configuration: Apodini.Configuration {
        TestSubcommandConfiguration<Self>()
        // ...
    }
```
would then be executable with the following arguments:
`--port 90 test --test 12`
while the normal web service startup would still work like this:
`--port 90`
or even with no arguments as the `port` has a default value.

This has to be integrated into the deployment providers and heavily tested in the #341 PR.

## :gear: Release Notes 
- Fixes startup crash when providing custom options or arguments to web service

This solution seems kind of hacky, but since this is was introduced with the new release, I think this should be fixed rather quickly. Any suggestions on how to improve this solution are appreciated.

### Related PRs
As mentioned, it fixes an issue introduced by #326 and discussed in #341

### Testing
Tests will be added in #341, including the needed support for the deployment provers to pass required arguments down to a web service, e.g., exporting the structure or starting a web service.


Sources/Apodini/WebService.swift
```diff
@@ -80,7 +85,45 @@ extension WebService {
...
+        // E.g. we have the following setup. We have a `test` subcommand with one arguemtn and a port argument for the main web service
+        // A call to the subcommand would result in the follwing arguments array:
+        // ["--port", "90", "test", "--test", "12"]
+        // down to
+        // ["--port", "90"]
+        // which is successful as port is the only argument
+        // We can then extract the subcommands and pass this to the configuration below
+        func parseAutomaticSubcommands(arguments: [String]) -> [ParsableCommand.Type] {
+            do {
+                print(arguments)
```
hendesi: I think this can be removed now.

hendesi (approved):
Thanks for providing such a fast solution to this 🚀 
I will take over the test case adding and the changes to the deployment providers 👍

Sources/Apodini/WebService.swift
PSchmiedmayer: Thanks, good catch, will be removed 🚀

philippzagar (approved):
Way more robust and cleaner solution than before, LGTM!

PR #345, opened 2021-08-24
ApodiniMigrator

eldcn:
# *ApodiniMigrator*

Currently Apodini does not provide a functionality to track the API of the Web Service over different versions

## :bulb: Proposed solution
This PR introduces `ApodiniMigration` target. It adds a dependency to `ApodinIMigrator` framework, which handles automated library generation, migration guide generation and automated migration of the client library by means of the migration guide. Relevant part of the functionality in `Apodini` is solely retrieval of API related information in form of simplified `Document` (similar to OpenAPI), which is further used to support the above mentioned tasks.

This PR additionally adds a documentation of `ApodiniMigrator` and corresponding links to an example project. Furthermore, a subcommand `migrator` is added, which runs and starts the Web Service to perform migration tasks.

### Testing
Testing of the core functionalities of `ApodiniMigrator` framework, have been added to its corresponding project. This PR adds tests to the functionalities that relate with `MigratorConfiguration` and `ApodiniMigratorInterfaceExporter`

### Reviewer Nudging
Have a look at the documentation markdown first, then to `ExportOptions` which are used for the CLI and the configuration in source code at the same code.


Sources/Apodini/Apodini.docc/ApodiniMigration/Apodini Migration.md
```diff
@@ -0,0 +1,184 @@
...
+                          the initial web service version
+  read                    A parsable command to export a local migration guide
+                          and the API document of the current version
+  compare                 A parsable command for generating the migration guide
+```
+
+For the initial version of the Web Service `document` subsubcommand can be used as follows to run the web service and expose the document of the current version at `/api-document` endpoint as `yaml`:
+
+```console
+$ swift run YourWebService migrator document --doc-endpoint=api-document --doc-format=yaml
```
PSchmiedmayer: I would suggest that the examples feature examples of exporting the migration guide/web service description to a file instead of offering it as an endpoint. I think it makes sense to offer this option as well but I would generally suggest the following:

Examples showing the configuration in the `WebService` feature exporting documents/migration guides at an endpoint and usage of the command line interface features exporting the documents to a directory. 
- In my opinion and vision the `migrator` subcommand is mostly used for automations (and therefore it makes most sense to export it to a file)
- The configuration in `WebService`s is used when actually using a web service and also exporting the document/migration guide in addition to starting the web service.

Sources/ApodiniMigration/CommandLine/MigratorDocument.swift
```diff
@@ -0,0 +1,34 @@
...
+            version: "0.1.0"
+        )
+    }
+    
+    func run(app: Application, mode: WebServiceExecutionMode) throws {
+        app.storage.set(
+            DocumentConfigStorageKey.self,
+            to: .init(exportOptions: export)
+        )
+        try Service.start(mode: mode, app: app)
```
PSchmiedmayer: I would be interested in your option but in general I would say that if I want to export an document (or migration guide) to a file I would not need to start up the complete web service but would be fine to just export the document and then exit the command. Or at least offer an option to not start up the web service which is enabled by default. What do you think? 

Somehow refers to my command in the DocC documentation where I lay out my vision when the subcommand and the configuration in a `WebService` is used.

TestWebService/Sources/TestWebService/TestWebService.swift
```diff
@@ -47,5 +48,9 @@ struct TestWebService: Apodini.WebService {
         }
         
         WebSocket()
+        
+        Migrator(
```
PSchmiedmayer: One thing that is currently missed as far as I can see is taking a look at the special case when a web service contains `@Arguments` or `@Options` from the argument parser. 

Similar to the #343 and #341 ... it would make sense to test this in here and maybe also in the example that the subcommand also works when you have arguments/options or additional things from the Swift Argument Parser in the main WebService.

I think you can take inspiration from #343 and #341 and the comments in there on how to address this in the subcommands.

PSchmiedmayer (approved):
Thanks for the nice PR @eldcn !!

All the functionality I requested and all the feedback I have given in the first feedback is incorporated. The code’s documentation, the code’s structure, and the overall quality look very good!

I only had three main comments, with one addressing the documentation, one questioning if the subcommand should fully start up the web service, and one that might result in the most work: We have to consider that `WebService`s themselves can also have `@Arguemt`s, `@Option`s and so on from the Swift Argument Parser.

I can already approve the PR, and when the three points are addressed, you can merge the PR. Let me know what you think about the second feedback regarding always starting up a web service with the migrator subcommand, which IMO might not be the rested default behavior.

Sources/ApodiniMigration/CommandLine/MigratorDocument.swift
eldcn: I included a flag at each of the subsubcommands to determine whether the web service should run or not. Makes totally sense, thank you 🙏

TestWebService/Sources/TestWebService/TestWebService.swift
eldcn: Solved it as done in #341

PR #346, opened 2021-08-26
Fix #318

### LINKED ISSUE #318: [MINOR FEATURE] Improve the `AutomaticCommunicationalPattern` Knowledge Source

theMomax:
# Add better heuristics to automatic detection of Communicational Pattern

See #318 

## :bulb: Proposed solution
As described in #318 and the Communicational Pattern proposal, the presence of `@State`, `AnyObservedObject`s, and `@Environment(\.connection)` is used to differentiate patterns in addition to the return-type.

## :gear: Release Notes 
* Automatic detection of `CommunicationalPattern` was improved. This might change the behavior of some endpoints, if they make use of `State`, `ObservedObject`, `Environment(\.connection)`, or any `Environment`/`EnvironmentObject` that contains a `ObservableObject`. In some cases manually setting the `CommunicationalPattern` via metadata may become redundant. In edge-cases it may happen that the previous, simple heuristic made the correct choice, whereas the new heuristic assumes the pattern to be a different one. In that case you have to explicitly specify the `Pattern` via metadata on your `Handler`.

### Testing
I added tests verifying the correct choice for some exemplary `Handler`s

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -2,7 +2,7 @@
   "object": {
     "pins": [
       {
-        "package": "apnswift",
+        "package": "APNSwift",
         "repositoryURL": "https://github.com/kylebrowning/APNSwift.git",
         "state": {
           "branch": null,
--- a/Sources/ApodiniAuthorization/Delegates/AuthorizationRequirementsChecker.swift
+++ b/Sources/ApodiniAuthorization/Delegates/AuthorizationRequirementsChecker.swift
@@ -38,7 +38,7 @@ struct AuthorizationRequirementsChecker<H: Handler, Element: Authenticatable>: H
         if !authenticatable.isAuthorized {
             switch type {
             case .optional:
-                return await try delegate.instance().handle()
+                return try await delegate.instance().handle()
             case .required:
                 throw authenticationRequired
             }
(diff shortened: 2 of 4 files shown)


bauer-andreas (approved):
👍

PR #348, opened 2021-08-28
OpenAPI TypeInformation

eldcn:
# *Integrate the standalone `ApodiniTypeInformation` package into `ApodiniOpenAPI`*

*The current type reflection approach does not consider `Fluent` property wrappers. Furthermore, there is no centralized solution to be used from different exporters for deriving type information of input and return types of the `Handler`s*

## :bulb: Proposed solution
*`ApodiniTypeInformation` package is the initial attempt to have a separate package for deriving type information, and adding additional meta data context to `Content` types.*

### Testing
*In order to guarantee unique type names for all content types of the Web Service, `TypeInformation` includes the absolute name in its `TypeName` property, considering also the nested types where a type has been defined in. Hence, adjustments to the existing OpenAPI tests were done by moving test types in the test target name space.*

### Reviewer Nudging
*Only `OpenAPIComponentsObjectBuilder.swift` has been trimmed down, and uses the new `TypeInformation`*


Package.swift
```diff
@@ -80,6 +80,10 @@ let package = Package(
         // restore original package url once https://github.com/wickwirew/Runtime/pull/93
         // and https://github.com/wickwirew/Runtime/pull/95 are merged
         // .package(url: "https://github.com/wickwirew/Runtime.git", from: "2.2.3"),
+        
+        // ApodiniTypeInformation
+        .package(url: "https://github.com/Apodini/ApodiniTypeInformation.git", from: "0.2.0"),
```
bauer-andreas: We might want to use `.upToNextMinor(from:)` to not automatically jump to 0.3.0, which might contain breaking changes.

bauer-andreas (approved):
Looks good to me 👍 Great work 🚀 

Lets just wait some hours until CI has run lol 🙃

Sources/ApodiniOpenAPI/Document/OpenAPIComponentsObjectBuilder.swift
```diff
@@ -101,102 +105,26 @@ class OpenAPIComponentsObjectBuilder {
...
+            let schema: JSONSchema = .from(typeInformation: typeInformation)
+            let schemaName = typeInformation.jsonSchemaName()
             if schema.isReference && !schemaExists(for: schemaName) {
-                var properties: [String: JSONSchema] = [:]
-                for child in node.children {
-                    if let propertyInfo = child.value.propertyInfo {
-                        properties[propertyInfo.name] = deriveSchemaFromReflectionInfo(child)
-                    }
+                let properties = typeInformation.objectProperties.reduce(into: [String: JSONSchema]()) { result, current in
+                    result[current.name] = .from(typeInformation: current.type)
```
bauer-andreas, line 93: Would we, in this case, need to recursively resolve `.reference` JSONSchemes here? Or is this dealt through the `allTypes` above anyways?

eldcn: Yes exactly, there is no need to because `allTypes` returns all distinct types contained in `rootTypeInformation`.

bauer-andreas: thanks for the clarification

PR #351, opened 2021-08-30
[Feature] IoT Deployment Provider

hendesi:
# IoT Deployment Provider

Currently Apodini supports the deployment to localhost and aws environments. As of now no deployment to IoT devices is possible.

## :bulb: Proposed solution
This PR adds the support for deployment of Apodini Web Service to IoT devices. The deployment is managed by the `DeploymentTargetIoT` library and handled by `IoTDeploymentProvider`. Latter can be configured via the initialiser: 
```
public init(
        searchableTypes: [String],
        productName: String,
        packageRootDir: String,
        deploymentDir: String,
        automaticRedeployment: Bool,
        additionalConfiguration: [ConfigurationProperty: Any] = [:],
        webServiceArguments: [String] = []
    ) {
       ...
    }
```
`searchableTypes` describes the identifier of device the web service should be deployed on. Example: If you want to deploy a web service on a raspberry pi, that publishes itself via avahi, you would want to pass `_workstation._tcp.` here.

The provider searches for devices using [SwiftDeviceDiscovery](https://github.com/hendesi/SwiftDeviceDiscovery). This allows to define `PostDiscoveryAction`s that will be executed on the found device. These actions can be used to search for end device that are connected the remote, e.g. LIFX lamps. The defined actions can be registered to the provider by calling `registerAction`. This requires to also pass a `DeploymentDeviceMetadata` which will be mapped to the action. The passed metadata can used to annotate handlers with. Annotated handlers will be exported when the web service is started remotely and the corresponding `PostDiscoveryAction` returned a positive value. 

`additionalConfiguration` allows to define custom `ConfigurationProperty`s that can be used in the actions. Such as:
Define it:
```
public extension ConfigurationProperty {
    static var deploymentDirectory = ConfigurationProperty("key_deploymentDirectory")
}
```
Use it:
```
@Configuration(.username)
public var username: String
```

The IoT deployment provider differs from the existing as it is no executable by itself, but rather offers the means to define a custom provider executable for the users needs. To do so one must follow these steps:

1. Create a new package that imports [SwiftDeviceDiscovery](https://github.com/hendesi/SwiftDeviceDiscovery) and define a custom action here. Normally you would use this to look for end devices connected to the device the web service is deployed on. A `PostDiscoveryAction` requires the implementation of `run` which should return the number of found devices.
2. Create a new target importing `DeploymentTargetIoTCommon`. This will contain the option. Import this target to your web service to annotate handlers with. This can look something like: 
```
extension DeploymentDevice {
    public static var lifx: Self {
        DeploymentDevice(rawValue: "lifx")
    }
}
```
3.  Create a new executable target importing `DeploymentTargetIoTCommon`, `DeploymentTargetIoT`, your new target containing the option and the package containing the `PostDiscoveryAction`. In this target you initialise the `IoTDeploymentProvider` and start it. Most importantly, register the action you defined in your package and map it to the option, such as:
```
provider.registerAction(
              scope: .all, 
              action: LIFXDeviceDiscoveryAction.self, 
              option: DeploymentDeviceMetadata(.lifx)
)
```  
After the provider is started, it begins to search for devices. If there are positive results of any of the define actions, the web service package is copied to the remote and build there. After that, the provider runs the `export-ws-structure` command to retrieve all endpoints that match the mapped options of the actions. If there are any endpoints, the web service is started in a tmux instance.

## :gear: Release Notes 
- Adds the `IoTDeploymentProvider` enabling the deployment of Apodini web service on to IoT devices
- Adds the `LifxIoTDeploymentTarget`, an implementation of the iot deployment provider that deploys a web service on devices to which are LIFX lamps connected

- This PR also adds a `KillSessionCommand` that allows to automatically kill all started instances of the deployed web service on the remote devices.
- I tried to explain shortly the functioning above, if there is anything this unclear, I added an extensive documentation as well. 
- As already said, I also added an implementation for LiFX devices.
- Although there is already  a parameter for the automatic re-deployment, this is __not__ added with this PR. I will create a separate PR once this feature is fully ready.
- 
### Related PRs
Might need some updates once #341 is merged

### Testing
I've updated `WebServiceStructureExportTests` to test the structure retrieval for the IoT deployer. But I'm still missing some extensive tests of the real deployment.
@PSchmiedmayer Is it possible to add a pi to the bamboo CI setup, similar to the arm mac books, so we could write a proper test case of the deployment?

### Reviewer Nudging
Probably best to start with the doc to get an overview and then have a look at the `LifxIoTDeploymentTarget` and [the project containing its PostDiscoveryAction](https://github.com/Apodini/Swift-NIO-LIFX-Impl) to see how it is actually implemented. After that one could look at the `IoTDeploymentProvider` itself.


hendesi:
The IoT deployment provider has been moved to a separate [repo](https://github.com/Apodini/ApodiniIoTDeploymentProvider).

Closing this.

PR #363, opened 2021-09-22
Fix Handler part

rstar900:
# *Name of the PR*

*Describe the current situation (if possible with and exemplary (or real) code snippet and/or where this is used)*

## :bulb: Proposed solution
*Describe the solution and how this affects the project and internal structure*

## :gear: Release Notes 
*Add a short summary of the feature as well as possible migration guides if this is a breaking change so this section can be added to the release notes.*
*Include code snippets that provide examples of the feature implemented if it appends or changes the public interface.*

*Provide some additional information if possible*

### Related PRs
*Reference the related PRs*

### Testing
*Are there tests included? If yes, which situations are tested and which corner cases are missing?*

### Reviewer Nudging

### DIFF
--- a/Sources/Apodini/Apodini.docc/GettingStarted/HelloWorld.md
+++ b/Sources/Apodini/Apodini.docc/GettingStarted/HelloWorld.md
@@ -70,7 +70,7 @@ Our `Greeter` defines a ``Parameter`` called `country`. The parameter must be of
 
 > Tip: See also: ``PathParameter`` express parameters that are exposed as part of the endpoint identifier (e.g. the HTTP URL).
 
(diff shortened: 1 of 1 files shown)


PSchmiedmayer:
@rstar900 Unfortunately, this `handle()-1xtna` is needed to directly reference the `handle()` function implementation we want to link in the DocC documentation. Therefore we can't remove this at the moment. Looking at this in the .md files in GitHub seems strange, but it looks better if you render the documentation in Xcode. We plan to automatically deploy the documentation to either GitHub Pages once DocC supports static HTML pages or a server we need to host if static site support doesn't come in the near future. After that, you can see the rendered documentation in the browser.

Issue #362, opened 2021-09-22
[BUG] Cannot set port when using HTTP2Configuration


Lerbert:
**Describe the bug**
The HTTP2Configuration does not support setting a port for the webservice. I also tried using it in conjunction with the HTTPConfiguration to set the port, but those two configurations seem to be incompatible.

**To Reproduce**
Create a webservice with this configuration:
```swift
var configuration: Configuration {
      HTTP2Configuration(cert: "/path/to/cert.pem", keyPath: "/path/to/key.pem")
      REST {
          OpenAPI()
      }
  }
```
Now try setting a port other than the default (8080).

When using the config with HTTPConfiguration like this:
```swift
var configuration: Configuration {
      HTTPConfiguration(port: 443)
      HTTP2Configuration(cert: "/path/to/cert.pem", keyPath: "/path/to/key.pem")
      REST {
          OpenAPI()
      }
  }
```
all the endpoints return an empty response.

**Expected behavior**
There is some possibility to set the port. Maybe additional parameters for HTTP2Configuration or making the two configurations compatible.

**Environment (enter your versions)**

 - Swift Version: 5.5
 - Xcode Version: none (using vscode and docker)

Issue #271, opened 2021-04-19
[BUG] Swagger UI is not able to send requests to web service


AndiBraimllari:
**Describe the bug**
Simply downloading and running the Apodini Template shows an error when accessing the Swagger-UI. When accessed through Chrome the following message is shown ```Failed to fetch. Possible Reasons: CORS Network Failure URL scheme must be "http" or "https" for CORS request.```. When ran in Safari another bug was thrown. On the other hand, using cURL worked fine with the endpoints.

**To Reproduce**
Download the Apodini Template and run it, then open Chrome and go to the Swagger-UI page and execute an endpoint.

**Expected behavior**
Executing endpoints of the Apodini Template from the Swagger-UI work without bugs.

**Environment (enter your versions)**

 - Swift Version: Apple Swift version 5.3.2
 - Xcode Version: Version 12.4 (12D4e)

**Additional context**
Recently started using Big Sur 11.2.3.

ridvan-cln:
Hi @AndiBraimllari @PSchmiedmayer,

I think this is a typical CORS issue that I have seen in many projects already.
I "solved" the issue by just adding `app.middleware.use(CORSMiddleware(configuration: .default()))` , before we start the vapor server in `Application+Vapor`.

I see two main possibilities to solve this issue:

- We add the CORS middleware by default to the swagger-ui requests only and close this issue for now
- We give developers a configuration to setup middlewares as they want (could be tricky to do this in an agnostic way, if we want to move from vapor), I think we don't have this option yet right?

What do you guys think?

ridvan-cln:
Hey guys,
little update from me on this topic. I have read a few articles on how to securely configure CORS and it is a very complex topic tbh. (https://www.pivotpointsecurity.com/blog/cross-origin-resource-sharing-security/)

But what I found out is that one can simply avoid this issue by opening Swagger UI from the URL http://127.0.0.1:8080/oas-ui instead of http://localhost:8080/oas-ui . (see screenshots below)
The reason is that for some reason 127.0.0.1 and localhost are not seen as the same Origin and CORS forbids that.

Do you think we can close this issue now? @PSchmiedmayer @AndiBraimllari 
![Bildschirmfoto 2021-09-20 um 17 41 20](https://user-images.githubusercontent.com/1402089/134031616-cd2f21da-a15c-4f48-84a5-a5307eb49807.png)
[...]

PR #365, opened 2021-09-25
HTTP configuration refactored

ridvan-cln:
# *HTTP configuration refactoring*

#271
#362

## :bulb: Proposed solution
Merged the two configs "HTTPConfiguration" and "HTTP2Configuration".
Changed default http values from "0.0.0.0:8080" to "localhost:80"
Removed unnecessary code from Application+HTTP

## :gear: Release Notes 
*Add a short summary of the feature as well as possible migration guides if this is a breaking change so this section can be added to the release notes.*
*Include code snippets that provide examples of the feature implemented if it appends or changes the public interface.*

*Provide some additional information if possible*

### Related PRs
*Reference the related PRs*

### Testing
*Are there tests included? If yes, which situations are tested and which corner cases are missing?*

### Reviewer Nudging


Sources/Apodini/Configurations/HTTPConfiguration.swift
```diff
@@ -53,7 +53,7 @@ public final class HTTPConfiguration: Configuration {
                 let port = components.last.flatMap { Int($0) }
                 self.address = .hostname(hostname, port: port)
             case let (hostname, port, .none, .none):
-                self.address = .hostname(hostname ?? Defaults.hostname, port: port ?? Defaults.port)
+                self.address = .hostname(hostname, port: port)
```
ridvan-cln, line 56: Looks like we covered the absence already in the first case (so no need to check?). I changed the first case to use the default values

TestWebService/Sources/TestWebService/TestWebService.swift
```diff
@@ -38,6 +38,8 @@ struct TestWebService: Apodini.WebService {
     }
     
     var configuration: Configuration {
+        HTTPConfiguration()
```
ridvan-cln, line 41: Any idea how we could avoid this call and use the default anyway? Or is this fine?

Sources/Apodini/Configurations/HTTPConfiguration.swift
```diff
@@ -15,7 +15,7 @@ import NIO
 /// function `address`
 public final class HTTPConfiguration: Configuration {
     enum Defaults {
-        static let hostname = "0.0.0.0"
+        static let hostname = "localhost"
         static let port = 8080
```
PSchmiedmayer: I think we can also move to port 80 by default.

Sources/ApodiniREST/RESTExporterConfiguration.swift
```diff
@@ -12,32 +12,35 @@ import Vapor
...
-        let bindAddress: Vapor.BindAddress
+        let bindAddress: Apodini.BindAddress
         let uriPrefix: String
 
-        init(_ configuration: HTTPServer.Configuration) {
-            self.configuration = configuration
-            self.bindAddress = configuration.address
+        init(_ app: Apodini.Application) {
+            // Default initialization
+            if app.http.address == nil {
```
PSchmiedmayer: As noted in the main review comment I don't think this is the right place to create an default version of the HTTPConfiguration here.

PSchmiedmayer (changes requested):
Hi @ridvan-cln first of all thanks for your first PR and being interested in the project!! Great to see.

As noted in the comment, we need to extend the configuration of an Apodini Web Service with additional functionality that is then reused in the exporters such as the REST and the OpenAPI export to correctly configure the exporters (Swagger-UI in the OpenAPI exporter and the links section in the REST exporter).

I think you correctly identified the HTTP Configuration as an excellent place to put all this functionality.
I would suggest doing the following:

### 1. Merge HTTPConfiguration & HTTP2Configuration
Merge the HTTPConfiguration and the HTTP2Configuration into a single configuration (also addresses issue #362). This merged HTTPConfiguration contains all the necessary configurations that are currently in the individual configurations as well as an [option set](https://developer.apple.com/documentation/swift/optionset) that details if the web service should support HTTP and/or HTTP2.

### 2. Default Values
The new HTTPConfiguration uses "localhost" and port 80 as the default values for the hostname and port. I would favor only HTTP2 as the default value for the HTTP version option set.

### 3. Use the Application.Storage directly and use it to create an default instance
The configuration stores itself in the app storage if it is part of the configuration (in the `public func configure(_ app: Application) { ... }` method. If it is not part of an Configuration e.g. a computed property accessor on application would then create a default instance. 
```swift
extension Application {
    var httpConfiguration: HTTPConfiguration {
        guard let httpConfiguration = self.storage[HTTPConfigurationStorageKey] else {
            HTTPConfiguration()
                .configure(self)
            return self. httpConfiguration
        }

        return httpConfiguration
    }
}

private struct HTTPConfigurationStorageKey: StorageKey {
    typealias Value = HTTPConfiguration
}
```
In this case, we should add a hint to the documentation that the order in the configuration result builder is relevant. But I would generally favor that, e.g., the links section in the REST exporter or the host's name in the SwaggerUI is not defined at startup time but is pulled from the app storage every time to always refer to the latest value. In addition, most of the content of `Application+HTTP.swift` can be removed.

### 4. Remove then unnecessary code from `Application+HTTP.swift`
Replace all the previous more or less unused extensions of application from the `Application+HTTP.swift` with access to `app. httpConfiguration ` that is then used to populate the SwaggerUI hostname ([as noted here](https://github.com/Apodini/Apodini/issues/271#issuecomment-924869327)) and the REST exporter in the Links section (e.g., here: https://github.com/Apodini/Apodini/blob/cf1d39d15145be58e27eecb6a7154eb866504c0f/Sources/ApodiniREST/RESTExporterConfiguration.swift#L17)

Last but not least, I would ask you to rework the PR description in GitHub to conform to our PR template.

ridvan-cln:
Hey @PSchmiedmayer,

thanks for your detailed explanations of every step, this really helped!
The current changes of the branch, already fix the issue #362 , but I will still merge the two configs 👍 .
For your second point I have questions left:

- You wanted to use port 80 as default, but that would require root permissions to run the project. This is possible (Xcode -> Product -> Scheme -> Edit Scheme), but I just want to be sure if you are aware of this.
- Also port 80 would mean (by default) http and not https, which is totally fine because not everyone should need to configure certificates etc.. 
But if you want HTTP2 as the default, I think this requires the use of https, right? At least that's what I understood from the code base and this [article](https://theswiftdev.com/a-simple-http2-server-using-vapor-4/). 
And with that decision every user would have to specify his certificate/key path to run a simple server and also port 80 would not make sense (because of https -> 443 as default).

Thanks again and I am happy to be contributing to the project 😃

PSchmiedmayer:
@ridvan-cln Great to hear that my feedback helped you and that you agree that fixing both issues and making the project nicer in the same PR is a good idea 👍

> You wanted to use port 80 as default, but that would require root permissions to run the project. This is possible (Xcode -> Product -> Scheme -> Edit Scheme), but I just want to be sure if you are aware of this.

I have tried to reproduce this and I don't need root privileges to run the web service at port 80. Might it be that there already is a web service running on port 80 on your machine that is killed when launching the new one which then requires root privileges?
I have tried to reproduce this when going to the [TestWebService](https://github.com/Apodini/Apodini/tree/develop/TestWebService) in this repo and adding `HTTPConfiguration(port: 80)` as a new configuration resulting in
```swift
    HTTPConfiguration(port: 80)
    // ...
}
```
doesn't require root access on my machine. I am running 12.0 Beta (21A5522h) on a MacBook Pro (15-inch, 2018).


> Also port 80 would mean (by default) http and not https, which is totally fine because not everyone should need to configure certificates etc..
> But if you want HTTP2 as the default, I think this requires the use of https, right? At least that's what I understood from the code base and this article.
> And with that decision every user would have to specify his certificate/key path to run a simple server and also port 80 would not make sense (because of https -> 443 as default).

Good points! I don't think we can create a reasonable default configuration for a certificate for localhost that would be the default can we? If we can do something like this this would be interesting but if not then let's stick with HTTP and port 80.

Sources/Apodini/Configurations/HTTPConfiguration.swift
```diff
@@ -37,46 +51,91 @@ public final class HTTPConfiguration: Configuration {
...
+    /// Sets the `.pem` file from which the key should be read.
+    public func key(_ filePath: String) -> Self {
+        guard self.keyURL == nil else {
             return self
         }
-        self.address = address
+        
+        self.keyURL = URL(fileURLWithPath: filePath)
         return self
     }
```
PSchmiedmayer, line 140: Two comments about these functions:
1. I would argue that the expected behavior in Swift and when using DSLs in Swift of 
```swift
HTTPConfiguration()
    .address(.hostname("localhost", port: 80))
    .address(.hostname("localhost", port: 8080))
```
Is that port `8080` is set as it is the last modified value. Therefore I would firstly remove the checks (e.g.: `guard self.bindAddress == .hostname(Defaults.hostname, port: Defaults.port) else { ... }` in the functions.

2. I would haven argue that these modifiers are not needed as all the arguments can bypassed in the initializer of the HTTPConfiguration and this makes discovery of the arguments a bit easier then using the modifier functions

Sources/Apodini/Configurations/HTTPConfiguration.swift
```diff
@@ -37,46 +51,91 @@ public final class HTTPConfiguration: Configuration {
...
-    private var address: BindAddress?
+    public var bindAddress: BindAddress
+    private var certURL: URL?
+    private var keyURL: URL?
+    public var supportVersions: Set<HTTPVersionMajor> = [.one]
+    public var tlsConfiguration: TLSConfiguration?
     
     /// initalize HTTPConfiguration
-    public init(hostname: String? = nil, port: Int? = nil, bind: String? = nil, socketPath: String? = nil) {
+    public init(hostname: String? = nil, port: Int? = nil, bind: String? = nil, socketPath: String? = nil, cert: String? = nil, key: String? = nil) {
```
PSchmiedmayer, line 61: I would suggest to model the arguments as only two arguments:
1. Let the user pass in the `BindAddress` that already exists and add a convenience static function for the `bind` argument that looks something like this so a developer can also use that way to create a `BindAddress`:
```swift
extension BindAddress {
    static func address(_ address: String) -> BindAddress {
        // ...
    }
}
```
so you can also create a `BindAddress` using `.address("localhost:80")`, this is quite a nice syntactic sugar improvement that is possible in Swift.

2. The TSLConfiguration needs to be wither passed in as a whole or not at all. So just a simple Struct that just expect a non optional cert and key.

Sources/Apodini/Configurations/HTTPConfiguration.swift
```diff
@@ -37,46 +51,91 @@ public final class HTTPConfiguration: Configuration {
...
+            if let certURL = certURL, let keyURL = keyURL {
+                let certificates = try NIOSSLCertificate.fromPEMFile(certURL.path)
+                let privateKey = try NIOSSLPrivateKey(file: keyURL.path, format: .pem)
+                
+                self.supportVersions.insert(.two)
+                self.tlsConfiguration = .makeServerConfiguration(
+                    certificateChain: certificates.map { .certificate($0) },
+                    privateKey: .privateKey(privateKey)
+                )
+
```
PSchmiedmayer, line 100: You can already do that work in the initializer, then you don't have to store the `certURL` and the `keyURL` as private properties in the class.

Sources/Apodini/Configurations/HTTPConfiguration.swift
```diff
@@ -8,15 +8,29 @@
...
+/// --cert=/some/path/cert.pem
+/// --key=/some/path/key.pem
+///
+/// Example of config via code:
+/// ```
+/// HTTPConfiguration()
+///     .address(.hostname("localhost", port: 80))
+///     .certificate("/some/path/cert.pem")
+///     .key("/some/path/key.pem")
+/// ```
```
PSchmiedmayer, line 28: I don't think that adding the certificate and key with command line arguments is implemented at the moment. This has to be done by the developer creating the web service using the Swift argument parser, you can see this being done in the last code example in the README: https://github.com/Apodini/Apodini

Sources/ApodiniOpenAPI/OpenAPIInterfaceExporter.swift
```diff
@@ -91,15 +91,12 @@ final class OpenAPIInterfaceExporter: InterfaceExporter {
     }
     
     private func setApplicationServer(from app: Apodini.Application) {
-        let isHttps = app.http.tlsConfiguration != nil
+        let isHttps = app.httpConfiguration.tlsConfiguration != nil
```
PSchmiedmayer, line 94: Instead of each exporter computing if the web service is using HTTPs we should make this a computed property of the HTTPConfiguration. In addition we should add a way to "force" the web service to believe that it is served over HTTPS even though there is no TLS Configuration in the HTTPConfiguration.
A web service most often only serves HTTP as there is a reverse proxy or a more sophisticated gateway before it that takes care of HTTPS for the web service. Ideally the traffic between the web service and reverse proxy/gateway is also encrypted but that is most often not the case by default and often left out (e.g. in a Docker network). 
If one uses such a setup then the links and e.g. the OpenAPI document should reflect the fact that the actual communication is performed using HTTPS.

Maybe name the argument of the HTTPConfiguration `httpsReverseProxyExists` or something like this.

PSchmiedmayer (changes requested):
Thanks for improving the PR! I think we go in the right direction, but there are still a few things here and there and should be improved. I have added some comments in code and want to address one of the biggest ones in this message here: The problem with orders in the Configuration: As you noted, if you put the REST configuration before the HTTPConfiguration, you and up with the state that the REST exporter doesn't note, e.g., the configured hostname.

To solve this, I would suggest the following:
Remove the `REST.Configuration` altogether. The functionality of creating the `uriPrefix` can be done in the HTTPConfiguration as a computed property that the REST and the OpenAPI exporter can then access when they need it. There is also no need to store the binding address as a duplicate. As a second step, I would suggest that the relevant parts of the exporters store a reference to the Apodini App instance and access the HTTPConfiguration on runtime when, e.g., the links are generated or, e.g., the OpenAPI exporter creates the document. That way, we guarantee a single source of truth, and the order of configurations should become irrelevant.


**Last but not least a few housekeeping things:**

I have given you write access to the repo so you can create a local feature branch. As we currently use Bamboo for the macOS 12 builds, we can't test macOS builds on forks at the moment. Would you please recreate the PR and make sure that you have a detailed PR description?

In addition, please also make sure that all the tests are passing for the PR. They seemed to have timed out after a few hours, and the GitHub Action runners were canceled: https://github.com/Apodini/Apodini/pull/365/checks

Sources/Apodini/Configurations/HTTPConfiguration.swift
ridvan-cln: For the first point, completely agree. That was also my initial thought and I had changed it, but there were some tests failing specifically testing for this case. And I didn't want to adjust the existing tests to my "needs", but I will do that now :)

For the second point, I didn't quite understand, do you want to get rid of the address, certificate and key functions and let the user only specify with the initializer?

Sources/Apodini/Configurations/HTTPConfiguration.swift
ridvan-cln: Good idea, but I think it was done this way, because there are the certificate and key functions, with which you can set the paths after initialization.
Depending on your reply to my other comment, we could implement this change once we deleted the certificate and key functions

EDIT: deleting the functions as stated above and moving that work 👍

Sources/Apodini/Configurations/HTTPConfiguration.swift
PSchmiedmayer: 2. Point: Yes, I would argue for that 👍

PR #370, opened 2021-10-26
HTTP configuration refactoring

### LINKED ISSUE #271: [BUG] Swagger UI is not able to send requests to web service

### LINKED ISSUE #362: [BUG] Cannot set port when using HTTP2Configuration

ridvan-cln:
# *HTTP configuration refactoring*

Currently there are problems and inconsistencies when specifying/modifying the HTTP configuration of your Application with Apodini. As described in #271 and #362.

## :bulb: Proposed solution
The PR merges the two configs `HTTPConfiguration` and `HTTP2Configuration` and offers three different configuration elements:
- `hostname`: The `Hostname` that is used for populate information in exporters, the default value is `localhost:80` if there is no TLS configuration passed in the `HTTPConfiguration`, port 443 otherwise.
- `bindAddress`: The `BindAddress` that is used for bind the web service to a network interface, the default value is `0.0.0.0:80` if there is no TLS configuration passed in the `HTTPConfiguration`, port 443 otherwise.
- `tlsConfigurationBuilder`: Information about the key and certificate needed to enable HTTPS.

## :gear: Release Notes 

The PR merges the two configs `HTTPConfiguration` and `HTTP2Configuration` and offers three different configuration elements:
- `hostname`: The `Hostname` that is used for populate information in exporters, the default value is `localhost:80` if there is no TLS configuration passed in the `HTTPConfiguration`, port 443 otherwise.
- `bindAddress`: The `BindAddress` that is used for bind the web service to a network interface, the default value is `0.0.0.0:80` if there is no TLS configuration passed in the `HTTPConfiguration`, port 443 otherwise.
- `tlsConfigurationBuilder`: Information about the key and certificate needed to enable HTTPS.

The REST and OpenAPI exporters and Vapor only use the httpConfiguration when needed. This ensures that the ordering of HTTPConfiguration and the above exporters doesn't matter.


Sources/Apodini/Application/Application+HTTP.swift
```diff
@@ -25,80 +25,48 @@ public enum HTTPVersionMajor: Equatable, Hashable {
 
 /// BindAddress
 public enum BindAddress: Equatable {
-    case hostname(_ hostname: String?, port: Int?)
+    case hostname(_ hostname: String? = HTTPConfiguration.Defaults.hostname, port: Int? = HTTPConfiguration.Defaults.port)
```
PSchmiedmayer: I would suggest to make the `hostname non-optional.
The port can be optional but I would then even make it nil it not specified to make sure that it is set to either 80 or 443 based on the presence of a valid TLS config as described in the comments below.

Sources/Apodini/Configurations/HTTPConfiguration.swift
```diff
@@ -8,75 +8,93 @@
...
+            
+            return httpProtocol + "+unix: " + path
         }
     }
     
-    private var address: BindAddress?
     
     /// initalize HTTPConfiguration
-    public init(hostname: String? = nil, port: Int? = nil, bind: String? = nil, socketPath: String? = nil) {
+    public init(bindAddress: BindAddress? = nil, tlsFilePaths: TLSFilePaths? = nil) {
```
PSchmiedmayer: I would suggest to make the `bindAddress` parameter non-optional and instead add an other initializer that doesn't have a `bindAddress` parameter and then uses the default values. It would be cool if that default value would be calculated based on the presence of a valid `tlsFilePaths` value (see commend below). E.g. if there is no bindingAddress specified but a valid TLS config then the default port is 443.

Sources/Apodini/Configurations/HTTPConfiguration.swift
```diff
@@ -8,75 +8,93 @@
...
 
+    
     /// Configure application
     public func configure(_ app: Application) {
-        if let address = self.address {
-            app.http.address = address
+        if supportVersions.contains(.two) {
+            app.logger.info("Using HTTP/2 and TLS.")
+        } else if tlsUnsuccessful {
+            app.logger.warning("Error while enabling HTTP/2. Starting without HTTP/2.")
```
PSchmiedmayer: This error message is a bit misleading. It could also be triggered if you want HTTP/1 but have configured the TLS setting wrong. I would suggest to remove the `tlsUnsuccessful` property and print out a TLS warning message in the initializer where we can direly check this (currently line 83 `self.tlsUnsuccessful = true`) or better in the `TLSFilePaths` type as detailed above.

Sources/Apodini/Configurations/HTTPConfiguration.swift
```diff
@@ -8,75 +8,93 @@
...
-                throw HTTPConfigurationError.incompatibleFlags
+            if let paths = tlsFilePaths {
+                let certificates = try NIOSSLCertificate.fromPEMFile(paths.certificatePath)
+                let privateKey = try NIOSSLPrivateKey(file: paths.keyPath, format: .pem)
+                
+                self.supportVersions.insert(.two)
+                self.tlsConfiguration = .makeServerConfiguration(
+                    certificateChain: certificates.map { .certificate($0) },
+                    privateKey: .privateKey(privateKey)
+                )
```
PSchmiedmayer: I would suggest to put this logic in an [failable Initializer](https://developer.apple.com/swift/blog/?id=17) of `TLSFilePaths` and store the server configuration as an internal property instead of storing the two paths. Maybe it makes sense to rename the type. 
As a result you can assume that in the initializer any non-optional value for the `tlsFilePaths` property is actually a valid configuration and you can make assumptions based on that (e.g. the default port). You can then also print an error message in the `TLSFilePaths`. type instead of here if the TLS config is wrong.

PSchmiedmayer (commented):
Thanks for the improvements! I think we are going in the right direction, I just had a few minor comments regarding the recent refactorings that you can find below.

Sources/Apodini/Application/Application+HTTP.swift
PSchmiedmayer: Further thinking about it, maybe hostname is not the right name here, it should probably be called `interface` as we should probably differentiate between the interface the web service binds to locally (e.g. which network interface in the docker container/host machine should be used) where `0.0.0.0` indicates all possible addresses (INADDR_ANY) and therefore bings it to all possible interfaces and the hostname that is available from the outside which then might be "localhost" or "ase.in.tum.de".

Sources/Apodini/Configurations/HTTPConfiguration.swift
PSchmiedmayer: As noted in my other comment in the second review: I think we should add a dedicated `hostname` parameter that has a default value of `"localhost"` and a port that is `nil` (set based on the TLS config). The port is then always the same as the `bindAddress` if one was specified there and no one in the `hostname` parameter or both were not specified.
In contrast to `BindAddress` this data structure only accepts a hostname and a port as a UNIX socket or something else is not possible for the outside value of an HTTP web service as the hostname would indicate the domain name or IP that the web service is reachable at and differentiate this from the binding address that defines the interfaces the web service binds to locally.

This `hostname` parameter is then used in the links of the REST exporter and the OpenAPI document and the OpenAPI document and can be changed to be anything else that might be needed (e.g., our web service is hosted at `ase.in.tum.de` then this would be the value) without any change to the local binding address (which would also make a lot of sense as the local binding address can't be resolved when you pass in `ase.in.tum.de`). 

Maybe this also solves the binding problems and permission errors on macOS.

Sources/Apodini/Application/Application+HTTP.swift
ridvan-cln: Why do you think that would be better? I think that it would be beneficial to have it optional in the case, you only want to change the port, and don't care about the hostname (keeping it default). What do you think?

Sources/Apodini/Configurations/HTTPConfiguration.swift
PSchmiedmayer: After a second lookI think this should also be possible with a single initializer 👍

Sources/Apodini/Application/Application+HTTP.swift
PSchmiedmayer: If we keep the default value we would end up always overriding the current hostname with the detail value anyways. It would only make a difference if the default value would be `nil`, allowing us to differentiate explicitly setting no value (and maybe keeping the old hostname set to something different before) and the default value explicitly being being passed in. But as far as I can see we don't support chaining the value after the fact or don't differentiate between these cases and therefore a non optional value would also allow us to express the same logic as it is there at the moment.

Sources/Apodini/Application/Application+HTTP.swift
```diff
@@ -25,80 +25,74 @@ public enum HTTPVersionMajor: Equatable, Hashable {
...
-extension Application {
-    /// Used to keep track of http related configuration
-    public var http: HTTP {
-        .init(application: self)
+/// Hostname
+public struct Hostname {
+    let address: String
+    let port: Int?
+    
+    /// Create a new `TLSFilePaths`
```
PSchmiedmayer:
```suggestion
    /// Create a new `Hostname`
```

Sources/Apodini/Application/Application+HTTP.swift
```diff
@@ -25,80 +25,74 @@ public enum HTTPVersionMajor: Equatable, Hashable {
...
-
-        /// HTTP Server address
-        public var address: BindAddress? {
-            get { storage.address }
-            set { storage.address = newValue }
+/// Builds the TLS configuration from given paths for use in HTTPConfiguration
+public struct TLSConfigurationBuilder {
+    let tlsConfiguration: TLSConfiguration
+    
+    /// Create a new `TLSFilePaths`
```
PSchmiedmayer:
```suggestion
    /// Create a new `TLSConfigurationBuilder`
```

Sources/Apodini/Configurations/HTTPConfiguration.swift
```diff
@@ -8,75 +8,96 @@
...
-    public func address(_ address: BindAddress) -> Self {
-        guard self.address == nil else {
-            return self
+    
+    /// Configure application
+    public func configure(_ app: Application) {
+        if supportVersions.contains(.two) {
+            app.logger.info("Using HTTP/2 and TLS.")
+        } else {
+            app.logger.info("Starting without HTTP/2. and TLS")
```
PSchmiedmayer:
```suggestion
            app.logger.info("Starting without HTTP/2 and TLS")
```

Sources/ApodiniMigration/ApodiniMigratorInterfaceExporter.swift
```diff
@@ -149,7 +149,7 @@ final class ApodiniMigratorInterfaceExporter: InterfaceExporter {
         }
 
         if let hostName = hostName, let port = port {
-            let serverPath = "http\(app.http.tlsConfiguration != nil ? "s" : "")://\(hostName):\(port)"
+            let serverPath = "http\(app.httpConfiguration.tlsConfiguration != nil ? "s" : "")://\(hostName):\(port)"
```
PSchmiedmayer: We can probably use the `uriPrefix` of the `HTTPConfiguration` here.

Sources/Apodini/Configurations/HTTPConfiguration.swift
```diff
@@ -8,75 +8,96 @@
...
+                httpProtocol = "http://"
+                if hostname.port != 80 {
+                    port = ":\(hostname.port)"
+                }
+            } else {
+                httpProtocol = "https://"
+                if hostname.port != 443 {
+                    port = ":\(hostname.port)"
+                }
             }
```
PSchmiedmayer: Please one level of indentation here.
```suggestion
    public var uriPrefix: String {
        let httpProtocol: String
        var port = ""
        
        if self.tlsConfiguration == nil {
            httpProtocol = "http://"
            if hostname.port != 80 {
                port = ":\(hostname.port)"
            }
        } else {
            httpProtocol = "https://"
            if hostname.port != 443 {
                port = ":\(hostname.port)"
            }
        }
```

Sources/Apodini/Configurations/HTTPConfiguration.swift
```diff
@@ -8,75 +8,96 @@
...
+        }
+        
+        if let bindAddress = bindAddress {
+            switch bindAddress {
+            case let .interface(address, port):
+                if port != nil {
+                    self.bindAddress = bindAddress
+                } else {
+                    self.bindAddress = .interface(address, port: defaultPort)
+                }
```
PSchmiedmayer: If you want you can take advantage of the power of Swift switch statements here: 
```suggestion
            switch bindAddress {
            case let .interface(address, nil):
                self.bindAddress = .interface(address, port: defaultPort)
            case .interface:
                self.bindAddress = bindAddress
```

Same apples to line 82 and following below

PSchmiedmayer (approved):
@ridvan-cln Thanks for the improvements! The PR looks good to me 🚀

There are only a few small comments that I have that should be addressed but once they are done we should be good to go to merge the PR. Be sure to merge the develop branch in the feature branch and fix any failing tests and then you can merge the PR.

PSchmiedmayer (approved):

PR #371, opened 2021-10-26
Replace Vapor with ApodiniNetworking

lukaskollmer:
# Replace Vapor with ApodiniNetworking

This PR removes Vapor --- throughout the entirely of Apodini, i.e. the "core" targets, as well as all interface exporters and all tests --- from the project, and replaces it with a custom HTTP server built on top of Swift-NIO.

Replacing Vapor with a custom HTTP component allows us greater flexibility and control when interacting with HTTP-based requests. Since we were using only a small subset of Vapor's functionality to begin with (essentially just the HTTP routing, their testing APIs, and, in the case of the WebSocketInterfaceExporter, the HTTP-to-WebSocket upgrade mechanism), and most of Apodini was already in a way that the concrete details of dealing with HTTP interactions were abstracted away from directly using the respective Vapor APIs, this change slims down Apodini by removing unnecessary dependencies and unifying HTTP-related behaviour across all targets and interface exporters.

More specifically, ApodiniNetworking replaces all Vapor components and APIs we were using (e.g.: a HTTP server, the ability to register routes on that server, the ability to define these individual routes' responders, access to raw request and response objects, etc) with somewhat more "low-level" APIs, insofar as that many things which previously were abstracted away by Vapor using APIs we weren't using are now replaced by more bare-bones  (and therefore also more flexible) APIs, which should also allow Apodini better control over how exactly specifics of HTTP interactions are handled.
Example: Vapor defined a pretty expansive API for accessing request or response content, via a set of custom decoders and encoders, etc. Apodini wasn't using any of that, but instead always directly accesssed the underlying body storage. We now always expose direct access to a request's or response's raw bytes, and can simply pipe that through to Apodini's en/decoding facilities. This more direct access also enables things like the below-mentioned HTTP/2 streaming responses.

Looking into the future, this more direct access to the underlying HTTP happenings (particularly things like HTTP/2 connections and streams) will be useful when (re-)implementing the gRPC and GraphQL interface exporters.

## :gear: Release Notes 
- Removed the `ApodiniVaporSupport` target, which means that `app.vapor.app` is no longer a thing. If you want to register custom HTTP routes, you can import the `ApodiniNetworking` target and use the `app.httpServer` property to access ApodiniNetworking's HTTP server.
So, what previously was
```swift
app.vapor.app.get("path") { req in res }
```
now becomes
```swift
app.httpServer.registerRoute(.GET, "path") { req in res }
```

I tried to limit this PR to only the changes required to replace Vapor, but there's a couple of additional things to note:
- New Features
  - The HTTP interface exporter now supports proper service-side-streaming HTTP, instead of the previous behaviour which consisted of simply waiting until all responses were sent by the handler, and then returning them all at once to the client.
  - This also means that service-side-streaming HTTP endpoints will no longer return a JSON-encoded array of objects of the handler's response type (e.g. a `[String]` for a streaming handler returning `String`s), but instead simply return a byte stream consisting of the (individually encoded) responses received from the handler)
  - Improved HTTP/2 support. Enabling TLS will now implicitly also enable HTTP/2. This kinda was sometimes already the case before, but now it's explicit and we have proper handling of it.
  - Adds type-safe APIs for defining and working with HTTP headers. This is pretty nice in fact, because it allows us to interpret the (at HTTP-header-level always string-based) header values using their respective native swift types, and it also allows us to differentiate between single-value and multi-value headers. (E.g. when accessing a header which is defined in the spec as only ever having at most one value, you won't get an array anymore but instead just that one value).
    - This means that we now also have two (currently separate) implementations of MIME types (or, to use the correct term, media types). At some point we probably should merge the two (the existing one is only ever used in one place, to define a `Blob` response's content type), but that isn't part of this PR since there's some dependency conflicts between the two targets involved.
- Changes:
  - We no longer allow registering multiple routes for the same HTTP Method / path combination. This was possible in the past because Vapor allowed it, but it doesn't make sense since there is no way to influence which of the multiple routes registered for a single endpoint should in fact get invoked when handling a request. This mainly is an issue when enabling the REST and HTTP interface exporters simultaneously, since these two will both register a web service's handlers onto the same paths.
- Removed Features
  - Removed FCM support from the ApodiniNotifications target, since that doesn't work w/out Vapor (this change has @PSchmiedmayer's blessing)
  - Temporarily broke the AWS Lambda Deployment Provider, since the "Lambda Events ↔ Apodini HTTP Server" integration was implemented entirely on the assumption that the underlying `Apodini.Application` would be using a vapor-based http server. This functionality will be restored in a future PR.

### Related PRs
#370 will probably be affected by this, since I rewrote a decent chunk of the HTTP-config-related code

### Testing
All tests were migrated from using the Vapor-based testing facilities (i.e. the stuff defined in XCTVapor) to using new ApodiniNetworking-specific testing facilities. These are pretty similar to what Vapor offered, meaning that the general structure of Apodini's test cases remains unaffected.
There are currently no test cases for explicitly testing ApodiniNetworking's functionality, but since for essentially all of the WebService- and InterfaceExporter-tests succeeding is contingent on the HTTP server functioning as expected, this isn't a problem (i.e. if the HTTP server doesn't work as expected, the other tests, which rely on the HTTP server, won't pass either).

### State of the PR
This isn't 100% complete yet, but it's complete enough that it can be reviewed. There's like a handful of TODOs currently still in the code, all of which will are small and trivial changes and will be resolved shortly, so that shouldn't impede on reviewing the general implementation.

### Reviewer Nudging
Essentially all of the new code (relating to the networking APIs) is located in the `Sources/ApodiniNetworking` target, so I guess that's a good starting point.
One specific thing I'd like explicit feedback on, because I'm not quite sure whether it's a good implementation, or whether it behaves the way we want it to in the first place, is in ServiceSideStreaming.swift, the implementation for handling service-side streams. It works, but I feel like it shouldnt? (I don't have a ton of experience with the new concurrency stuff, so I can't really judge whether this is a good / good enough implementation)


Sources/ApodiniNetworking/URI.swift
```diff
@@ -0,0 +1,123 @@
...
+//
+// SPDX-License-Identifier: MIT
+//
+
+import Foundation
+import ApodiniUtils
+
+
+/// A URL
+public struct URI: LosslessStringConvertible {
```
moritzsternemann: URL or URI? 😜

Sources/ApodiniNetworking/URI.swift
```diff
@@ -0,0 +1,123 @@
...
+        self.path = path
+        self.rawQuery = rawQuery.hasPrefix("?") ? String(rawQuery.dropFirst()) : rawQuery
+        self.fragment = fragment
+        self.queryItems = .init(
+            uniqueKeysWithValues: (URLComponents(string: "?\(self.rawQuery)")?.queryItems ?? []).map { ($0.name, $0.value) }
+        )
+    }
+    
+    public init?(string: String) {
+        guard let nsUrl = URL(string: string) else {
```
moritzsternemann: it's not a `NSURL` 🤨

Sources/ApodiniNetworking/URI.swift
```diff
@@ -0,0 +1,123 @@
...
+        }
+        if !fragment.isEmpty {
+            retval.append("#\(fragment)")
+        }
+        return retval
+    }
+    
+    
+    /// Constructs, from this `URL`, a `Fondation.URL`
+    public func toNSURL() -> URL {
```
moritzsternemann: Again, not a `NSURL`.

Sources/ApodiniNetworking/MediaType.swift
```diff
@@ -0,0 +1,139 @@
...
+    /// gRPC media type subtype suffix options
+    public enum GRPCEncodingOption: String {
+        case proto = "proto"
+        case json = "json"
+    }
+    
+    /// Creates a gRPC media type with the specified encoding appended as the subtype's suffix
+    public static func gRPC(_ encoding: GRPCEncodingOption) -> HTTPMediaType {
+        HTTPMediaType(type: "application", subtype: "grpc+\(encoding.rawValue)")
+    }
```
moritzsternemann, line 138: I don't know that much about Apodini's modularization yet, but I think maybe this gRPC-specific stuff shouldn't be in the "general HTTP module" but in the `ApodiniGRPC` module instead?

Sources/ApodiniNetworking/Utils.swift
```diff
@@ -0,0 +1,87 @@
...
+    /// Reads all data currently in the byte buffer, without moving the reader index (i.e. non-consuming).
+    public func getAllData() -> Data? {
+        self.getData(at: 0, length: self.writerIndex)
+    }
+}
+
+
+extension HTTPResponseStatus {
+    /// Creates a `Vapor``HTTPStatus` based on an `Apodini` `Status`.
+    /// - Parameter status: The `Apodini` `Status` that should be transformed in a `Vapor``HTTPStatus`
```
moritzsternemann:
```suggestion
    /// Creates a `HTTPResponseStatus` based on an `Apodini` `Status`.
    /// - Parameter status: The `Apodini` `Status` that should be transformed in a `HTTPResponseStatus`
```

Sources/ApodiniNetworking/Utils.swift
```diff
@@ -0,0 +1,87 @@
...
+        case .redirect:
+            self = .seeOther
+        }
+    }
+}
+
+
+extension HTTPMethod {
+    /// Creates a `Vapor``HTTPMethod` based on an `Apodini` `Operation`.
+    /// - Parameter operation: The `Apodini` `Operation` that should be transformed in a `Vapor``HTTPMethod`
```
moritzsternemann:
```suggestion
    /// Creates a `HTTPMethod` based on an `Apodini` `Operation`.
    /// - Parameter operation: The `Apodini` `Operation` that should be transformed in a `HTTPMethod`
```

Sources/ApodiniNetworking/Utils.swift
```diff
@@ -0,0 +1,87 @@
...
+        case .delete:
+            self =  .DELETE
+        }
+    }
+}
+
+
+extension HTTPHeaders {
+    /// Creates a `Vapor``HTTPHeaders` instance based on an `Apodini` `Information` array.
+    /// - Parameter information: The `Apodini` `Information` array that should be transformed in a `Vapor``HTTPHeaders` instance
```
moritzsternemann:
```suggestion
    /// Creates a `HTTPHeaders` instance based on an `Apodini` `Information` array.
    /// - Parameter information: The `Apodini` `Information` array that should be transformed in a `HTTPHeaders` instance
```

Package.swift
```diff
@@ -107,7 +104,9 @@ let package = Package(
         .package(url: "https://github.com/Apodini/ApodiniMigrator.git", .upToNextMinor(from: "0.1.0")),
 
         // TypeInformation
-        .package(url: "https://github.com/Apodini/ApodiniTypeInformation.git", .upToNextMinor(from: "0.2.0"))
+        .package(url: "https://github.com/Apodini/ApodiniTypeInformation.git", .upToNextMinor(from: "0.2.0")),
+        .package(url: "https://github.com/swift-server/async-http-client.git", from: "1.0.0"),
```
bauer-andreas: Might add a empty line so the `// TypeInformation` line isn't incorrect interpreted to apply to those two lines as well.

Sources/Apodini/Configurations/Configuration.swift
```diff
@@ -71,14 +71,10 @@ public struct EmptyCommand: ParsableCommand {
 
 extension Array: Configuration where Element == Configuration {
     public func configure(_ app: Application) {
-        forEach {
-            $0.configure(app)
-        }
+        forEach { $0.configure(app) }
```
bauer-andreas, line 74: Maybe a nitty, but I'm not sure, therefore I ask. Is this in line with our style guide? Same below.

Sources/Apodini/Semantic Model Builder/Model/Endpoint.swift
```diff
@@ -59,12 +59,8 @@ public struct Endpoint<H: Handler>: _AnyEndpoint {
     }
     
     public subscript<S>(_ type: S.Type) -> S where S: KnowledgeSource {
-        get {
-            self.blackboard[type]
-        }
-        nonmutating set {
-            self.blackboard[type] = newValue
-        }
+        get { blackboard[type] }
```
bauer-andreas, line 62: Same as above, just for the record.

Sources/ApodiniOpenAPI/Document/OpenAPIComponentsObjectBuilder.swift
```diff
@@ -11,7 +11,7 @@ import Apodini
 import ApodiniREST
 import ApodiniUtils
 import ApodiniTypeInformation
-import ApodiniVaporSupport
+//import ApodiniVaporSupport
```
bauer-andreas: Overlooked?

Sources/ApodiniREST/RESTInterfaceExporter.swift
```diff
@@ -90,25 +96,33 @@ final class RESTInterfaceExporter: InterfaceExporter, TruthAnchor {
         let absolutePath = endpoint.absoluteRESTPath
         absolutePath.build(with: &pathBuilder)
 
-        let routesBuilder = pathBuilder.routesBuilder(app)
+        //let routesBuilder = pathBuilder.routesBuilder(app)
```
bauer-andreas: forgotten?

Sources/ApodiniREST/RESTInterfaceExporter.swift
```diff
@@ -70,15 +73,18 @@ extension REST {
...
     
     /// Initialize `RESTInterfaceExporter` from `Application`
     init(_ app: Apodini.Application,
-         _ exporterConfiguration: REST.ExporterConfiguration = REST.ExporterConfiguration()) {
-        self.app = app.vapor.app
-        self.configuration = REST.Configuration(app.vapor.app.http.server.configuration)
+         _ exporterConfiguration: REST.ExporterConfiguration = REST.ExporterConfiguration()
+    ) {
+        self.app = app
+        //self.app = app.vapor.app
```
bauer-andreas: forgotten?

Sources/ApodiniREST/RESTInterfaceExporter.swift
```diff
@@ -70,15 +73,18 @@ extension REST {
 final class RESTInterfaceExporter: InterfaceExporter, TruthAnchor {
     static let parameterNamespace: [ParameterNamespace] = .individual
     
-    let app: Vapor.Application
+    //let app: Vapor.Application
```
bauer-andreas: forgotten?

Sources/ApodiniREST/RESTInterfaceExporter.swift
```diff
@@ -146,7 +162,8 @@ final class RESTInterfaceExporter: InterfaceExporter, TruthAnchor {
         }
         
         // Set option to activate case insensitive routing, default is false (so case-sensitive)
-        self.app.routes.caseInsensitive = self.exporterConfiguration.caseInsensitiveRouting
+        //self.app.routes.caseInsensitive = self.exporterConfiguration.caseInsensitiveRouting
```
bauer-andreas: forgotten?

Sources/ApodiniREST/RESTPathBuilder.swift
```diff
@@ -26,11 +27,11 @@ struct RESTPathBuilder: PathBuilder {
...
     mutating func append<Type>(_ parameter: EndpointPathParameter<Type>) {
-        pathComponents.append(.parameter(parameter.pathId))
+        pathComponents.append(.namedParameter(parameter.pathId))
         pathString.append("{\(parameter.name)}")
     }
 
-    func routesBuilder(_ app: Vapor.Application) -> Vapor.RoutesBuilder {
-        app.routes.grouped(pathComponents)
-    }
+//    func routesBuilder(_ app: Vapor.Application) -> Vapor.RoutesBuilder {
```
bauer-andreas: forgotten?

Sources/ApodiniREST/ResponseContainer.swift
```diff
@@ -56,27 +59,30 @@ public struct ResponseContainer: Encodable, ResponseEncodable {
...
-    public func encodeResponse(for request: Vapor.Request) -> EventLoopFuture<Vapor.Response> {
-        let response = Vapor.Response()
-        response.headers = HTTPHeaders(information)
+    public func encodeResponse(for request: HTTPRequest) -> EventLoopFuture<HTTPResponse> {
+        let response = HTTPResponse(
+            version: request.version,
+            status: .ok,
+            headers: HTTPHeaders(information)
+        )
+        //response.headers = HTTPHeaders(information)
```
bauer-andreas: forgotten?

Sources/ApodiniREST/ResponseContainer.swift
```diff
@@ -56,27 +59,30 @@ public struct ResponseContainer: Encodable, ResponseEncodable {
...
+            response.status = containsNoContent ? .noContent : .ok
         }
         
         do {
             if !containsNoContent {
-                try response.content.encode(self, using: self.encoder)
+                var buffer = ByteBuffer()
+                try self.encoder.encode(self, to: &buffer, headers: &response.headers)
+                response.bodyStorage = .buffer(buffer)
+                    //try response.content.encode(self, using: self.encoder)
```
bauer-andreas: forgotten?

Sources/ApodiniWebSocket/Infrastructure/ConnectionResponsible.swift
```diff
@@ -148,41 +144,31 @@ class ConnectionResponsible: Identifiable {
     
     private func processOpenMessage(from rootObject: [String: Any], retrieving context: inout UUID?) throws {
         let openMessage = try OpenContextMessage(json: rootObject)
-
```
bauer-andreas, line 151: not sure if all the empty line removals improve readability(?)

Sources/XCTApodini/Mocks/MockExporter.swift
```diff
@@ -8,8 +8,8 @@
 
 #if DEBUG || RELEASE_TESTING
 import Foundation
-import class Vapor.Application
-import class Vapor.Request
+//import class Vapor.Application
+//simport class Vapor.Request
```
bauer-andreas: forgotten?

Tests/ApodiniDeployTests/LambdaDeploymentProviderTests.swift
```diff
@@ -19,15 +20,13 @@ import SotoApiGatewayV2
...
     private var taskStdioObserverToken: AnyObject?
     
     
     func testLambdaDeploymentProvider() throws { // swiftlint:disable:this function_body_length cyclomatic_complexity
-        guard Self.shouldRunDeploymentProviderTests else {
-            print("Skipping test case '\(#function)'.")
-            return
-        }
+        throw XCTSkip() // Remove this once the lambda DP is fixed
+        //try XCTSkipUnless(Self.shouldRunDeploymentProviderTests)
```
bauer-andreas, line 29: forgotten?

Tests/ApodiniTests/ComponentTests/ComponentTests.swift
```diff
@@ -50,18 +52,14 @@ class ComponentTests: ApodiniTests {
         TestWebService().start(app: app)
         
         
-        try app.vapor.app.test(.GET, "/v1/") { res in
+        //try app.vapor.app.test(.GET, "/v1/") { res in
```
bauer-andreas: forgotten?

Tests/ApodiniTests/MigratorTests/ApodiniMigratorTests.swift
```diff
@@ -99,7 +101,8 @@ final class ApodiniMigratorTests: ApodiniTests {
         XCTAssertEqual(Self.migratorConfig.command._commandName, "migrator")
         XCTAssertEqual(Self.migratorConfig.command.configuration.subcommands.count, 3)
         
-        XCTAssert(app.vapor.app.routes.all.isEmpty)
+        //XCTAssert(app.vapor.app.routes.all.isEmpty)
```
bauer-andreas: forgotten?

Tests/ApodiniTests/MigratorTests/ApodiniMigratorTests.swift
```diff
@@ -136,9 +139,10 @@ final class ApodiniMigratorTests: ApodiniTests {
         
         start()
         
-        try app.vapor.app.test(.GET, path) { response in
+        //try app.vapor.app.test(.GET, path) { response in
```
bauer-andreas: forgotten?

Tests/ApodiniTests/DatabaseTests/DatabaseHandlerTests.swift
```diff
@@ -81,16 +78,12 @@ final class DatabaseHandlerTests: ApodiniTests {
...
-            method: .GET,
-            url: uri,
-            on: app.eventLoopGroup.next()
-        )
+        let url = URI("http://example.de/test/id")
+        let request = HTTPRequest(method: .GET, url: url, eventLoop: app.eventLoopGroup.next())
         
         let idParameter = try pathParameter(for: handler)
-        request.parameters.set("\(idParameter.id)", to: "\(birdId)")
+        //request.parameters.set("\(idParameter.id)", to: "\(birdId)")
```
bauer-andreas: forgotten?

Sources/ApodiniNetworking/URI.swift
```diff
@@ -0,0 +1,123 @@
...
+            retval.append("?\(rawQuery)")
+        }
+        if !fragment.isEmpty {
+            retval.append("#\(fragment)")
+        }
+        return retval
+    }
+    
+    
+    /// Constructs, from this `URL`, a `Fondation.URL`
```
bauer-andreas:
```suggestion
    /// Constructs, from this `URL`, a `Foundation.URL`
```

Sources/ApodiniNetworking/HTTPServer.swift
```diff
@@ -0,0 +1,402 @@
...
+                    do {
+                        sslContext = try NIOSSLContext(configuration: tlsConfig)
+                    } catch {
+                        logger.error("Unable to configure TLS: \(error)")
+                        return channel.close(mode: .all)
+                    }
+                    let tlsHandler = NIOSSLServerHandler(context: sslContext)
+                    return channel.pipeline.addHandler(tlsHandler)
+                        .flatMap { () -> EventLoopFuture<Void> in
+                            return channel.configureHTTP2SecureUpgrade { channel in
```
bauer-andreas: Swiftlint yield a `implicit_return` warning here for me 🤔

Sources/ApodiniNetworking/HTTPServer.swift
```diff
@@ -0,0 +1,402 @@
...
+                            } http1ChannelConfigurator: { channel in
+                                channel.addApodiniNetworkingHTTP1Handlers(responder: self)
+                            }
+                        }
+                        .flatMapError { error in
+                            channel.eventLoop.makeFailedFuture(error)
+                        }
+                } else {
+                    if enableHTTP2 {
+                        // NOTE this doesn't make sense and (probably) doesn't work
```
bauer-andreas: May this be an outdated comment? Otherwise, why would it not work? Will this be tested? Will this be resolved in this PR? Otherwise I would suggest to track this with an Issue providing a bit more detail about the problem situation.

Sources/ApodiniNetworking/HTTPServer.swift
```diff
@@ -0,0 +1,402 @@
...
+                        return channel.addApodiniNetworkingHTTP2Handlers(responder: self)
+                    } else {
+                        return channel.addApodiniNetworkingHTTP1Handlers(responder: self)
+                    }
+                }
+            }
+            .childChannelOption(ChannelOptions.socket(IPPROTO_TCP, TCP_NODELAY), value: 1)
+            .childChannelOption(ChannelOptions.socket(SocketOptionLevel(SOL_SOCKET), SO_REUSEADDR), value: 1)
```
bauer-andreas: First of all the `SO_REUSEADDR` seems to be duplicating the channel option above?
Secondly, why would be enable tcp no Delay and most importantly reuseaddr out of the box? The second one doesn't really make sense to me particularly. Might be beneficial in the future to provide some means of configuration of socket options?

Sources/ApodiniNetworking/ChannelHandlers/UpgradeHandler.swift
```diff
@@ -0,0 +1,149 @@
...
+import NIO
+import NIOHTTP1
+import NIOHTTP2
+import NIOWebSocket
+import WebSocketKit
+import Logging
+
+
+/// A removable channel handler which can be used to upgrade HTTP1 requests to a different protocol.
+/// Heavily insired by Vapor's equivalent of this
```
bauer-andreas:
```suggestion
/// Heavily inspired by Vapor's equivalent of this
```

bauer-andreas: Might want to check the Vapor Licensing if there are any restrictions or duties, like including the original license info.

Sources/ApodiniNetworking/HTTPRouter.swift
```diff
@@ -0,0 +1,184 @@
...
+            return getRoute(for: request, overridingMethod: .GET)
+        } else {
+            return nil
+        }
+    }
+    
+    
+    private func getRoute(for request: HTTPRequest, overridingMethod: HTTPMethod?) -> Route? {
+        guard let candidates = routes[request.method] else {
+            logger.error("[Router] Unable to find a route for \(request.method) request to '\(request.url)'")
```
bauer-andreas: not sure if we want to log on error level on every encountered 404. Same for the last logger statement in this `func`.

Sources/ApodiniNetworking/URI.swift
```diff
@@ -0,0 +1,123 @@
...
+        fragment: String = ""
+    ) {
+        self.scheme = scheme
+        self.hostname = hostname
+        self.port = port
+        self.path = path
+        self.rawQuery = rawQuery.hasPrefix("?") ? String(rawQuery.dropFirst()) : rawQuery
+        self.fragment = fragment
+        self.queryItems = .init(
+            uniqueKeysWithValues: (URLComponents(string: "?\(self.rawQuery)")?.queryItems ?? []).map { ($0.name, $0.value) }
```
bauer-andreas, line 43: Am I correct that the [Precent-Encoding](https://en.wikipedia.org/wiki/Percent-encoding) (for encoding and decoding) through Foundations API, right?

Sources/ApodiniNetworking/HTTPPathMatcher.swift
```diff
@@ -0,0 +1,171 @@
...
+        case 1:
+            return potentialMatches.first!
+        default:
+            let matchesSorted = potentialMatches.sorted(by: \.penaltyScore, ascending: true)
+            if matchesSorted[0].penaltyScore < matchesSorted[1].penaltyScore {
+                return matchesSorted[0]
+            } else {
+                fatalError("Multiple match results with same penalty.")
+            }
+            //return potentialMatches.min { $0.score < $1.score }
```
bauer-andreas: code comment

Sources/ApodiniNetworking/HTTPPathMatcher.swift
```diff
@@ -0,0 +1,171 @@
...
+    }
+    
+    
+    private enum MatchOneResult {
+        case abort
+        case successAndContinue
+        case successAndReturn
+    }
+    
+    private mutating func match() -> MatchResult? {
```
bauer-andreas, line 73: Out of interest, how does the matcher implementation compare to the Vapor one (I don't now that one). Just remember, that Vapor said about themselves that they got a pretty fast routing system. Would just be curious how those compare and how those two implementations compare performance wise.

Sources/ApodiniNetworking/HTTP/Request.swift
```diff
@@ -0,0 +1,161 @@
...
+            self.value = rawValue
+        }
+    }
+}
+
+
+/// A HTTP request
+public final class HTTPRequest: RequestBasis, Equatable, Hashable, CustomStringConvertible, CustomDebugStringConvertible {
+    struct ParametersStorage: Hashable {
+        /// The requst's parameters that are expressed via a clear key-value mapping (e.g.: explicitly named path parameters)
```
bauer-andreas: *request

Sources/ApodiniNetworking/HTTP/Request.swift
```diff
@@ -0,0 +1,161 @@
...
+        var namedParameters: [String: String] = [:]
+        /// The request's parameters that are the result from matching the request againat a ``HTTPRouter.Route`` which contained single-path-component wildcards
```
bauer-andreas: against* (same typo below).

Sources/ApodiniNetworking/HTTP/Request.swift
```diff
@@ -0,0 +1,161 @@
...
+    
+    public var debugDescription: String {
+        self.description
+    }
+    
+    
+    var loggingMetadata: [LoggingMetadataInformation] {
+         [
+            LoggingMetadataInformation(key: .init("ApodiniNetworkingRequestDescription"), rawValue: .string(self.description)),
+            //LoggingMetadataInformation(key: .init("HTTPBody"), rawValue: .string(self.bodyData.count < 32_768 ? String(decoding: self.bodyData, as: UTF8.self) : "\(String(decoding: self.bodyData, as: UTF8.self).prefix(32_715))... (further bytes omitted since HTTP body too large!")),
```
bauer-andreas: any reason those got commented out?

Sources/ApodiniNetworking/HTTP/Response.swift
```diff
@@ -0,0 +1,192 @@
...
+    /// This response's HTTP version
+    public var version: HTTPVersion
+    /// The HTTP status code to be sent alongside this response
+    public var status: HTTPResponseStatus
+    /// The HTTP headers to be sent alongside this response
+    public var headers: HTTPHeaders
+    /// This response's body. This can be either a buffer-, or a stream-based body, see the `BodyStorage` type for more info.
+    public var bodyStorage: BodyStorage
+    /// Whether the request handler should ignore the value in the `version` property, and instead use the request's version.
+    /// This option should be avoided if possible, and it really just here to accomodate very specific edge cases in which a response
```
bauer-andreas:
```suggestion
    /// This option should be avoided if possible, and it really just here to accommodate very specific edge cases in which a response
```

Sources/ApodiniNetworking/HTTP/Headers.swift
```diff
@@ -0,0 +1,498 @@
...
+public protocol __ANNIOHTTPHeadersType {
+    /// Initialises the headers type from a sequence of key-value pairs
+    init(_ headers: [(String, String)])
+    /// Checks whether at least one header entry exists for the specified key
+    func contains(name: String) -> Bool
+    /// Adds a header entry
+    mutating func add(name: String, value: String, indexing: HPACKIndexing)
+    /// Removes a header entry
+    mutating func remove(name: String)
+    /// Adds a header entry, removing any existing enties with the same key if necessary
```
bauer-andreas:
```suggestion
    /// Adds a header entry, removing any existing entities with the same key if necessary
```

Sources/ApodiniNetworking/HTTP/Headers.swift
```diff
@@ -0,0 +1,498 @@
...
+
+
+extension __ANNIOHTTPHeadersType {
+    /// Creates a new headers struct, giving the caller the opportunity to initialise it via the closure.
+    /// - Note: The reason this initialiser exists is to offer a type-safe way of declaring immutable headers objects.
+    ///         The underlying problem here is that Swift doesn't support variadic generics, meaning that (since the type-safe header
+    ///         names are implemented via generics) you wouldn't be able to pass more than one generic key-value pair to the initialiser.
+    ///         This initialiser wors around that by giving the caller the ability to access a mutable version of the headers
+    ///         struct (which can be modified using the type-safe API), while still allowing the caller to store the resulting value into an immutable object.
+    ///         Sinice the block is non-escaping, immediately evaluated, and evaluated only once, this is semantically equivalent to declaring a mutable headers object and modifying that.
```
bauer-andreas:
```suggestion
    /// - Note: The reason this initializer exists is to offer a type-safe way of declaring immutable headers objects.
    ///         The underlying problem here is that Swift doesn't support variadic generics, meaning that (since the type-safe header
    ///         names are implemented via generics) you wouldn't be able to pass more than one generic key-value pair to the initializer.
    ///         This initializer works around that by giving the caller the ability to access a mutable version of the headers
    ///         struct (which can be modified using the type-safe API), while still allowing the caller to store the resulting value into an immutable object.
    ///         Since the block is non-escaping, immediately evaluated, and evaluated only once, this is semantically equivalent to declaring a mutable headers object and modifying that.
```

Sources/ApodiniNetworking/HTTP/Headers.swift
```diff
@@ -0,0 +1,498 @@
...
+        self.init([])
+        try block(&self)
+    }
+    
+    /// Checks whether at least one entry exists for the specified name
+    public func isPresent<T>(_ name: HTTPHeaderName<T>) -> Bool {
+        self.contains(name: name.rawValue)
+    }
+    
+    /// Sets the spefified value for `name`, if no entry for that key already exists
```
bauer-andreas:
```suggestion
    /// Sets the specified value for `name`, if no entry for that key already exists
```

Sources/ApodiniNetworking/HTTP/Headers.swift
```diff
@@ -0,0 +1,498 @@
...
+    /// - Throws: When reading: if the field is present, but there was an error decoding its value. When writing: if there was an error encoding the new value
+    /// /// - Note: Using this subscript to set HTTP2 header values will default the HPACKIndexable property to .indexable. If you don't want this, use the `add` or `replaceOrAdd` functions
+    public subscript<T: HTTPHeaderFieldValueCodable>(name: HTTPHeaderName<[T]>) -> [T] {
+        get {
+            switch name {
+            case .setCookie:
+                // Set-Cookie requires special handling because we can't split or join its values in a comma separator
+                return self[name.rawValue].map { T(httpHeaderFieldValue: $0)! }
+            default:
+                // All other headers can be split on commmas
```
bauer-andreas:
```suggestion
                // All other headers can be split on commas
```

Sources/ApodiniNetworking/HTTP/Headers.swift
```diff
@@ -0,0 +1,498 @@
...
+        case .bearer(let token):
+            return "Bearer \(token)"
+        case let .other(type, credentials):
+            return "\(type) \(credentials)"
+        }
+    }
+}
+
+
+public enum ETagHTTPHeaderValue: HTTPHeaderFieldValueCodable {
```
bauer-andreas, line 414: Those typed implementations of Header values currently duplicated equivalent ones of those implemented via the `Information` system. See `ApodiniHTTPProtocol/Information` (e.g. the `ETag.swift`). Might make sense to rely on a single implementation for those(?).

Sources/ApodiniHTTP/InterfaceExporter.swift
```diff
@@ -54,57 +53,71 @@ public final class HTTP: Configuration {
...
     
     init(_ app: Apodini.Application, _ configuration: HTTP.ExporterConfiguration) {
         self.app = app
         self.configuration = configuration
         self.logger = app.logger
-        
-        // Set option to activate case insensitive routing, default is false (so case-sensitive)
-        self.app.vapor.app.routes.caseInsensitive = configuration.caseInsensitiveRouting
+        // Note: a thing to consider here is that we might want to have per-route case sensitivity.
+        // For example, if the REST IE wants case-insensitive routing, but the HTTP IE does not, that'd somehow need to be supported.
```
bauer-andreas, line 64: Is this currently not supported, right?

Sources/ApodiniHTTP/ClosureBuilders/HTTP1Compatibility/ServiceSideStreaming.swift
```diff
@@ -9,24 +9,70 @@
...
+        using defaultValues: DefaultValueStore
+    ) -> (HTTPRequest) throws -> EventLoopFuture<HTTPResponse> where H.Response.Content == Blob {
+        _buildServiceSideStreamingClosure(for: endpoint, using: defaultValues) { (response: Blob?) -> Data? in
+            precondition(response?.byteBuffer.readerIndex == 0)
+            return response?.byteBuffer.getAllData()
+        }
+    }
+    
+    
+    private func _buildServiceSideStreamingClosure<H: Handler>(
```
bauer-andreas, line 64: I'm also not really knowledgeable on the core of Apodinis Exporter API and async stuff. But, to my limited knowledge, this looks fine. Not sure if this helps at all 😅

bauer-andreas (approved):
To my best effort, this PR looks good to me (assuming the few ToDos still present get fixed accordingly) 👍 great work there 🚀 

A good majority of my comments are about so spelling and styling 😛 maybe its worth to enable the spell checker in Xcode some time 🙃
I looked at the `ServiceSideStreaming.swift` as you requested and it looks good and as expected to me. Though, as mentioned below, I'm also not the most knowledgeable on that topic.

Feel free to reach out if you have a different opinion on individual points.

Sources/ApodiniNetworking/HTTP/Headers.swift
lukaskollmer: You're right, same also applies to the MediaType stuff, we have (with this PR, but also before) two implementations of that: we were always using Vapor's implementation for things like setting Content-Type headers for encoded output, but then also had a custom implementation which is used exclusively for specifying a `Blob`'s content type.

I'm planning on merging these two, and would then also merge the ETag stuff, although I'd like this to be a separate PR since it'll most likely require making changes to the package structure.

Sources/ApodiniNetworking/HTTP/Headers.swift
lukaskollmer: Yeah sry for all of these. I'm using one of the macbooks w/ the broken keyboards so there's missing or duplicated letters all over the place and I sadly don't always catch them all

bauer-andreas: classic 😅

Sources/ApodiniNetworking/HTTPServer.swift
lukaskollmer: Nope, very much not outdated. Basically no one supports plaintext HTTP/2 (which is what you'd end up configuring in that branch). Technically speaking, plaintext h2 is a thing, but there seems to be a consensus that everyone refuses to implement it (see eg also Nginx and (iirc) Chrome).

[Previously, this would've been a crash](https://github.com/vapor/vapor/blob/a8bd13fa10b14d2c014f002a4eed0558f66bcf6f/Sources/Vapor/HTTP/Server/HTTPServer.swift#L369). If you think that's a more appropriate response we can also add a fatal error here, I'd be up for that.

Sources/ApodiniWebSocket/Infrastructure/ConnectionResponsible.swift
lukaskollmer: I only removed lines where the two lines surrounding the removed line either were directly related to each other (eg: defining a `Data` variable in one line, and then trying to decode some type from it in the next), or in cases where the blocks of code previously separated by the now-removed consisted of statements that would also indicate a visual separation between the blocks (eg: removing a blank line between an if statement (which terminates with a line consisting only of a single character) and some other block. in that case the almost-empty line will (IMO) still be good enough in terms of readability to indicate that these are two separate blocks)

Tests/ApodiniDeployTests/LambdaDeploymentProviderTests.swift
lukaskollmer: Nope. As mentioned in the PR description, the lambda DP is currently broken, hence we have to disable the test.
(This will be fixed in a later PR, since it's not that trivial, but @PSchmiedmayer agreed that it'd be ok to commit the temporarily not-working DP)

Sources/ApodiniNetworking/ChannelHandlers/UpgradeHandler.swift
lukaskollmer: @PSchmiedmayer what's your take on this, wrt the Licensing?

Generally speaking, I didn't copy any code from Vapor, but at the end of the day there's only so many ways you can implement a HTTP server so there's bound to be some overlap.
In this specific case I did in fact (to a certain extent, mainly in terms of handling the protocol upgrade w/in the NIO handler) base my implementation on Vapor's. How should this be indicated?

Sources/ApodiniNetworking/URI.swift
lukaskollmer: That was my assumption, yes. I would've expected the tests to fail if it wasn't.

Sources/ApodiniNetworking/HTTPRouter.swift
lukaskollmer: good point, will change that

Sources/ApodiniNetworking/HTTPServer.swift
lukaskollmer: Yeah that's a leftover from a previous version of the code. I think that it might be a good idea to expose these through a config option, either at interface exporter level (i.e. that IEs can set this) or even at client level (i.e. that a WebServer could set this)

Sources/ApodiniNetworking/HTTPPathMatcher.swift
lukaskollmer: I haven't done any performance testing, but my gut response would be that it's probably worse.
There's some ways this probably could be improved (somehow caching past routing results comes to mind), but we're currently not doing any of that.

Sources/ApodiniHTTP/InterfaceExporter.swift
lukaskollmer: In the Vapor-based version? No, not that I'm aware. There's only one Vapor application, and if multiple IEs set a property on that application's http config, the last one wins.

Sources/Apodini/Configurations/Configuration.swift
lukaskollmer: SwiftLint didn't complain so I'd assumed it'd be fine. (tbh if this were up to me we'd be using classic `for` loops instead of `forEach` where ever possible, but I'm not gonna make that change since it's not at all up to me to make that decision)

Sources/ApodiniNetworking/HTTP/Request.swift
lukaskollmer: Added some of these (HTTP version, content type) back.
The main issue here is that we simply don't have access to some of this information anymore. Or, rather, we can't access it the same way as Vapor would let us. For example, we now have a much clearer separation between buffer- and stream-based request bodies, which means that we can't just "get some bytes" to display in the logging metadata.

Sources/ApodiniNetworking/HTTPPathMatcher.swift
lukaskollmer: @Supereg addendum: routing performance should depend primarily on whether or not you're using routes w/ multi-component wildcards. if you are not, it's at worst O(n) w.r.t. the number of routes registered for the request's HTTP method (and, in the case of HEAD requests, also the number of GET routes registered[^1]). there are, of course, early exits for cases where a URL obviously doesn't match a route, see the non-private function above, but at worst it'll have to compare all path components in the URL and the route. 

for routes with multi-component wildcards, performance is by definition significantly worse (compared to my non-wildcard matching, not compared to Vapor, although that probably as well). the reason being that there's multiple ways a URL could be matched onto a multi-component wildcard, so we have to evaluate all of them and then compare the results to determine which mapping is the best. though afaik we currently don't allow `@Parameters` that world map to (multi-component) wildcards, and also don't use them internally in Apodini. 


[^1]: this is something i added to match vapor's behavior: if there's an incoming HEAD request but no matching route exists we also look for a GET route that might match the request and use that if possible. what's your opinion on this? do you think this is desirable to have?

Sources/ApodiniNetworking/HTTPServer.swift
moritzsternemann: I agree with @lukaskollmer here. HTTP/2 without TLS is not really a thing and I think this should throw an error or even crash the service.

Sources/ApodiniExtension/AsyncSequenceHelpers/ConvenienceFunctions.swift
```diff
@@ -39,7 +39,7 @@ public extension AsyncSequence {
         promise.completeWithAsync {
             try await self.first(where: { _ in true })
         }
-    
+
```
moritzsternemann: nit: Don't think this change is according to the style guide, is it? There are also a few other instances of empty lines that contain spaces. Maybe we can automatically format this as well.

Sources/ApodiniWebSocket/WebSocketInterfaceExporter.swift
```diff
@@ -41,36 +42,31 @@ final class WebSocketInterfaceExporter: LegacyInterfaceExporter {
...
         let factory = endpoint[DelegateFactory<H, WebSocketInterfaceExporter>.self]
-        
-        self.router.register({(clientInput: AnyAsyncSequence<SomeInput>, eventLoop: EventLoop, request: Vapor.Request) -> (
+        self.router.register({(clientInput: AnyAsyncSequence<SomeInput>, eventLoop: EventLoop, request: HTTPRequest) -> (
                     defaultInput: SomeInput,
                     output: AnyAsyncSequence<Message<H.Response.Content>>
                 ) in
             // We need a new `Delegate` for each connection
             let delegate = factory.instance()
+            print("WS request", request, clientInput)
```
moritzsternemann: I think this can be removed, looks like a debug print 🤔

moritzsternemann (approved):
First of all, I think this is a super exciting improvement of Apodini! 🚀 

The changes in this PR look good and I especially like the HTTP headers implementation 🤩 
I also took a thorough look at `ServiceSideStreaming.swift` and don't see any problems with your approach (though I don't have much experience with Swift Structured Concurrency yet as well).

Sources/ApodiniNetworking/URI.swift
lukaskollmer: Fun fact: [it actually is, under the hood](https://github.com/apple/swift-corelibs-foundation/blob/main/Sources/Foundation/URL.swift#L435). You're right, though, and I'll rename it to just `url`.

Sources/ApodiniNetworking/URI.swift
lukaskollmer: I intentionally called the type `URI` instead of `URL` to avoid naming ambiguities with Foundation's URL type. IIRC a URL and a URI are effectively the same thing; will update the comment though.

(The reason why this type exists in the first place it that Foundation's URL type implements a lot of unnecessary functionality we're not interested in, and also because the URL type defines all of its properties as optional, whereas we want things scheme, hostname, or path to be non-optional.)

Sources/ApodiniNetworking/MediaType.swift
lukaskollmer: I'm currently reworking the gRPC stuff so in my other local branch this has in fact already been moved to a different target.

Sources/ApodiniNetworking/URI.swift
moritzsternemann: Sounds good 👍

moritzsternemann (approved):
Everything resolved, looks good to me 👍

Sources/ApodiniNetworking/ChannelHandlers/UpgradeHandler.swift
PSchmiedmayer: @lukaskollmer @Supereg Yes, I would suggest to add a reference to the Vapor License similar to how we did it in the Application+... extensions. I don't think there is any harm in giving credit to other open-source work and even if you did not direly copy code I would somehow consider this an "indirect citation" and think we can credit the Vapor authors for doing thorough the struggles of implementing the functionality first and giving us a change to use their implementation as a reference.

Sources/ApodiniNetworking/HTTPPathMatcher.swift
PSchmiedmayer: Nice explanation. And I didn't know you could use footnotes this way in md, very nice, learned two things here 😅

Sources/ApodiniNetworking/HTTP/Headers.swift
PSchmiedmayer: I agree with @Supereg here, would be good to merge these implementation. A separate PR is fine with me as well, we should nevertheless document all open issues (including this one) in GitHub Issues so we don't forget about them 👍

PSchmiedmayer:
Thanks a lot @lukaskollmer, a great addition to Apodini! 🚀

PR #379, opened 2021-11-16
Remove the `ServiceType` type in favour of `CommunicationalPattern`

### LINKED ISSUE #377: [INTERNAL CHANGE]: Duplicate implementation of communicational pattern metadata
### Problem

There are currently two seemingly independent definitions and implementations of metadata types specifying whether a handler is a unary, client-side-streaming, service-side-streaming, or bidirectional handler.

https://github.com/Apodini/Apodini/blob/3738aa97daa1ceac2c8b6979c996e3b86e6eaedc/Sources/Apodini/Metadata/BuiltinMetadata/ServiceTypeMetadata.swift#L11-L22

https://github.com/Apodini/Apodini/blob/3738aa97daa1ceac2c8b6979c996e3b86e6eaedc/Sources/Apodini/Metadata/BuiltinMetadata/CommunicationalPatternMetadata.swift#L11-L24

It seems like the first one is used by the (old) gRPC interface exporter and the OpenAPI interface exporter, whereas the second one is used by the HTTP interface exporter (and potentially also other places, via the knowledge source stuff).

This is a problem insofar as:
1. it's generally a bad idea to have multiple implementations of the same thing, and
2. having two separate types means that a web service which uses one of the two will be exported using different sevice types by different interface exporters. (e.g.: if a handler uses `ServiceType(.bidirectional)`, the gRPC interface exporter will export the handler as a bidirectional service, but the HTTP interface exporter will look for the `CommunicationalPattern` metadata field, which it won't find and then use `.unary` as a default value.)

### Solution

Ideally, there would be only one such definition, and that one definition would be used throughout all of Apodini

### Additional context

n/a

lukaskollmer:
Fixes #377.

Removes the `ServiceType` type in favour of `CommunicationalPattern`Remove the `ServiceType` type in favour of `CommunicationalPattern`.

Short summary: there are currently two separate types for defining whether a handler is an unary/bidirectional/etc endpoint.
This PR removes one of these types and replaces all uses with the other one.

See the associated issue (#377) for more info.

<details>
<summary>Migrating existing handlers</summary>

A handler which previously had its communicational pattern expressed using the `.serviceType` modifier:

```swift
struct MyWebService: WebService {
    var content: some Component {
        SomeHandler()
            .serviceType(.bidirectional)
    }
}
```

should now use the `.pattern` modifier:

```swift
struct MyWebService: WebService {
    var content: some Component {
        SomeHandler()
            .pattern(.bidirectionalStream)
    }
}
```

The mapping for the enum case names is as follows:
| ServiceType (old) | CommunicationalPattern (new) |
| :----: | :----: |
| `unary` | `requestResponse` |
| `clientStreaming` | `clientSideStream` |
| `serviceStreaming` | `serviceSideStream` |
| `bidirectional` | `bidirectionalStream` |
</details>

### :gear: Release Notes 
- Removed redundant types

### Testing
not affected


Sources/Apodini/Apodini.docc/Advanced/Semantic Model.md
```diff
@@ -174,7 +173,7 @@ public required init<B>(_ blackboard: B) throws where B: Blackboard {
 - ``EndpointSource``
 - ``AnyEndpointSource``
 - ``Blackboards``
-- `Context`
+- ``Context``
```
bauer-andreas: Context doesn’t reside in the main target anymore (instead as far as I remember, it’s now part of the MetadataSystem repo). Therefore this has single Ticks, as reference cannot be resolved cross targets/projects

bauer-andreas (approved):
Thanks 👍

bauer-andreas:
Might want to add a migration guide e.g. showing how to migrate the Handler modifiers.

Sources/Apodini/Apodini.docc/Advanced/Semantic Model.md
lukaskollmer: Ah ok didn't know this. I assumed this was just a typo. Will fix.

PSchmiedmayer (approved):

philippzagar (approved):

PR #380, opened 2021-11-16
Remove the MimeType type

### LINKED ISSUE #378: [INTERNAL CHANGE]: Duplicate implementation of HTTP Media Types
### Problem

As a result of merging #371, there are now two separate implementations of HTTP media types in Apodini, one in the Apodini target (and used only by the `Blob` data type), and one in the ApodiniNetworking target (and used by everything else).

These two implementations should be merged into one that is used throughout all of Apodini.

### Solution

refactor into a separate target.

### Additional context

_No response_

lukaskollmer:
Fixes #378

# Remove the MimeType type

Apodini currently has two separate implementations of HTTP media types (formerly known as MIME types). This PR removes the old one (which was used only by the `Blob` type) and replaces all of its uses with the new one.

## :gear: Release Notes 
- Remove the `MimeType` type. Use the new `HTTPMediaType` instead. Existing uses should be unaffected since the two type's public APIs are identical (the new one is a superset of the old one, to be precise).

- Consolidating these two types also included refactoring part of ApodiniNetworking out into a separate module (ApodiniNetworkingHTTPSupport). This new module currently contains only the HTTP Headers API as well as the Media Type.
- I also took the liberty to rename the `Blob` type's coding key representing the contents of the blob from `byteBuffer` to `data` (since we're storing a `Data` object in that field), and to avoid storing a nil data object (instead we now construct an empty one). This closer matches the actual behaviour of the `Blob` type, since the blob's byteBuffer property is itself non-nil, and the only reason we would've gotten a nil data object would've been because of NIO's internal behaviour, not because of the buffer actually being empty.

### Testing
The existing tests have been adopted to test the new API instead.
Note that the actual data we're testing against weren't changed, so we do in fact know that the new API behaves exactly the same way as the old one.

### Reviewer Nudging
Blob.swift is probably the file containing the most significant changes here, everything else is essentially just renaming types and variables from the old one to the new one.


PSchmiedmayer (approved):
Looks good, thanks for fixing the issue 🚀

PR #384, opened 2021-11-26
Re-implement the gRPC interface exporter and the ProtobufferCoding target

lukaskollmer:
# Re-implement the gRPC interface exporter and the ProtobufferCoding target

The current implementations both of the gRPC interface exporter as well as of Apodini's protobuffer coding facilities, are imperfect insofar as that they do not conform to their respective standards, and do not fully implement the APIs and functionalities associated with these two features (gRPC and Protobuffers).

## :bulb: Proposed solution
This PR fixes the above-mentioned issues, by completely re-writing both the gRPC interface exporter as well as the ProtobufferCoding target from the ground up.

- The gRPC interface exporter now operates directly on a NIO-channel-level (as opposed to being implemented as HTTP routes on ApodiniNetworking's HTTPServer). This gives us way more control over an RPC connection's underlying HTTP/2 stream, which in turn allows us to have our implementation conform to the [official gRPC protocol spec](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md)
  - The gRPC interface exporter now supports proper gRPC streams (within the limits of Apodini's general streaming support). Incoming messages are handled directly when they are received, instead of waiting until the client closes the stream. Analogously, outgoing (i.e. client-bound) messages are sent immediately when they are returned from the handler, instead of waiting until the handler closes the stream.
  - The gRPC interface exporter now supports the [gRPC reflection protocol](https://github.com/grpc/grpc/blob/master/doc/server-reflection.md), which allows clients (such as e.g. [grpcurl](https://github.com/fullstorydev/grpcurl)) to interact with a gRPC-exported Apodini web service without needing to statically know the service's schema, or having to statically generate client code based on the schema.
  - The gRPC interface exporter now properly handles the lifecycle and behaviour of gRPC services as described in the spec. (E.g.: we now properly handle all the different HTTP frames that can be exchanged between a client and a server. Previously we would've handled only actual DATA frames.)
- The `ApodiniProtobuffer` target has been removed entirely. Its functionalities are now part of the `ProtobufferCoding` target and the `ApodiniGRPC` target.
- The `ProtobufferEncoder` and the `ProtobufferDecoder` now properly follow the [proto encoding rules](https://developers.google.com/protocol-buffers/docs/encoding)
  - This includes support for proto packages
  - We now have support for handling handler request and response types in a well-defined way. (What this means is that e.g. if you define a handler which contains a single parameter where the type is a struct that is already a protobuf message in its own right, the generated endpoint will expect simply that type as its input. But if you define a handler where the parameter is an Int or a String (types which can't be "top-level" protobuf types), or a handler which has multiple parameters, the underlying proto schema will be smart enough to recognise that, and generate a new proto message type wrapping the parameter(s), which will then be used as the endpoint's request type. Same for responses.)
  - Message and enum types can now define whether they use the `proto2` or `proto3` coding behaviour, and the encoder and decoder will adopt their behaviour accordingly.
  - This even supports compositions of proto2 and proto3 types, e.g. a proto3 type can define a field where the field's type is a proto2 type.
  - The protobuffer coders now will verify that a Swift type can in fact be represented as a valid protobuffer type, and throw an error if that requirement is not satisfied
    - This is mostly (i.e. where ever possible) handled via protobuffer-coding-specific protocols types must conform to (e.g.: `ProtobufEnum`). This allows us to have the compiler reject programs that would produce non-protobuf-valid types.
    - In other cases, these checks are done at run-time (e.g.: rejecting types that contain a property of invalid types such as `[T?]` (i.e. an array of optional values))
  - The protobuffer encoder now properly supports optional and non-optional values, especially the encoding differences between proto2 and proto3.

## :gear: Release Notes 
- Rewrote the ApodiniGRPC and ProtobufferCoding targets, to follow the gRPC and protobuf specs more closely.
- Web Services no longer need to import and enable the ApodiniProtobuffer interface exporter
- Added a `.close` case to the `ConnectionState` enum.
  - This case indicates that a connection was closed after the most recent request, but that the close event didn't have a request associated with it.
  - The connection states are now as follows (from the perspective of a client-side-streaming Handler):
    - `open`: The connection just received a request, and the client did not signal that it wants the connection be closed.
    - `end`: The connection just received a request, and the client signalled that this was the last request and that the connection should be closed afterwards
    - `close`: The client signalled that the connection should be closed, but did not send a request alongside the close request. In this case the handler is invoked with the most-recent request (to make sure that e.g. accessing `@Parameter` properties doesn't result in a crash), but should ignore that request and instead return its `.final` response.
  - It is guaranteed that the last connection state sent to a client-side-streaming handler is either `end` or `close`, and that the handler will only ever receive one of the two.
  - This change requires corresponding changes to streaming handlers, since the previous pattern of simply using an if statement to check whether the connection is closing now no longer covers all possible connection states:
    - Previous:
      ```swift
      struct NameCollector: Handler {
          @Environment(\.connection) var connection
          @Parameter var name: String
          @State private var names: [String] = []
          
          func handle() -> Response<String> {
              if connection.state == .end {
                  return .final("Received names: \(names.joined(separator: ", ")).")
              } else {
                  names.append(name)
                  return .send()
              }
          }
      }
      ```
    - New:
      ```swift
      struct NameCollector: Handler {
          @Environment(\.connection) var connection
          @Parameter var name: String
          @State private var names: [String] = []
          
          func handle() -> Response<String> {
              switch connection.state {
              case .open:
                  names.append(name)
                  return .send()
              case .end:
                  names.append(name)
                  fallthrough
              case .close:
                  return .final("Received names: \(names.joined(separator: ", ")).")
              }
          }
      }
      ```
- Changed the localhost deployment provider's default ports from 5000 (and up) to 52000 (and up), to avoid conflicts with macOS 12's "Airplay Receiver" feature (which, unless disabled, also listens on 5000)

### Testing
The ProtobufferCoding target's tests have been adopted for testing the new protobuffer coding implementation.

The gRPC interface exporter also has tests, which use the [`grpcurl`](https://github.com/fullstorydev/grpcurl) command line tool to make requests to a gRPC service started using Apodini. (Note that I'm intentionally using a third-party tool to make these test requests, firstly we simply don't have a gRPC client in Apodini, and secondly this means that we're testing our server against a "proper" client implementation, which we can assume to be sound and to properly follow the gRPC spec. Were we to use both a custom server and a custom client for our tests, we might not catch errors in cases where we have incorrectly implemented some aspect of the spec in both of them).
The gRPC tests are skipped when the grpcurl executable is not found in the $PATH, and when running the tests inside Xcode (there were  some weird bugs where the tests would just randomly halt and wait forever, probably caused by the ChildProcess API (Xcode will try to hook its debugger into child processes launched from a test, which has also caused some problems with some other test cases in the past)).
We also have one test case that doesn't use grpcurl, but instead creates a NIO `EmbeddedChannel` to test the behaviour of the gRPC interface exporter's channel handlers. This test case exists mainly as a proof of concept, since it might be a good idea to, at some point in the future, move away from the dependency on grpcurl.

### Reviewer Nudging
The gRPCInterfaceExporter probably is the best starting point for looking at this. ~~The ProtobufferCoding stuff is currently still somewhat all over the place,~~ but the main three points there are the Encoding and Decoding folders (which contain the encoding and decoding implementations, respectively), and `SchemaManager.swift` (which contains the Swift type → Proto type mapping stuff).

This branch currently contains a `LKTestWebService` target, which will be removed before merging. You can run that web service using `swift run LKTestWebService`. It'll start a gRPC service on `localhost:50001`. You can use the [grpcurl](https://github.com/fullstorydev/grpcurl) command line tool to interact with that web service. For example, `grpcurl localhost:50001 describe` will return an overview of the available services obtained via the reflection service. You'll need to create some local TLS certificates, [this](https://ryanparman.com/posts/2019/how-to-create-local-tls-certificates-for-development-on-macos/) might be useful.


Package.swift
```diff
@@ -56,7 +55,10 @@ let package = Package(
 
         // Test Utils
         .library(name: "XCTApodini", targets: ["XCTApodini"]),
-        .library(name: "XCTApodiniObserve", targets: ["XCTApodiniObserve"])
+        .library(name: "XCTApodiniObserve", targets: ["XCTApodiniObserve"]),
+
+        // Lukas' playground
+        .executable(name: "LKTestWebService", targets: ["LKTestWebService"])
```
bauer-andreas: Will this be removed?

EDIT: nevermind mind, read the reviewer nudging. Ideally you keep it until I finished my review so I can experiment with it :)

Sources/Apodini/Apodini.docc/Advanced/CommunicationPattern/PatternImplementation/ClientSideStream.md
```diff
@@ -20,7 +20,7 @@ You can access the connection state using the `Connection`'s `state` property th
 ```swift
 struct SingleParameterHandler: Handler {
     @Parameter var name: String
-    @Environemnt(\.connection) var connection: Connection
+    @Environment(\.connection) var connection: Connection
```
bauer-andreas, line 23: Might want to properly declare this renaming as a breaking change in the PR migration guide and tag the PR accordingly.

Sources/Apodini/Application/Application+Connection.swift
```diff
@@ -29,27 +30,68 @@ extension Application {
...
+    /// The previous request was the last request in the stream,
+    /// and the client has now requested the stream be closed.
+    /// - Note: This state exists to handle cases where a connection is closed after the last request has been received.
+    ///         With some protocols (e.g. gRPC), it is not guaranteed that a client stream is always closed with a final request,
+    ///         but instead the stream can also be kept open for some time after sending the last requesst, with a client then
+    ///         eventually closing the stream by sending a request-less HTTP frame to the server.
+    ///         In such cases (i.e. a stream being closed w/out there being an accompanying final request), the handler is invoked
+    ///         once again, with this connection state.
+    ///         It is guaranteed that a handler will be invoked with only the `end` state or with only the `close` state.
+    ///         A hander will never be called for both of these states.
```
bauer-andreas, line 48: Why is this state need then when only ever one state is used? Maybe its because I'm just starting the review.

How much does the new state "break"? Like, would any Handler relying on the ConnectionState need a migration? How would handlers need to be designed if they need to handle both e.g. grpc and web sockets?
Is the communication pattern documentation adapted to accommodate for the new state?

EDIT: Like I somehow get the difference between the two states. But is it really needed from a Handler implementations perspective? All implementations I saw in this PR, always implemented the same thing for both states.

Sources/Apodini/Application/Application+Connection.swift
```diff
@@ -29,27 +30,68 @@ extension Application {
...
+    ///         comparisons be done using switch statements, but that is impossible since enums (regardless of whether or not they have raw values),
+    ///         will always get a compiler-synthesized `Equatable` implementation.
+    ///         Why do we want to deprecate this function? The issue here is that we want handlers to correctly handle all possible connection states.
+    ///         If a handler simply uses a `if connection.state == .end {} else {}` check, it'd miss `close` states, which would result
+    ///         in the client not receiving the handler response. (The same also applies the other way around, i.e. a handler only checking for `state == .close`
+    ///         and missing `end` states.
+    ///         Requiring the handler to use a switch statement, on the other hand, means that (unless the handler uses a default case, which it shouldn't be doing anyway),
+    ///         all possible connection states will be explicitly checked for in the handler, and responded to accordingly.
+    @available(*, deprecated, message: "Use a switch statement instead!")
```
bauer-andreas, line 61: Ah okay, so this automatically makes existing Handlers force to migrate to the new `close` state.

Sources/ApodiniNetworking/ChannelHandlers/RequestDecoder.swift
```diff
@@ -51,7 +51,16 @@ class HTTPServerRequestDecoder: ChannelInboundHandler, RemovableChannelHandler {
             if req.headers[.contentLength] == bodyBuffer.readableBytes {
                 req.bodyStorage = .buffer(bodyBuffer)
                 state = .awaitingEnd(req)
+            } else if req.headers[.transferEncoding].contains(.chunked) {
+                // The request body is chunked, so we need to take care of that.
+                fatalError("Not yet implemented: Incoming requests with 'Transfer-Encoding: chunked' not yet supported")
```
bauer-andreas: I would advocate to not crash the server in such scenarios but return a proper protocol error. This is basically a denial of service vector.

EDIT: is there an issue to track this feature?

Sources/ApodiniNetworking/ChannelHandlers/RequestDecoder.swift
```diff
@@ -51,7 +51,16 @@ class HTTPServerRequestDecoder: ChannelInboundHandler, RemovableChannelHandler {
...
             } else {
+                // Either there is no Content-Length header, or it has a size that doesn't match the body we were sent
+//                let stream = BodyStorage.Stream()
+//                req.bodyStorage = .stream(stream)
+                print("Headers", req.headers as Any)
+                print("Content-Length: \(req.headers[.contentLength] as Any)")
+                print("bodyBuffer: \(bodyBuffer.getString(at: 0, length: bodyBuffer.writerIndex))")
```
bauer-andreas: seems like some left over debugging?

Sources/ApodiniNetworking/HTTPRouter.swift
```diff
@@ -26,9 +26,19 @@ public enum HTTPPathComponent: Equatable, ExpressibleByStringLiteral {
         if string.hasPrefix(":") {
             self = .namedParameter(.init(string.dropFirst()))
         } else if string == "*" {
-            self = .wildcardSingle
```
bauer-andreas, line 29: Not entirely sure if it is too nit picky, but seems like the initializers and some parts of the file aren't really tested. As its such a core component, would it make sense to add some test cases, to avoid any regressions in the future?

Sources/ApodiniNetworkingHTTPSupport/Headers.swift
```diff
@@ -6,13 +6,13 @@
...
-import NIOHTTP1
+@_exported import NIOHTTP1
 import NIOHTTP2
-import NIOHPACK
+@_exported import NIOHPACK
 import ApodiniUtils
 import Foundation
 
-// swiftlint:disable redundant_string_enum_value
+// swiftlint:disable redundant_string_enum_value large_tuple
```
bauer-andreas: Instead of formulating disable rules for the whole file I would advocate for annotating the exact line it should be disable, such that any newly added violation is carefully considered. In the case of the `large_tuple` rule I counted four violations. Not sure if the overhead is worth it though.

Sources/ApodiniUtils/Other.swift
```diff
@@ -142,6 +153,42 @@ extension CharacterSet {
 }
 
 
+/// Assert that an implication holds
+public func precondition(
```
bauer-andreas, line 157: Seeing the `Other.swift` file growing. Would it maybe be beneficial for code readability purposes to start splitting some things out into separate files? E.g. `Box`, `Boxed`, `Weak` are all widely used types, which no one would expect to find in the `Other` file(?).

Sources/ApodiniUtils/Swift.swift
```diff
@@ -55,6 +55,11 @@ extension Array {
         copy.append(contentsOf: elements)
         return copy
     }
+    
+    /// Sorts the array in-place, using a key path
+    public mutating func sort<T: Comparable>(by keyPath: KeyPath<Element, T>) {
```
bauer-andreas, line 60: Isn't there already a mutating sort func https://developer.apple.com/documentation/swift/array/2296801-sort

Sources/ApodiniUtils/Swift.swift
```diff
@@ -76,6 +81,45 @@ extension Sequence {
...
+            if predicate(element) {
+                retval += 1
+            }
+        }
+        return retval
+    }
+    
+    
+    /// Creates an array by consuming all of the sequence's elements.
+    public func intoArray() -> [Element] {
```
bauer-andreas, line 120: Do we really need a custom overload for calling Array.init?

Tests/ApodiniHTTPTests/EndToEndTests.swift
```diff
@@ -84,6 +84,7 @@ class EndToEndTests: XCTApodiniTest {
             if counter == start {
                 return .final(Blob("🚀🚀🚀 Launch !!! 🚀🚀🚀\n".data(using: .utf8)!, type: .text(.plain)))
             } else {
+                //return .send("\(start - counter)...")
```
bauer-andreas: forgotten comment

bauer-andreas (commented):
Okay so, I know I'm by far not finished with my review. Pretty much skipped the important parts, the new GRPC exporter target and the ProtobufferCoding target. However, as I won't be in town the next few days, I won't be able to finish the review until next week.
So instead of holding back my current points, I decided to already push the comments of those files I have reviewed and continue with the review of the two core targets in a later point in time. I hope this is acceptable, but I didn't really expect an 18k diff 😅 🚀

Sources/ApodiniUtils/Swift.swift
lukaskollmer: i realize that this one is subjective, but IMO it is much nicer when working with chained operations on a sequence, and you can just call `.intoArray()` at the end, rather than having to wrap the whole function chaining expression in an `Array()` call. 

(it also makes more sense semantically, since the array construction happens at the end, but would otherwise appear at the very beginning of the expression)

Sources/Apodini/Apodini.docc/Advanced/CommunicationPattern/PatternImplementation/ClientSideStream.md
lukaskollmer: I only fixed the misspelling in the documentation, the actual type wasn't renamed

Sources/Apodini/Application/Application+Connection.swift
lukaskollmer: There are some protocols (e.g. gRPC) where a client-side-streaming connection can be closed w/out an actual request/message. In this case, we still need to invoke the handler one last time to give it the opportunity to send a response.

Consider an example handler that collects names from a client, and then, when the client indicates that there are no further names, responds with a concatenated list of all these names. In gRPC, the handler would receive each name with an `open` connection state, and the client would then send an empty HTTP frame indicating that the connection should be closed. The issue is that we only know retroactively that the most recent request was in fact the last one, so we need to somehow be able to tell the handler that it should now send its final response. This is what the close state is for (and also why a handler will only ever receive one of close and end, but never both.)

One additional thing (and this was @PSchmiedmayer's idea) is that this final `.close` handler invocation will receive the parameters from the most recent `.open` invocation. This means that the handler could, in theory, still access its parameters without crashing.

You can have a look at the playground test web service, there's an example of one such handler in there. In most cases the differentiation between close and end shouldn't matter too much, and can be handled pretty elegantly using a switch statement with a fallthrough

Sources/Apodini/Application/Application+Connection.swift
lukaskollmer: Yes, exactly. I was unable to completely remove the `==` function (enums automatically get a synthesized `Equatable` conformance, which can't be disabled), but we can mark it as deprecated which will at least result in a compiler warning telling people to use a switch instead.

I had noticed that essentially all existing streaming handlers were implemented using a `if state == .end { ... } else { ... }` pattern, in which case it'd be really easy to miss the close state. Hence the switch, which should hopefully result in people explicitly enumerating all cases

Sources/ApodiniUtils/Swift.swift
lukaskollmer: The difference between this one and the one you linked is that mine takes a KeyPath, whereas the one in the stdlib takes a closure.
(So you could do `names.sort(\.count)`, rather than `names.sort { $0.count < $1.count }`)

Sources/ApodiniNetworking/HTTPRouter.swift
lukaskollmer: Yeah I think that there's some functions in here that were added as part of the initial ApodiniNetworking PR, but then ended up being unused. I'll go through that and add some tests

Sources/ApodiniUtils/Other.swift
lukaskollmer: Agree with you, though I think that this should be a separate PR. Utility/helper types and extensions are currently scattered all over the place, so there is definitely potential for a big utils cleanup PR.

Example: We have ApodiniUtils, which I originally had intended to be the central place for all extensions/helpers that are not limited to just one specific target, but then we also have places like Apodini/Utilities which defines e.g. some EventLoopFuture extensions, and  ApodiniNetworking/Utils, which adds yet some more extensions.

Sources/ApodiniGRPC/Server.swift
```diff
@@ -0,0 +1,289 @@
...
+
+
+struct GRPCServerError: Swift.Error {
+    let message: String
+}
+
+
+/// The gRPC server manages a set of gRPC services (as well as their methods),
+/// and implements the logic for handling incoming requests to these methods.
+class GRPCServer {
```
moritzsternemann, line 26: Would be nice if we could rename this file to `GRPCServer.swift` took me a bit to find it 🙈

Edit: I tried out Xcode's new code review feature but apparently it doesn't support pending comments 😆

Sources/ApodiniGRPC/GRPCInterfaceExporter.swift
```diff
@@ -1,231 +1,324 @@
...
+                inputType: reflectionInputType,
+                outputType: reflectionOutputType,
+                streamRPCHandlerMaker: { [unowned self] in
+                    ServerReflectionInfoRPCHandler(server: self.server)
+                }
+            )
+        )
+        
+        // Makes the types managed by the schema ready for use by the reflection API
+        try! server.schema.finalize()
```
moritzsternemann, line 159: I guess we want this to crash when the schema is invalid, right?

Sources/ApodiniGRPC/GRPCInterfaceExporter.swift
```diff
@@ -1,231 +1,324 @@
...
+            // - communication pattern type (req-res, client-side stream, server-side stream, bidirectional stream). this is probaly the least useful one
+            let methodName = endpoint.absolutePath.reduce(into: "") { partialResult, pathComponent in
+                switch pathComponent {
+                case .root:
+                    break
+                case .string(let value):
+                    partialResult.append(value.capitalisingFirstCharacter)
+                case .parameter(let parameter):
+                    // TODO how should this be handled?
+                    // Maybe pull an objc and turn this into a "withX" schema?
```
moritzsternemann: Not a terrible idea imo. But I think it might make more sense to just drop them.

In general maybe something like combining the operation type and the last string path component could give a something like the example from the ProtoBuf style guide: https://developers.google.com/protocol-buffers/docs/style#services

Sources/ApodiniGRPC/GRPCInterfaceExporter.swift
```diff
@@ -1,231 +1,324 @@
...
-            // of the GRPC exporter.
-            // Each parameter should get a default field tag assigned
-            // above in the export() function.
-            fatalError("No default or explicit field tag available")
+    
+    
+    /// Registers some HTTP routes for accessing the proto reflection schema
+    private func setupReflectionHTTPRoutes() {
+        /// Make a JSON version of the whole gRPC schema available as a regular HTTP GET endpoint.
+        /// - NOTE this might not necessarily be the most dessirable thing, since it might expose internal data. But then again the OpenAPI interface exporter works the exact same way...
```
moritzsternemann:
```suggestion
        /// - NOTE this might not necessarily be the most desirable thing, since it might expose internal data. But then again the OpenAPI interface exporter works the exact same way...
```

Maybe it make sense to add a toggle for the reflection endpoints to the configuration 🤔

Sources/ProtobufferCoding/Encoding/Encoder.swift
```diff
@@ -0,0 +1,166 @@
...
+    public func encode<T: Encodable>(
+        _ value: T,
+        into buffer: inout ByteBuffer,
+        asField field: ProtoType.MessageField
+    ) throws {
+        do {
+            // We (currently) don't care about the actual result of the schema, but we want to ensure that the type structure is valid
+            try validateTypeIsProtoCompatible(T.self)
+        } catch let error as ProtoValidationError {
+            // Note that in this function (the one encoding values into fields, instead of encoding entire messages), our requirements to the
```
moritzsternemann: to the ...?

Sources/ProtobufferCoding/Encoding/Encoder.swift
```diff
@@ -0,0 +1,166 @@
...
+        for codingPath in codingPaths {
+            unmarkAsRequiredOutput(codingPath)
+        }
+    }
+    
+    func isMarkedAsRequiredOutput(_ codingPath: [CodingKey]) -> Bool {
+        fieldsMarkedAsRequiredOutput.contains(Self.codingPathToInts(codingPath))
+    }
+    
+    private static func codingPathToInts(_ codingPath: [CodingKey]) -> [Int] {
```
moritzsternemann: just a nit
```suggestion
    private static func codingPathToFieldNumbers(_ codingPath: [CodingKey]) -> [Int] {
```

Sources/ProtobufferCoding/Encoding/Encoder.swift
```diff
@@ -0,0 +1,166 @@
...
+        codingPath.map { $0.getProtoFieldNumber() }
+    }
+}
+
+
+class _ProtobufferEncoder: Encoder { // swiftlint:disable:this type_name
+    let codingPath: [CodingKey]
+    let userInfo: [CodingUserInfoKey: Any]
```
moritzsternemann, line 139: I don't think this is used anywhere, is it?

Sources/ProtobufferCoding/Encoding/UnkeyedEncodingContainer.swift
```diff
@@ -0,0 +1,103 @@
...
+    private(set) var count: Int = 0
+    
+    init(codingPath: [CodingKey], dstBufferRef: Box<ByteBuffer>, context: EncoderContext) {
+        self.codingPath = codingPath
+        self.dstBufferRef = dstBufferRef
+        self.context = context
+    }
+    
+    mutating func encodeNil() throws {
+        fatalError("Not implemented")
```
moritzsternemann, line 28: Are all these just not implemented yet or are they not supported in the unkeyed context? In the latter case, I think we should throw an error instead.

Sources/ProtobufferCoding/Encoding/KeyedEncodingContainer.swift
```diff
@@ -0,0 +1,428 @@
...
+
+let protobufferUnsupportedNumericTypes = Set(
+    Int8.self, UInt8.self, Int16.self, UInt16.self
+)
+
+
+func throwUnsupportedNumericTypeEncodingError(value: Any, codingPath: [CodingKey]) throws -> Never {
+    precondition(
+        protobufferUnsupportedNumericTypes.contains(type(of: value)),
+        "Asked to throw an \"unsupported numeric type\" error for a type that it not, in fact, an unsupported numeric type."
```
moritzsternemann:
```suggestion
        "Asked to throw an \"unsupported numeric type\" error for a type that is not, in fact, an unsupported numeric type."
```

Sources/ProtobufferCoding/Encoding/KeyedEncodingContainer.swift
```diff
@@ -0,0 +1,428 @@
...
+            return true
+        case .proto3:
+            let isRequiredOutput = context.isMarkedAsRequiredOutput(codingPath.appending(key))
+            return isRequiredOutput ? true : !isDefaultZeroValue
+        }
+    }
+    
+    mutating func encode(_ value: Bool, forKey key: Key) throws {
+        if shouldEncodeNonnilValue(forKey: key, isDefaultZeroValue: value == false) {
+        //if value || context.isMarkedAsOptional(codingPath.appending(key)) {
```
moritzsternemann: What's up with the commented if/guards in this and the following methods?

Sources/ProtobufferCoding/Decoding/Decoder.swift
```diff
@@ -0,0 +1,88 @@
...
+//
+// SPDX-License-Identifier: MIT
+//
+
+import NIO
+import Foundation
+import ApodiniUtils
+
+
+/// The `ProtobufferEncoder` decodes `Decodable` types from protocol buffers
```
moritzsternemann:
```suggestion
/// The `ProtobufferDecoder` decodes `Decodable` types from protocol buffers
```

Sources/ProtobufferCoding/Decoding/Decoder.swift
```diff
@@ -0,0 +1,88 @@
...
+        let decoder = _ProtobufferDecoder(codingPath: [], buffer: buffer)
+        let keyedDecoder = try decoder.container(keyedBy: FixedCodingKey.self)
+        return try keyedDecoder.decode(T.self, forKey: .init(intValue: fieldInfo.fieldNumber))
+    }
+}
+
+
+class _ProtobufferDecoder: Decoder { // swiftlint:disable:this type_name
+    let codingPath: [CodingKey]
+    let userInfo: [CodingUserInfoKey: Any]
```
moritzsternemann, line 61: This is unused as well, I think.

Sources/DeploymentTargetLocalhost/LocalhostDeploymentProvider.swift
```diff
@@ -32,7 +32,7 @@ private struct LocalhostDeploymentProviderCLI: ParsableCommand {
     var port: Int = 80
     
     @Option(help: "The port number for the first-launched child process")
-    var endpointProcessesBasePort: Int = 5000
+    var endpointProcessesBasePort: Int = 52000
```
moritzsternemann: Just curious what the reason for the change is 🤓

Sources/ProtobufferCoding/Proto/Descriptor.swift
```diff
@@ -0,0 +1,1115 @@
...
+//
+// SPDX-License-Identifier: MIT
+//
+
+import Foundation
+
+// swiftlint:disable identifier_name todo missing_docs line_length
+
+
+// TODO we probably can remove a decent amount of the optionals here??? (<<-- NO; we actually have to bring back the ones we removed TODO do this!)
```
moritzsternemann: Is this TODO relevant or not? 😅

Sources/ProtobufferCoding/Proto/Descriptor.swift
```diff
@@ -0,0 +1,1115 @@
...
+
+public struct ExtensionRangeOptions: Codable, Hashable, ProtobufMessageWithCustomFieldMapping, _ProtoPackage_Google_Protobuf {
+    /// The parser stores options it doesn't recognize here. See above.
+//  repeated UninterpretedOption uninterpreted_option = 999;
+    public let uninterpretedOptions: [UninterpretedOption]
+
+
+//  // Clients can define custom options in extensions of this message. See above.
+//  extensions 1000 to max;
+
```
moritzsternemann: Should these be doc comments?

Sources/ProtobufferCoding/SchemaManager.swift
```diff
@@ -0,0 +1,1315 @@
...
+//            for enumTypename in self.allEnumTypes.keys {
+//                var prevMapping = retval.updateValue(enumTypename.packageName, forKey: enumTypename.fullyQualified)
+//                precondition(prevMapping == nil, "Duplicate fully qualified typename")
+//            }
+//            for msgTypename in self.allMessageTypes.keys {
+//                var prevMapping = retval.updateValue(msgTypename.packageName, forKey: msgTypename.fullyQualified)
+//                precondition(prevMapping == nil, "Duplicate fully qualified typename")
+//            }
+//            return retval
+//        }()
```
moritzsternemann: Can we remove this and the following commented blocks?

Tests/ApodiniTests/GRPCTests/GRPCServiceTests.swift
```diff
@@ -0,0 +1,84 @@
...
+//
+// SPDX-License-Identifier: MIT
+//              
+
+//import XCTest
+//@testable import Apodini
+//@testable import ApodiniGRPC_old
+//import XCTApodiniNetworking
+//
+//final class GRPCServiceTests: ApodiniTests {
```
moritzsternemann: I guess these tests won't do any good 😅

Tests/ProtobufferCodingTests/TestTypes.swift
```diff
@@ -0,0 +1,11 @@
+//
```
moritzsternemann, line 1: Will you add some tests in here?

Tests/ProtobufferCodingTests/ProtobufferCodingTests.swift
```diff
@@ -0,0 +1,1161 @@
...
+        struct SimpleStructWithIntProperty: Codable, Equatable {
+            let value: Int
+        }
+//        let input = SimpleStructWithIntProperty(value: 52)
+//        let encoded = try ProtobufferEncoder().encode(input)
+//        try XCTAssertEqual(encoded, [0b1000, 52])
+//        XCTAssertEqual(try ProtobufMessageLayoutDecoder.getFields(in: encoded), ProtobufFieldsMapping([
+//            1: [ProtobufFieldInfo(tag: 1, keyOffset: 0, valueOffset: 1, valueInfo: .varInt(52), fieldLength: 2)]
+//        ]))
+//        try assertDecodedMatches(encoded, input)
```
moritzsternemann:
```suggestion
```

Sources/ApodiniUtils/Shell.swift
```diff
@@ -27,17 +27,13 @@ public enum ShellCommand {
 /// A helper function to run custom shell commands, e.g. on app launch
 @discardableResult
 public func runShellCommand(_ command: ShellCommand) -> String {
-    shell(command.method)
-}
-
-@discardableResult
-private func shell(_ command: String) -> String {
+    //(try? ChildProcess.runZshShellCommandSync(command.method).output) ?? ""
```
moritzsternemann:
```suggestion
```

Sources/ApodiniUtils/Shell.swift
```diff
@@ -27,17 +27,13 @@ public enum ShellCommand {
...
     let task = Process()
     let pipe = Pipe()
-
+
```
moritzsternemann: Can only recommend to enable "Automatically trim trailing whitespaces" in Xcode Preferences > Text Editing > Editing, if you don't have it enabled already 😜

Sources/Apodini/Application/Application+HTTP.swift
```diff
@@ -67,7 +67,7 @@ public struct TLSConfigurationBuilder {
         do {
             let certificate = try NIOSSLCertificate.fromPEMFile(certificatePath)
             let privateKey = try NIOSSLPrivateKey(file: keyPath, format: .pem)
-
+
```
moritzsternemann, line 70: 🤔

Sources/ApodiniUtils/Swift.swift
moritzsternemann: I think Lukas's reasoning is very valid, though I don't like the method name at all, doesn't feel swifty to me 😋. Ideas I had were `.toArray()` or `.asArray()`, `.array()`, or even `.array`. I thought there was something about this in the Swift API Design Guidelines but I couldn't find it unfortunately.

moritzsternemann (changes requested):
Following Andi's example, I partially reviewed the PR for now, focusing on the `ProtobufferCoding` target. I'm not very familiar with the inner workings of Protocol Buffers and gRPC, but I did my best to check everything by common sense and for consistency. The addition of support for the gRPC Reflection Protocol is a super cool in my opinion.

A general thing I noticed with the encoding and decoding is a lot of `"Not implemented"` fatal errors. I think it would be nice to throw errors instead, or at least change the message to something that indicates that proto doesn't support the type instead of a missing implementation.

Sources/ApodiniGRPC/GRPCInterfaceExporter.swift
lukaskollmer: Yes, that's intentional. If the schema is invalid, it fails to finalize, the `try!` will crash, and the error message logged as part of the `try!` will contain an explanation as to what went wrong. (You can have a look at the `ProtoValidationError`, it has a ton of different cases listing all the possible issues that might occur in a schema)

Sources/ProtobufferCoding/Encoding/Encoder.swift
lukaskollmer: Nope, but it's a requirement of the `Encoder` protocol. (I guess the `Codable` implementation might use this internally somewhere?)

Sources/ProtobufferCoding/Decoding/Decoder.swift
lukaskollmer: Same as for the encoder

Sources/DeploymentTargetLocalhost/LocalhostDeploymentProvider.swift
lukaskollmer: I think i added a note about this in the changes section; since macOS 12, the Control Centre will use this port for the "Airplay Receiver" feature

Sources/ProtobufferCoding/Proto/Descriptor.swift
lukaskollmer: The entire Descriptors file was essentially just copied from [the source](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/descriptor.proto), and i then commented out the proto code and added matching Swift code directly below. So the commented parts aren't really documentation, but rather the corresponding proto code. Since not all proto features are supported by us atm (extensions being one of the missing pieces), I deciced to keep them in anyway, to make it directly visible that there's missing stuff

Sources/ProtobufferCoding/SchemaManager.swift
lukaskollmer: YEs, I missed that one. Will push a fix tomorrow

Sources/ProtobufferCoding/Encoding/Encoder.swift
moritzsternemann: Ohh right, I didn't think of this at all somehow.. my bad 🤦‍♂️

Sources/ProtobufferCoding/Encoding/UnkeyedEncodingContainer.swift
lukaskollmer: (Same applies to the `SingleValueEncodingContainer`, as well as the `UnkeyedDecodingContainer`, as well as the `SingleValueDecodingContainer`)

---

Having these crash instead of throwing errors was a deliberate decision.

The thing about en/decoding protobuffer messages is that you should , in theory, only ever need the `Keyed{Encoding,Decoding}Container`s, and never the SingleValue- or Unkeyed- containers. The reason being that all data in protubuffer messages is encoded using a key-value format. (This includes things like arrays, which, in other formats (e.g. JSON), are the primary use case for the unkeyed coding containers.)

There are some edge cases, however, where, in order to implement the en/decoding of proto messages, it is necessary to fall back back to the SingleValue/Unkeyed containers (e.g.: you're decoding a custom type, have already consumed the field's key, and are now invoking the type's `init(from:)` initializer. The type may then call `.singleValueContainer()` on the decoder, in order to read its value from that.) The important thing about this is that accessing a SingleValue/Unkeyed Container will only ever happen within the context of en/decoding a key-value field.

Meaning that if you end up in one of these crashing functions, something went horribly wrong (since you should never end up there, based on how the proto format is defined), and there really isn't a good way to recover from that (apart from just ignoring what you were trying to en/decode).

We could of course replace it with errors, although tbh the way i see it throwing an error should be used in cases where we were unable to en/decode the data due to some runtime problem, and just straight-up crashing should be used in cases where you should never ever end up in that function in the first place.

One more thing: The proto en/decoders do in fact, before en/decoding something, check whether the type can be represented as a proto message, and throw if that is not the case ([see e.g. here]( https://github.com/Apodini/Apodini/blob/00069cb634b7ee7c4d5db8ad54bdf39c264a89c9/Sources/ProtobufferCoding/Encoding/Encoder.swift#L30)), meaning that we do effectively already throw in these edge cases, and the fatalErrors serve as an indication that something in the proto type validation is implemented incorrectly. Hence why I preferred crashing, since that guarantees that you'll actually be made aware of the issue, otherwise it'd be too easy to just silently ignore the error somewhere.

@PSchmiedmayer what are your thoughts on this?

Sources/ProtobufferCoding/Encoding/KeyedEncodingContainer.swift
lukaskollmer: That was the previous implementation for checking whether encoding the value should be skipped, will remove that.

Sources/ApodiniUtils/Swift.swift
lukaskollmer: @moritzsternemann it should definitely be a function and not a computed property, because it's not an O(1) operation. As for the name, I chose `intoArray()` because it consumes the elements in the sequence and puts them into the array (it's not as extreme as in Rust, which is where I got the name from, but in Swift Sequences don't necessarily guarantee that you can iterate over them multiple times, so you might in fact end up with this call having side-effects (though it's incredibly unlikely). Hence the into.) If you want to rename it to `toArray()` that'd be absolutely fine with me, though I'd be against `asArray()` since that sounds like a type cast, which is not what's going on here.

Sources/ApodiniNetworking/ChannelHandlers/RequestDecoder.swift
lukaskollmer: Yeah good point, I'll change it to a 500 error response.

Thing is, supporting chunked requests isn't actually required anymore, since the HTTP interface exporter's streaming is implemented on top of HTTP/2, and Transfer-Encoding is an HTTP/1-only thing.

Sources/DeploymentTargetLocalhost/LocalhostDeploymentProvider.swift
PSchmiedmayer: See #390 for a bit more context. I have made this as a separate PR to also unblock other PRs.

Sources/ApodiniGRPC/Server.swift
lukaskollmer: I mean, we can of course rename it, though if i use the cmd-shift-option thing to search for either "grpcserver" or just "server", it always is within the first 5 results...

Sources/ProtobufferCoding/Encoding/Encoder.swift
lukaskollmer: The integers in the array aren't meant to represent the field numbers, they just happen to do so. In practice we don't care about the actual contents of the array, we just need an injective mapping function, so that we can compare whether a coding path was previously registered with the context. Will still rename it, though

Sources/ProtobufferCoding/Encoding/UnkeyedEncodingContainer.swift
PSchmiedmayer: Can this fatal error be triggered by an incoming Protocol Buffer message after the web service did pass compile time and startup time checks?
If it can potentially be reached (e.g. by a malformed message) at runtime then we should throw and error and make sure that this is propagated to the client. Otherwise this would be a very easy attack vector so kill server instances. So this would be hard yes for proper error handling.

If this is logically impossible to reach I can see that a fatal error could be a valid option. Nevertheless we must assume our check logic is implemented correctly this is probably hard to verify? The question would be if this is any harm to properly throw an error here and where would this error send up? Could it potentially be passed to the client?

lukaskollmer: @PSchmiedmayer 

> Can this fatal error be triggered by an incoming Protocol Buffer message after the web service did pass compile time and startup time checks?

Technically speaking this is completely independent from the web service, since the ProtobufferEncoder could be used from anywhere. In reality, client requests sent to a gRPC service are (apart from the tests) the only source of incoming proto messages.

Whether or not these `fatalError`s are reached is completely independent from the actual message received by the web service (or the data being passed to the decoder), since the actual program execution path within the en/decoder (i.e. which functions are called, which if/etc branches are taken, etc) is determined solely by the concrete `Codable` implementation of the Swift type being en/decoded.
Meaning that if we end up reaching one of these fatalErrors, its not because the input data was invalid, but rather because the type being en/decoded has an invalid `Codable` implementation[^1].

If you prefer throwing errors instead, I'll change it to that, though tbh I don't see how a web service (or rather Apodini operating w/ an invalid web service) could reasonably recover from these errors, since ending up in one of these functions would mean that either no incoming requests could ever be decoded, or that all outgoing responses would fail to encode. At which point the web service would already be effectively unusable.


> The question would be if this is any harm to properly throw an error here and where would this error send up? Could it potentially be passed to the client?

I think this is a different question entirely. Whether or not the error reaches the client depends on where exactly it's thrown. From how I understand the code, this would actually be fully outside the control of either ApodiniGRPC or ApodiniProtobuffer: The decoding of a gRPC input message happens in the gRPC IE's `GRPCEndpointParameterDecodingStrategy`, which gets called from somewhere w/in Apodini. So the question of whether these errors could ever reach a client would be the same as whether e.g. JSON decoding errors in the REST IE could ever propagate back to the client.


[^1]: To provide an example of this: If a type's `encode(to:)` function were to be called w/ a `ProtobufEncoder`, and this type then were to try to encode its properties into this encoder's unkeyedEncodingContainer, that'd result in an invalid proto output, because all proto fields need to be key-value pairs. So it's not like the issue would be that it failed to encode this time, but might work at some point in the future, but rather the issue is that this should never be possible at all. And also, the en/decoder will only try to en/decode if the validation succeesed (which throws if it fails), so this should already be taken care of.

Sources/ApodiniUtils/Shell.swift
lukaskollmer: Isn't there a difference between trailing whitespace at the end of an otherwise non-empty line, and lines that consist only of whitespace? In the first case, you obviously don't want that,  but in the second case I feel like the indentation of an empty line should match the indentation of the lines around it, right?

Sources/ProtobufferCoding/Encoding/UnkeyedEncodingContainer.swift
PSchmiedmayer: @lukaskollmer Thanks for the clarifications! 
Based on your insights I would be fine with sticking with fatalErrors and but would suggest to improve the error message with a hing why this is an fatalError and maybe a request to the developer to file a bug in this project if the error would ever be encountered.
Encoding errors would be able to reach the client but as you mentioned that it is impossible for a any specific client request to reach the web service and trigger this function this is probably also not an error message that a client should be able to see and would be useful for a client to see.

Sources/ApodiniGRPC/GRPCInterfaceExporter.swift
lukaskollmer: Done. Added a config option to enable reflection (on by default)

Sources/Apodini/Application/Application+Connection.swift
bauer-andreas: Thanks for taking the time to explain this in a bit more detail. Helped a lot to understand the situation.

So, `close` and `end` are pretty much the same thing (from a Handler implementation perspective), with the difference, that with `close` the Handler isn't updated (e.g. Parameters hold the value from the previous request). 

Does this always hold? Is the Handler state always equal to that from the previous request when receiving a `close`?

As `close` and `end` pretty much model the same thing, with the difference that `close` is caused by an empty request, would it make sense to merge those two cases and encode this property via an associated value, which indicates if the request was empty or not? Note, this builds around my assumptions outlined above. But to me, this would maybe model the situation a bit better? And Handlers which do not care about it, don't need to expose those "edge case" in their implementation. Also Handlers that need that (like the `StreamingGreeter_CS` Handler in the example web service) could benefit from it, as a `if` syntax is more quickly recognized by a glance than `fallthrough`(?). Lastly, we wouldn't need the Equatable workaround (EDIT: especially as its only a warning and doesn't eliminate programming mistakes, see https://github.com/Apodini/Apodini/pull/384#discussion_r768068374).
But not sure if such an approach was already discussed and I'm missing something essential.

Sources/Apodini/Application/Application+Connection.swift
bauer-andreas: > [...] compiler warning [...]

Isn't there a `removed` syntax with `@available` which also results in a compiler error? Making this a bit more safe?

Sources/ApodiniUtils/Other.swift
bauer-andreas: I'm fine with moving this to another PR. Would be great though, if this would be tracked by an issue 👍

Sources/ApodiniUtils/Swift.swift
bauer-andreas: Yes, I know. The non-mutating KeyPath based on is a custom extension located a few lines below. Would have envisioned a similar version for the mutating one.
I would have envisioned that doing a mutating sort may be less expensive than doing a memory copy and sorting on that. But haven't done any measurements.

Sources/ProtobufferCoding/Proto/Descriptor.swift
bauer-andreas: The proto file was turned to swift code by hand? Is there a particular reason why we didn't just use code generation for that? Seems a bit more sensible for future maintenance then comparing changes by hand.
Otherwise, the `swift-protobuf` has the same file generated in their project: [descriptor.pb.swift](https://github.com/apple/swift-protobuf/blob/cf07a8889859ef191721094fa52713c834cf1481/Sources/SwiftProtobuf/descriptor.pb.swift) or their API wrapper around it [Descirptor.swift](https://github.com/apple/swift-protobuf/blob/cf07a8889859ef191721094fa52713c834cf1481/Sources/SwiftProtobufPluginLibrary/Descriptor.swift). Not sure if we want to pull in that dependency just for that file.

Sources/Apodini/Application/Application+Connection.swift
lukaskollmer: IIRC I initially tried `@available(*, unavailable)` to disable it entirely, but that didn't work because it'd leave the Equatable protocol's requirements not entirely satisfied

bauer-andreas: thanks for the clarification.

Sources/ApodiniUtils/Swift.swift
lukaskollmer: Oh yeah I'll probably be able to fix that by simply implementing them all in terms of the mutating one in the stdlib. I'll see if i can get that working

Sources/ProtobufferCoding/Proto/Descriptor.swift
lukaskollmer: It's a one-off, so it doesn't really make sense to automate it. Sure it might be changed at some point in the future, but if that were to happen a) it wouldn't be too problematic since the proto format is backward-compatible enough that our definitions would still be compatible, and b) the only way i'd see them changing the descriptors types would be as part of e.g. a proto4 spec, which would then probably break our coding implementation anyway.

Additionally, swift-protobuf would've turned it into Swift code compatible with their interfaces and protocols, whereas we need it to be compatible with the protocols used by ProtobufferCoding. It wouldn't be possible to simply import and use the proto definition types from swift-protobuf, since they're using an entirely different set of protocols and conventions to do their proto message type <-> swift type definition mapping.

Sources/ApodiniUtils/Swift.swift
moritzsternemann: Makes sense, I'm fine with keeping `intoArray()` then 👍

Sources/ApodiniGRPC/ChannelHandlers/GRPCRequestDecoder.swift
```diff
@@ -0,0 +1,192 @@
...
+            fatalError("Got .ping(ack=\(ack)): \(pingData)")
+        case let .goAway(lastStreamID, errorCode, opaqueData): // HTTP2StreamID, HTTP2ErrorCode, ByteBuffer?
+            fatalError("Got .goAway(lastStreamId: \(lastStreamID), errorCode: \(errorCode), opaqueData: \(opaqueData as Any))")
+        case let .windowUpdate(windowSizeIncrement): // Int
+            fatalError("Got .windowUpdate: \(windowSizeIncrement)")
+        case let .alternativeService(origin, field): // String?, ByteBuffer?
+            fatalError("Got .alternativeService(origin: \(origin as Any), field: \(field as Any)")
+        case let .origin(origins): // [String]
+            fatalError("Got .origin(\(origins))")
+        }
```
moritzsternemann: Maybe this is similar to the fatalErrors in the protobuf decoding.. but as far as I understand, an attacker could crash the server by sending any of these unsupported HTTP2 frames 🤔

Sources/ApodiniGRPC/StreamHandlers/StreamHandlerBase.swift
```diff
@@ -0,0 +1,81 @@
...
+        switch self.endpointContext.endpointResponseType! {
+        case .primitive, .enumTy, .refdMessageType:
+            fatalError("Encountered invalid proto type: gRPC method return type must be a message. Got: \(endpointContext.endpointResponseType!)")
+        case let .message(name: _, underlyingType, nestedOneofTypes: _, fields):
+            if let underlyingType = underlyingType {
+                precondition(underlyingType == type(of: responseContent))
+                // If there is an underlying type, we're handling a response message that is already a message type, so we simply encode that directly into the message payload
+                let payload = try ProtobufferEncoder().encode(responseContent)
+                return payload
+                //return .singleMessage(headers: headers, payload: payload, closeStream: true)
```
moritzsternemann:
```suggestion
```

moritzsternemann (commented):
Sorry again that this took so long.. I finally got around to review the remaining parts 💪

Sources/ApodiniGRPC/ChannelHandlers/GRPCRequestDecoder.swift
lukaskollmer: You're right, I forgot to replace this with a proper error-handling implementation. Now fixed

moritzsternemann (approved):
Thanks for all the changes! 👍 Looks good to me

PR #388, opened 2021-12-04
Tracing Configuration for ApodiniObserve

moritzsternemann:
# Tracing Configuration for ApodiniObserve

Currently ApodiniObserve only includes Logging and Metrics capabilities for the `WebService`, Tracing functionality is still missing.

## :bulb: Proposed solution
This PR extends the ApodiniObserve module with the initial Tracing configuration capabilities based on [swift-distributed-tracing](https://github.com/apple/swift-distributed-tracing). The newly introduced `TracingConfiguration` and `InstrumentConfiguration` types allow to configure the `InstrumentationSystem` through Apodini's DSL.
Additionally, another target, `ApodiniObserveOpenTelemetry`, introduces convenience methods to configure the swift-distributed-tracing [opentelemetry-swift](https://github.com/slashmo/opentelemetry-swift) backend on `InstrumentConfiguration`.
Lastly, the `InstrumentationSystem`'s unscoped `Tracer` is exposed through the `Application`.

The tracing configuration works as follows:
```swift
import Apodini
import ApodiniObserve
import ApodiniObserveOpenTelemetry
import ApodiniREST
import OpenTelemetry
import OtlpGRPCSpanExporting
import Tracing

    var content: some Component {
        Greeter()
    }

        REST()
        
        TracingConfiguration(
            // Placeholder for any backend implementing the `Instrument` protocol
            InstrumentConfiguration(NoOpTracer()),
            
            // Convenience factory for an OpenTelemetry `InstrumentConfiguration` with default configuration options
            .defaultOpenTelemetry(serviceName: "ExampleWebService"),

            // Convenience factory for an OpenTelemetry `InstrumentConfiguration` with exposed configuration options
            // This example shows the defaults            
            .openTelemetryWithConfig(
                serviceName: "ExampleWebService",
                resourceDetection: .automatic(additionalDetectors: []),
                idGenerator: OTel.RandomIDGenerator(),
                sampler: OTel.ParentBasedSampler(rootSampler: OTel.ConstantSampler(isOn: true)),
                processor: {
                    OTel.SimpleSpanProcessor(exportingTo: OtlpGRPCSpanExporter(config: .init(eventLoopGroup: $0))
                },
                propagator: OTel.W3CPropagator(),
                logger: Logger(label: "org.apodini.observe.OTel")
            )
        )         
    }
}
```

When tracing is configured, `Handler`s can be traced like this:
```swift
struct Greeter: Handler {    
    @Environment(\.tracer) var tracer: Tracer

    func handle() -> String {
        let span = tracer.startSpan(operationName: "Greeter.handle()", context: context)
        defer { span.end() }

        return "Hello World!"
    }
}
```

## :gear: Release Notes 
- Enables configuration of tracing backends in the Apodini Configuration DSL through `TracingConfiguration` and `InstrumentConfiguration`
- Adds convenience `InstrumentConfiguration`s for OpenTelemetry backends
- Enables tracing of `WebService`s using the `Tracer` exposed through the `Application`

### Testing
You can try out tracing on the adapted TestWebService `TraditionalGreeter` endpoint.

First, run the Jaeger service with an OpenTelemetry collector. This can be easily done using the `jaegertracing/all-in-one` Docker container; find the instructions [in the Jaeger documentation](https://www.jaegertracing.io/docs/1.29/getting-started/#all-in-one).

Now you can run the TestWebService and call the TraditionalGreeter endpoint for example using curl: `curl -v "http://localhost:80/v1/greet?age=21"`.

Opening `http://localhost:16686` in your browser, you can select the *TestWebService* from the *Services* list, and see a trace for each call to the endpoint. You can inspect each span individually for collected metadata:
Screen Shot 2021-12-11 at 18 54 32

### Reviewer Nudging
Start by looking at `InstrumentConfiguration.swift` as it's the foundation for instrument configuration. Next, in `TracingConfiguration.swift` you can see how `InstrumentConfiguration`s are handled. Lastly, the `InstrumentConfiguration+OpenTelemetry.swift` extension provides the convenience methods for opentelemetry-swift.


Sources/ApodiniObserveOpenTelemetry/InstrumentConfiguration+OpenTelemetry.swift
```diff
@@ -0,0 +1,64 @@
...
+import OtlpGRPCSpanExporting
+import Tracing
+
+/// Extension that allows for easier setup of the ``InstrumentConfiguration`` for an OpenTelemetry backend.
+extension InstrumentConfiguration {
+    /// Default OpenTelemetry `Instrument`, exporting spans to an OpenTelemetry collector via gRPC in the OpenTelemetry protocol (OTLP).
+    /// - Parameter serviceName: The name of the `WebService` being traced.
+    public static func defaultOpenTelemetry(serviceName: String) -> InstrumentConfiguration {
+        openTelemetryWithConfig(
+            serviceName: serviceName,
```
moritzsternemann, line 21: It would be cool if we could somehow get a default value for this from Apodini, though I didn't really find anything appropriate.

PSchmiedmayer:
@moritzsternemann regarding the build Issues:

This is the relevant error log for the main Apodini builds:
```
gnu/debug/ApodiniObserve.build/Tracing/TracingConfiguration/TracingConfiguration.swift.o -index-store-path /__w/Apodini/Apodini/.build/x86_64-unknown-linux-gnu/debug/index/store -index-system-modules
1.	Swift version 5.5.1 (swift-5.5.1-RELEASE)
2.	
3.	While evaluating request ASTLoweringRequest(Lowering AST to SIL for file "/__w/Apodini/Apodini/Sources/ApodiniObserve/Tracing/TracingConfiguration/TracingConfiguration.swift"
4.	While silgen emitFunction SIL function "@$s14ApodiniObserve20TracingConfigurationC9configureyy0A011ApplicationCF". for 'configure(_:)' (at /__w/Apodini/Apodini/Sources/ApodiniObserve/Tracing/TracingConfiguration/TracingConfiguration.swift:43:12)
5.	While verifying SIL function "@$s14ApodiniObserve20TracingConfigurationC9configureyy0A011ApplicationCF". for 'configure(_:)' (at /__w/Apodini/Apodini/Sources/ApodiniObserve/Tracing/TracingConfiguration/TracingConfiguration.swift:43:12Stack dump without symbol names (ensure you have llvm-symbolizer in your PATH or set the environment var `LLVM_SYMBOLIZER_PATH` to point to it):
```
The error log above points to issues with KeyPaths in the `configure(_:)` function in `TracingConfiguration.swift`: https://github.com/Apodini/Apodini/runs/4493957121?check_suite_focus=true#step:9:1773
My first idea would be to try avoiding KeyPaths there and see if this solves the problem. 

The test web service builds fail due to a corrupted Package.resolved file. Try deleting it and generate it again e.g. using `swift package update` in the TestWebService folder.

The Bamboo builds: Should be resolved by #390

Sources/ApodiniObserveOpenTelemetry/InstrumentConfiguration+OpenTelemetry.swift
PSchmiedmayer: A thing that would come to my mind would be the `hostname ` in the `HTTPConfiguration`: https://github.com/Apodini/Apodini/blob/fac5afd188640b794ec3dda6283d873ea6dd59d0/Sources/Apodini/Configurations/HTTPConfiguration.swift . It should ideally be set to e.g. `ase.in.tum.de`. The only tricky thing is that the InstrumentConfiguration could come after the HTTP configuration and therefore the name might not by populated at this point. But this is something that might not be easily solvable and is something that we should highlight in the documentation that the order is relevant.

PSchmiedmayer: A thing that would come to my mind would be the `hostname ` in the `HTTPConfiguration`: https://github.com/Apodini/Apodini/blob/fac5afd188640b794ec3dda6283d873ea6dd59d0/Sources/Apodini/Configurations/HTTPConfiguration.swift . It should ideally be set to e.g. `ase.in.tum.de`. The only tricky thing is that the InstrumentConfiguration could come after the HTTP configuration and therefore the name might not by populated at this point. But this is something that might not be easily solvable and is something that we should highlight in the documentation that the order is relevant.

moritzsternemann: Hmm, right.. I'll mention it in the docs and try to populate the name from the hostname when it's available.

Also, it might be interesting to think about a mechanism to enforce the order of some configurations because for example [Apple also recommends to bootstrap Logging, Metrics, and Tracing in this specific order](https://github.com/apple/swift-distributed-tracing#recommended-bootstrap-order).

Sources/ApodiniObserve/Tracing/TracingConfiguration/TracingConfiguration.swift
```diff
@@ -0,0 +1,60 @@
...
+        
+        // Bootstrap the instrumentation system
+        InstrumentationSystem.bootstrap(
+            MultiplexInstrument(constructedInstruments.map { $0.instrument })
+        )
+        
+        // Add instrument shutdown lifecycle hooks to app
+        constructedInstruments
+            .compactMap { $0.instrumentShutdown }
+            .map { InstrumentConfiguration.Lifecycle(instrumentShutdown: $0) }
```
lukaskollmer: not that it would result in a noticeable performance change, but you can combine the `compactMap` and `map` operations into, which should halve the number of iterations in the loop...

Sources/ApodiniObserveOpenTelemetry/InstrumentConfiguration+OpenTelemetry.swift
lukaskollmer: Is the serviceName here the same the hostname, though? I'd imagine the "name" of the web service to be the product (e.g. "Xpense") rather than the hostname on which it just so happens to reside (e.g. xpense.ast.in.tum.de).

This is actually something I've been thinking about a bit, since we currently have several places where the "name" of the web service being exported is somehow used/required. The first, iirc was the OpenAPIInterfaceExporter (bc the OpenAPI definition contains a name field). The gRPC IE also has configuration options for "package name" and "service name" (the first being protobuf-specific). And GraphQL might end up needing something like this as well.

It'd probably be a good idea to have a top-level (i.e. not belonging to any specific IE) configuration option for specifying these kinds of metadata about a web service. This would need to come at the very beginning of the configuration, and then make this available to all other interface exporters.

(If your comments were in fact about the actual hostname, rather than the end-user-legible name of the service, I guess you can ignore my comment here...)

Sources/ApodiniObserveOpenTelemetry/InstrumentConfiguration+OpenTelemetry.swift
```diff
@@ -0,0 +1,64 @@
...
+                eventLoopGroup: group,
+                resourceDetection: resourceDetection,
+                idGenerator: idGenerator,
+                sampler: sampler,
+                processor: processor(group),
+                propagator: propagator,
+                logger: logger
+            )
+
+            try! otel.start().wait()
```
lukaskollmer: do we want the crash here?

Sources/Apodini/Apodini.docc/Ecosystem/Observe.md
```diff
@@ -16,9 +16,9 @@ SPDX-License-Identifier: MIT
...
+`ApodiniObserve` heavily relies on Apple's observability libraries, especially [swift-log](https://github.com/apple/swift-log), [swift-metrics](https://github.com/apple/swift-metrics), and [swift-distributed-tracing](https://github.com/apple/swift-distributed-tracing), to provide instrumentation possibilities to developers. It's important to note that these packages just define an observability API ("observability frontend"), but don't actually process and transmit the insturmented telemetry information to an observability software stack. This functionality is left to individual observability integrations ("observability backends") that conform to the processing APIs of Apple’s observability libraries. A list of compatible observability integrations can be found on the [GitHub page of swift-log](https://github.com/apple/swift-log), the [GitHub page of swift-metrics](https://github.com/apple/swift-metrics), as well as the [GitHub page of swift-distributed-tracing](https://github.com/apple/swift-distributed-tracing).
```
lukaskollmer:
```suggestion
`ApodiniObserve` heavily relies on Apple's observability libraries, especially [swift-log](https://github.com/apple/swift-log), [swift-metrics](https://github.com/apple/swift-metrics), and [swift-distributed-tracing](https://github.com/apple/swift-distributed-tracing), to provide instrumentation possibilities to developers. It's important to note that these packages just define an observability API ("observability frontend"), but don't actually process and transmit the instrumented telemetry information to an observability software stack. This functionality is left to individual observability integrations ("observability backends") that conform to the processing APIs of Apple’s observability libraries. A list of compatible observability integrations can be found on the [GitHub page of swift-log](https://github.com/apple/swift-log), the [GitHub page of swift-metrics](https://github.com/apple/swift-metrics), as well as the [GitHub page of swift-distributed-tracing](https://github.com/apple/swift-distributed-tracing).
```

lukaskollmer (approved):

Sources/ApodiniObserveOpenTelemetry/InstrumentConfiguration+OpenTelemetry.swift
moritzsternemann: Right, I agree with Lukas here. The service names in the tracing context tend to be something like "frontend", or "customer". That's why I initially thought about using the actual name of the type that is implementing the `WebService` protocol.

PSchmiedmayer (approved):
As noted in our meeting it would be good to add two test cases, otherwise this looks good:
1. A short test case asserting that the configuration correctly creates the instruments
2. A test case testing the Tracer in the environment and that e.g. a span is created

Apart from this the PR looks good, thanks for the new functionality🚀

moritzsternemann:
Since I had to make more changes than expected to enable testing, it would be nice if you could do a quick sanity check again, @PSchmiedmayer.

Package.swift
```diff
@@ -106,6 +107,8 @@ let package = Package(
         // Use a forked repository of the https://github.com/apple/swift-metrics-extras repository that
         // is versioned and already contains test functionality
         .package(url: "https://github.com/Apodini/swift-metrics-extras.git", .upToNextMinor(from: "0.1.0")),
+        .package(url: "https://github.com/apple/swift-distributed-tracing.git", .upToNextMinor(from: "0.1.2")),
+        .package(url: "https://github.com/slashmo/opentelemetry-swift.git", from: "0.1.1"),
```
PSchmiedmayer: Probably would also make sense to use `.upToNextMinor(from:` here.

PSchmiedmayer (approved):
Looks good to me! Only noticed one small thing I missed before in the Package.swift file. 
The code quality, comments and overall structure is really nice, good job! Great to see the progress and getting the first Tracing PR merged in Apodini 🚀

Package.swift
moritzsternemann: Agreed, thanks for pointing it out!

PR #390, opened 2021-12-12
Use a port range for the LocalhostDeploymentProvider that can not be registered at IANA

PSchmiedmayer:
# Use a port range for the LocalhostDeploymentProvider that can not be registered at IANA

As noted in #384 we currently rely on a port range that is e.g. used by Apple AirPlay and might be potentially be registered at the IANA for the LocalhostDeploymentProvider. @lukaskollmer changed this in his PR (https://github.com/Apodini/Apodini/pull/384/commits/4b5914a52725de725000ce1127e2011fc2591086), this PR only incorporates this change as a single PR to block other non-building PRs such as #388.

## :bulb: Proposed solution
Moves the LocalhostDeploymentProvider ports to the ["Dynamic, private or ephemeral ports"-range](https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers) of 52000 and onwards.

## :gear: Release Notes 
Moves the LocalhostDeploymentProvider ports to the ["Dynamic, private or ephemeral ports"-range](https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers) of 52000 and onwards.

### Related PRs
#384 & #388

### DIFF
--- a/Sources/DeploymentTargetLocalhost/LocalhostDeploymentProvider.swift
+++ b/Sources/DeploymentTargetLocalhost/LocalhostDeploymentProvider.swift
@@ -32,7 +32,7 @@ private struct LocalhostDeploymentProviderCLI: ParsableCommand {
     var port: Int = 80
     
     @Option(help: "The port number for the first-launched child process")
-    var endpointProcessesBasePort: Int = 5000
+    var endpointProcessesBasePort: Int = 52000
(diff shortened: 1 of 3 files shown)


PaulsAutomationBot (approved):
🤖:+1:

PR #402, opened 2021-12-24
Add Basic GraphQL Support

lukaskollmer:
# Add Basic GraphQL Support

Currently, Apodini does not support making web services available via GraphQL.

## :bulb: Proposed solution
This PR would address the aforementioned problem, by adding support for making web services available via GraphQL.
This will be supported via a dedicated new `GraphQLInterfaceExporter`.

## :gear: Release Notes 
- Added a new `GraphQLInterfaceExporter`
- Protobuffer: Added support for the `Foundation.Date`, `Foundation.URL`, and `Foundation.UUID` types
- Added a new "endpoint name" API, which allows users to specify endpoint names, either at handler-type-level (via the `metadata` property), or at DSL-level (via the `endpointName(_:)` modifier. Endpoint names are made available to interface exporters, which can use them to generate user-facing names for the individual endpoints in the exported API.
- Added a few convenience functions and types to make `Configuration`s conditional. This allows enabling/disabling configuration entries based on certain conditions such as whether HTTPS is enabled, the build type (debug vs release), or others.

This PR ended up also including some changes to the ProtobufferCoding target (and therefore the gRPC interface exporter). One could argue that these changes should really be a separate PR, but IMO it's fine to have them in this one (most of the proto changes came about from implementing features for GraphQL and realising that the gRPC IE would also need to support something equivalent. Examples: things like encoding/decoding `Date`/`URL`/`UUID` objects.)

### Related PRs
n/a

### Testing
All new features should have tests, this includes both the GraphQL interface exporter, as well as additional tests for the new things added to the ProtobufferCoding target.
I also fixed some issues in the TestWebService, enabled HTTPS, and added the GraphQL interface exporter, so that it should now be using all of Apodini's interface exporters.

### Reviewer Nudging
The `GraphQLInterfaceExporter` and the `GraphQLSchemaBuilder` are probably the best entry points here. There's testing for all new features, so i guess looking at the tests would also be a good way to see how the new stuff is used.

(Also, don't be confused by the number of commits listed on GitHub for this PR, this branch originated from the gRPC branch before that one was merged, so it lists all of the gRPC commits as well...)


Sources/ApodiniGraphQL/GraphQLInterfaceExporter.swift
```diff
@@ -0,0 +1,150 @@
...
+    
+    private struct Wrapped<T: Codable>: Codable {
+        let value: T
+    }
+    
+    let name: String
+    
+    func decode(from input: Input) throws -> T {
+        guard let value = try input.dictionaryValue(converting: true)[name] else {
+            throw Error(message: "Unable to find parameter named '\(name)' (T: \(T.self))")
```
PSchmiedmayer: I would suggest to thrown an `ApodiniError` instead of a custom error struct here.

Sources/ApodiniNetworking/HTTPServer.swift
```diff
@@ -437,6 +437,7 @@ extension Channel {
         responder: HTTPResponder
     ) -> EventLoopFuture<Void> {
         pipeline.addHandlers([
+            //HTTP2ServerRequestDecoder(),
```
PSchmiedmayer: Should be removed

TestWebService/Sources/TestWebService/TestWebService.swift
```diff
@@ -41,6 +51,29 @@ struct TestWebService: Apodini.WebService {
     }
     
     var configuration: Configuration {
+        switch httpsConfig {
+        case .none:
+            HTTPConfiguration(
+                hostname: .init(address: "localhost", port: port),
```
PSchmiedmayer: I think you can omit `"localhost"` and `"0.0.0.0"` as both arguments should have them as the default values. Also applies to the other cases.

TestWebService/Sources/TestWebService/TestWebService.swift
```diff
@@ -16,20 +16,30 @@ import ApodiniObserve
...
 @main
 struct TestWebService: Apodini.WebService {
     private static let greeterRelationship = Relationship(name: "greeter")
-
+    
     @Argument(help: "Endpoint to expose OpenAPI specification")
     var openApiEndpoint: String = "oas"
     
+    @Option(help: "The TestWebService's HTTPS config. Omit to disable HTTPS, specify 'builtin' to use buitin self-signed certificates, or pass a path to a custom certificate and key.") // swiftlint:disable:this line_length
+    var httpsConfig: HTTPSConfig = .none
```
PSchmiedmayer: Nice addition!

TestWebService/Sources/TestWebService/TestWebService.swift
```diff
@@ -56,9 +89,48 @@ struct TestWebService: Apodini.WebService {
         
         Migrator()
         
+        GRPC(packageName: "org.apodini", serviceName: "TestWebService")
+            .skip(if: !.isHTTPSEnabled)
```
PSchmiedmayer, line 99: 🚀

Tests/ApodiniHTTPTests/EndToEndTests.swift
```diff
@@ -84,6 +84,7 @@ class EndToEndTests: XCTApodiniTest {
             if counter == start {
                 return .final(Blob("🚀🚀🚀 Launch !!! 🚀🚀🚀\n".data(using: .utf8)!, type: .text(.plain)))
             } else {
+                //return .send("\(start - counter)...")
```
PSchmiedmayer:
```suggestion
```

Tests/ApodiniTests/GRPCTests/GRPCServiceTests.swift
```diff
@@ -0,0 +1,84 @@
+//                   
+// This source file is part of the Apodini open source project
+//
+// SPDX-FileCopyrightText: 2019-2021 Paul Schmiedmayer and the Apodini project authors (see CONTRIBUTORS.md) <paul.schmiedmayer@tum.de>
+//
+// SPDX-License-Identifier: MIT
+//              
+
+//import XCTest
```
PSchmiedmayer: Do you plan to uncomment the tests here? Or are they related to the test failures that  wee have seen and that's why there are commented out? If so, I would suggest to use `XCTSkip` instead.

PSchmiedmayer (approved):
Very cool job with the addition of the GraphQL exporter, thanks @lukaskollmer!! 🚀
Only had a few comments here and there, the overall state of the PR looks good to me 👍

Tests/ApodiniTests/GRPCTests/GRPCServiceTests.swift
lukaskollmer: this file has already been removed in the other PR, so merging develop back into this branch should remove it from here as well

Sources/ApodiniGraphQL/GraphQLInterfaceExporter.swift
lukaskollmer: I intentionally didn't use the `ApodiniError` type because I thought it was only for errors thrown in handlers? (at least that's what it says in the documentation)

TestWebService/Sources/TestWebService/TestWebService.swift
lukaskollmer: You're right that they could be removed. I kept them in here intentionally so that, if someone were to run the TestWebService, it'd be easier to see where the web service is running. Also, specifying just a port w/out an address did look kinda weird. I'll remove them if you want me to do so.

Sources/ApodiniGraphQL/GraphQLInterfaceExporter.swift
PSchmiedmayer: I would say that `ApodiniError`s should be used everywhere whey they might be returned to the client, so they should be everywhere when they are part of the request/response pipeline including the handler execution but also including the encoding/decoding. This is also especially relevant for #405 and #410 after that.
Would also be good to update that quote in the documentation there. The Docs are more user focused but this is also an information that would be useful for internal development.

TestWebService/Sources/TestWebService/TestWebService.swift
PSchmiedmayer: I am fine with making it explicit if you want to do so but would then suggest to avoid specifying the string literal three times each. Maybe a static property of the `TestWebService` that is used in all switch cases would make sense to change the binding address and the hostname in a single location 👍

Sources/ApodiniGraphQL/GraphQLInterfaceExporter.swift
lukaskollmer: Done.

TestWebService/Sources/TestWebService/TestWebService.swift
lukaskollmer: I've changed it so that both `hostname` and `bindAddress` are now CLI options with a default value. that way they're defined in a central place, can also easily be changed, and additionally also will show up in the `--help` output.

Sources/ApodiniGraphQL/GraphQLInterfaceExporter.swift
PSchmiedmayer: Thanks! 💪

TestWebService/Sources/TestWebService/TestWebService.swift
PSchmiedmayer: Nice, very cool! 🚀

PR #405, opened 2022-01-05
Decoding, Evaluation, and Result Transformation Error Forwarding

moritzsternemann:
# Decoding, Evaluation, and Result Transformation Error Forwarding

For the support of automatic tracing functionality, we want to record errors and attach additional metadata to traces when errors occur. An inspiration for the implementation is ApodiniObserve's automatic error rate metric. However, the issue here is that only errors occurring during the actual `Handler` implementation of each endpoint are captured. Errors that are thrown in the remaining parts of the handler evaluation chain, namely decoding and result transformation, are ignored.

## :bulb: Proposed solution
This PR introduces functionality to catch and forward errors that happen while decoding, evaluating, and transforming the result of `Handler`s. It does so by introducing an `ErrorForwarder` local knowledge source which basically just wraps a closure that receives a captured error. These forwarders are now used in all interface exporters to capture and forward errors from the decoding pipeline and the result transformation (which also captures handler evaluation errors) transparently.

## :gear: Release Notes

Adds support for transparent forwarding of errors that occur during the `Handler` evaluation chain (decoding, evaluation, result transformation) for all interface exporters. The feature can be used by configuring an `ErrorForwarder` for an `Endpoint`.

Example usage:
```swift
struct ErrorForwarderExporter: InterfaceExporter {
    func export<H>(_ endpoint: Endpoint<H>) -> () where H : Handler {
        endpoint[ErrorForwarder.self] = try! ErrorForwarder { error in
            print("received error", error)
        }
    }

    // ...
}
```

In a follow-up PR I plan to make use of this functionality to receive the errors and attach them to tracing Spans where appropriate. Finally, the captured errors can also be used to determine what metadata should be attached to a Span.

### Testing
The PR contains unit tests for forwarding of decoding, evaluation, and result transforming errors in the PRs.

### Reviewer Nudging
As I haven't worked with this part of Apodini before, I'm especially interested in feedback on the approach with the `ErrorForwarder` knowledge source. Also, please let me know if I missed capturing relevant errors in any of the interfaces exporters.

Lastly, one interesting addition might be to allow developers to attach multiple forwarding closures to a single endpoint, in order to enable multiple receivers of each error. To reduce complexity, I left this out for now, but I'm happy to hear your thoughts on this.


Sources/ApodiniExtension/KnowledgeSources/ErrorForwarder.swift
```diff
@@ -0,0 +1,34 @@
...
+    let forward: Key.Value?
+
+    public init(from forward: Key.Value?) throws {
+        self.forward = forward
+    }
+
+    /// Forward an error using the forwarders forwarding closure, if it is non-nil.
+    ///
+    /// - Parameter error: The error to forward.
+    public func forwardError(_ error: Error) {
```
lukaskollmer: Nit, but wouldn't `forward(error:)` be a better name here?

Sources/ApodiniExtension/RequestProcessing/DecodingErrorForwarding.swift
```diff
@@ -0,0 +1,67 @@
...
+public struct DecodingErrorForwardingRequest: WithRequest {
+    public let request: Request
+    let forward: (Error) -> Void
+
+    init(request: Request, forward: @escaping (Error) -> Void) {
+        self.request = request
+        self.forward = forward
+    }
+
+    public func retrieveParameter<Element>(_ parameter: Parameter<Element>) throws -> Element where Element: Decodable, Element: Encodable {
```
lukaskollmer: why not just do `func retrieveParameter<Element: Codable>` instead? That'd save you the `where` clause at the end (which btw is also somewhat redundant, since `where Element: Decodable, Element: Encodable` is the same as `where Element: Codable`, which then again of course is the same as simply having `<Element: Codable>` in the parameter list)

Sources/ApodiniExtension/RequestProcessing/DecodingErrorForwarding.swift
```diff
@@ -0,0 +1,67 @@
...
+        }
+    }
+}
+
+extension ErrorForwarder {
+    func forwardDecodingErrors(_ request: Request) -> DecodingErrorForwardingRequest {
+        if let forward = forward {
+            return DecodingErrorForwardingRequest(request: request, forward: forward)
+        }
+        return DecodingErrorForwardingRequest(request: request, forward: { _ in })
```
lukaskollmer:
```suggestion
        DecodingErrorForwardingRequest(request: request, forward: forward ?? { _ in })
```

lukaskollmer (approved):
Had a look at this, and it looks good to me (though tbh I haven't done extensive testing).

Regarding the question of catching errors thrown in interface exporters, that's probably a bit more tricky since each IE is kinda its own independent thing, so it probably wouldn't really be possible to have a one-size-fits-all approach to error tracing there.
The current approach (i.e. what's in this PR) will probably already catch quite a lot of errors, but in some cases an IE might throw, catch, and handle an error in a way that it wouldn't be recorded using this PR's approach. E.g. if an error occurs while handling a request (and the error occurs within the IE, rather than w/in the handler), the IE in some cases would just immediately respond with an error to the client, rather than going through the Apodini evaluation mechanism, meaning that this error would evade recording.
Two examples for this: incoming HTTP requests that can't be resolved to a registered route. In this case the HTTPServer will immediately return a 404 response, and Apodini's evaluation mechanism won't be activated at all. Same for the gRPC NIO infrastructure when dealing with incoming messages to non-existent methods. I'm not sure how this could be integrated into the error tracing stuff in a good way, though, and also not sure whether these kinds of errors are something you'd be interested in in the first place (since they're not happening due to some internal error when dealing with a request, but rather because the request was invalid to begin with...)

Sources/ApodiniExtension/KnowledgeSources/ErrorForwarder.swift
moritzsternemann: I thought this was a recommendation from the Swift API design guidelines but after having another look, I feel like it might be even better to just name it `forward(_:)`. The meaning will be clear from the `ErrorForwarder` type name and the name of the passed property at the point of use. See "Omit needless words" in [the API design guidelines](https://www.swift.org/documentation/api-design-guidelines/). What do you think?

bauer-andreas (approved):
This PR looks good to me. Great work 🚀 While it was pretty tricky to find a completely transparent approach, this solution fits well with the current Exporter API pipeline-style design.
The only fear is of Exporter implementations forgetting to implement this feature. I don't think we can do anything about that really, apart from maybe adding some documentation about this (either in the DocC article or in the inline docs maybe?). Just so we can't blame it on missing documentation if someone forgets to support this?

To add onto @lukaskollmer though:
> The current approach (i.e. what's in this PR) will probably already catch quite a lot of errors, but in some cases an IE might throw, catch, and handle an error in a way that it wouldn't be recorded using this PR's approach.

I stumbled across the same though when reviewing the REST part of the PR. I think there is a use case of ErrorForwarders (or error reporters) of errors which aren't associated with an specific endpoints. In the case of REST this could e.g. be stuff like the `RESTDefaultRootHandler` (though a bit of a bad example as this http endpoint never returns any errors). Or as discussed 404 style errors, or other non-endpoint related endpoints advertising e.g. web service descriptions (OpenAPI, proto schema). Maybe there are some uses cases for the deployment providers where errors could be routed through the metrics framework?

> Lastly, one interesting addition might be to allow developers to attach multiple forwarding closures to a single endpoint, in order to enable multiple receivers of each error. To reduce complexity, I left this out for now, but I'm happy to hear your thoughts on this.

I think this could be an interesting addition, where the developer supplies the type of Error they want to catch and the respective closure (with `Swift.Error` being the standard error aka. catch all). To support this you probably need to create another layer, where there is one default/main ErrorForwarder which reads from a list of typed ErrorForwarders, checks if the error matches the expected type and calls the closure in that case.
This should be easily doable, but not sure if there is really a use case for it?

PR #406, opened 2022-01-07
Restore AWS Lambda Compatibility

lukaskollmer:
# Restore AWS Lambda Compatibility

The ability to deploy to AWS Lambda was temporarily removed in #371, because the old implementation was using a Vapor-based AWS Lambda integration component, which was incompatible with ApodiniNetworking.

## :bulb: Proposed solution
This PR restores the functionality of the AWS Lambda Deployment Provider, by adding a custom adapter to handle incoming lambda events in Apodini.
Additionally, this PR fixes some problems in the Lambda Deployment Provider which were caused by [recent changes](https://aws.amazon.com/de/blogs/compute/coming-soon-expansion-of-aws-lambda-states-to-all-functions/) AWS made to the lambda APIs.

## :gear: Release Notes 
- Restored the "deploy to AWS Lambda" functionality

The way the AWS Lambda integration is implemented is that the `HTTPServer` now has a `shouldBindOnStart` property, which is set to `true` by default, and can be set to `false` to prevent the server from binding to its specified bind address on startup. This is necessary because you can't bind to e.g. localhost on AWS Lambdas (which also wouldn't ne necessary/useful in the first place because there's a dedicated API for handling incoming events).

### Related PRs
- #371, where we removed AWS Lambda support

### Testing
The old tests have been re-activated.

### Reviewer Nudging
The interesting things are in `LambdaIntegration.swift`, that's where the Lambda Event API <-> ApodiniNetworking's HTTP Server integration resides.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -608,7 +608,6 @@ let package = Package(
                 .target(name: "ApodiniNetworking"),
                 .product(name: "AWSLambdaRuntime", package: "swift-aws-lambda-runtime"),
                 .product(name: "AWSLambdaEvents", package: "swift-aws-lambda-runtime")
-                //.product(name: "VaporAWSLambdaRuntime", package: "vapor-aws-lambda-runtime")
             ]
         ),
         
--- a/Package.swift
+++ b/Package.swift
@@ -114,7 +114,8 @@ let package = Package(
         .package(url: "https://github.com/Apodini/ApodiniMigrator.git", .upToNextMinor(from: "0.2.0")),
 
         // TypeInformation
-        .package(url: "https://github.com/Apodini/ApodiniTypeInformation.git", .upToNextMinor(from: "0.3.0"))
+        .package(url: "https://github.com/Apodini/ApodiniTypeInformation.git", .upToNextMinor(from: "0.3.0")),
+        .package(url: "https://github.com/swift-server/swift-aws-lambda-runtime.git", .upToNextMajor(from: "0.3.0"))
     ],
     targets: [
         .target(name: "CApodiniUtils"),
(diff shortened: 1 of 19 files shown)


Sources/DeploymentTargetAWSLambdaRuntime/LambdaIntegration.swift
```diff
@@ -0,0 +1,128 @@
...
+// This source file is part of the Apodini open source project
+//
+// SPDX-FileCopyrightText: 2019-2021 Paul Schmiedmayer and the Apodini project authors (see CONTRIBUTORS.md) <paul.schmiedmayer@tum.de>
+//
+// SPDX-License-Identifier: MIT
+//
+//
+// This code is based on the vapor-aws-lambda-runtime project: https://github.com/vapor-community/vapor-aws-lambda-runtime
+//
+// SPDX-FileCopyrightText: TODO @Paul what should I put here?
```
lukaskollmer: @PSchmiedmayer question: I'm referencing the [vapor-aws-lambda-runtime](https://github.com/vapor-community/vapor-aws-lambda-runtime) project here, since the code of the Lambda integration is based on that. But I'm not really sure what I should use as the FileCopyrightText? Should this just be "the vapor-aws-lambda authors" or something like that? They don't have license headers in their files, otherwise i'd just have used that as a reference point...

PSchmiedmayer: Yes, I think that "the vapor-aws-lambda authors" sounds right. The license file also just contains a placeholder and it gives credit to the right persons.

Package.swift
```diff
@@ -603,7 +604,10 @@ let package = Package(
             dependencies: [
                 .target(name: "DeploymentTargetAWSLambdaCommon"),
                 .target(name: "ApodiniDeployRuntimeSupport"),
-                .target(name: "ApodiniOpenAPI")
+                .target(name: "ApodiniOpenAPI"),
+                .target(name: "ApodiniNetworking"),
+                .product(name: "AWSLambdaRuntime", package: "swift-aws-lambda-runtime"),
+                .product(name: "AWSLambdaEvents", package: "swift-aws-lambda-runtime")
                 //.product(name: "VaporAWSLambdaRuntime", package: "vapor-aws-lambda-runtime")
```
PSchmiedmayer: This line should be removed.

PSchmiedmayer (approved):
@lukaskollmer Thanks for reenabling and fixing the AWS Lambda deployment, great to see the extensibility of the Apodini Networking component 💪

bauer-andreas (approved):
Looks good from my side 🚀

PR #409, opened 2022-01-11
 Configure the Root Path for HTTP-based Exporters

PSchmiedmayer:
# Configure the Root Path for HTTP-based Exporters

The current REST and HTTP exporters can't be used side by side as they both register endpoints at the same path.

## :bulb: Proposed solution
The Interface Exporters get a `rootPath` configuration option to define the root path.
The exporters also no longer use the version as a default root path.

## :gear: Release Notes 
The REST and HTTP Exporter don't use the version as a URI prefix for all endpoints exported. You can use the `rootPath` configuration to get the previous behavior:
```swift
HTTP(rootPath: .version)
```
```swift
REST(rootPath: .version)
```

Additionally also fixes the following more minor bugs:
- The Migrator CLI options were overwritten by the configuration defined in the compiled code. The CLI arguments now take prescience over the predefined configuration points
- The Migrator CLI options were not parsed when passing arguments for the API Docs and the Migration Guide. Improved the parsing by using unique property names
- An unknown HTTP upgrade header could crash the web service, now only logs an error
- Unexpected input in a client-side stream in the HTTP Exporter crashed the web service. It now returns an appropriate error.

### Testing
Added some tests for the HTTP configuration


Sources/ApodiniMigration/ApodiniMigratorInterfaceExporter.swift
```diff
@@ -208,7 +208,7 @@ final class ApodiniMigratorInterfaceExporter: InterfaceExporter {
             }
         }
         
-        if exportOptions.directory == nil && exportOptions.endpoint == nil {
+        if exportOptions.directory == nil && exportOptions.directory == nil {
```
bauer-andreas: `if` with twice the same condition?

bauer-andreas (approved):
Looking good apart from one minor comment 🚀

Sources/ApodiniMigration/ApodiniMigratorInterfaceExporter.swift
PSchmiedmayer: Thanks, that is a mistake, I will fix this and merge the PR after that 👍

PR #416, opened 2022-01-25
Improve Parameter Handling

### LINKED ISSUE #415: [Bug]: HTTP and REST interface supporters don't support `Date` objects as path/query parameters
### Description

Because `Date` doesn't conform to `LosslessStringConvertible` (for good reasons, btw), date objects cannot be used as path or query parameters in HTTP/REST endpoints. It is possible to work around this, by adding a dummy `extension Date: LosslessStringConvertible` conformance somewhere in the web service, but that still produces invalid parameter values: All `Date` query parameters are interpreted as the amount of seconds passed since `Foundation`'s reference date of 00:00:00 UTC on 1 January 2001.

### Reproduction

```swift
struct DateDiffCalculator: Handler {
    @Parameter(.http(.query)) var start: Date
    @Parameter(.http(.query)) var end: Date
    
    func handle() -> TimeInterval {
        return end.timeIntervalSince(start)
    }
}


    var content: some Component {
        DateDiffCalculator()
    }
    
        REST()
    }
}
```

### Expected behavior

The web service should be able to handle `Date` objects as path or query parameters.

Additionally, it should be possible to specify how `Date` path/query parameters should be interpreted (e.g. whether they should be decoded as UNIX timestamps, Foundation timestamps, ISO8601, or something else entirely).

All of this is actually already implemented, we (i.e. I) just forgot to connect the date decoding implementation with the parameter decoding components.

### Additional context

_No response_

### Code of Conduct

lukaskollmer:
# Improve Parameter Handling

Because Date doesn't conform to LosslessStringConvertible (for good reasons, btw), date objects cannot be used as path or query parameters in HTTP/REST endpoints. It is possible to work around this, by adding a dummy extension Date: LosslessStringConvertible conformance somewhere in the web service, but that still produces invalid parameter values: All Date query parameters are interpreted as the amount of seconds passed since Foundation's reference date of 00:00:00 UTC on 1 January 2001.

## :bulb: Proposed solution
We allow `Date` objects as lightweight parameters (despite the fact that they don't conform to `LosslessStringConvertible`, and without adding such a conformance), and expose configuration options (with reasonable default values) to Interface Exporters (and interface exporter configurations), in order to allow Handlers to use `Date` objects as lightweight parameters.

## :gear: Release Notes 
- Added support for `Date` objects as path or query parameters in the HTTP and REST interface exporters

I decided to set the default date decoding strategy to `secondsSince 1970`, since that seems like a more reasonable default than using Foundation's reference date.

### Related PRs
n/a

### Testing
There are tests included to check that the date decoding works properly.

### Reviewer Nudging
There really isn't too much code, so this should be straightforward, basically I just added a `dateDecodingStrategy` parameter to all functions/types between the REST/HTTP IEs and ApodiniNetworking's `HTTPRequest` class.

Fixes #415

### DIFF
--- a/Sources/Apodini/Semantic Model Builder/Model/EndpointParameter.swift
+++ b/Sources/Apodini/Semantic Model Builder/Model/EndpointParameter.swift
@@ -188,15 +188,18 @@ public struct EndpointParameter<Type: Codable>: _AnyEndpointParameter, Identifia
         let httpOption = options.option(for: PropertyOptionKey.http)
         switch httpOption {
         case .path:
-            precondition(Type.self is LosslessStringConvertible.Type, "Invalid explicit option .path for '\(description)'. Option is only available for wrapped properties conforming to \(LosslessStringConvertible.self).")
+            //precondition(Type.self is LosslessStringConvertible.Type, "Invalid explicit option .path for '\(description)'. Option is only available for wrapped properties conforming to \(LosslessStringConvertible.self).")
+//            precondition(Self.isValidPathOrQueryParamType, "Invalid option .path for \(description): Parameter type \(T.self) must ")
+            Self.assertIsValidLightweightParamType(forParameterMode: .path, description: description)
             parameterType = .path
         case .query:
-            precondition(Type.self is LosslessStringConvertible.Type, "Invalid explicit option .query for '\(description)'. Option is only available for wrapped properties conforming to \(LosslessStringConvertible.self).")
+//            precondition(Type.self is LosslessStringConvertible.Type, "Invalid explicit option .query for '\(description)'. Option is only available for wrapped properties conforming to \(LosslessStringConvertible.self).")
+            Self.assertIsValidLightweightParamType(forParameterMode: .query, description: description)
             parameterType = .lightweight
         case .body:
             parameterType = .content
         default:
-            parameterType = Type.self is LosslessStringConvertible.Type ? .lightweight : .content
+            parameterType = Self.isValidLightweightParamType ? .lightweight : .content
         }
     }
 
@@ -214,6 +217,16 @@ public struct EndpointParameter<Type: Codable>: _AnyEndpointParameter, Identifia
 
     func exportParameter<I: InterfaceExporter>(on exporter: I) -> I.ParameterExportOutput {
         exporter.exportParameter(self)
+    }
+    
+    private static var isValidLightweightParamType: Bool {
+        Type.self is LosslessStringConvertible.Type || Type.self == Foundation.Date.self
+    }
+    
+    private static func assertIsValidLightweightParamType(forParameterMode parameterMode: HTTPParameterMode, description: String) {
+        if !isValidLightweightParamType {
+            fatalError("Invalid option .\(parameterMode) for \(description): Parameter type \(Type.self) must conform to \(LosslessStringConvertible.self), or be \(Foundation.Date.self).")
+        }
     }
 }
 
--- a/Sources/Apodini/Semantic Model Builder/Model/EndpointParameter.swift
+++ b/Sources/Apodini/Semantic Model Builder/Model/EndpointParameter.swift
@@ -188,12 +188,9 @@ public struct EndpointParameter<Type: Codable>: _AnyEndpointParameter, Identifia
         let httpOption = options.option(for: PropertyOptionKey.http)
         switch httpOption {
         case .path:
-            //precondition(Type.self is LosslessStringConvertible.Type, "Invalid explicit option .path for '\(description)'. Option is only available for wrapped properties conforming to \(LosslessStringConvertible.self).")
-//            precondition(Self.isValidPathOrQueryParamType, "Invalid option .path for \(description): Parameter type \(T.self) must ")
             Self.assertIsValidLightweightParamType(forParameterMode: .path, description: description)
             parameterType = .path
         case .query:
-//            precondition(Type.self is LosslessStringConvertible.Type, "Invalid explicit option .query for '\(description)'. Option is only available for wrapped properties conforming to \(LosslessStringConvertible.self).")
             Self.assertIsValidLightweightParamType(forParameterMode: .query, description: description)
             parameterType = .lightweight
         case .body:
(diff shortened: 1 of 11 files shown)


Tests/ApodiniNetworkingTests/ApodiniNetworkingTests.swift
```diff
@@ -69,4 +70,44 @@ class ApodiniNetworkingTests: XCTestCase {
...
+            XCTAssertThrowsError(try URLQueryParameterValueDecoder().decode(Bool.self, from: invalidBoolInput))
+        }
+    }
+    
+    
+    func testQueryParamDateDecoding() throws {
+        func imp(_ strategy: DateDecodingStrategy, input: String) throws -> Date {
+            try URLQueryParameterValueDecoder(dateDecodingStrategy: strategy).decode(Date.self, from: input)
+        }
+        for _ in 0...1000 {
```
philippzagar, line 102: Is that actually effective/necessary?

Sources/ApodiniREST/RESTExporterConfiguration.swift
```diff
@@ -19,6 +19,8 @@ extension REST {
         public let encoder: AnyEncoder
         /// The to be used `AnyDecoder` for decoding requests to the `RESTInterfaceExporter`
         public let decoder: AnyDecoder
+        /// How `Date` objects passed as query or path parameters should be decoded.
```
philippzagar: Really picky: Remove the dot at the end, just for consistency

Sources/ApodiniNetworking/URLQueryParamDecoder.swift
```diff
@@ -10,69 +10,78 @@ import Foundation
...
+        var errorDescription: String? {
+            "Unable to decode date from raw input '\(rawInput)'. Used strategy: \(decodingStrategyDesc)"
         }
     }
     
+    /// Decodes `Date` objects from an ISO8601 string
+    case iso8601
+    /// Decodes `Date` objects from a UNIX timestamp
+    case secondsSince1970
+    /// Decodes `Date` objects from a timestamp relative to Foundation's reference date of 00:00:00 UTC on 1 January 2001.
```
philippzagar: Really picky again: Remove dots for consistency

Sources/Apodini/Semantic Model Builder/Model/EndpointParameter.swift
```diff
@@ -188,15 +188,18 @@ public struct EndpointParameter<Type: Codable>: _AnyEndpointParameter, Identifia
         let httpOption = options.option(for: PropertyOptionKey.http)
         switch httpOption {
         case .path:
-            precondition(Type.self is LosslessStringConvertible.Type, "Invalid explicit option .path for '\(description)'. Option is only available for wrapped properties conforming to \(LosslessStringConvertible.self).")
+            //precondition(Type.self is LosslessStringConvertible.Type, "Invalid explicit option .path for '\(description)'. Option is only available for wrapped properties conforming to \(LosslessStringConvertible.self).")
```
philippzagar: Should we actually leave this in there?

Sources/ApodiniHTTP/Configuration.swift
```diff
@@ -19,6 +20,8 @@ extension HTTP {
         public let encoder: AnyEncoder
         /// The `AnyDecoder` to be used for decoding requests
         public let decoder: AnyDecoder
+        /// How `Date` objects passed as query or path parameters should be decoded.
```
philippzagar: Dot stuff again

Sources/ApodiniNetworking/URLQueryParamDecoder.swift
```diff
@@ -10,69 +10,78 @@ import Foundation
...
+    case secondsSinceReferenceDate
```
philippzagar, line 41: Maybe a more meaningful name? Suggestion: `secondsSinceFoundationReferenceDate`

Sources/ApodiniNetworking/URLQueryParamDecoder.swift
```diff
@@ -10,69 +10,78 @@ import Foundation
...
+    /// Decodes `Date` objects using a custom closure.
+    case custom((String) throws -> Date)
+    
+    /// The "default" date decoding strategy, which interprets dates as UNIX timestamps.
```
philippzagar: Dots

philippzagar (approved):
LGTM! Very minor suggestions. Thanks for the quick PR!

bauer-andreas (approved):
Looks great to me. Thanks for addressing this issue 👍 

Just, as Philipp already pointed out, removed those two occasions where there is still commented out code 👍

Sources/ApodiniREST/RESTExporterConfiguration.swift
bauer-andreas: I would argue to put dots everywhere to properly mark End of Sentence 🙃

Tests/ApodiniNetworkingTests/ApodiniNetworkingTests.swift
lukaskollmer: There were some inconsistencies where the `Date` decoding would randomly fail (due to how Foundation deals with timestamps; this is also the reason why I'm comparing the dates' `timeIntervalSince1970`, instead of comparing the dates directly), so I added the loop to repeat the test a bunch of times. This increases the possibility of us detecting a failure.

Sources/Apodini/Semantic Model Builder/Model/EndpointParameter.swift
lukaskollmer: nope, of course not, will remove

Sources/ApodiniNetworking/URLQueryParamDecoder.swift
lukaskollmer: reference date is fine, since that's a well-established term familiar to anyone who ever worked with Apple's date APIs. Also, it's explained in the documentation.

PSchmiedmayer (approved):

PR #417, opened 2022-01-28
Remove `___` suffixes from implicitly created wrapper proto types

lukaskollmer:
# Remove `___` suffixes from implicitly created wrapper proto types

The gRPC interface exporter sometimes needs to generate proto message types for things like collecting handler inputs or wrapping handler response values.
These types currently were created using a `{HandlerName}___{Suffix}` pattern, with the suffix being for example `Input`.
The initial reason for this was to avoid conflicts with user-defined types that may have the same name.

## :bulb: Proposed solution
We remove the underscores, and instead have a proper check for such name clashes. They can be avoided by setting, via the handler's metadata, an explicit typename to be used by the gRPC interface exporter. Such metadata definitions are available for both input and output wrapper types.
Alternatively, this issue of conflicting type names can also be avoided by changing the name of the not-auto-generated type, e.g. by directly changing its name within the Swift type system, or via a conformance to the `ProtoTypeWithCustomProtoName` protocol. (This isn't something new, but an existing solution to this problem.)

## :gear: Release Notes 
- gRPC Interface Exporter:
  - Changed protobuffer message typename generation to no longer use `___` underscores for auto-generated wrapper types.
  - Added `HandlerInputProtoMessageName` and `HandlerResponseProtoMessageName` metadata definitions for explicitly specifying the typenames of a `Handler`'s implicitly generated input and/or output proto message types.
- GraphQL Interface Exporter:
  - Added custom scalar types for commonly-used Swift types such as `Date`, `URL`, or `Data`
    - `Date` objects are represented as ISO8601-encoded strings
    - `URL` objects are represented as strings containing the full and absolute URL
    - `Data` objects are represented as base-64 encoded strings
  - Added custom scalar types for Swift's integer types with > 32 bits
    - the `Int64` scalar type maps to Swift's `Int` and `Int64` types
    - the `UInt64` scalar type maps to Swift's `UInt` and `UInt64` types
    - (we're assuming that Apodini is running on a 64-bit system here, which should be fine since the probability of someone running Apodini on an old Apple Watch is rather low...)
    - There is a flag to disable these custom 64-bit Int scalars, which will cause the GraphQL interface exporter to throw an error instead. (This option is exposed via the `enableCustom64BitIntScalars` flag in the `GraphQL` configuration type)

n/a

### Related PRs
n/a

### Testing
The existing gRPC tests have been adjusted to contain a handler with the new metadata options set, thus testing that the user-defined options, if specified, are picked up correctly.

### Reviewer Nudging
The code.

### DIFF
--- a/Tests/ApodiniTests/WebSocketInterfaceExporterTests.swift
+++ b/Tests/ApodiniTests/WebSocketInterfaceExporterTests.swift
@@ -297,17 +297,24 @@ class WebSocketInterfaceExporterTests: XCTApodiniTest {
         let errorForwardingExporter = ErrorForwardingInterfaceExporter {
             forwardedError = $0
         }
+        print("register exporter")
         app.registerExporter(exporter: errorForwardingExporter)
 
         let testCollection = TestWebSocketExporterCollection()
+        print("configure app")
         testCollection.configuration.configure(app)
 
+        print("create STV")
         let visitor = SyntaxTreeVisitor(modelBuilder: SemanticModelBuilder(app))
+        print("accept STV")
         testService.accept(visitor)
+        print("finish parsing")
         visitor.finishParsing()
 
+        print("start app")
         try app.start()
 
+        print("create client")
         let client = StatelessClient(on: app.eventLoopGroup.next())
 
         let userId = "1234"
@@ -317,7 +324,9 @@ class WebSocketInterfaceExporterTests: XCTApodiniTest {
             let userId: String
         }
 
+        print("client.resolve")
         XCTAssertThrowsError(try client.resolve(one: InvalidUserHandlerInput(userId: userId), on: "user.:userId:").wait() as User)
+        print("check")
         let forwardedApodiniError = try XCTUnwrap(forwardedError as? ApodiniError)
         XCTAssertEqual(forwardedApodiniError.option(for: .errorType), .badInput)
     }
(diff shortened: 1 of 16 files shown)


PSchmiedmayer (approved):
Thanks for improving the names of the protocol buffer messages @lukaskollmer! 🚀
Also thanks for addressing the memory leaks and the retain cycle in the Web Socket exporter. The PR and the changes look good to me 👍

Tests/ApodiniTests/WebSocketInterfaceExporterTests.swift
```diff
@@ -297,17 +297,24 @@ class WebSocketInterfaceExporterTests: XCTApodiniTest {
         let errorForwardingExporter = ErrorForwardingInterfaceExporter {
             forwardedError = $0
         }
+        print("register exporter")
```
bauer-andreas: As I'm just randomly checking over this PR (to ensure this doesn't break anything on my side), might remove those debug print statements before merging 🙃

lukaskollmer: yeah i forgot to add a note on here but the PR is currently on-hold bc there's some test failures caused by the WebSocketIE which i'm also trying to fix via this PR, the print stmt you found was part of that there's probably some more

PSchmiedmayer (approved):

PR #426, opened 2022-02-09
Fix some ProtobufferCoding bugs

### LINKED ISSUE #424: [Bug]: ProtobufferCoding doesn't add explicit presence information to optional message properties.
### Description

To my testing, with the current state of `ApdoiniGRPC`/`ProtobufferCoding`, the `optional` label isn't added to message files where the Swift type is declared with an optional scalar value for the respective field (e.g. the message type `Foo` has a property like `var bar: String`).

### Reproduction

1. Write a minimal WebService using a type as described above as e.g. a response type:
```swift
struct FooResponse {
  var bar: String?
}
```
2. The generated proto file will contain something like the following:
```proto
message FooResponse {
  string bar = 1;
}
```

### Expected behavior

I would expect something like the following to be generated:
```proto
message FooResponse {
  optional string bar = 1;
}
```

I know there was a time, with initial versions of proto3 where this wasn't supported. But afaik this syntax is supported with current versions of proto3 syntax (e.g. refer to https://stackoverflow.com/a/62566052/15337360).

### Additional context

CC @lukas

### Code of Conduct

### LINKED ISSUE #425: [Bug]: Not sending required parameters results in server crash
### Description

When ProtobufferCoding expects a given field in a message to be present but it is indeed not part of the request, it will fatal out.

### Reproduction

Given the following `Handler` implementation:

```swift
struct AddReviewHandler: Handler {
    @Parameter(.http(.body))
    var review: ReviewForm
    
    @Parameter
    var eventID: Event.ID
    
    @Parameter
    var userID: User.ID
    
    @Parameter
    var displayName = false
    
    func handle() throws -> ReviewForm {
        try .instance(or: notFound)
    }
}
```

This will yield the following proto file.

```proto
service SomeService {
  rpc AddReviewHandler(.QONECTIQ2.AddReviewHandlerInput) returns (.QONECTIQ2.ReviewForm);
}

message AddReviewHandlerInput {
  .QONECTIQ2.ReviewForm review = 1;
  string eventID = 2;
  string userID = 3;
  bool displayName = 4;
}
```

Sidenote: `displayName` is wrongfully not declared as optional. This is tracked with https://github.com/Apodini/Apodini/issues/424. But that doesn't matter here.

You note, that all four fields are declared as required. Now if you not e.g. `review` in your request to the web service, the web service will crash:
```
ProtobufferCoding/KeyedDecodingContainer.swift:206: Fatal error: Unable to get field info
```

Source line: https://github.com/Apodini/Apodini/blob/95e67fbb85aac175e5f0dbfb3f17a79d363d9b83/Sources/ProtobufferCoding/Decoding/KeyedDecodingContainer.swift#L206

### Expected behavior

Instead of crashing, creating a potential denial of service vector, the error should be gracefully handled and returned to the client.

### Additional context

_No response_

### Code of Conduct

lukaskollmer:
# Fix some ProtobufferCoding bugs

The ProtobufferCoding target currently contains some bugs, for example #424 and #425.

## :bulb: Proposed solution
We should fix the bugs

## :gear: Release Notes 
- ApodiniGRPC:
  - Fix issue where the generated proto schema files would mark optional fields as such only for proto2 types, but not proto3 types
  - Fix issue where attempting to decode a non-existent field would, in some circumstances, result in a crash, thus shutting down the entire web service. Now throws an error instead.
  - Fix issue where Apodini constructing an ApodiniError object with an error message containing (via string interpolation) another error object, which happens to be a `DecodingError` object with a `CodingKey` property where this `CodingKey` is a `FixedCodingKey`, would result in the `DecodingError`'s `debugDescription` property getting called, which in turn would try to produce a string description for the contained `FixedCodingKey`, which would go through `CodingKey`'s conformance to `DebugStringConvertible`, where the default implementation for `debugDescription` would try to access the `FixedCodingKey`'s `stringValue` property, which we had replaced with a `fatalError`, would result in a crash.
  - Improved quality of errors thrown when attempting to decoding non-existent fields, by also including the field's name in the error message (instead of just the field number)

n/a

### Related PRs
- Fixes #424
- Fixes #425 

### Testing
n/a

### Reviewer Nudging
n/a

### DIFF
--- a/Sources/ProtobufferCoding/Helpers.swift
+++ b/Sources/ProtobufferCoding/Helpers.swift
@@ -8,20 +8,36 @@
 
 import Foundation
 
-/// A helper type conforming to `CodingKey` that does not support string-based coding keys,
+/// A helper type conforming to `CodingKey` that does not necessarily support string-based coding keys,
 /// and will always return the same `intValue`.
 struct FixedCodingKey: CodingKey {
-    /// Guaranteed to be non-nil, but has to be nullable to satisfy the `CodingKey` protocol
-    let intValue: Int?
+    private let _intValue: Int
+    private let _stringValue: String?
     
     init(intValue: Int) {
-        self.intValue = intValue
+        self._intValue = intValue
+        self._stringValue = nil
+    }
+    
+    init(intValue: Int, stringValue: String) {
+        self._intValue = intValue
+        self._stringValue = stringValue
     }
     
     init?(stringValue: String) {
         fatalError("Not supported. Provide an integer")
     }
-    var stringValue: String {
-        fatalError("Not supported")
+    
+    /// Guaranteed to be non-nil, but has to be nullable to satisfy the `CodingKey` protocol
+    var intValue: Int? { _intValue }
+    /// Has to be non-nil to satisfy the `CodingKey` protocol. Will crash if there is no underlying string value.
+    var stringValue: String { _stringValue! }
+    
+    var description: String {
+        "\(Self.self)(intValue: \(_intValue), stringValue: \(_stringValue))"
+    }
+    
+    var debugDescription: String {
+        description
     }
 }
--- a/Sources/ApodiniGRPC/ProtoPrinter.swift
+++ b/Sources/ApodiniGRPC/ProtoPrinter.swift
@@ -176,8 +176,12 @@ struct ProtoPrinter {
         case .LABEL_REPEATED:
             write("repeated ")
         case .LABEL_REQUIRED:
-            precondition(self.fileDescriptor!.syntax == "proto2")
+            precondition(self.fileDescriptor?.syntax == ProtoSyntax.proto2.rawValue)
             write("required ")
+        }
+        
+        if self.fileDescriptor?.syntax == ProtoSyntax.proto3.rawValue && descriptor.proto3Optional {
+            write("optional ")
         }
         
         switch descriptor.type {
--- a/Sources/ProtobufferCoding/Decoding/Decoder.swift
+++ b/Sources/ProtobufferCoding/Decoding/Decoder.swift
@@ -51,7 +51,7 @@ public struct ProtobufferDecoder {
         }
         let decoder = _ProtobufferDecoder(codingPath: [], buffer: buffer)
         let keyedDecoder = try decoder.container(keyedBy: FixedCodingKey.self)
-        return try keyedDecoder.decode(T.self, forKey: .init(intValue: fieldInfo.fieldNumber))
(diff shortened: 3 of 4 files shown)


Sources/ProtobufferCoding/Helpers.swift
```diff
@@ -8,20 +8,36 @@
...
     init?(stringValue: String) {
         fatalError("Not supported. Provide an integer")
     }
-    var stringValue: String {
-        fatalError("Not supported")
+    
+    /// Guaranteed to be non-nil, but has to be nullable to satisfy the `CodingKey` protocol
+    var intValue: Int? { _intValue }
+    /// Has to be non-nil to satisfy the `CodingKey` protocol. Will crash if there is no underlying string value.
+    var stringValue: String { _stringValue! }
```
bauer-andreas, line 34: SwiftLint force unwrapping.
Instead of force unwrapping, can we at least use an optional binding guard and provide a small error message in the `fatalError`? As we have control over every `init` this is only a minor issue and we don't need to find a way to handle that gracefully.

Sources/ProtobufferCoding/Helpers.swift
```diff
@@ -8,20 +8,36 @@
...
+        self._stringValue = nil
+    }
+    
+    init(intValue: Int, stringValue: String) {
+        self._intValue = intValue
+        self._stringValue = stringValue
     }
     
     init?(stringValue: String) {
         fatalError("Not supported. Provide an integer")
```
bauer-andreas, line 28: Instead of fataling out, can we use the fact, that this init is declared as optional?
Meaning, just setting `self = nil` and rely on the party using it to generate the error message (which potentially includes more context?). I think, doing so could avoid similar errors as the once we are fixing right now (were some StringConvertible conformance access the properties).
For reference: https://developer.apple.com/documentation/swift/codingkey/2893595-init.

bauer-andreas (approved):
Thanks for addressing those issues.

Sources/ProtobufferCoding/Helpers.swift
lukaskollmer: Sure, I will change that.

Sources/ProtobufferCoding/Helpers.swift
lukaskollmer: We could, but since this is an internal type and only intended to be used with a non-nil int value, i figured it'd be the best approach to make sure you can in fact only use it that way.
It would of course be possible to delegate error handling here to the caller, but i don't see how there is any more context that could be added. Either an int value is provided or not. (IMO the error we're fixing right now is in fact an argument in favour of using a crashing `fatalError` instead of just returning `nil`.)

lukaskollmer:
Closed in favour of #429, which now contains this PR's changes as well

PR #429, opened 2022-02-27
[WIP] Various Small Improvements

### LINKED ISSUE #424: [Bug]: ProtobufferCoding doesn't add explicit presence information to optional message properties.

### LINKED ISSUE #425: [Bug]: Not sending required parameters results in server crash

lukaskollmer:
# Various Small Improvements

A variety of small changes and improvements to a couple of places in Apodini.
This also contains some preparatory groundwork for potentially implementing proper HTTP/2-based streaming in the HTTP Interface Exporter.
Also fixes a bug where the WebSocket Interface Exporter would send the last request in a connection twice, by using `.end` for the last handler invocation, instead of `.close`. (This change might affect the behaviour of the other interface exporters, but the tests are all ok so i'm assuming that it's fine.)

## :bulb: Changes
- Make the initialisers for `AnyComponent` and `AnyHandler` public. (This is to match `SomeHandler`, which already has a public initialiser.)
- Improve the HTTP Route Registration API support optionally adding communication pattern information to HTTP routes. (This information is currently not used, but it will allow ApodiniNetworking to determine whether an incoming HTTP request should be handled as a "normal" request or as a stream. The problem here is that there is no way to determine this by looking at just the header of the HTTP request, so we need the information from elsewhere.)
- Change the StatefulEvaluation APIs to send a `.close` event instead of `.end` when a stream is closed without an accompanying request
- Refactor the HTTP Route Registration and HTTP Route Responding APIs out of HTTPServer.swift
- Make the `registerRoute` function throw if there was an error registering a route. (None of these errors are currently being properly caught, this is to match the previous behaviour since it used to call `fatalError` instead of throwing.)
- Rename the HTTP Interface Exporter to `HTTPInterfaceExporter`
- Make the HTTPInterfaceExporter a class (it was a struct until now)
- Make the `InterfaceExporter` protocol class-bound, so that all of Apodini can operate on the assumption that instances of interface exporters are references
- HTTPUpgradeHandler: Remove the ChannelHandler from the pipeline if it received a non-upgrade request
- HTTPServer: improve logging of protocol/network-level errors
- ApodiniGRPC:
  - Fix issue where the generated proto schema files would mark optional fields as such only for proto2 types, but not proto3 types
  - Fix issue where attempting to decode a non-existent field would, in some circumstances, result in a crash, thus shutting down the entire web service. Now throws an error instead.
  - Fix issue where Apodini constructing an ApodiniError object with an error message containing (via string interpolation) another error object, which happens to be a `DecodingError` object with a `CodingKey` property where this `CodingKey` is a `FixedCodingKey`, would result in the `DecodingError`'s `debugDescription` property getting called, which in turn would try to produce a string description for the contained `FixedCodingKey`, which would go through `CodingKey`'s conformance to `DebugStringConvertible`, where the default implementation for `debugDescription` would try to access the `FixedCodingKey`'s `stringValue` property, which we had replaced with a `fatalError`, would result in a crash.
  - Improved quality of errors thrown when attempting to decoding non-existent fields, by also including the field's name in the error message (instead of just the field number)

## :gear: Release Notes 
- ApodiniGRPC:
  - Improve handling of enum fields and optional enum fields when en/decoding from/to Protocol Buffers
  - Improved error messages for encoding/decoding issues
- Handlers are now sent a `.close` event instead of `.end` when a stream is closed without an accompanying request
- Various internal changes and bug fixes which should not directly affect clients

None of the changes here are particularly important or useful on their own, but especially the ApodiniNetworking stuff should make it easier to implement proper HTTP/2 streaming for the HTTP interface exporter.

### Related PRs
- Supersedes #426, which contained some changes to the ProtobufferCoding target, which were merged into this PR
- Fixes #424
- Fixes #425

### Testing
n/a since no new features were added.

### Reviewer Nudging
The change to StatefulEvaluation.swift in particular is something I'd like some feedback on, since this will probably have repercussions on the other Interface Exporters, but I'm not sure what/where exactly.


PSchmiedmayer (approved):
@lukaskollmer Thanks a lot for the PR, the changes look good. Thanks for improvements 🚀

PR #433, opened 2022-04-13
First version of ApodiniAudit

simonbohnen:
# First version of ApodiniAudit

## :gear: Release Notes 
This is the first version of ApodiniAudit, which enables developers to automatically test their APIs against common best practices for REST and HTTP. It introduces the `RESTAuditor` and `HTTPAuditor` Configurations which register the `audit` CLI command. This command performs the audit, currently checking for simple best practices such as adequate length of URL path segments.

### Changes outside ApodiniAudit
The concept of `DependentStaticConfigurations` was generalized. The `HTTPExporter` now accepts dependent configurations too, and the `REST` and `OpenAPI` configurations had to be adapted slightly to fit the new model.

### Testing
I am going to implement testing for all the best practices at a later stage.

### Reviewer Nudging
`AuditStaticConfiguration` is a good entry point. It registers the `AuditInterfaceExporter`, which manages the auditing process, and is available as a static configuration for the `HTTP` and `REST` exporters.


Sources/Apodini/Configurations/DependentStaticConfigurationBuilder.swift
```diff
@@ -0,0 +1,48 @@
...
+/// A function builder used to aggregate multiple `DependentStaticConfiguration`s
+@resultBuilder
+public enum DependentStaticConfigurationBuilder<ParentConfiguration: ConfigurationWithDependents> {
+    
+    /// A method that transforms multiple `DependentStaticConfiguration`s
+    ///
+    /// - Parameter staticConfigurations: A variadic number of `DependentStaticConfiguration`
+    ///
+    /// - Returns: An `AnyConfigurationCollection` which consists of `ConfigurationConvertible`s
+    public static func buildBlock(_ staticConfigurations: [AnyDependentStaticConfiguration]...) -> [AnyDependentStaticConfiguration] {
```
PSchmiedmayer, line 28: I think using variadic parameters and in generally `AnyDependentStaticConfiguration` works in a way that one can pass `ConfigurationWithDependents` to the builder but not only restricting them to a specific configuration type by constraining it to a specific `InternalConfiguration`.
I think that `ParentConfiguration` does not need to conform to `ConfigurationWithDependents` and that the constraint should then be used with concrete `ConfigurationWithDependents` in the build functions using the Swift `where` clause. Otherwise one can pass any `ConfigurationWithDependents` to the builder from the outside and we have the same level of type safety as we have internally using the type erased version.

```swift
@resultBuilder
public struct DependentStaticConfigurationBuilder<Configuration> {
    public static func buildIf<T: ConfigurationWithDependents>(_ component: T?) -> T? where T.Configuration == Configuration {
        return component
    }
    
    public static func buildBlock<T: ConfigurationWithDependents>(_ arg0: T) -> T where T.Configuration == Configuration {
        return arg0
    }

    public static func buildEither<T: ConfigurationWithDependents>(first: T) -> T where T.Configuration == Configuration {
        return first
    }
    
    public static func buildEither<T: ConfigurationWithDependents>(second: T) -> T where T.Configuration == Configuration {
        return second
    }

    public static func buildIf<T: ConfigurationWithDependents>(first: T) -> T where T.Configuration == Configuration {
        return first
    }

    public static func buildBlock<T0: ConfigurationWithDependents, T1: ConfigurationWithDependents >(_ arg0: T0, _ arg1: T1) -> [AnyDependentStaticConfiguration] where T0.Configuration == Configuration, T1.Configuration == Configuration {
        return [arg0, arg1]
    }
}
```
We will need more buildBlock functions with more types, as [Swift sadly does not yet support variadic generics (but this is on the way)](https://forums.swift.org/t/variadic-generics/54511)

Sources/ApodiniAudit/BestPractices/BestPractice.swift
```diff
@@ -0,0 +1,33 @@
...
+}
+
+extension BestPractice {
+    static func audit(_ app: Application, _ endpoint: AnyEndpoint) -> Audit {
+        let auditReport = check(app, endpoint)
+        return Audit(report: auditReport, endpoint: endpoint, bestPracticeType: Self.self)
+    }
+}
+
+public enum BestPracticeCategory {
```
PSchmiedmayer: Probably an [OptionSet](https://developer.apple.com/documentation/swift/optionset) might make sense for the category and especially for the scope. Some best practices might lie in multiple categories and the scope might be across multiple exporter types.

Sources/ApodiniHTTP/InterfaceExporter.swift
```diff
@@ -58,6 +63,35 @@ public final class HTTP: Configuration {
...
+    ///    - caseInsensitiveRouting: Indicates whether the HTTP route is interpreted case-sensitively
+    ///    - rootPath: Configures the root path for the HTTP endpoints
+    ///    - staticConfigurations: A result builder that allows passing dependent static Exporters like the OpenAPI Exporter
+    public convenience init(
+        encoder: JSONEncoder = defaultEncoder as! JSONEncoder,
+        decoder: JSONDecoder = defaultDecoder as! JSONDecoder,
+        urlParamDateDecodingStrategy: ApodiniNetworking.DateDecodingStrategy = .default,
+        caseInsensitiveRouting: Bool = false,
+        rootPath: RootPath? = nil,
+        @DependentStaticConfigurationBuilder<HTTP> staticConfigurations: () -> [AnyDependentStaticConfiguration] = { [] }
```
PSchmiedmayer: As noted above, as far as I can see, the constraint currently doesn't have any effect here.

Sources/ApodiniHTTP/InterfaceExporter.swift
```diff
@@ -15,8 +15,11 @@ import ApodiniNetworking
 // MARK: HTTP Declaration
 
 /// Public Apodini Interface Exporter for basic HTTP
-public final class HTTP: Configuration {
+public final class HTTP: Configuration, ConfigurationWithDependents {
+    public typealias InternalConfiguration = ExporterConfiguration
+    
     let configuration: ExporterConfiguration
+    public var staticConfigurations = [AnyDependentStaticConfiguration]()
```
PSchmiedmayer, line 22: Do the static configurations have to be public? What would be the reason for making them available to the outside? Same applies to the REST exporter.

Sources/ApodiniREST/RESTInterfaceExporter.swift
```diff
@@ -172,6 +173,10 @@ final class RESTInterfaceExporter: InterfaceExporter, TruthAnchor {
                 app.logger.info("  - links to: \(destination.destinationPath.asPathString())")
             }
         }
+        
+        // TODO Use this to only run audit when wanted
+        // app.logger.info("\(app.storage[AuditStorageKey.self])")
+        // Auditor.audit(app, endpoint)
```
PSchmiedmayer: Please remove comments before merging the PR

TestWebService/Sources/TestWebService/TestWebService.swift
```diff
@@ -80,9 +81,12 @@ struct TestWebService: Apodini.WebService {
             )
         }
         
-        HTTP(rootPath: "http")
+        HTTP(rootPath: "http") {
+            HTTPAuditor()
```
PSchmiedmayer: I would suggest to name the auditors the same for all exporters. As noted on chat.tum.de: I think it makes sense to even use the same `InternalConfiguration` for the REST and HTTP Exporters.

Sources/Apodini/Configurations/DependentStaticConfigurationBuilder.swift
simonbohnen: resultBuilders allow for the conversion to an intermediate type using the `buildExpression` method, as explained [here](https://docs.swift.org/swift-book/ReferenceManual/Attributes.html#ID634). All subsequent methods can take `[AnyDependentStaticConfiguration]`s, as the correct type was already ensured by `buildExpression`. This can be verified by trying to pass an `OpenAPI` Configuration into a `HTTP` Configuration, which will not compile. Swift is actually being quite handy with types here for once!

Sources/ApodiniHTTP/InterfaceExporter.swift
simonbohnen: As `ConfigurationWithDependents` is defined in `Apodini`, it has to be public to be accessible from `ApodiniHTTP` and `ApodiniREST`. Swift seems to force me to make the property public too unfortunately. Removing the property from the protocol doesn't seem like a good option, because it is being used in the protocol's extension (see [here](https://github.com/Apodini/Apodini/pull/433/files#diff-5cb1ee72a2e05fe5677b3af34c6d22f81b964f8a72a53c562f2a656719336129)).

TestWebService/Sources/TestWebService/TestWebService.swift
simonbohnen: I've merged the two `InternalConfigurations`, which were already the same anyways.

Currently, the type parameter to a `DependentStaticConfiguration` is the outward facing `REST` or `HTTP` configuration, not the internal configuration. Thus, merging the `HTTPAuditor` and `RESTAuditor` is currently not possible, as this would require conforming to the same protocol twice.

If we were to take the internal configuration as a type parameter to `DependentStaticConfiguration` (which was your suggestion anyways @PSchmiedmayer 😅), then that would force us to make the `OpenAPI` configuration available as a dependent of the `HTTP` as well as the `REST` configuration. That's fine by me, and @PSchmiedmayer suggested it too.

Do we need to make any modifications to the OpenAPI exporter for it to work as a dependent of the HTTP exporter? Who would I best talk to for this @PSchmiedmayer?

Sources/Apodini/Configurations/DependentStaticConfigurationBuilder.swift
PSchmiedmayer: Nice, learned something new, very cool! Thanks for sharing the link and the insight 🚀

TestWebService/Sources/TestWebService/TestWebService.swift
PSchmiedmayer: Thanks for the response. I took a look at the diff and the current state and think that making the `OpenAPI` configuration available as a dependent configuration of the `HTTP` and the `REST` Exporter is definitely a good idea and would not require too many changes to the Exporter as far as I can see. 

I see two main differences that would need to be addressed:
1. The response type of an endpoint: The REST Exporter wraps a response in the `ResponseContainer` to add the HATEOAS information to the responses, which the HTTP Exporter does not do. Therefore it would make sense to add the way a response is packaged at the end to the `HTTPExporterConfiguration`. The REST Exporter would share with the OpenAPI Exporter that it uses the `ResponseContainer`, and the HTTP Exporter would share that it doesn't use a wrapping. Moving the `ResponseContainer` type to the ApodiniHTTPProtocol target would then probably be a reasonable idea.
2. The HTTP Exporter should Exporter all request-response Handlers the same way as the REST Exporter does. The exporters currently use different mechanisms to register the endpoints but should result in the same final product. That this is the same could only be guaranteed if they would share the code for registering the routes, e.g., the ApodiniHTTPProtocol target. The only area where this registration differs is in the streaming cases for Handlers (e.g., client-side stream). I am not sure if and how OpenAPI could reflect HTTP/2 streaming behavior, but this might then require some additional code in the OpenAPI Exporter. But this feature is currently still on the roadmap and not implemented, so it would not require immediate changes in the Exporter at the moment, and when you tackle this, we can see how and if this can be incorporated into the OpenAPI Exporter 👍

simonbohnen: That sounds good! Point 1 makes a lot of sense.

Would you include moving the route registration to `ApodiniHTTPProtocol` in this pull request? It seems like that is a bigger task which might deserve its own PR to me.

simonbohnen: I implemented the correct schema exporting for HTTP with OpenAPI! It looked to me like the modifications I made were the only ones necessary, right?

Sources/ApodiniAudit/InterfaceExporter/AuditStaticConfiguration.swift
```diff
@@ -0,0 +1,54 @@
...
+
+public final class APIAuditorConfiguration<Service: WebService>: DependentStaticConfiguration {
+    public typealias InteralParentConfiguration = HTTPExporterConfiguration
+    
+    public var command: ParsableCommand.Type? {
+        getAuditCommand(AuditCommand<Service>.self)
+    }
+    
+    public func configure(_ app: Apodini.Application, parentConfiguration: HTTPExporterConfiguration) {
+        registerInterfaceExporter(app, mode: .rest)
```
PSchmiedmayer: I think it would be needed to make the mode dependent on the `HTTPExporterConfiguration`.

Sources/ApodiniAudit/InterfaceExporter/AuditInterfaceExporter.swift
```diff
@@ -0,0 +1,52 @@
...
+}
+
+extension AuditInterfaceExporter {
+    static let bestPractices: [BestPractice.Type] = [
+        AppropriateLengthForURLPathSegments.self,
+        NoUnderscoresInURLPathSegments.self
+    ]
+}
+
+public enum AuditMode {
```
PSchmiedmayer: Would it be possible to replace this with the `HTTPExporterConfiguration` instead of a separate AuditMode enum as both could express the same functionality and potentially more?

Sources/ApodiniAudit/InterfaceExporter/AuditStaticConfiguration.swift
```diff
@@ -0,0 +1,54 @@
...
+    }
+    
+    public init() { }
+}
+
+// FUTURE write test that _commands works
```
PSchmiedmayer: I would suggest to make the types and the functions static functions of the `APIAuditorConfiguration` to give them sone sort of scope and not add too many global functions and properties.

Sources/ApodiniOpenAPI/Document/OpenAPIComponentsObjectBuilder.swift
```diff
@@ -33,18 +38,27 @@ class OpenAPIComponentsObjectBuilder {
...
         let schemaName = "\(title)Response"
-        let schemaObject: JSONSchema = .object(
-            title: schemaName,
-            properties: [
-                ResponseContainer.CodingKeys.data.rawValue: schema,
-                ResponseContainer.CodingKeys.links.rawValue: try buildSchema(for: ResponseContainer.Links.self)
-            ])
+        let schemaObject: JSONSchema
+        
+        if configuration.parentConfiguration.useResponseContainer {
```
PSchmiedmayer: Nice, great to see that the changes to make the OpenAPI exporter support the HTTP export were quite small and easy to do 🚀

PSchmiedmayer (approved):
Thanks for the first Apodini Auditor PR! The changes look good, I only had a few small comments. This PR can be merged once these are addressed, the test code coverage is increased, and the SwiftLint errors are resolved.
Thanks for the work that went into the PR 🚀

PR #434, opened 2022-04-13
Refine the Hostname and Port Configuration Selection Logic

PSchmiedmayer:
# Refine the Hostname and Port Configuration Selection Logic

The Apodini Migrator currently uses the bind address as the hostname for the API document and, therefore, the client library generation. This is, e.g., a problem when a developer uses a web service in a docker container (requires the docker IP address or wildcard 0.0.0.0) or does not provide a bind address that is not the wildcard address (0.0.0.0) the Apodini Migrator uses this bind address which is not usable for outside connections.

In addition, the port number specified by the bind address does not take prescience over the default HTTP port if there is no explicit port defined by the hostname.

## :bulb: Proposed solution

The Apodini HTTP Configuration offers the hostname configuration for this exact purpose. We should use this configuration. This PR proposes to change the hostname port assignment by using the bind address port before the default port if provided by the developer. 

## :gear: Release Notes 

- Changes the behavior that the bind address port does take prescience over the default HTTP port if there is no explicit port defined by the hostname. 
- The Apodini Migrator uses the hostname for the hostname in the API document instead of the bind address.

### Related PRs
A problem occurred in https://github.com/Apodini/ApodiniMigratorExample/pull/3 lead to the detection of this problem.

### Testing
Added an additional unit test.

### DIFF
--- a/Sources/Apodini/Configurations/HTTPConfiguration.swift
+++ b/Sources/Apodini/Configurations/HTTPConfiguration.swift
@@ -75,7 +75,7 @@ public final class HTTPConfiguration: Configuration {
             self.bindAddress = .interface(Defaults.bindAddress, port: defaultPort)
         }
         
-        // We use the port from the bindAddress as the default port for the hostname, taking presicence over the defailt HTTP port
+        // We use the port from the bindAddress as the default port for the hostname, taking precedence over the default HTTP port.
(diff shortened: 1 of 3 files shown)


bauer-andreas (approved):
Thanks for addressing this 👍

PR #442, opened 2022-07-12
HTTP/2 Streaming

simonbohnen:
# HTTP/2 Streaming

The streaming `CommunicationPattern`s were previously not supported for the HTTPInterfaceExporter. The array-based streaming via HTTP/1.1 does not allow the server or client to dynamically react to the other peer's messages.

## :bulb: Solution
A lightweight protocol of length-prefixed messages via HTTP/2 streams. The implementation includes:
- Combining HTTP/1.1 array-based streaming with HTTP/2 length-prefixed streaming (see [here](https://github.com/Apodini/Apodini/pull/442/files#diff-8c61fcf38099ab9ea4e96b88af511e5f12c070dfe2b52173714aa3b1658760ad) and [here](https://github.com/Apodini/Apodini/pull/442/files#diff-2092aeab627896031cd28bdf232fc8abf9de3a0a4a3d0a27512f5f9cf4d48ca1))
- Providing a client which speaks this length-prefixed protocol: see [here](https://github.com/Apodini/Apodini/pull/442/files#diff-bc07c94403572167fdfa38342915f242872b62111695e53afa222a75f77a0862) and [here](https://github.com/Apodini/Apodini/pull/442/files#diff-8bab2a5d67c09812c2b6c799b4672598d24cb9be4e9149852fec1e7d989043e1)

### Testing
- Tests for each stream type, e.g. [bidirectional streaming](https://github.com/Apodini/Apodini/pull/442/files#diff-638376e8e760cc0586fd891f97d7cda4a135c9dbd37149de02a0c80a18b53b85)
- Tests for [error handling](https://github.com/Apodini/Apodini/pull/442/files#diff-35bd74dbc8a002afe9a15527adc8c298ea63ee1eb5132b24bfba1f093959b1a6)

### Reviewer Nudging
[Bidirectional streaming](https://github.com/Apodini/Apodini/blob/simonbohnen/http2/Sources/ApodiniHTTP/ClosureBuilders/BidirectionalStreaming.swift) as the most complete type of streaming
[AsyncSequence](https://github.com/Apodini/Apodini/blob/simonbohnen/http2/Sources/ApodiniHTTP/ClosureBuilders/HTTPRequestStreamAsyncSequence.swift) for reading length-prefixed messages from a stream


Sources/ApodiniHTTP/ClosureBuilders/HTTP1Compatibility/BidirectionalStreaming.swift
```diff
@@ -19,90 +19,127 @@ extension HTTPInterfaceExporter {
...
-            }
-            return data
+        _buildBidirectionalStreamingClosure(for: endpoint, using: defaultValues) { (response: Blob?) -> Data? in
+            precondition(response?.byteBuffer.readerIndex == 0)
+            return response?.byteBuffer.getAllData()
         }
     }
     
     
     private func _buildBidirectionalStreamingClosure<H: Handler>(
```
simonbohnen, line 42: This is a good place to start reviewing.

Sources/ApodiniHTTP/ClosureBuilders/HTTP1Compatibility/HTTPRequestStreamAsyncSequence.swift
```diff
@@ -0,0 +1,100 @@
...
+
+import Apodini
+import ApodiniNetworking
+import Foundation
+
+/// An ``AsyncSequence`` which emits ``Data`` objects of serialized JSON.
+/// The ``Data`` objects are created by reading length-prefixed blocks from an ``HTTPRequest``'s stream body storage.
+///
+/// This is used for streaming ``CommunicationPatterns`` via HTTP/2 DATA frames.
+class HTTPRequestStreamAsyncSequence: AsyncSequence, AsyncIteratorProtocol {
```
simonbohnen, line 17: This is also a good entry point as it is the heart of the streaming.

Sources/Apodini/Apodini.docc/Advanced/CommunicationPattern/PatternImplementation/PatternImplementation.md
```diff
@@ -30,12 +30,126 @@ Currently, most exporters do not support all communication patterns:
...
+    func handleInbound(response: String, serverSideClosed: Bool) {
+        // As a String cannot be decoded into JSON directly, this method won't be called
+    }
+    
+    func handleInboundNotDecodable(buffer: ByteBuffer, serverSideClosed: Bool) {
+        guard let countryString = buffer.getString(at: 0, length: buffer.readableBytes) else {
+            return
+        }
+        if countryString != countries[nextExpectedIndex] {
+            fatalError("Got the wrong country!")
```
PSchmiedmayer, line 124: Wouldn't this code example always run into an `fatalError` as the `Handler` always emits something like `"Hello, Germany!"` and not the country?
You would need to check if the string contains the country or the specific concatenation of `"Hello, \(country)!"`.
I might also just miss something there, let me know if I have a wrong understanding of the code here!

Tests/ApodiniTests/HTTP2Tests/HTTP2BidirectionalTests.swift
```diff
@@ -0,0 +1,89 @@
+//
```
PSchmiedmayer: I would suggest moving all HTTP2 Test into the ApodiniHTTPTests target 👍

PSchmiedmayer (approved):
Thanks again for bringing HTTP/2 streaming to Apodini @simonbohnen 🚀
I have taken a look at the PR and only had a question in the documentation, as well as a request to move a few test files to a different folder. 
One thing that would be really great would be if you could add a `curl` call in the TestWebService `test.sh` script to verify that the HTTP/2 connection is working there. Doesn't need to be something too special, e.g., a `curl` call to the `Countdown` service side streaming handler would be a great addition.

One additional thing that would be great if you could take a look at this in this PR: The GitHub Action builds seem to have failed after your latest PR was merged in: https://github.com/Apodini/Apodini/runs/7993377945?check_suite_focus=true ... seems like this is a setup issue in the GitHub Action that runs when something is pushed in the main branch regarding the Python setup. Do you think that you can take a look at this in this PR?

Apart from these comments, there is nothing that would speak against merging the PR from my side 👍

Sources/Apodini/Apodini.docc/Advanced/CommunicationPattern/PatternImplementation/PatternImplementation.md
simonbohnen: Yep true, resolved that 👌🏼

TestWebService/testhttp2.sh
```diff
@@ -0,0 +1,11 @@
...
+#
+# This source file is part of the Apodini open source project
+#
+# SPDX-FileCopyrightText: 2019-2021 Paul Schmiedmayer and the Apodini project authors (see CONTRIBUTORS.md) <paul.schmiedmayer@tum.de>
+#
+# SPDX-License-Identifier: MIT
+#
+
+set -e
+curl --fail --insecure --request GET --data-binary "@teststreamingrequest" -H "Content-Type: application/data" --http2-prior-knowledge --output /dev/null https://localhost/http/countdown
```
simonbohnen: Added this in a separate sh file because the Docker Compose & Traefik stuff is running via port 80, which I didn't want to mess with. swift-nio-http2 only supports HTTP/2 via HTTPS, so that's why I had to use HTTPS on port 443.

PR #444, opened 2022-09-03
Fix NIOExtras dependency

simonbohnen:
The ApodiniAuditLinguaWebService builds failed as NIOExtras wasn't avaliable from the XCTApodiniNetworking target. I am not sure why the builds were working normally here. Anyways, this should fix the error.


PSchmiedmayer (approved):
@simonbohnen Thanks a lot for the fixes, greatly appreciated!

